PR #10, opened 2016-04-20
Update README.md

ryanmaclean:


README.md
```diff
@@ -19,7 +19,7 @@ All code in this repository is Python 3
 # Doing a local build
 
 ## Dependencies
-  1. Linux distribution:
+  1. You will need to use a Linux distribution, as:
```
cmaloney: This makes this have a really different form than everything else in the list (all of which just are the name of the needed thing).

PR #11, opened 2016-04-20
Use curl instead of wget

lingmann:
Fixes issue with installer failing to build with the following:

```
wget: error getting response: Connection reset by peer
The command '/bin/sh -c wget
https://github.com/andyshinn/alpine-pkg-glibc/releases/download/2.22-r5/glibc-2.22-r5.apk
&& apk --allow-untrusted add glibc-2.22-r5.apk && rm glibc-2.22-r5.apk
&& rm -rf /var/cache/apk/*' returned a non-zero code: 1
```

This is due to busybox wget in Alpine not supporting https. Use the installed curl instead.


spahl:
:ship:

PR #12, opened 2016-04-21
Add ssh_tunnel.py to ssh module

ghost:
Allows for simple, synchronous, persistent coordination

The purpose of this PR is to provide the utility for cleaner test implementation. Specifically:
-replacing parallel ssh implementation in test_util/installer_api_test.py
-replacing ssh async code thats executed synchronously in test_util/test_installer_ccm.py
-allowing the for the creation of a test_runner utility that runs test coordination commands on an arbitrary remote host with network visibility of a DCOS cluster

### DIFF
--- a/pytest/test_ssh_integration.py
+++ b/pytest/test_ssh_integration.py
@@ -5,15 +5,19 @@ import os
 import random
 import socket
 import subprocess
+import tempfile
 import time
 import uuid
-from contextlib import contextmanager
+from contextlib import closing, contextmanager
+from functools import partial
 
 import pytest
 from retrying import retry
 
 import pkgpanda.util
 from ssh.ssh_runner import MultiRunner, Node
+from ssh.ssh_tunnel import (SSHTunnel, TunnelCollection, run_scp_cmd,
+                            run_ssh_cmd)
 from ssh.utils import AbstractSSHLibDelegate, CommandChain
 
 
@@ -312,3 +316,52 @@ def test_tags_async(sshd_manager, loop):
                     "sleep",
                     "1"
                 ]
+
+
+def tunnel_write_and_run(remote_write_fn, remote_cmd_fn):
+    """write random data across the tunnel with a write function, then run a
+    remote command to read that same random data. Finally assert the returned
+    random data is the same
+    """
+    with tempfile.NamedTemporaryFile() as tmp_fh:
+        rando_text = str(uuid.uuid4())
+        tmp_fh.write(rando_text.encode())
+        tmp_fh.flush()
+        remote_tmp_file = '/tmp/' + str(uuid.uuid4())
+        remote_write_fn(src=tmp_fh.name, dst=remote_tmp_file)
+        returned_text = remote_cmd_fn(cmd=['cat', remote_tmp_file])
+        assert returned_text == rando_text
+
+
+def test_ssh_tunnel(sshd_manager):
+    with sshd_manager.run(1) as sshd_ports:
+        tunnel_args = {
+                'ssh_user': getpass.getuser(),
+                'ssh_key_path': sshd_manager.key_path,
+                'host': '127.0.0.1',
+                'port': sshd_ports[0]}
+        with closing(SSHTunnel(**tunnel_args)) as tunnel:
+            tunnel_write_and_run(tunnel.write_to_remote, tunnel.remote_cmd)
+
+
+def test_ssh_tunnel_collection(sshd_manager):
+    with sshd_manager.run(10) as sshd_ports:
+        tunnel_args = {
+                'ssh_user': getpass.getuser(),
+                'ssh_key_path': sshd_manager.key_path,
+                'host_names': ['127.0.0.1:'+str(i) for i in sshd_ports]}
+        with closing(TunnelCollection(**tunnel_args)) as tunnels:
+            for tunnel in tunnels.tunnels:
+                tunnel_write_and_run(tunnel.write_to_remote, tunnel.remote_cmd)
+
+
+def test_ssh_one_offs(sshd_manager):
+    with sshd_manager.run(1) as sshd_ports:
+        ssh_args = {
+                'ssh_user': getpass.getuser(),
+                'ssh_key_path': sshd_manager.key_path,
+                'host': '127.0.0.1',
+                'port': sshd_ports[0]}
+        scp = partial(run_scp_cmd, **ssh_args)
+        ssh = partial(run_ssh_cmd, **ssh_args)
+        tunnel_write_and_run(scp, ssh)
--- a/ssh/ssh_tunnel.py
+++ b/ssh/ssh_tunnel.py
@@ -1,0 +1,128 @@
+"""
+Module for creating persistent SSH connections for use with synchronous
+commands. Typically, tunnels should be invoked with a context manager to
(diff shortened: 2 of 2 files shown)


ssh/ssh_tunnel.py
```diff
@@ -0,0 +1,138 @@
...
+from contextlib import closing
+from subprocess import CalledProcessError, TimeoutExpired, check_call, check_output
+
+logger = logging.getLogger(__name__)
+
+SOCKET_DIR = '$HOME/.cache/ssh'
+SSH_CMD = [
+    '/usr/bin/ssh',
+    '-oConnectTimeout=10',
+    '-oControlMaster=auto',
```
malnick: 👍

malnick:
LGTM!

pytest/test_ssh_integration.py
```diff
@@ -63,6 +67,8 @@ def run(self, count):
         # Run sshd servers in parallel, cleaning up when the yield returns.
         subprocesses = []
         for port in sshd_ports:
+            print(self.key_path)
+            print(' '.join(['/usr/sbin/sshd', '-p{}'.format(port), '-f{}'.format(self.sshd_config_path), '-e', '-D']))
             subprocesses.append(subprocess.Popen(
                 ['/usr/sbin/sshd', '-p{}'.format(port), '-f{}'.format(self.sshd_config_path), '-e', '-D'],
```
kartick: Just being pedantic:
```
ssh_cmd = [ '/usr/sbin/sshd', .... ]
print(' '.join(ssh_cmd))
subprocesses.append(subprocess.Popen(ssh_cmd))
```

pytest/test_ssh_integration.py
```diff
@@ -312,3 +318,48 @@ def test_tags_async(sshd_manager, loop):
                     "sleep",
                     "1"
                 ]
+
+
+def tunnel_write_and_run(remote_write_fn, remote_cmd_fn):
```
kartick: Doc string here, please. Also will be nice if you can provide a default `remote_write_fn` and `remote_cmd_fn` that an importer can re-use.

ssh/ssh_tunnel.py
kartick: Love this control master. Have to point a small issue I have experienced when using persistent connections. If the network connection to the ssh goes down for whatever reason and recovers only after timeout, a subsequent ssh call will hang. The fix for it will be to remove the control file and re-run ssh. Not sure if we will hit that situation here, but will be nice to code around it.

pytest/test_ssh_integration.py
kartick: Nvm, spoke too soon. `remove_cmd` from context should be reusable default.

ssh/ssh_tunnel.py
```diff
@@ -0,0 +1,139 @@
+"""
+Module for creating persistent SSH connections for use with synchronous
+commands. Typically, tunnels should be invoked with a context manager to
+ensure proper cleanup. E.G.:
+with contextlib.closing(SSHTunnel(*args, **kwargs)) as tunnel:
+    ssh.write_to_remote('/usr/local/usrpath/testfile.txt', 'test_file.txt')
```
kartick: `tunnel.`

kartick:
Some minor-isa comments, other than that, looks good overall @mellenburg-mesosphere.

mellenburg:
Still necessary and still waiting on this to get in, just commenting that ghost is me, @mellenburg

ssh/ssh_tunnel.py
cmaloney: Code wise this is duplicating stuff that is in the async ssh module. Can we make it not be duplicated? 

Is there any particular reason why we don't want to turn on using managed connections for all of the installer SSH Library? Seems like a good global optimization

ssh/ssh_tunnel.py
```diff
@@ -0,0 +1,139 @@
...
+        self.port = port
+        start_tunnel = SSH_CMD + [
+            '-fnN',
+            '-i',  ssh_key_path,
+            '-p', str(port), self.target]
+        start_tunnel = ' '.join(start_tunnel)
+        logger.debug('Starting SSH tunnel: ' + start_tunnel)
+
+        # Make sure we can cache the socket before starting
+        check_call('mkdir -p '+SOCKET_DIR, shell=True)
```
cmaloney: None of this code should ever be using `shell=True`. Should always code the parameter list directly as a list of strings, in this case `['mkdir', '-p', SOCKET_DIR]`. For recursively making directories we also already have some helper functions which would be better to use

ssh/ssh_tunnel.py
```diff
@@ -0,0 +1,139 @@
...
+        try:
+            check_call(start_tunnel, shell=True)
+        except CalledProcessError:
+            logger.critical('SSHTunnel could not be established with cmd: ' + start_tunnel)
+            raise AssertionError('Failed to create SSHTunnel with: ' + start_tunnel)
```
cmaloney: Why is an AssertionError the right sort of error here? IT seems like the CalledPRocessError is more correct unless you're making a local exception type so people can catch exceptions coming from just ssh tunnel (And someone is actually doing that)

cmaloney:
Gave some initial / base comments which go for a lot of the code, this is going to need a couple more rounds of back and forth / work.

The big open questions are
1. Why don't we just do this in the core ssh library / async ssh library? Why do we need to duplicate so much of the command code. How can we minimize the duplication
2. Code shouldn't use `shell=True`

ssh/ssh_tunnel.py
malnick: @kartick How about trapping or using context manager to cleanup sockets on specified signals?

mellenburg:
@cmaloney 
1. Raising an AssertionError in the case of a CalledProcessError was a fossil and has been removed
2. shell=True has been removed everywhere
3. Reconciling this with the async ssh lib is a little bit tricky. The goal of this module was to provide a very simple ssh wrapper for doing synchronous commands on a small number of nodes so that we dont have a zoo of subprocess invocations to do the same thing . The current async implementation is quite complex and was not designed for use with a persistently open connection that must be synchronously created only once before executing a command chain and self destructs upon completion. Furthermore, I think the command chain mechanism is robust enough that I would not want to mess with it in a PR designed to provide a simple ssh API. Lastly, async uses its own brand of subprocesses, so you can't create a module that does that kind of busy work without throwing a wrench into how async is handling the various subroutines. E.G. the shared module would be a collection of of SSH args and maybe some kind of state function to say if the tunnel was created, and then another synchronous API module would have to be made that utilizes that.

Anyway, I think ssh/ssh_runner.py could benefit from this methodology and that the sync and async libs should ideally share an invocation of ssh, but doing so in this PR would greatly increase complexity, bit rot other code based on this API and introduce instability into the solid async implementation

lingmann:
I'm a 👍 

Seb and I chatted and this has decent value in being separate code and I'm in favor of not going overboard on deduplication efforts. As is, this makes it easier for people to write simple integration tests using synchronous SSH calls, which is often all you need.

@spahl can you chime in?

spahl:
I'm :ship: on this too. This logic is simple and readable for now.

We can refactor it later to see if we can avoid code repetition. I doubt it would make the code much simpler in this case with all the asyncio stuff.

PR #17, opened 2016-04-22
add public agent role

darkonie:
- add public agent role to installer.
- refactor installer code.
- update integration / unit tests.
- make agent_list and agent_public_list not required to pass validation.

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -228,7 +228,7 @@ def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_del
         state_file_path = os.path.join(state_json_dir, 'deploy.json')
         log.debug('retry executed for a state file deploy.json')
         for _host in hosts:
-            _remove_host(state_file_path, _host)
+            _remove_host(state_file_path, '{}:{}'.format(_host.ip, _host.port))
 
         # We also need to update total number of hosts
         json_state = _read_state_file(state_file_path)
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -16,6 +16,31 @@ class ExecuteException(Exception):
     """Raised when execution fails"""
 
 
+def nodes_count_by_type(config):
+    total_agents_count = len(config.get('agent_list', [])) + len(config.get('public_agent_list', []))
+    return {
+        'total_masters': len(config['master_list']),
+        'total_agents': total_agents_count
+    }
+
+
+def get_full_nodes_list(config):
+    def add_nodes(nodes, tag):
+        return [Node(node, tag) for node in nodes]
+
+    node_role_map = {
+        'master_list': 'master',
+        'agent_list': 'agent',
+        'public_agent_list': 'public_agent'
+    }
+    full_target_list = []
+    for config_field, role in node_role_map.items():
+        if config_field in config:
+            full_target_list += add_nodes(config[config_field], {'role': role})
+    log.debug("full_target_list: {}".format(full_target_list))
+    return full_target_list
+
+
 @asyncio.coroutine
 def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block=False, state_json_dir=None,
                   async_delegate=None, retry=False, options=None):
@@ -29,14 +54,7 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     if not os.path.isfile(pf_script_path):
         log.error("genconf/serve/dcos_install.sh does not exist. Please run --genconf before executing preflight.")
         raise FileNotFoundError('genconf/serve/dcos_install.sh does not exist')
-    targets = []
-    for host in config['master_list']:
-        s = Node(host, {'role': 'master'})
-        targets += [s]
-
-    for host in config['agent_list']:
-        s = Node(host, {'role': 'agent'})
-        targets += [s]
+    targets = get_full_nodes_list(config)
 
     pf = get_async_runner(config, targets, async_delegate=async_delegate)
     chains = []
@@ -61,15 +79,9 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     # Setup the cleanup chain
     cleanup_chain = ssh.utils.CommandChain('preflight_cleanup')
     add_post_action(cleanup_chain)
-
     chains.append(cleanup_chain)
-    master_agent_count = {
-        'total_masters': len(config['master_list']),
-        'total_agents': len(config['agent_list'])
-    }
-
     result = yield from pf.run_commands_chain_async(chains, block=block, state_json_dir=state_json_dir,
-                                                    delegate_extra_params=master_agent_count)
+                                                    delegate_extra_params=nodes_count_by_type(config))
     return result
 
 
@@ -151,35 +163,37 @@ def _remove_host(state_file, host):
 
 
 @asyncio.coroutine
-def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_delegate=None, try_remove_stale_dcos=False,
-                 roles=None, **kwargs):
-    if roles is None:
-        roles = ['master', 'agent']
+def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_delegate=None, try_remove_stale_dcos=False,
+                 **kwargs):
+    if hosts is None:
+        hosts = []
 
     # Role specific parameters
     role_params = {
         'master': {
             'tags': {'role': 'master', 'dcos_install_param': 'master'},
-            'hosts': hosts if hosts else config['master_list']
+            'hosts': config['master_list']
         },
         'agent': {
             'tags': {'role': 'agent', 'dcos_install_param': 'slave'},
-            'hosts': hosts if hosts else config['agent_list']
+            'hosts': config.get('agent_list', [])
+        },
+        'public_agent': {
+            'tags': {'role': 'public_agent', 'dcos_install_param': 'slave_public'},
+            'hosts': config.get('public_agent_list', [])
         }
     }
 
     bootstrap_tarball = _get_bootstrap_tarball()
-
     log.debug("Local bootstrap found: %s", bootstrap_tarball)
 
     targets = []
-    for role in roles:
-        default_params = role_params[role]
-        for host in default_params['hosts']:
-            node = Node(host, default_params['tags'])
-            targets += [node]
+    if hosts:
+        targets = hosts
+    else:
+        for role, params in role_params.items():
+            targets += [Node(node, params['tags']) for node in params['hosts']]
 
-    log.debug('Got {} hosts'.format(targets))
     runner = get_async_runner(config, targets, async_delegate=async_delegate)
     chains = []
     if try_remove_stale_dcos:
@@ -208,10 +222,7 @@ def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_deleg
     )
 
     # UI expects total_masters, total_agents to be top level keys in deploy.json
-    delegate_extra_params = {
-        'total_masters': len(config['master_list']),
-        'total_agents': len(config['agent_list'])
-    }
+    delegate_extra_params = nodes_count_by_type(config)
     if kwargs.get('retry') and state_json_dir:
         state_file_path = os.path.join(state_json_dir, 'deploy.json')
         log.debug('retry executed for a state file deploy.json')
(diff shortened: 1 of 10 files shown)


darkonie:
@cmaloney this is ready for code review

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -16,6 +16,26 @@ class ExecuteException(Exception):
     """Raised when execution fails"""
 
 
+def add_nodes(nodes, tag):
```
cmaloney: This is only used / called once. Why not make it a local or inline function?

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -65,7 +78,7 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     chains.append(cleanup_chain)
     master_agent_count = {
         'total_masters': len(config['master_list']),
-        'total_agents': len(config['agent_list'])
+        'total_agents': len(config['agent_list']) + len(config['public_agent_list'])
```
cmaloney: This code comes up twice. Why not have a single function which you call to go from config -> nodes_count_by_type?

This could also potentially be more cleanly written using a python groupby (https://docs.python.org/3/library/itertools.html#itertools.groupby). But that is probably out of scope for this PR.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -151,10 +164,10 @@ def _remove_host(state_file, host):
...
 
 @asyncio.coroutine
-def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_delegate=None, try_remove_stale_dcos=False,
-                 roles=None, **kwargs):
-    if roles is None:
-        roles = ['master', 'agent']
+def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_delegate=None, try_remove_stale_dcos=False,
+                 **kwargs):
+    if hosts is None:
+        hosts = []
```
cmaloney: Why did the default for hosts change to None? Shouldn't there just be an assertion that hosts is an instance of a list?

packages/dcos-installer-ui/buildinfo.json
```diff
@@ -3,7 +3,7 @@
   "single_source": {
     "kind": "git",
     "git": "https://github.com/dcos/dcos-installer-ui.git",
-    "ref": "ba16865174fdf51a3f487112f5c390d500c155c1",
+    "ref": "b2b5ace30f33dbe3d983e6db91c10ecb0974b334",
```
cmaloney: This should be updated / changed independently / not in this PR

test_util/run-all
```diff
@@ -113,7 +114,7 @@ RET=$?
 mkdir -p $WORKDIR/test/vagrant-logs
 if [ $RET -ne 0 ]; then
   echo "TEST FAILED"
-  for i in boot m1 m2 m3 a1 a2; do
+  for i in boot m1 m2 m3 a1 a2 p2; do
```
cmaloney: why not p1?

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -164,22 +177,21 @@ def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_deleg
         },
         'agent': {
             'tags': {'role': 'agent', 'dcos_install_param': 'slave'},
-            'hosts': hosts if hosts else config['agent_list']
+            'hosts': hosts if hosts else config.get('agent_list', [])
+        },
+        'public_agent': {
+            'tags': {'role': 'public_agent', 'dcos_install_param': 'slave_public'},
+            'hosts': hosts if hosts else config.get('public_agent_list', [])
```
cmaloney: This seems to say that if "hosts" is supplied we should always make them public_agents as well as private_agents

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: done

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: done

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: basically it is a bad practice to have mutable default parameter, this is a workaround to make sure if `hosts` wasn't passed, we should always default to a new list.

test_util/run-all
darkonie: done

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: done

darkonie:
@cmaloney this is ready for review. See a second commit. Once it is ready to ship i'll squash all commits.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -151,35 +163,37 @@ def _remove_host(state_file, host):
 
 
 @asyncio.coroutine
-def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_delegate=None, try_remove_stale_dcos=False,
-                 roles=None, **kwargs):
-    if roles is None:
-        roles = ['master', 'agent']
+def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_delegate=None, try_remove_stale_dcos=False,
+                 **kwargs):
+    if hosts is None:
```
cmaloney: Can we add an assert that everything in the hosts list is an instance of Nodes if hosts is given? (This code will make really odd error messages if that isn't the case)

ssh/validate.py
```diff
@@ -27,32 +27,35 @@ def validate_ssh_port(ssh_port):
...
+
+    if public_agent_list:
+        compare_lists(master_list, public_agent_list)
 
 
 def check_duplicates(arg_list):
     assert isinstance(arg_list, list), 'only lists can be verified for duplicates'
     dups = list(filter(lambda x: arg_list.count(x) > 1, arg_list))
-    assert len(dups) == 0, 'List cannot contain duplicates: {}'.format(', '.join(set(dups)))
+    assert not dups, 'List cannot contain duplicates: {}'.format(', '.join(set(dups)))
```
cmaloney: nit: This (and a number of these condition checks) is an unrelated change

ssh/validate.py
```diff
@@ -63,30 +66,55 @@ def check_ipv4_addrs(ips):
...
-        'master_list': lambda master_list: validate_master_agent_lists(master_list, config.get('agent_list'))
-    }
+    if public_agent_list:
+        validate_hosts_list(public_agent_list, 'public_agent_list')
+
+    if agent_list and public_agent_list:
+        compare_lists(agent_list, public_agent_list)
 
+
+def run_validate_config(value_func_map, config, required=True):
```
cmaloney: `dev validate_config_chunk(key_validate_fn_map, config, keys_required)`
 - Make the function name closer to what the function is doing
 - Always require setting required. We don't call this from many places, defaults should be avoided

ssh/validate.py
```diff
@@ -63,30 +66,55 @@ def check_ipv4_addrs(ips):
...
     errors = {}
-    for ssh_key, validate_func in ssh_keys_checks_map.items():
+    for ssh_key, validate_func in value_func_map.items():
         input_value = config.get(ssh_key)
```
cmaloney:
```python
if ssh_key not in config:
  if required:
    errors[ssh_key] = ...
try:
  validate_func(config[ssh_key])
except:
...
```

ssh/validate.py
```diff
@@ -63,30 +66,55 @@ def check_ipv4_addrs(ips):
...
+def validate_config(config):
+    assert isinstance(config, dict)
+
+    ssh_keys_checks_map_required = {
+        'ssh_user': validate_ssh_user,
+        'ssh_port': validate_ssh_port,
+        'ssh_key_path': validate_ssh_key_path,
+        'master_list': lambda master_list: validate_master_agent_lists(
+            master_list,
+            config.get('agent_list'),
```
cmaloney:
```
config.get('agent_list', [])
config.get('public_agent_list', [])
```

ssh/validate.py
```diff
@@ -63,30 +66,55 @@ def check_ipv4_addrs(ips):
...
-
         try:
             validate_func(input_value)
         except AssertionError as err:
             errors[ssh_key] = str(err)
     return errors
+
+
+def validate_config(config):
+    assert isinstance(config, dict)
```
cmaloney: Might make sense when loading config to do a 
```
config.setdefault('agent_list', [])
config.setdefault('public_agent_list', [])
```

That way all the code everywhere can just assume master_list, public_agent_list, and agent_ilst always exist.

test_util/integration_test.py
```diff
@@ -1004,7 +1008,7 @@ def test_3dt_nodes(cluster):
         assert len(response) == 1, 'nodes response must have only one field: nodes'
         assert 'nodes' in response
         assert isinstance(response['nodes'], list)
-        assert len(response['nodes']) == len(cluster.masters + cluster.slaves), (
+        assert len(response['nodes']) == len(cluster.masters + cluster.slaves) + len(cluster.public_slaves), (
```
cmaloney: len(cluster.all_slaves) + len(cluster.masters)?

test_util/run-all
```diff
@@ -113,7 +114,7 @@ RET=$?
 mkdir -p $WORKDIR/test/vagrant-logs
 if [ $RET -ne 0 ]; then
   echo "TEST FAILED"
-  for i in boot m1 m2 m3 a1 a2; do
+  for i in boot m1 m2 m3 a1 a2 p1; do
```
cmaloney: note: Will need to update the CI log collector as well that outputs to the actual TeamCity log in chunks

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: done

ssh/validate.py
darkonie: done

test_util/run-all
mellenburg: @cmaloney I just went ahead and checked in the CI log collection: https://github.com/dcos/dcos/pull/77

spahl:
This is good now.

:ship:

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -151,35 +163,38 @@ def _remove_host(state_file, host):
 
 
 @asyncio.coroutine
-def install_dcos(config, block=False, state_json_dir=None, hosts=[], async_delegate=None, try_remove_stale_dcos=False,
-                 roles=None, **kwargs):
-    if roles is None:
-        roles = ['master', 'agent']
+def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_delegate=None, try_remove_stale_dcos=False,
```
cmaloney: When this is called via a retry in async_server.py, hosts is passed through as a raw list of hosts, not a list of Node objects, which means later the `targets=` sometimes results in targets being set to a list of Node instances, and sometimes a list of whatever was in the json state.

cmaloney:
@spahl @darkonie NAK for landing now, retrying is broken.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
spahl: Nice catch. I'm gonna back it out of the current train.

darkonie:
@spahl @cmaloney eventually it will be converted into Node https://github.com/dcos/dcos/blob/master/ssh/ssh_runner.py#L78

this may be very implicit and not documented very well, but the idea was to be able to pass a list of ip addresses or a list of Node objects. So this will work correctly. Let me know if you want to me to fix it or leave it as is.

cmaloney:
That is a very complicated / confusing code path... Please simplify. We want code to be followable / easily understood when reading it.

Also note that final / late `Node()` conversion doesn't attach the `master` or `slave` tag to the hosts, which I'm suspect would mean https://github.com/dcos/dcos/pull/17/files#diff-6c2af57e1ff0e55764fc9f3f45ab4ce5R222 would go wrong / break.

darkonie:
@cmaloney @spahl done, ready for review

cmaloney:
This is looking good to me. :ship: when you're happy with it.

spahl:
@darkonie you can rebase and merge

PR #18, opened 2016-04-25
Set tcp_be_liberal as part of Minuteman initialization

sargun:


sargun:
@spahl @cmaloney

cmaloney:
We don't own these settings globally / don't own sysctl settings. For AWS / CloudFormation clusters we can / should set these (in a way that is persistent across reboots and the like). The service cannot set them however.

sargun:
Why can't we take them over? I don't think that people will actually set this setting based on the documentation, and they'll see issues when they're hairpinning traffic between Docker containers. Given that DC/OS runs on edge systems, and you're typically not using conntrack for any kind of security / hole-punch limiting, it seems like a fairly reasonable, conservative option. 

The kernel gets confused because it has an earlier conntrack entry for this traffic since it creates one before it gets from the Docker container to the forwarding table. 

If you think that we should avoid touching this, can you create a PR to set this in the AWS / Cloudformation images, and I'll do the PR so it gets into ACS?

lingmann:
It sounds like this should be replaced with:
1. A pre-flight check to validate that sysctl is configured appropriately for minuteman
2. An update to the Azure templates to configure sysctl
3. An update to the AWS templates to configure sysctl

lingmann:
Per some conversations with @spahl and @sargun I think we are going to claim ownership of certain default sysctl settings. People will be able to override, but at their own risk. Do we have enough consensus here to move forward with the patch as is @cmaloney?

spahl:
+1

bajohns:
Thanks for this patch - this solved a number of vexing connection issues from the MySQL driver.

damhau:
This also solved my problem with Kibana, it was unable to connect to Elasticsearch.
I mention it so that google can pick this answer if somebody else has the same problem.

PR #21, opened 2016-04-26
dcos-installer: fix postflight bug

darkonie:
postflight may finish successfully when the cluster is not fully up and
running because we consider systemd unit states: active, activating healthy.
The cluster is really healthy when endpoint /acs/api/v1/logout in a leader host
returns a sane status code.

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -253,15 +253,24 @@ def run_postflight(config, dcos_diag=None, block=False, state_json_dir=None, asy
 # Run the DC/OS diagnostic script for up to 15 minutes (900 seconds) to ensure
 # we do not return ERROR on a cluster that hasn't fully achieved quorum.
 T=900
-until OUT=$(/opt/mesosphere/bin/./3dt -diag) || [[ T -eq 0 ]]; do
-    sleep 1
-    let T=T-1
+err_msg="Waited $T seconds and DC/OS did not come up"
+test_endpoint="http://leader.mesos/acs/api/v1/logout"
+while [ $T -ne 0 ]; do
+  OUTPUT=$(/opt/mesosphere/bin/3dt -diag 2>/dev/null)
+  if [ $? -eq 0 ]; then
+    STATUS=$(curl -s -o /dev/null -w "%{http_code}" $test_endpoint)
+    if [ $STATUS -eq 401 -o $STATUS -eq 200 -o $STATUS -eq 201 ]; then
+      # if 3dt -diag returns success and  http://leader.mesos/acs/api/v1/logout
+      echo "DC/OS is up and running, $test_endpoint status code $STATUS"
+      echo "3dt output: $OUTPUT"
+      exit 0
+    fi
+  fi
+  let T=T-1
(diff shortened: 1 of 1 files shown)


ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -253,15 +253,24 @@ def run_postflight(config, dcos_diag=None, block=False, state_json_dir=None, asy
...
+  fi
+  let T=T-1
+  sleep 1
 done
-RETCODE=$?
-for value in $OUT; do
-    echo $value
-done
-exit $RETCODE"""
+echo "$err_msg, endpoint $test_endpoint returned status code: $STATUS" >&2
```
malnick: Thanks for moving this outside the loop

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -253,15 +253,24 @@ def run_postflight(config, dcos_diag=None, block=False, state_json_dir=None, asy
...
+  if [ $? -eq 0 ]; then
+    STATUS=$(curl -s -o /dev/null -w "%{http_code}" $test_endpoint)
+    if [ $STATUS -eq 401 -o $STATUS -eq 200 -o $STATUS -eq 201 ]; then
+      # if 3dt -diag returns success and  http://leader.mesos/acs/api/v1/logout
+      echo "DC/OS is up and running, $test_endpoint status code $STATUS"
+      echo "3dt output: $OUTPUT"
+      exit 0
+    fi
+  fi
+  let T=T-1
```
kartick: ((T--)) should do it.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -253,15 +253,24 @@ def run_postflight(config, dcos_diag=None, block=False, state_json_dir=None, asy
...
+  sleep 1
```
kartick: Instead of 900 sleeps, why don't you do an exponential back-off?

darkonie: i think exponential back-off is not a good use case here. I think we want to notify user as soon as cluster is up and running. @malnick might be a right person to ask since he wrote up the initial code for postflight

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: `((T--))` and `let` behave exactly in the same way

cmaloney:
That endpoint isn't a strong check the cluster is actually up.

https://github.com/dcos/dcos/blob/master/test_util/integration_test.py#L186 is the most comprehensive "wait for the cluster to be up" that we have.

We ideally want to share that code / check logic between all of our launches so the AWS, Azure, GUI, CLI, and Advanced all give the same quality of check / use the same code rather than building more one-off code.

spahl:
We should talk about a better solution for this one. I'm still not keen on having every node do that curl. I'm going to move this aside unless we don't have something better in mind.

malnick:
@spahl Sounds good to me. I think the better implementation would be to have the installer execute the request to the acs endpoint once.

PR #22, opened 2016-04-26
installer:fail preflight check if SELinux disabled

darkonie:
DCOS-6230

### DIFF
--- a/gen/installer/bash.py
+++ b/gen/installer/bash.py
@@ -172,6 +172,7 @@ function check_selinux() {
   fi
 
   print_status $RC "Is SELinux disabled?"
+  (( OVERALL_RC += $RC ))
   return $RC
 }
 


malnick:
👍

cmaloney:
Shouldn't this be: fail preflight check if SE Linux is *enabled*?

I'm also not the greatest at bash, but the rest of these "check" functions all do fine with the `return $RC`. Why is this one so different? Why doesn't it work like the other `check` lines: https://github.com/darkonie/dcos/blob/mnaboka/fix-dcos-install/gen/installer/bash.py#L320?

cmaloney:
@kartick do you have any input here? I know you originally wrote the check code here.

gen/installer/bash.py
```diff
@@ -172,6 +172,9 @@
   fi
 
   print_status $RC "Is SELinux disabled?"
+  if [[ $RC -ne 0 ]]; then
+    exit $RC
+  fi
```
kartick: The rest of this script uses `OVERALL_RC` and checks cummulative return code in the caller. Exiting here will cause the preflight to stop when it hits this check instead of producing a summary (for subsequent checks)

darkonie:
@kartick fixed, please review

kartick:
Looks good @darkonie. Please test it out, and merge.

darkonie:
PR tested, works as expected, thanks @kartick

kartick:
Ready to merge, @cmaloney

cmaloney:
:ship:

cmaloney:
Merged in #28

PR #24, opened 2016-04-27
Remove `--skip-build` and simplify `do_build_packages` to help prepare for "upstreams"

cmaloney:
This removes the `--skip-build` option and gets `release create` to use the result of `pkgpanda.build.build_tree` directly rather than calling a helper which scrapes up artifacts.

General shape of the upstream work for pkgpanda:

1. Return structured information about the packages built and place that in the metadata.json
  1. this PR
  1. Returning more dictionaries of exactly what was built from build_tree (all the packages, what variants, what those package ids were, what the bootstrap variants are, what those bootstrap ids are). Broken into a separate step so that we have testing over the return structure change.
2. Add the notion of a "backing" / "upstream" repository. `pkgpanda` can refer to pre-built packages in this "upstream" repository just like it can local filesystem packages. It does not know how to build these packages (does not have access to the `build` and `buildinfo.json` files.

### DIFF
--- a/release/__init__.py
+++ b/release/__init__.py
@@ -440,9 +440,14 @@ def do_build_packages(cache_repository_url):
     def get_build():
         # TODO(cmaloney): Stop shelling out
         package_dir = os.getcwd() + '/packages'
-        pkgpanda.build.build_tree(package_dir, True, cache_repository_url, None)
+        result = pkgpanda.build.build_tree(package_dir, True, cache_repository_url, None)
+        last_set = pkgpanda.build.get_last_bootstrap_set(package_dir)
+        assert last_set == result, \
+            "Internal error: get_last_bootstrap_set doesn't match the results of build_tree: {} != {}".format(
+                last_set,
+                result)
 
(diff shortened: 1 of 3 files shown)


release/__init__.py
```diff
@@ -440,9 +440,14 @@ def do_build_packages(cache_repository_url):
     def get_build():
         # TODO(cmaloney): Stop shelling out
         package_dir = os.getcwd() + '/packages'
-        pkgpanda.build.build_tree(package_dir, True, cache_repository_url, None)
-
-        return pkgpanda.build.get_last_bootstrap_set(package_dir)
+        result = pkgpanda.build.build_tree(package_dir, True, cache_repository_url, None)
+        last_set = pkgpanda.build.get_last_bootstrap_set(package_dir)
+        assert last_set == result, \
+            "Internal error: get_last_bootstrap_set dooesn't match the results of build_tree: {} != {}".format(
```
cmaloney: doesn't

PR #27, opened 2016-04-27
Switch ELB port 80 from HTTP to TCP mode.

brndnmtthws:
cc @spahl

### DIFF
--- a/gen/aws/templates/advanced/advanced-master.json
+++ b/gen/aws/templates/advanced/advanced-master.json
@@ -240,9 +240,9 @@
           },
           {
             "LoadBalancerPort" : "80",
-            "Protocol" : "HTTP",
+            "Protocol" : "TCP",
             "InstancePort" : "80",
-            "InstanceProtocol" : "HTTP"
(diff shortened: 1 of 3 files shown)


cmaloney:
What about Azure?

brndnmtthws:
Looks like Azure's already configured as TCP:

https://github.com/dcos/dcos/blob/master/gen/azure/templates/azuredeploy.json#L773-L790
https://github.com/dcos/dcos/blob/master/gen/azure/templates/acs.json#L981-L1000

lingmann:
LGTM. I was initially concerned about the changes to InternalMasterLoadBalancer, since it may introduce subtle issues for internal services which poll adminrouter... But since the health checks are still explicitly HTTP, polling services won't see a difference.

@brndnmtthws how confident are you in this from a testing perspective?

brndnmtthws:
@lingmann I've tested this, and we're using TCP on Azure already anyway, so it will be transparent to the user.

brndnmtthws:
Also rebased on master.

cmaloney:
:ship:

PR #15, opened 2016-04-21
DCOS-4

malnick:
https://dcosjira.atlassian.net/browse/DCOS-4

This PR fixes the `__init__.py` for the installer to ensure we do not execute make_default_dir() when importing the installer library. This fixes the issue of tox requiring root access to run as well as the requirement to have /genconf available at test time. 

We refactored the installer in a way which required /genconf to exist if you import installer lib. Because of that directory requirement, tox needed to be root to run, and it should not. This fixes the code to ensure you don’t need /genconf to exist to import the lib fixes that.

PR #2, opened 2016-04-19
Remove SSH Keys from Minuteman

sargun:

PR #28, opened 2016-04-27
Merge Train 0

cmaloney:
Contains #2, #15, #22


cmaloney:
Dropped #2 since it broke the build

PR #29, opened 2016-04-28
Better error for conflicting files in packages

cmaloney:
Needs a test


cmaloney:
:ship:, not going to get to a test soon, and this has been hand tested

PR #34, opened 2016-04-29
ssh: add stage field to process output result

darkonie:
DCOS-6431
The result will include a comment from a command chain. This should give
a human readable description of a command being executed on a host.

### DIFF
--- a/ssh/ssh_runner.py
+++ b/ssh/ssh_runner.py
@@ -103,7 +103,7 @@ class MultiRunner():
         return shared_opts
 
     @asyncio.coroutine
-    def run_cmd_return_dict_async(self, cmd, host, namespace, future):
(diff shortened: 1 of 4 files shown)


ssh/ssh_runner.py
```diff
@@ -103,7 +103,7 @@ def _get_base_args(self, bin_name, host):
         return shared_opts
 
     @asyncio.coroutine
-    def run_cmd_return_dict_async(self, cmd, host, namespace, future):
+    def run_cmd_return_dict_async(self, cmd, host, namespace, future, comment):
```
cmaloney: Why does the parameter `comment` become the field `stage`? Please make them consistent, probably using `stage` everywhere if that's what the API in the json should be.

cmaloney:
:ship:

PR #37, opened 2016-05-03
Adding analytics tracking to installer

malnick:
@cmaloney @spahl @darkonie

Add analytics tracking to installer CLI (GUI will be next). 

Refactored the CLI a bit to use a dispatcher for the action arguments. This way we can call the tracker once in a clean method. I also moved the dispatcher to a new file cli.py as the groundwork to refactor the __init__.py (future work).

### DIFF
--- a/ext/dcos-installer/dcos_installer/cli.py
+++ b/ext/dcos-installer/dcos_installer/cli.py
@@ -1,0 +1,105 @@
+import asyncio
+import logging
+import sys
+
+from dcos_installer import action_lib, backend
+from dcos_installer.util import CONFIG_PATH
+from dcos_installer.installer_analytics import InstallerAnalytics
+from dcos_installer.action_lib.prettyprint import print_header, PrettyPrint
+from ssh.utils import AbstractSSHLibDelegate
+
+
+log = logging.getLogger(__name__)
+
+
+class CliDelegate(AbstractSSHLibDelegate):
+    def on_update(self, future, callback_called):
+        chain_name, result_object, host = future.result()
+        callback_called.set_result(True)
+        log.debug('on_update executed for {}'.format(chain_name))
+
+    def on_done(self, name, result, host_status=None):
+        print_header('STAGE {}'.format(name))
+
+
+def run_loop(action, options):
+    assert callable(action)
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
+
+    print_header('START {}'.format(action.__name__))
+    try:
+        config = backend.get_config()
+        cli_delegate = CliDelegate()
+        result = loop.run_until_complete(action(config, block=True, async_delegate=cli_delegate, options=options))
+        pp = PrettyPrint(result)
+        pp.stage_name = action.__name__
+        pp.beautify('print_data')
+
+    finally:
+        loop.close()
+    exitcode = 0
+    for host_result in result:
+        for command_result in host_result:
+            for host, process_result in command_result.items():
+                if process_result['returncode'] != 0:
+                    exitcode += 1
+    print_header('ACTION {} COMPLETE'.format(action.__name__))
+    pp.print_summary()
+    return exitcode
+
+
+def tall_enough_to_ride():
+    choices_true = ['yes', 'y']
+    choices_false = ['no', 'n']
+    while True:
+        do_uninstall = input('This will uninstall DC/OS on your cluster. You may need to manually remove '
+                             '/var/lib/zookeeper in some cases after this completes, please see our documentation '
+                             'for details. Are you ABSOLUTELY sure you want to proceed? [ (y)es/(n)o ]: ')
+        if do_uninstall.lower() in choices_true:
+            return
+        elif do_uninstall.lower() in choices_false:
+            sys.exit(1)
+        else:
+            log.error('Choices are [y]es or [n]o. "{}" is not a choice'.format(do_uninstall))
+
+
+def get_new_analytics(config_path=CONFIG_PATH):
+    return InstallerAnalytics(backend.get_config(config_path))
+
+
+def dispatch_action(args):
+    installer_analytics = get_new_analytics()
+
+    def preflight(args):
+        print_header("EXECUTING PREFLIGHT")
+        return action_lib.run_preflight
+
+    def postflight(args):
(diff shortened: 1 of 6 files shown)


malnick:
@cmaloney Could you review, this is no longer WIP.

malnick:
@cmaloney Could you review when you get a minute?

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                             'for details. Are you ABSOLUTELY sure you want to proceed? [ (y)es/(n)o ]: ')
+        if do_uninstall.lower() in choices_true:
+            return
+        elif do_uninstall.lower() in choices_false:
+            sys.exit(1)
+        else:
+            log.error('Choices are [y]es or [n]o. "{}" is not a choice'.format(do_uninstall))
+
+
+def get_new_analytics(config_path=CONFIG_PATH):
```
cmaloney: why is this a helper function? It's only called once, not a lot of code, and never given a different parameter. Having a helper function vs. just doing the little bit of code inline isn't providing a lot of value

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -0,0 +1,34 @@
+import analytics
+import os
+import uuid
+
+from dcos_installer import backend
+
+
+class InstallerAnalytics():
+    def __init__(self, gen_config):
```
cmaloney: Why take gen_config as a parameter? It doesn't appear to be used...

cmaloney:
Some overarching things that need to be looked at / worked on:

1. This seems to conflate two things in one patch. Ideally these should be two different patches. The first is much simpler to review separately, just code motion is much easier to verify / review than code motion + some sprinkled in changes to make things work. The two parts:
  a. Move a bunch of code into a cli.py / refactor, adding a new dispatch_action() function
  b. Add a new InstallerAnalytics class, use it to add analytics to GUI and CLI installer
2. There needs to be a way to opt out of installer analytics / tracking (Likely a command line flag)
3. Functions like send() in InstallerAnalytics shouldn't get defaults for all their parameters. There is value in making the programmer who calls the API think about what they're doing. Also, we only call it from one place passing all the parameters, so all the defaults are effectively extra dead code.
4. I don't like the new "dispatch_action" refactoring. Especially for the getattr(args, action) thing. It takes a lot of mental work to comprehend / understand what all is going on in it, and the old code was much simpler to understand and follow. I can provide some sample other ways you could factor it if you really want to avoid writing out the commonality of having to do the installer analytics send more.
5. If analytics-python is now a dependency of dcos-installer, than it also needs to get added as a dependency to the dcos-installer package in pkgpanda: https://github.com/dcos/dcos/blob/master/packages/dcos-installer/buildinfo.json. Otherwise this code when you actually use the installer won't work (dcos_generate_config.sh won't contain that package).

malnick:
On 1: I could break this out, but it makes sense to me not to have a PR which adds dead code (just the Analytics class). It makes more sense to me to add live code whose baseline callers are fully implemented. This ensures our integration tests pass still and that we're not accidentally merging code that has bugs. 

The complete implementation of this also includes the UI sending metrics, which I saved for a future PR to keep this one as minimal as possible without having dead code. 

On 4: Yes, that would be great if you had an idea around not re-writing the analytics call for every if statement. I like the dispatcher as it allows us to have a more generic way to call these methods which can be wrapped more easily.

cmaloney:
1: I'm not saying add dead code. I'm saying separate concerns out. The first PR is purely moving code around, not adding any dead code. The second is adding InstallerAnalytics and using it. At no point is any of that "dead" or unused code which is committed.

When you mix "moving" code with "Changing" code I have to read the code, and in my mind reconstruct what the exact moving happens to look like, then figure out the delta between that and what is written on the page. It is a lot easier to review code which purely does moving code to other files / changing function names / etc. I just have to check "Every line is the same, just in a different place".

malnick:
I understand it's not optimal. What would you like as an alternative? I have the dispatcher refactor in there to ensure the code isn't repetitive (calling analytics). I could place individual calls in the if statements, but that's not clean.

malnick:
On number 5, I'm not seeing any of our other installer deps in that file, did you mean the top level setup.py for the installer?

ext/dcos-installer/dcos_installer/cli.py
malnick: We'll be adding several calls to this function when we add the analytics tracking to the UI.

malnick:
@spahl This is ready to go in.

mellenburg:
Nak: if you squash commits, can you please make sure the remaining commit title is representative of the change? In this case, your analytics addition is masquerading as my already merged change for updating python-requests

malnick:
The complete commit log shows both the original commit message and your commit message.

spahl:
Ok this has been here long enough. Lets not split this out even more just for the sake of it now. Next time at least split into multiple commits. Having to squashs does not mean it has to be a single one.

The getattr might not be ideal but it's not unreadable either. Lets merge this and if someone wants to get rid of it later fine.

I'm a :ship:

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -0,0 +1,34 @@
+import analytics
+import os
+import uuid
+
+
+class InstallerAnalytics():
+    def __init__(self, config):
+        self.uuid = self.new_uuid()
+        self.bootstrap_version = os.getenv("BOOTSTRAP_ID", "")
```
cmaloney: This should have been a hard fail if it's not set. This will swallow when we make programming mistakes

ext/dcos-installer/dcos_installer/cli.py
cmaloney: There isn't value in having a never changed default parameter to a constant. Should just fold the constant into the use place.

ext/dcos-installer/dcos_installer/installer_analytics.py
spahl: I'm going to change that in the train branch. Makes sense, there is no way this ever needs to runs without one

PR #38, opened 2016-05-03
Teach pkgpanda to refer to a "package store" to get packages rather than filesystem assumptions

cmaloney:
This is part of the groundwork to enable having a "variant" / "spin" of DC/OS which lives in a separate repository.

The end goal is that we can have a upstream source (which is specified the same as a "source" in a buildinfo) which will get cloned / checked out so that all the "build" and "buildinfo.json" files are available. The local "packages" directory will look like it is "overlaid" on top of that repository, allowing treeinfo.json and dependencies to cross the boundary. The local repository will also be able to replace at will packages in the upstream repository.

This consists of three main parts so far
 - Adding the "package store" abstraction
 - Increasing reliance on the abstraction for getting details rather manually traversing the filesystem
 - Teaching `mkpanda tree` to be a little more intelligent with what it is building.

### DIFF
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -72,7 +72,7 @@ def get_variants_from_filesystem(directory, extension):
             # Should be foo. since we've moved the extension.
             if variant[-1] != '.':
                 raise BuildError("Invalid variant filename {}. Expected a '.' separating the "
-                                 "variant name and extension.".format(filename))
+                                 "variant name and extension '{}'.".format(filename, extension))
             variant = variant[:-1]
 
         results.add(variant)
@@ -468,7 +468,7 @@ def get_tree_package_tuples(package_store, tree_variant):
     to_visit = list(package_tuples)
     while len(to_visit) > 0:
         name, variant = to_visit.pop()
-        requires = package_store.packages[(name, variant)]['requires']
+        requires = package_store.get_buildinfo(name, variant)['requires']
         for require in requires:
             require_tuple = expand_require(require)
             if require_tuple not in package_tuples:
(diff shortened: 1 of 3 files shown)


pkgpanda/build/__init__.py
```diff
@@ -51,6 +51,92 @@ def run(self, cmd):
...
+                filename, extension))
+
+        # Empty / default variant is represented as 'None'.
+        if variant == '':
+            variant = None
+        else:
+            # Should be foo. since we've moved the extension.
+            if variant[-1] != '.':
+                raise BuildError("Invalid variant filename {}. Expected a '.' separating the "
+                                 "variant name and extension.".format(filename))
```
cmaloney: nit: include the extracted extension in the error message

pkgpanda/build/__init__.py
```diff
@@ -400,7 +468,7 @@ def include_package(name, variant):
     to_visit = list(package_tuples)
     while len(to_visit) > 0:
         name, variant = to_visit.pop()
-        requires = package_requires[(name, variant)]['requires']
+        requires = package_store.packages[(name, variant)]['requires']
```
cmaloney: Use get_buildinfo here rather than the tuple indexing.

pkgpanda/build/__init__.py
```diff
@@ -456,61 +518,66 @@ def visit(pkg_tuple):
...
 
+    def visit_packages_in_tree(variant):
+        all_tuples = get_tree_package_tuples(package_store, variant)
+        for pkg_tuple in sorted(all_tuples, key=key_func):
+            if pkg_tuple in visited:
+                continue
+            visit(pkg_tuple)
+
     # Build everything if no variant is given
     if tree_variant is None:
```
cmaloney: instead of doing this `if tree_variant is None` twice should switch to just making a ordered list to visit

pkgpanda/build/__init__.py
```diff
@@ -456,61 +518,66 @@ def visit(pkg_tuple):
...
+        built_packages[name][variant] = build(
+            package_store,
+            name,
+            variant,
+            repository_url,
+            True)
 
     def make_bootstrap(variant):
-        print("Making bootstrap variant:", variant or "<default>")
+        print("Making bootstrap variant:q", variant or "<default>")
```
cmaloney: unnecessary q insertion

pkgpanda/build/__init__.py
```diff
@@ -599,7 +647,9 @@ def pkg_abs(name):
     # Build up the docker command arguments over time, translating fields as needed.
     cmd = DockerCmd()
 
-    buildinfo = load_buildinfo(package_dir, variant)
+    assert (name, variant) in package_store.packages, \
+        "Programming error: name, variant should have been validated to be valid before calling build()."
+    buildinfo = copy.deepcopy(package_store.packages[(name, variant)])
```
cmaloney: use get_buildinfo

PR #43, opened 2016-05-04
Cleanup Installer __init__.py [depends on #37]

malnick:
Groundwork for cleaning up the __init__.py for the installer:

- Remove DcosInstaller class
- Clean out the __init__.py file to not have any installer logic
- Create a entrypoint function

**Notice** this PR is rebased on https://github.com/dcos/dcos/pull/37 and requires it as a dependency.


malnick:
@darkonie @cmaloney Please take a look if you get a minute. Keep in mind all changes associated with analytics are rebases from #37

bhack:
Do you plan to refactor https://github.com/dcos/dcos/blob/master/ext/dcos-installer/dcos_installer/action_lib/__init__.py then?

malnick:
@bhack Definitely, but not in this PR.

bhack:
OK glad to know that it is on the roadmap

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -0,0 +1,164 @@
...
+def start_installer(args=None):
+    """
+    The web based installer leverages Flask to present end-users of
+    dcos_installer with a clean web interface to configure their
+    site-based installation of DC/OS.
+    """
+    # If no args are passed to the class, then we're calling this
+    # class from another library or code so we shouldn't execute
+    # parser or anything else
+    if args:
```
malnick: This can be removed now that the class construct is gone

malnick:
@spahl Can you update the label to "In Review" please?

spahl:
Can you please split this back into the original 3 commits? You outlined them in the description.

I can't see any of the code changes due to the move. See comments from the other PR:-)

malnick:
Let's merge in 37 and see what this looks like after master has that; I think this PR will be more clear once those changes are in since it's rebased on it.

spahl:
rebase please

malnick:
@spahl This is rebased now.

ext/dcos-installer/dcos_installer/cli_dispatcher.py
```diff
@@ -0,0 +1,168 @@
...
+            print_validation_errors(validation_errors)
+            sys.exit(1)
+        sys.exit(0)
+
+    def genconf(args):
+        print_header("EXECUTING CONFIGURATION GENERATION")
+        code = backend.do_configure()
+        if code != 0:
+            sys.exit(1)
+        sys.exit(0)
```
malnick: Can drop this `sys.exit(0)` to the bottom of the parent def(), and only call once.

ext/dcos-installer/dcos_installer/cli_dispatcher.py
```diff
@@ -0,0 +1,168 @@
...
+
+    for action in ['web', 'hash_password', 'validate_config', 'genconf']:
+        if getattr(args, action):
+            locals()[action](args)
```
malnick: Should return the exit code here and handle sys.exit() in once place like the action dispatcher does.

ext/dcos-installer/dcos_installer/__init__.py
```diff
@@ -1,235 +1 @@
...
-                    'bold': True,
-                },
-            },
-            fmt='%(message)s',
-            isatty=True
-        )
-
-        log.debug("Logger set to DEBUG")
-        return options
+__version__ = "1.7"
```
malnick: We should figure out a better way to strictly pin this to the overall version of DCOS Image rather than relying on hard coding it here.

malnick:
@spahl This is ready.

ext/dcos-installer/dcos_installer/__init__.py
cmaloney: Please just don't put it here at all. It isn't adding significant value, and the value in it is already incorrect / doesn't match the string we use in the UI

pytest/test_installer_backend.py
```diff
@@ -74,13 +74,11 @@ def test_do_validate_config(tmpdir):
 
     expected_output = {
         'ssh_user': 'required parameter ssh_user was not provided',
-        'ssh_key_path': 'could not find ssh private key: /genconf/ssh_key',
```
cmaloney: how is this related / why is this here?

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
+            "source": self.source,
             "install_method": install_method,
             "stage": action,
             "errors": num_errors,
-            "customer_key": self.customer_key,
         })
         analytics.flush()
 
     def new_uuid(self):
-        return str(uuid.uuid4())
```
cmaloney: str definitely seems much more likely to be the format we want it in. (then things show up as a uuid, instead of the bits that make up the uuid)

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
+from dcos_installer import backend
+
 
 class InstallerAnalytics():
-    def __init__(self, config):
+    def __init__(self):
         self.uuid = self.new_uuid()
-        self.bootstrap_version = os.getenv("BOOTSTRAP_ID")
-        self.customer_key = config["customer_key"] if 'customer_key' in config else ""
+        self.bootstrap_version = os.getenv("BOOTSTRAP_ID", "")
```
cmaloney: this should hard fail if it isn't set / that is our internal code being inconsistent if it isn't set. A set bootstrap_id is mandatory

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
         variant: string | open or enterprise
         action: string | preflight, deploy, or postflight
         install_method: string | gui, cli or advanced
         """
         analytics.write_key = "39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP"
 
+        # We set customer key from config to avoid loading the config during class init
+        self.customer_key = backend.get_config()["customer_key"] if 'customer_key' in backend.get_config() else ""
+
         analytics.track(user_id=self.customer_key, anonymous_id=self.uuid, event=action, properties={
```
pyronicide: Please leave user_id out if `customer_key` is empty. Segment needs it that way internally.

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
             "install_id": self.uuid,
             "bootstrap_id": self.bootstrap_version,
-            "provider": self.source,
-            "source": "installer",
+            "source": self.source,
             "install_method": install_method,
             "stage": action,
             "errors": num_errors,
-            "customer_key": self.customer_key,
```
pyronicide: This is important and needs to be in the body as well.

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
+
 
 class InstallerAnalytics():
-    def __init__(self, config):
+    def __init__(self):
         self.uuid = self.new_uuid()
-        self.bootstrap_version = os.getenv("BOOTSTRAP_ID")
-        self.customer_key = config["customer_key"] if 'customer_key' in config else ""
+        self.bootstrap_version = os.getenv("BOOTSTRAP_ID", "")
+        self.customer_key = ""
```
cmaloney: I don't see value in having this member. We always just get the customer_key out of the configuration (https://github.com/dcos/dcos/pull/43/files?diff=split#diff-2067a16f4c753c685193f0c92a648ea8R24), so we never actually use this member's value, all it could be is old / outdated.

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
+    def send(self, action="", install_method="", num_errors=0):
         """Sends analytics track data to segmentIO.
         variant: string | open or enterprise
         action: string | preflight, deploy, or postflight
         install_method: string | gui, cli or advanced
         """
         analytics.write_key = "39uhSEOoRHMw6cMR6st9tYXDbAL3JSaP"
 
+        # We set customer key from config to avoid loading the config during class init
+        self.customer_key = backend.get_config()["customer_key"] if 'customer_key' in backend.get_config() else ""
```
pyronicide: `backend.get_config().get("customer_key", "")`

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
 import os
 import uuid
 
+from dcos_installer import backend
+
 
 class InstallerAnalytics():
-    def __init__(self, config):
+    def __init__(self):
         self.uuid = self.new_uuid()
```
cmaloney: We do this exactly once, and the new_uuid() code is really short / simple `uuid.uuid4()`. Lets just put that inline here. Adding the abstraction of calling a function isn't adding a lot of value.

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
-        self.bootstrap_version = os.getenv("BOOTSTRAP_ID")
-        self.customer_key = config["customer_key"] if 'customer_key' in config else ""
+        self.bootstrap_version = os.getenv("BOOTSTRAP_ID", "")
+        self.customer_key = ""
         self.source = "onprem"  # If you're using the installer you're not in AWS or Azure
 
-    def send(self, action, install_method, num_errors):
+    def send(self, action="", install_method="", num_errors=0):
```
cmaloney: we never don't set all these parameters / use the defaults, so we aren't getting value out of having them. All they're doing is hiding programmer errors where we writing the code forget to set some of the parameters.

ext/dcos-installer/dcos_installer/cli_dispatcher.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def install_prereqs(args):
+        print_header("EXECUTING INSTALL PREREQUISITES")
+        return action_lib.install_prereqs
+
+    for action in ['deploy', 'preflight', 'postflight', 'uninstall', 'install_prereqs']:
+        if getattr(args, action):
+            errors = run_loop(locals()[action](args), args)
+            installer_analytics.send(
+                action=action,
```
pyronicide: How about `"installer_{}".format(action)`? I'd like to make sure we're not gonna clash on things like "deploy" in the future.

ext/dcos-installer/dcos_installer/installer_analytics.py
pyronicide: Note that it is `customerKey`, not `customer_key`.

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
+
         analytics.track(user_id=self.customer_key, anonymous_id=self.uuid, event=action, properties={
             "install_id": self.uuid,
             "bootstrap_id": self.bootstrap_version,
-            "provider": self.source,
-            "source": "installer",
+            "source": self.source,
             "install_method": install_method,
             "stage": action,
             "errors": num_errors,
```
pyronicide: Can we get `environmentVersion`?

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,36 @@
...
 
+        # We set customer key from config to avoid loading the config during class init
+        self.customer_key = backend.get_config()["customer_key"] if 'customer_key' in backend.get_config() else ""
+
         analytics.track(user_id=self.customer_key, anonymous_id=self.uuid, event=action, properties={
             "install_id": self.uuid,
             "bootstrap_id": self.bootstrap_version,
-            "provider": self.source,
-            "source": "installer",
+            "source": self.source,
```
pyronicide: `provider: "onprem"`
`source: "installer"`

cmaloney: This is both open and enterprise. This is the open source repository. Should include `BOOTSTRAP_VARIANT`, which will give that information.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: In the config.yaml it's `customer_key`, this isn't the key that will be sent to segmentIO.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: I'm not going to run exit out here, since that logic is taken care of upstream (unless you see a reason to dup that code here as well). I'll remove the default value.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: I'm going to default it to None instead of "" since that's the module default: https://github.com/segmentio/analytics-python/blob/master/analytics/client.py#L64

This way we avoid two tracks in `if` statements.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: I'll move this in as `customerKey` per above comment.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: This is the bootstrap_id

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: I thought source was `onprem`?

pytest/test_installer_backend.py
malnick: I'm not sure why this wasn't removed before. But this no longer returns the same output so I removed it.

malnick:
@cmaloney @pyronicide @spahl This is ready for another pass. Thanks!

spahl:
What's up with the test?

spahl:
Oh and can you also rebase as well? There are conflicts

ext/dcos-installer/dcos_installer/installer_analytics.py
pyronicide: Here's what you're sending me right now for `health`:

https://www.dropbox.com/s/1704qt0mf08v4y4/Screenshot%202016-05-27%2008.09.04.png?dl=0

malnick:
Rebased to the top three better commits from the last round of PR comments. Fixed a bug in the analytics class loading which should fix the test.

spahl:
You have commited binary files (built pkgpanda packages)

malnick:
Indeed. Someday, I'll figure out how computers work.

malnick:
@spahl Let's try this one more time...

darkonie:
https://github.com/dcos/dcos/pull/43/files#diff-b1a486deebfabe07c0790a5737a04c79L24
you should have the `prepare_status` 
here is an abstract class definition https://github.com/dcos/dcos/blob/master/ssh/utils.py#L43

malnick:
@cmaloney @spahl Green again!

ext/dcos-installer/dcos_installer/installer_analytics.py
```diff
@@ -2,33 +2,33 @@
...
-            "install_id": self.uuid,
-            "bootstrap_id": self.bootstrap_version,
-            "provider": self.source,
+        # We set customer key from config to avoid loading the config during class init
+        customer_key = backend.get_config().get("customer_key", None)
+
+        analytics.track(user_id=customer_key, anonymous_id=self.uuid, event=action, properties={
+            "provider": "onprem",
             "source": "installer",
+            "variant": os.getenv("BOOTSTRAP_VARIANT"),
```
cmaloney: os.environ['BOOTSTRAP_VARIANT']. `getenv()` doesn't error if the key doesn't exist. Same goes for BOOTSTRAP_ID. https://docs.python.org/3/library/os.html#os.getenv

ext/dcos-installer/dcos_installer/cli_dispatcher.py
```diff
@@ -0,0 +1,171 @@
...
+        else:
+            log.error('Choices are [y]es or [n]o. "{}" is not a choice'.format(do_uninstall))
+
+
+def log_warn_only():
+    """Drop to warning level and down to get around gen.generate() log.info
+    output"""
+    coloredlogs.install(
+        level='WARNING',
+        level_styles={
```
cmaloney: This code could be replaced with a call to https://github.com/dcos/dcos/pull/43/files#diff-b1a486deebfabe07c0790a5737a04c79R12 that sets level = 'WARNING'. Right now most of it is just copy / paste.

Probably makes sense to change setup_logger to take the level as the parameter (rather than an options object). IT doesn't really use anything but that one field. In the one case we call it with an options object, just pass in (`'DEBUG' if options.verbose else 'INFO'`)

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -1,108 +1,163 @@
...
+                'bold': True,
+            },
+        },
+        fmt='%(message)s',
+        isatty=True
+    )
+    log.debug("Logger set to DEBUG")
+
+
+def start_installer(args=None):
```
cmaloney: don't give a default to args here. We don't use it, it's dead code that could potentially hide programmer mistakes, and in theory makes it so every function this code passes args() to should actually check for `None` or list() to be correct.

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -1,108 +1,163 @@
...
+    def print_usage():
+        return """
+Install Mesosophere's Data Center Operating System
+
+dcos_installer [-h] [-f LOG_FILE] [--hash-password HASH_PASSWORD] [-v]
+[--web | --genconf | --preflight | --deploy | --postflight | --uninstall | --validate-config | --test]
+
+Environment Settings:
+
+PORT                  Set the :port to run the web UI
```
cmaloney: nit/  pre-existing: CHANNEL_NAME, BOOTSTRAP_ID shouldn't be set by end users. CHANNEL_NAME no longer exists as a configurable for this stuff.

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -1,108 +1,163 @@
...
+        action='store_true',
+        default=False,
+        help='Performs tests on the dcos_installer application')
+
+    options = parser.parse_args(args)
+    return options
+
+
+def main():
+    if len(sys.argv[1:]) == 0:
```
cmaloney: why not just `if len(sys.argv) == 1`? That seems more straightforward than slicing off the first element and checking if the length is 0.

cmaloney:
Much closer! Just need to change the os.getenv() calls to be something that hard fails if the variables aren't set (I'd suggest just indexing to os.environ, `os.environ['BOOTSTRAP_VARIANT']`), that way if we ever produce an artifact which changes them or doesn't set them, we'll know rather than get weird runtime errors.

ext/dcos-installer/dcos_installer/installer_analytics.py
malnick: My thinking was this code will never run unless the CLI is working, and the CLI already has a check on this ENV var. That being said, better safe than sorry.

malnick:
@cmaloney Big things are fixed

cmaloney:
:ship: Good improvement, still more to do in future refactors

PR #44, opened 2016-05-05
Restrict AWS public agents to mesos port ranges

lingmann:
AWS public agents were previously allowing traffic to ports up to 65535.
This change adjusts the public agent security group to only allow
traffic to ports controlled by the Mesos agent (ports below 32000).

This fixes a potential security issue, where internal DC/OS services
which listen these high range ports (including minuteman) could be
exposed to the outside world on the public agents.


cmaloney:
:ship:

PR #50, opened 2016-05-10
Remove unused files / packages

cmaloney:
This is trying to land the changes in #19 minus the "removing dnsypthon" which caused that to need to be reverterd by #33


cmaloney:
Launched a CloudFormation from the PR, everything seems to be functional.

PR #51, opened 2016-05-10
split lashup into separate service [WIP]

sargun:


packages/minuteman/build
```diff
@@ -16,7 +16,9 @@ cat <<EOF > $service
 Description=Layer 4 Load Balancer: DC/OS Layer 4 Load Balancing Service
 After=dcos-gen-resolvconf.service
 After=dcos-epmd.service
+After=dcos-lashup.service
 BindsTo=dcos-epmd.service
+BindsTo=dcos-lashup.service
```
cmaloney: nak using BindsTo. It's derived from Requires. Requires causes bad behavior

sargun: Should I get rid of the other bindsto?

cmaloney: Yea... We shouldn't have landed any BindsTo... 

Before/After also give no real guarantee of ordering (Although aren't actively harmful like Wants/Requires and friends)

PR #52, opened 2016-05-10
Teach packaging about "upstream" packages

cmaloney:
This is useful for being able to have a distributon of DC/OS derived from the
open source DC/OS.

The distribution of DC/OS could do things like
  - Add new DC/OS system packages which should be installed on every host
  - Replace DC/OS components with alternate implementations

In another patch set the ability to have gen spit out more configuration files
will be added.

TODO: 
 - [x] Write a test that upstreams work at all
 - [ ] Write documentation on how to build a derivative of DC/OS
 - [ ] Write code to allow adding new configuration parameters, changing configuration defaults / values in the gen library
 - [ ] Make `release create` work (this makes `mkpanda tree` work)

### DIFF
--- a/packages/.gitignore
+++ b/packages/.gitignore
@@ -2,6 +2,7 @@
 *.active.json
 *.bootstrap.tar.xz
 bootstrap.latest
+cache/*
 
 # Per-pacage artifacts
 */*.tar.xz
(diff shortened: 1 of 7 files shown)


cmaloney:
Tested upstreams out of band, and they are definitely working / I can have an independent "packages" directory and things get built. The moving of the "cache" to out of the individual packages folders makes it so the cache is kept even though we do a new checkout of the upstream each build.

PR #55, opened 2016-05-10
Ubuntu compatibility

iamthemuffinman:
I haven't seen any issues or PR's regarding Ubuntu support so I thought I'd get the ball rolling on this and start making some progress. I've fixed a few issues that I believe are affecting Ubuntu support.

- Everything is hardcoded to `/usr/bin` (which is fine, but we need symlinks similar to how azure is done because binaries in Ubuntu are generally in `/bin`.)
- The install prereqs stage only had CentOS/RHEL in mind. I've fixed that to include Ubuntu/Debian by looking at `/etc/os-release` and acting accordingly.

Let me know what you guys think.

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -41,6 +41,21 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     pf = get_async_runner(config, targets, async_delegate=async_delegate)
     chains = []
 
+    binary_chain = ssh.utils.CommandChain('preflight_compatibility')
+
+    # for Ubuntu/Debian derivatives
+    binary_chain.add_execute(
+        'sudo ln -s /bin/rm /usr/bin/rm',
+        'sudo ln -s /bin/mkdir /usr/bin/mkdir',
+        'sudo ln -s /bin/tar /usr/bin/tar',
+        'sudo ln -s /bin/ln /usr/bin/ln',
+        'sudo ln -s /bin/cp /usr/bin/cp',
+        'sudo ln -s /bin/systemctl /usr/bin/systemctl',
+        'sudo ln -s /bin/mount /usr/bin/mount',
+        'sudo ln -s /bin/bash /usr/bin/bash'
+    )
+    chains.append(binary_chain)
+
     preflight_chain = ssh.utils.CommandChain('preflight')
     # In web mode run if no --offline flag used.
     if options.web:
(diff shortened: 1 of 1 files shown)


iamthemuffinman:
Squashed to fix the code-quality tests.

bhack:
See also https://dcosjira.atlassian.net/projects/DCOS/issues/DCOS-25?filter=allopenissues

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -310,52 +325,96 @@ def _add_prereqs_script(chain):
 # setenforce is in this path
 PATH=$PATH:/sbin
 
-dist=$(cat /etc/*-release | sed -n 's@^ID="\(.*\)"$@\\1@p')
-
-if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
-  echo "$dist is not supported. Only RHEL and CentOS are supported" >&2
-  exit 0
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
```
bhack: What do you think to use python platform.linux_distribution call?

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -41,6 +41,21 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     pf = get_async_runner(config, targets, async_delegate=async_delegate)
     chains = []
 
+    binary_chain = ssh.utils.CommandChain('preflight_compatibility')
+
+    # for Ubuntu/Debian derivatives
+    binary_chain.add_execute(
```
malnick: Ideally this would execute only on Ubuntu derivatives of a specific release. I would rather this end up in a separate, Debian or Ubuntu only preflight. Generally speaking, we're about to implement a big refactor on action_lib. I think it would be nice to break out distro-specific code into their own code-bases (preflight-debian.py?) and we have a top level caller which would execute the correct action based on distro.

bhack:
@malnick what do you think to use yum and apt Python API?

malnick:
@iamthemuffinman I think this is a great start, and thank your for making the contribution. At the cost of making this preflight script more complicated, I think we should start a discussion around the following actions to make extending this a bit more organized:

1. Refactor action_lib to break out the __init__.py into individual actions (preflight.py, deploy.py, etc)
1. Make the actions distro-specific (preflight-centos.py, preflight-ubuntu.py, etc)
1. Make a top level action dispatcher in the __init__.py which does the logic around figuring out which distro we're on, it could be as simple as passing in the supported distro explicitly in the config.yaml (preferred) or maybe it executes `uname` on the host (more complicated, but extendable to heterogenous clusters).

cmaloney:
@bhack I much prefer using the command line commands for each distro than trying to use a "generic" library or some special API which likely isn't as maintained / developed as the command line which people interact with and use all the time.

malnick:
@bhack For client-side package installation?

bhack:
@malnick Yes @cmaloney ok.. but are official supported by yum and apt. Using api could let to better control errors and exception  if you need to notify users

malnick:
@bhack I agree with @cmaloney on this. This would require python, a specific distro of python, on the host to execute. That's a dependency we shouldn't have for something as general as package installation.

bhack:
Ok if we are without python go ahead with the (s)hell

ext/dcos-installer/dcos_installer/action_lib/__init__.py
iamthemuffinman: I agree. Breaking out a preflight for every distro or even just Debian/Ubuntu derivatives seems like it would get a little messy.

iamthemuffinman:
@malnick You're welcome! I agree with 1 and 3 wholeheartedly. 2 seems like it would get a little messy. I think refactoring into separate steps (ie: deploy.py, preflight.py, postflight.py, etc) and then decide from there whether or not distro-specific steps need to be broken out into separate files (ie: preflight-centos.py) is the way to go.

cmaloney:
There's some large pieces (both inside and outside this repo) which need to be looked at for this as well. Most notable being some way to test this in CI reliably (There is dcos-vagrant for CentOS, that doesn't work for ubuntu). The CI image also needs to have pre-requisites pre-installed (It's amazing how flaky things like `yum install` are inside of CI. It has been our leading install failure until we started baking them into the base machines for CI).

We also need to start looking at building out the dcos.io docs for this.

I don't things like refactoring to break out `__init__.py` into `preflight.py` and co at this point is actually very valuable or is valuable code motion as part of this PR. It's going to add a lot of noise.

I would argue that we need to make the bash script more maintainable with this it passes a critical threshold for maintainability (And it was arguably over that before but it skipped code review...). 

For that either need to split the bash into two pieces for the preflight (one for Ubuntu, one for CentOS) and make some way for the user specify ahead of time which to use (Slightly worse usability, but with an advantage of not guessing).

Better would be to do a little bit of introspection since the GUI installer is supposed to be the "easy" button. Ideally inspect all the hosts. Make sure they're all the "same" distribution, set that bit of config for the user, and use the appropriate prerequisite installer, as well as disallow alternate-OS nodes from joining the cluster (I think the potential issues mixing base OS inside a cluster probably aren't worth dealing with at this point unless there is a strong customer use case. DC/OS is supposed to make all machines look the same).

We also need to work on getting frameworks / DC/OS Universe packages and tutorials tested out on Ubuntu, and fixed up so we know when it's shipped day 1 things actually work.

Probably should actualy build out a big design doc / checklist of tasks... But that is a task for tomorrow

iamthemuffinman:
Sure. There should be some sort of design doc, checklist, etc to understand and track all the pieces that would need to be touched.

I don't think refactoring in this PR is valuable, but in general it is valuable. It will make the code easier to understand and probably easier to test.

I believe introspection is the best way to do that. Maybe with an optional override in case of bugs. (I could see someone in the slack channel having issues with their deployment because the wrong OS is being detected.) Disallowing alternate base OS nodes could be somewhat difficult. Presumably that info would be stored on the bootstrap node and whichever OS is detected first would become the only allowable base. I agree that there isn't really a use case for mixing base OS's inside a cluster.

A design doc/checklist would be extremely helpful. I am more than happy to work towards getting Ubuntu support ASAP. In my organization, we use Ubuntu for everything so getting DC/OS on Ubuntu is huge.

Thank you for taking the time to look at and review this!

bhack:
Why we cannot remove hardcoded commando path?

cmaloney:
Systemd mandates absolute paths to commands. In some cases we could switch
to some internal implementation of the same command.

isavcic:
$(readlink -f symlink) can't be used? It would provide an absolute link.

bhack:
Could we rely on "which" command or is not supposed to be installed?

iamthemuffinman:
We could abuse `/usr/bin/env`. ;)

isavcic:
I'm afraid I don't know anything about those checks that haven't completed, can someone give me a rundown on whether there's any news regarding this? Thanks!

cmaloney:
@isavcic The code is still a work in progress and there are bigger open design questions about how to do this, and landing the necessary bits across DC/OS.

mellenburg:
@isavcic the three uncompleted integraion-test/* checks are currently lacking proper documentation (coming soon). What these tests all do is spin up a minimal on-prem DCOS cluster using the installer and run this: https://github.com/dcos/dcos/blob/master/test_util/integration_test.py

For this particular PR, those tests won't provide meaningful feedback as they don't run on Ubuntu hosts. One of the many bits needed to land this ( as @cmaloney points out)  would be expanding test coverage.

bhack:
Is there a roadmap?

bhack:
Why?

spahl:
We can continue the discussion here: https://dcosjira.atlassian.net/browse/DCOS-25.

We are planning on adding ubuntu support. This PR is just not enough to get us there. I can't give a definitive date but will come back with some news soon.

It is of course entirely our intention to support ubuntu as a first class base OS

PR #56, opened 2016-05-11
Make `release create` with an upstream work

cmaloney:
Moves the critical resources used by `release create` to use `pkg_resources` to load from the python package rather than assuming they're relative to the current working directory.

With this patch successfully ran a `release create` in a derived repository.


spahl:
:ship:

PR #57, opened 2016-05-11
Remove old local_cp code in release storage. The local storage can do the same thing

cmaloney:


cmaloney:
Revisiting, I think we should actually just drop the `artifacts` directory stuff for this. We have the per-commit per-channel artifacts now which serve the same purpose inside our CI, and if a local copy is needed, can just use a local storage provider

PR #58, opened 2016-05-11
DCOS-6050: integration tests

kartick:
Depends on PR12 and a whole bunch of other stuff by @mellenburg.


test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
     r_data['Properties']['clusterId'] = ''
 
     assert r_data == exp_data
+
+
+class AgentManipulator:
+
+    def __init__(self, cluster, ssh_user, key_path):
```
lingmann: Add some docs to the init method for clarification (`cluster` is a bit too broad for end-user to know what to pass here).

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+        self._cluster = cluster
+        self._ssh_user = cluster.ssh_user
+        self._ssh_key_path = cluster.ssh_key_path
+
+    def _multi_run(self, cmds):
+        with contextlib.closing(ssh_tunnel.SSHTunnel(
+            self._cluster.ssh_user,
```
lingmann: Should this be `self._cluster._ssh_user`?

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+
+    def _run_volume_discovery(self, delete_only=False):
+        steps = (
+            'rm /var/lib/dcos/mesos-resources',
+            'systemctl start dcos-vol-discovery-priv-agent'
+        )
+        return self._multi_run(steps[:1] if delete_only else steps)
+
+    def _make_local_volume(self, size_mb, image_file):
+        return self._run('dd of={} if=/dev/null blocksize={}'.format(image_file, size_mb))
```
lingmann: I don't think `blocksize` is a valid argument. Probably want something like this instead:
`'dd of={} if=/dev/null bs=1M count={}'.format(image_file, size_mb)`

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+        yield data
+
+
+def test_add_volume_noop(cluster):
+    agentm = AgentManipulator(cluster)
+    agentm._stop_mesos_agent()
+    added_volumes = agentm._add_volumes_to_agent((200, 200))
+    agentm._run_volume_discovery()
+    agentm._start_mesos_agent(clear_state=False)
+    agentm._remove_volumes_from_agent(added_volumes)
```
lingmann: I was a bit confused about why this line was here, until I realized you are doing some pre-assert cleanup work. So a comment might be nice... perhaps 'clean up mounted volumes before running test to ensure test does not mutate agent state'.

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+    def _clear_agent_state(self):
+        # TODO: check this path
+        self._run('rm -rf /var/lib/mesos/slave')
+
+    def _run_volume_discovery(self, delete_only=False):
+        steps = (
+            'rm /var/lib/dcos/mesos-resources',
+            'systemctl start dcos-vol-discovery-priv-agent'
+        )
+        return self._multi_run(steps[:1] if delete_only else steps)
```
lingmann: Probably better to split _run_volume_discovery into two simple methods:
* `_start_volume_discovery(self)` to match the start_mesos_agent naming convention
* `_clear_volume_discovery_state(self)` to match the clear_agent_state naming convention

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+    assert 'error' in started.lower()
+
+
+def test_add_volume_works(cluster):
+    agentm = AgentManipulator(cluster)
+    agentm._stop_mesos_agent()
+    added_volumes = agentm._add_volumes_to_agent((200, 200))
+    agentm._run_volume_discovery(delete_only=True)
+    started = agentm._start_mesos_agent(clear_state=True)
+    agentm._remove_volumes_from_agent(added_volumes)
```
lingmann: See above comment. In this case I think we actually need to do more thorough cleanup (stop agent, remove volumes, run volume discovery, start agent), since the agent will be registered with mount volumes which don't exist when this test finally exits.

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+        for size, _, vol in added_volumes:
+            assert vol not in d
+
+
+def test_missing_disk_resource_file(cluster):
+    agentm = AgentManipulator(cluster)
+    agentm._stop_mesos_agent()
+    agentm._run_volume_discovery(delete_only=True)
+    started = agentm._start_mesos_agent(clear_state=False)
+    assert 'error' in started.lower()
```
lingmann: Cleanup here, by re-running volume discovery after the test?

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+            assert vol in d
+
+
+def test_vol_discovery_fails_due_to_size(cluster):
+    agentm = AgentManipulator(cluster)
+    agentm._stop_mesos_agent()
+    added_volumes = agentm._add_volumes_to_agent((50,))
+    agentm._run_volume_discovery()
+    res_file = agentm._run('file /var/lib/dcos/mesos-resources')
+    agentm._remove_volumes_from_agent(added_volumes)
```
lingmann: Cleanup finalizer: (remove volumes from agent, re-run volume discovery service)

test_util/integration_test.py
```diff
@@ -1340,3 +1349,148 @@ def wait_for_endpoint():
...
+    assert 'No such file or directory' in res_file
+
+
+def test_vol_discovery_non_json_mesos_resources(cluster):
+    agentm = AgentManipulator(cluster)
+    agentm._run('echo "MESOS_RESOURCES=ports:[1025-2180]" > /etc/mesos-slave')
+    discovery_status = agentm._run_volume_discovery()
+    assert 'error' in discovery_status
```
lingmann: Some sort of cleanup finalizer here as well.

lingmann:
@kartick this LGTM once all of the integration tests are passing.

kartick:
Closed in favor of https://github.com/dcos/dcos/pull/126

PR #59, opened 2016-05-12
[WIP] Added DC/OS overlay network packages and its dependencies to the installer.

asridharan:
Added two packages:
  * mesos-overlay
  * cni

Also made the following modifications to the installer:
  * Introduced user configurable parameters `dcos_overlay_enable` and
   `dcos_overlay_network`.
  * Modified the `dcos-config.yaml` template to enable DC/OS overlay
    specific Mesos modules when the user enables DC/OS overlay by
    setting `dcos_overlay_enable` parameter to "true".
  * Added generation of `overlay_config.json` to specify the overlay
    networks  specified by user using the `dcos_overlay_network`
    parameter.

### DIFF
--- a/gen/aws/templates/cloudformation.json
+++ b/gen/aws/templates/cloudformation.json
@@ -71,10 +71,10 @@
         "stable": "ami-05bc0164"
       },
       "us-west-1": {
-        "stable": "ami-27553a47"
+        "stable": "ami-652c5505"
       },
       "us-west-2": {
-        "stable": "ami-00ebfc61"
+        "stable": "ami-f5bc4e95"
       }
     },
     "NATAmi" : {
--- a/gen/aws/templates/cloudformation.json
+++ b/gen/aws/templates/cloudformation.json
@@ -47,34 +47,34 @@
   "Mappings": {
     "RegionToAmi": {
       "ap-northeast-1": {
-        "stable": "ami-84e0c7ea"
+        "stable": "ami-802dcde1"
       },
       "ap-southeast-1": {
-        "stable": "ami-da67a0b9"
+        "stable": "ami-4c97462f"
       },
       "ap-southeast-2": {
-        "stable": "ami-f35b0590"
+        "stable": "ami-55f6d936"
       },
       "eu-central-1": {
-        "stable": "ami-fdd4c791"
+        "stable": "ami-3b1cf054"
       },
       "eu-west-1": {
-        "stable": "ami-55d20b26"
+        "stable": "ami-2a118659"
       },
       "sa-east-1": {
-        "stable": "ami-154af179"
+        "stable": "ami-ee1b9382"
       },
       "us-east-1": {
-        "stable": "ami-37bdc15d"
+        "stable": "ami-3cf20751"
       },
       "us-gov-west-1": {
-        "stable": "ami-05bc0164"
+        "stable": "ami-9041fef1"
       },
       "us-west-1": {
-        "stable": "ami-652c5505"
+        "stable": "ami-61e99101"
       },
       "us-west-2": {
-        "stable": "ami-f5bc4e95"
+        "stable": "ami-c30af5a3"
       }
     },
     "NATAmi" : {
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -86,6 +86,22 @@ def calculate_mesos_slave_modules_json(mesos_slave_modules):
     return json_multiline.replace('\n', '\n' + injected_indent)
 
 
+def calculate_mesos_slave_modules(dcos_overlay_enable):
+    if dcos_overlay_enable == 'false':
+        return calculate_mesos_slave_modules_json(__default_mesos_slave_modules)
+    else:
+        return \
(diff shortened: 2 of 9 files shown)


gen/aws/templates/cloudformation.json
```diff
@@ -71,10 +71,10 @@
         "stable": "ami-05bc0164"
       },
       "us-west-1": {
-        "stable": "ami-27553a47"
```
cmaloney: Why are we changing these AMIs?

gen/calc.py
```diff
@@ -86,6 +86,22 @@ def calculate_mesos_slave_modules_json(mesos_slave_modules):
     return json_multiline.replace('\n', '\n' + injected_indent)
 
 
+def calculate_mesos_slave_modules(dcos_overlay_enable):
```
cmaloney: This really breaks adding more downstream modules... Would it be reasonable to wait to land this until the modules dir stuff @karya0 has been working on lands? That makes this really simple to do.

gen/calc.py
```diff
@@ -296,7 +366,18 @@ def calculate_exhibitor_static_ensemble(master_list):
         'ui_banner_header_content': 'null',
         'ui_banner_footer_content': 'null',
         'ui_banner_image_path': 'null',
-        'ui_banner_dismissible': 'null'
+        'ui_banner_dismissible': 'null',
+        'dcos_overlay_network': '{                    \
```
cmaloney: This doesn't seem particularly safe to default. Probably we want to just lay out in docs waht all needs to be set and how to set it. That this is not configurable at cluster runtime is likely to also be an issue.

As a code style thing, if we want the multi-line format here, should use a python multi-line comment `"""` and declare the value earlier, then just reference that declared value here.

packages/mesos/buildinfo.json
```diff
@@ -3,9 +3,9 @@
   "docker": "dcos-builder",
   "single_source" : {
     "kind": "git",
-    "git": "https://github.com/mesosphere/mesos",
-    "ref": "76c87d55a78e8cbb380fc0c71388ccec53ed1dc8",
-    "ref_origin" : "dcos-1.7"
+    "git": "git@github.com:apache/mesos.git",
```
cmaloney: can't land changing mesos like this here. We bump DC/OS Mesos after a Mesos release

gen/aws/templates/cloudformation.json
asridharan: For DC/OS overlay we need docker 1.9 and higher. The cloud formation template has an older CoreOS AMI that comes with Docker 1.7. So in order to test the overlay features with cloud formation, we needed a newer AMI. 

I will move this to a separate commit so that it can be discarded in case we plan to upgrade the AMI using a separate commit.

gen/calc.py
asridharan: We need this in 1.8. So if the infrastructure for adding modules is ready in that time frame, we can discard this PR and just use that infrastructure. Will check with @karya0 on this.

packages/mesos/buildinfo.json
asridharan: Will Make this into a separate commit, and discard the commit once we are done testing here.

gen/aws/templates/cloudformation.json
cmaloney: We need to at the very least give notice to support that we're changing the base requirements if we're going to update our Docker requirement. We also need to get from support if upgrading docker will be a dealbreaker for any customers (We really don't want to strand customers on old DC/OS).

As far as upgrading CoreOS, we need to do it in all regions/AMIs if we're doing it. And do it for other providers than AWS. Also note we intentionally haven't bumped CoreOS in a while: https://dcosjira.atlassian.net/browse/DCOS-38. We need to resolve that issue before we can update / change the CoreOS version.

asridharan: @spahl @cmaloney has valid points here. We probably should looking at updating the base AMI ASAP else this might become a blocker?

gen/calc.py
asridharan: Had a chat with @karya0. Looks like he should be adding the `--modules_dir` flag shortly. Once its merged upstream, I will go ahead and restructure this PR to use the `--modules_dir` flag on the Master and the Agent.

gen/calc.py
malnick: I echo @cmaloney and would add that it might be nice to have the parameters configurable, that way in the future we could configure subsets of the defaulted JSON (though this is purely an armchair observation, it would make validating the contents of the config more attainable as well if we had a complete data structure and not a string blob).

sargun: @cmaloney Why don't you think it's safe to default?

cmaloney: It states we own / customers don't use several IP spaces.

It also says `'dcos_overlay_enable': 'true'` which both doesn't seem right (and breaks standard json practice...)

sargun: "44.128.0.0/16" is a safe set of IPs.
We could choose a different set for VTEPs. Like from 198.18.0.0/15 (maybe 198.19.0/20?)

PR #61, opened 2016-05-13
WIP: cli only packages compatible cosmos

tamarrow:


cmaloney:
Newer cosmos has been merged in, binary cli is being linked to by default in #376

PR #68, opened 2016-05-16
Build docker images locally

mellenburg:
Refactors how VPC intgration tests are run
-creates pluggable python module for setting up and running
tests on bare remote hosts (or mesos masters)
-Adds CLI for running on arbitrary clusters as a tool
-bakes arguments and integration test into docker image
before every test => custom depdencies can be used in integration +
huge string of arguments does not need to be copied to rerun test

### DIFF
--- a/test_util/test_runner.py
+++ b/test_util/test_runner.py
@@ -3,13 +3,16 @@
 Note: ssh_user must be able to use docker without sudo priveleges
 """
 import logging
+import tempfile
 import time
 from contextlib import contextmanager
 from multiprocessing import Process
 from os.path import join
-from subprocess import CalledProcessError, TimeoutExpired
+from subprocess import CalledProcessError, TimeoutExpired, check_call
 
 import pkg_resources
+
+from pkgpanda.util import write_string
 
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
@@ -65,9 +68,7 @@ def prepare_test_registry(tunnel, test_dir):
     """
     test_server_docker = pkg_filename('docker/test_server/Dockerfile')
     test_server_script = pkg_filename('docker/test_server/test_server.py')
-    pytest_docker = pkg_filename('docker/py.test/Dockerfile')
     log.info('Setting up integration_test.py to run on ' + tunnel.host)
-    tunnel.remote_cmd(['mkdir', '-p', test_dir])
 
     tunnel.remote_cmd(['mkdir', '-p', join(test_dir, 'test_server')])
     tunnel.write_to_remote(test_server_docker, join(test_dir, 'test_server/Dockerfile'))
@@ -90,12 +91,6 @@ def prepare_test_registry(tunnel, test_dir):
     tunnel.remote_cmd(['docker', 'push', '127.0.0.1:5000/test_server'])
     log.debug('Cleaning up test_server files')
     tunnel.remote_cmd(['rm', '-rf', join(test_dir, 'test_server')])
-    log.info('Building base integration_test.py container on test host')
-    tunnel.remote_cmd(['mkdir', '-p', join(test_dir, 'py.test')])
-    tunnel.write_to_remote(pytest_docker, join(test_dir, 'py.test/Dockerfile'))
-    tunnel.remote_cmd([
-        'cd', join(test_dir, 'py.test'), '&&', 'docker', 'build', '-t', 'py.test', '.'])
-    tunnel.remote_cmd(['rm', '-rf', join(test_dir, 'py.test')])
 
 
 def integration_test(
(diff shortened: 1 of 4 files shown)


mellenburg:
@lingmann This is ready for review (note it is currently dependent upon https://github.com/dcos/dcos/pull/122 )

test_util/test_runner.py
```diff
@@ -117,31 +126,43 @@ def integration_test(
...
+                aws_secret_access_key=aws_secret_access_key,
+                region=region))
+
+    # image uses ONBUILD so put the appropriate resources in the temp/build folder
+    log.info('Building and exporting integration test docker image')
+    pytest_docker = pkg_filename('docker/py.test/Dockerfile')
     test_script = pkg_filename('integration_test.py')
+    check_call(['docker', 'build', '-t', 'py.test:noop', '-'], stdin=open(pytest_docker))
+    shutil.copy(test_script, join(temp_dir, 'integration_test.py'))
+    check_call(['docker', 'build', '-t', 'py.test:run', '.'], cwd=temp_dir)
```
cmaloney: I don't see what the advantage of having this be a separate / second build step is. From what I can see here, we always build them in sequence, and only build the `:run` one once. Docker will do layer caching that makes just doing a single build the repeated layers will cache just like splitting it into two parts does.

test_util/test_runner.py
```diff
@@ -1,28 +1,52 @@
...
 
 import pkg_resources
 
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
 log = logging.getLogger(__name__)
+log.setLevel(logging.DEBUG)
+
+TEST_DOCKERFILE = """FROM py.test:noop
+ENV DCOS_DNS_ADDRESS=http://{dcos_dns}
```
cmaloney: Every one of these ENV lines adds another docker layer, and each layer adds a bit of time.  Would be better to do a "wrapper file" that is copied in which sets all of these. See: https://github.com/dcos/dcos/blob/master/gen/installer/bash/installer_internal_wrapper.in for what we do in dcos_generate_config.sh's docker.

test_util/test_runner.py
```diff
@@ -34,71 +58,57 @@ def proxy_registry_as_local(tunnel, agent_list, test_dir):
...
+    log.info('Building and exporting '+test_server_image)
+    temp_dir = tempfile.mkdtemp()
     test_server_docker = pkg_filename('docker/test_server/Dockerfile')
     test_server_script = pkg_filename('docker/test_server/test_server.py')
-    pytest_docker = pkg_filename('docker/py.test/Dockerfile')
-    log.info('Setting up integration_test.py to run on ' + tunnel.host)
+    shutil.copy(test_server_docker, join(temp_dir, 'Dockerfile'))
+    shutil.copy(test_server_script, join(temp_dir, 'test_server.py'))
+    check_call(['docker', 'build', '-t', test_server_image, '.'], cwd=temp_dir)
+    check_call(['docker', 'save', '-o', join(temp_dir, 'test_server.tar'), test_server_image])
```
cmaloney: Why do we need to do this export to tar, ship across, then import from tar? We should be able to just push from the local over the network.

If we can't, than it will be very beneficial to just copy the Dockerfile and test_server.py, and build the docker remotely.

test_util/test_runner.py
```diff
@@ -117,31 +126,43 @@ def integration_test(
...
-        '-e', 'AWS_SECRET_ACCESS_KEY='+aws_secret_access_key,
-        '-e', 'AWS_REGION='+region,
-        '--net=host', '--name='+test_container_name, 'py.test', 'py.test',
-        '-vv', ci_flags, '/integration_test.py']
-    log.info('Running integration test...')
+    test_cmd = ['docker', 'run', '--net=host', '--name='+test_container_name, 'py.test:run',
+                'py.test', '-vv', ci_flags, '/integration_test.py']
     try:
-        with proxy_registry_as_local(tunnel, agent_list, test_dir):
+        with remote_port_forwarding(tunnel, agent_list, join(test_dir, 'ssh_Key')):
```
cmaloney: ssh_Key -> ssh_key

test_util/test_runner.py
mellenburg: The point of leaving it in two build steps is that this docker image is used test_util/run-all so and I did not want to disturb the flow there. Would you prefer to see them both use this approach? I would like if they could keep sharing the same Dockerfile

test_util/test_runner.py
mellenburg: Good point, I will rework towards that pattern

test_util/test_runner.py
mellenburg: I'm not a huge fan of keeping the registry on the teamcity host and tunneling to the test host which then maintains a tunnel to all the dcos nodes.

The reason that I changed this from doing your favored approach was that I was hoping to utilize the fact that the test_server and registry layers have likely already been built and cached on the test runner node, whereas if we build on remote we will definitely have to pull in all the deps.

Also, just for kicks and giggles I thought it would be cool if this test was made in such a way that none of the test nodes needed a connection outside the local network. E.G. if you wanted to use this as a CLI tool to trigger tests against arbitrary clusters, it would be nice if your test node could be in a restricted network

So, that was the logic, but those features aren't currently being asked for, so I can definitely revert this if desired

mellenburg:

test_util/test_runner.py
```diff
@@ -3,18 +3,30 @@
...
 import pkg_resources
 
+import gen.template
+
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
 log = logging.getLogger(__name__)
 
+BASH_TEMPLATE = """
+#!/bin/bash
```
cmaloney: I'd add a `set -euo pipefail; set -x` here to make it harder for us to make bash mistakes and miss them.

test_util/test_runner.py
```diff
@@ -90,12 +100,12 @@ def prepare_test_registry(tunnel, test_dir):
...
     tunnel.remote_cmd(['rm', '-rf', join(test_dir, 'test_server')])
-    log.info('Building base integration_test.py container on test host')
-    tunnel.remote_cmd(['mkdir', '-p', join(test_dir, 'py.test')])
-    tunnel.write_to_remote(pytest_docker, join(test_dir, 'py.test/Dockerfile'))
-    tunnel.remote_cmd([
-        'cd', join(test_dir, 'py.test'), '&&', 'docker', 'build', '-t', 'py.test', '.'])
-    tunnel.remote_cmd(['rm', '-rf', join(test_dir, 'py.test')])
+
+
+def compressed_transfer(tunnel, src, dst):
```
cmaloney: rather than manually gzipping / unpacking, why not use `ssh -C`?

test_util/test_runner.py
```diff
@@ -120,37 +130,53 @@ def integration_test(
...
-        '-e', 'PUBLIC_MASTER_HOSTS='+','.join(master_list),
-        '-e', 'SLAVE_HOSTS='+','.join(agent_list),
-        '-e', 'PUBLIC_SLAVE_HOSTS='+','.join(public_agent_list),
-        '-e', 'REGISTRY_HOST=127.0.0.1',
-        '-e', 'DCOS_VARIANT='+variant,
-        '-e', 'DNS_SEARCH='+dns_search,
-        '-e', 'AWS_ACCESS_KEY_ID='+aws_access_key_id,
-        '-e', 'AWS_SECRET_ACCESS_KEY='+aws_secret_access_key,
-        '-e', 'AWS_REGION='+region])
+    test_env = [
```
cmaloney: We only use BASH_TEMPLATE once, and we only use it for this stuff. Why not make all these formattings part of it. Something like:

```

"""
#!/bin/bash
set -euo pipefail
set -x
export DCOS_DNS_ADDRESS=http://{dcos_dns}
export MASTER_HOSTS={master_hosts}
export PUBLIC_MASTER_HOSTS={master_hosts}
export SLAVE_HOSTS={slave_hosts}
# ...

{test_cmd}
""".format(
  dcos_dns=dcos_dns,
  master_hosts=','.join(master_list),
  slave_hosts=','.join(agent_list)
)
```

(Still leaving it as the template up above, .format down below). Then we're doing one less layer of string formatting (It all just happens at once into cmd_script, and we should be fine living just with python format strings, unless there's some particular reason you want to use gen.template, which is also fine to use).

test_util/test_runner.py
```diff
@@ -120,37 +130,53 @@ def integration_test(
...
-    test_cmd.extend(['py.test', 'py.test', '-vv', ci_flags, '/integration_test.py'])
+            extra_env = key + '=' + value
+            test_env.append(extra_env)
+    test_cmd = 'py.test -vv ' + ci_flags + ' /integration_test.py'
+    cmd_script = gen.template.parse_str(BASH_TEMPLATE).render({'test_env': test_env,
+                                                               'test_cmd': test_cmd})
+    log.info('Building py.test image')
+    # Make a clean docker context
+    temp_dir = tempfile.mkdtemp()
+    with open(join(temp_dir, 'test_wrapper.sh'), 'w') as fh:
```
cmaloney:
```
from pkgpanda.util import write_string

write_string(join(temp_dir, 'test_wrapper.sh'), cmd_script)
```
I think that's a little cleaner (hides more API details which aren't the focus here), but fine either way.

mellenburg:
@cmaloney Thanks for the review, this is ready for another pass!

cmaloney:
code-quality failure is spurious, rebasing on master will fix (picks up #222)

mellenburg:
@cmaloney Is this ready to ship? Once we get this in, we can unblock https://github.com/dcos/dcos/pull/135
and
https://github.com/dcos/dcos/pull/206

cmaloney:
🚢

PR #74, opened 2016-05-17
Create test_runner module

mellenburg:
-transfers generic test-orchestration code from test_installer_ccm
-switches from awkward synchronous async ssh usage to sync ssh
-configures agents via single control node for easy compatibility
  with clusters behind load balancers

### DIFF
--- a/ssh/ssh_tunnel.py
+++ b/ssh/ssh_tunnel.py
@@ -54,7 +54,7 @@ class SSHTunnel():
         check_call(start_tunnel)
         logger.debug('SSH Tunnel established!')
 
-    def remote_cmd(self, cmd, raw_output=False, timeout=None):
+    def remote_cmd(self, cmd, timeout=None):
         """
         Args:
             cmd: list of strings that will be interpretted in a subprocess
@@ -67,14 +67,11 @@ class SSHTunnel():
         logger.debug('Running socket cmd: ' + ' '.join(run_cmd))
         try:
             output = check_output(run_cmd, timeout=timeout)
-            if raw_output:
-                return output
-            else:
-                return output.decode('utf-8').rstrip('\r\n')
+            return output
         except TimeoutExpired as e:
-            logging.error('{} timed out after {} seconds'.format(cmd, timeout))
+            logging.exception('{} timed out after {} seconds'.format(cmd, timeout))
             logging.debug('Timed out process output:\n' + e.output)
-            raise e
+            raise
 
     def write_to_remote(self, src, dst):
         """
--- a/ssh/ssh_tunnel.py
+++ b/ssh/ssh_tunnel.py
@@ -54,7 +54,7 @@ class SSHTunnel():
         check_call(start_tunnel)
         logger.debug('SSH Tunnel established!')
 
-    def remote_cmd(self, cmd, timeout=None):
+    def remote_cmd(self, cmd, raw_output=False, timeout=None):
         """
         Args:
             cmd: list of strings that will be interpretted in a subprocess
@@ -66,11 +66,15 @@ class SSHTunnel():
         run_cmd = self.ssh_cmd + ['-p', str(self.port), self.target] + cmd
         logger.debug('Running socket cmd: ' + ' '.join(run_cmd))
         try:
-            return check_output(run_cmd, timeout=timeout).decode('utf-8').rstrip('\r\n')
+            output = check_output(run_cmd, timeout=timeout)
+            if raw_output:
+                return output
+            else:
+                return output.decode('utf-8').rstrip('\r\n')
         except TimeoutExpired as e:
             logging.error('{} timed out after {} seconds'.format(cmd, timeout))
             logging.debug('Timed out process output:\n' + e.output)
-            raise TimeoutExpired
+            raise e
 
     def write_to_remote(self, src, dst):
         """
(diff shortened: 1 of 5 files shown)


mellenburg:
@kartck @lingmann This is my MVP for major test refactoring. With this PR, there is a python binding for abitrary tests (e.g. azure and AWS), multi agent configuration through single host (azure tests require this), and completely constructing the py test docker image locally which allows adding arbitrary dependencies to  integration_test.py (e.g. ssh module for volumes test)

test_util/test_runner.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python3
+"""Module for running integration_test.py inside of a remote cluster
```
malnick: How hard would it be to leverage this in postflight during the installation process?

malnick:
This is a great refactor! 👍

test_util/test_runner.py
mellenburg: It would not be too difficult, but I think we wold need to put a little more effort into how we run tests rather than just plugging in py.test into the installer. I.E. racy tests should have better retry tolerance and the output would need to be restructured.

That being said, I think it would be awesome if postflight was just a thin wrapper for integration_test.py

mellenburg:
@cmaloney, @spahl 
This refactor is ready for review

spahl:
This is a big one. Can we split it into multiple logical commits?

mellenburg:
@spahl I split into into 4 chunks, but I do not think its too feasible/useful to get the big -317 one (replace ssh.ssh_runner with test_util.test_runner) to be much smaller. Thoughts?

ssh/ssh_tunnel.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, ssh_user, ssh_key_path, host, port=22):
         check_call(start_tunnel)
         logger.debug('SSH Tunnel established!')
 
-    def remote_cmd(self, cmd, timeout=None):
+    def remote_cmd(self, cmd, raw_output=False, timeout=None):
```
cmaloney: why do we ever want to run without raw output mode? It seems like generally the better option for running the commands

ssh/ssh_tunnel.py
```diff
@@ -66,11 +66,15 @@ def remote_cmd(self, cmd, timeout=None):
...
         logger.debug('Running socket cmd: ' + ' '.join(run_cmd))
         try:
-            return check_output(run_cmd, timeout=timeout).decode('utf-8').rstrip('\r\n')
+            output = check_output(run_cmd, timeout=timeout)
+            if raw_output:
+                return output
+            else:
+                return output.decode('utf-8').rstrip('\r\n')
         except TimeoutExpired as e:
             logging.error('{} timed out after {} seconds'.format(cmd, timeout))
```
cmaloney: Should use https://docs.python.org/3/library/logging.html#logging.Logger.exception

Also note, should just do a `raise`, that way you re-throw with all the attached stack history the most easily (otherwise you need to do a `raise ex from ex` thing).

test_util/test_runner.py
cmaloney: Note for this, we don't have SSH in a bunch of our postflights. AWS, Azure don't have SSH, but ideally they all should work about hte same way.

test_util/test_installer_ccm.py
```diff
@@ -123,153 +83,10 @@ def remote(path):
...
-    host_prep_chain.add_execute(['sudo', 'groupadd', '-g', '65500', 'nogroup'])
-    host_prep_chain.add_execute(['sudo', 'usermod', '-aG', 'docker', 'centos'])
-
-    if minuteman_enabled:
-        host_prep_chain.add_execute(['sudo', 'mkdir', '-p', '/etc/mesosphere/roles'])
-        host_prep_chain.add_execute(['sudo', 'touch', '/etc/mesosphere/roles/minuteman'])
-
-    check_results(run_loop(ssh_runner, host_prep_chain))
+    tunnel.write_to_remote(ip_detect_script, remote('ip-detect.sh'))
+    local_ip = tunnel.remote_cmd(['bash', remote('ip-detect.sh')])
```
lingmann: This method (get_local_address) will not work with clusters which have already been provisioned, since the ip detect script is unique depending on the environment. Can we just default to calling `/opt/mesosphere/bin/detect_ip` and only upload and run a custom ip detect script if the consumer of this method overrides the default behavior by passing in a custom ip detect script? That should make it more generic.

ssh/ssh_tunnel.py
mellenburg: I agree, this formatting is a holdover from trying to parse information retrieved with remote_cmd through the azure load balancer. For some reason, even though you are on *nix, because you go through the azure LB everything has windows returns. :( Anyway, I just need to remember to account for that when I plug in azure.

mellenburg:
@spahl , @cmaloney  this is ready for review again

lingmann:
After a pretty detailed review with @mellenburg and @kartick yesterday, I feel like this has rough edges, but is clearly a big improvement over the refactored integration test codebase. @mellenburg can you please confirm that your latest changes pass against our Azure test infrastructure? If that looks good then I am a 👍

malnick:
If nobody else disagrees, I like the commits as they are, with the exception of https://github.com/dcos/dcos/pull/74/commits/2a7cbe8e9dfa47f5987769ba66eeb93693fac485 can probably be squashed. 

👍 Nice work @mellenburg

kartick:
LGTM.

PR #77, opened 2016-05-18
Check in CI log collection

mellenburg:
Currently stored in TC, but needs to edited with every test configuration change. Rather, it should just live with the test

### DIFF
--- a/test_util/run-all
+++ b/test_util/run-all
@@ -113,8 +113,16 @@ RET=$?
 mkdir -p $WORKDIR/test/vagrant-logs
 if [ $RET -ne 0 ]; then
   echo "TEST FAILED"
+  echo "Attempting to gather logs"
+  export LOG_REPLAY_LINES=${LOG_REPLAY_LINES:=20}
   for i in boot m1 m2 m3 a1 a2; do
-    vagrant ssh $i -c 'sudo journalctl -o short' > $WORKDIR/test/vagrant-logs/$i
+    if [ "$TEAMCITY_BUILD_ENV" == "true" ]; then
+      echo "##teamcity[blockOpened name='$i']"
+      vagrant ssh $i -c "'sudo journalctl -n $LOG_REPLAY_LINES'"
+      echo "##teamcity[blockClosed name='$i']"
+    else
+      vagrant ssh $i -c "'sudo journalctl -n $LOG_REPLAY_LINES'"
+    fi
   done
   exit 1
 fi


malnick:
👍

spahl:
:ship:

test_util/run-all
```diff
@@ -113,8 +113,16 @@ RET=$?
 mkdir -p $WORKDIR/test/vagrant-logs
 if [ $RET -ne 0 ]; then
   echo "TEST FAILED"
+  echo "Attempting to gather logs"
+  export LOG_REPLAY_LINES=${LOG_REPLAY_LINES:=1000}
```
cmaloney: why does this need to be configurable? Why is it configurable only for one of the two methods?

test_util/run-all
```diff
@@ -113,8 +113,16 @@ RET=$?
...
   for i in boot m1 m2 m3 a1 a2; do
-    vagrant ssh $i -c 'sudo journalctl -o short' > $WORKDIR/test/vagrant-logs/$i
+    if [ "$TEAMCITY_BUILD_ENV" == "true" ]; then
+      echo "##teamcity[blockOpened name='$i']"
+      vagrant ssh $i -c "'sudo journalctl -n $LOG_REPLAY_LINES'"
+      echo "##teamcity[blockClosed name='$i']"
+    else
+      vagrant ssh $i -c 'sudo journalctl -o short' > $WORKDIR/test/vagrant-logs/$i
```
cmaloney: Why is this one short but the other one not?

I also don't see value in keeping this dumping to the vagrant-logs folder. We don't pick it up with our CI as artifacts.

test_util/run-all
mellenburg: This doesnt need to be configurable as much as it is set at 10000 in TC right now, and that seemed a bit excessive to hard code in

test_util/run-all
mellenburg: This commit was just merging the build steps on TC into this script to avoid situations of "something merged, so immediately go fix the build system", thus I didnt want to change the status quo.

As for what we should do, I would agree on dumping the folder. I think using short should be the default (most people dont want thousands of lines of stuff dumped to their terminal) and if LOG_REPLAY_LINES  is set (as it is in TC), then use that value

spahl: Ok, can you update you PR please?

dump the folder, change to short and keep LOG_REPLAY_LINES since it's nicer than just to have a default

mellenburg:
@spahl failed due to CCM being down. rerunning is all thats needed

test_util/run-all
```diff
@@ -20,6 +20,14 @@ cleanup_vagrant_leftovers() {
   set -e
 }
 
+replay_logs() {
```
spahl: Lets simplify this and only do `vagrant ssh $1 -c "'sudo journalctl -n $LOG_REPLAY_LINES'"` further down, removing the function. This is not particularly useful.

mellenburg: Done

spahl:
:ship:

PR #80, opened 2016-05-18
Update options for test_installer_ccm.py

mellenburg:
-Minuteman enabled is a deprecated configuration option
-Collect AWS env vars in the same manner as everything else
-Add ability to set cluster size/composition

### DIFF
--- a/test_util/test_installer_ccm.py
+++ b/test_util/test_installer_ccm.py
@@ -54,8 +54,6 @@ import test_util.installer_api_test
 from ssh.ssh_runner import MultiRunner
 from ssh.utils import CommandChain, SyncCmdDelegate
 
-AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID', '')
-AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY', '')
 
 DEFAULT_AWS_REGION = 'us-west-2'
 
@@ -200,7 +198,7 @@ def test_setup(ssh_runner, registry, remote_dir, use_zk_backend):
 
 def integration_test(
         ssh_runner, dcos_dns, master_list, agent_list, public_agent_list, region, registry_host,
-        test_dns_search, ci_flags):
+        test_dns_search, ci_flags, aws_access_key_id, aws_secret_access_key):
     """Runs integration test on host
     Note: check_results() will raise AssertionError if test fails
 
@@ -227,8 +225,8 @@ def integration_test(
         '-e', 'REGISTRY_HOST='+registry_host,
         '-e', 'DCOS_VARIANT=default',
         '-e', 'DNS_SEARCH='+dns_search,
-        '-e', 'AWS_ACCESS_KEY_ID='+AWS_ACCESS_KEY_ID,
-        '-e', 'AWS_SECRET_ACCESS_KEY='+AWS_SECRET_ACCESS_KEY,
+        '-e', 'AWS_ACCESS_KEY_ID='+aws_access_key_id,
+        '-e', 'AWS_SECRET_ACCESS_KEY='+aws_secret_access_key,
         '-e', 'AWS_REGION='+region,
         '--net=host', 'py.test', 'py.test',
         '-vv', ci_flags, '/integration_test.py']
@@ -296,6 +294,7 @@ def make_vpc(use_bare_os=False):
 
 def check_environment():
     """Test uses environment variables to play nicely with TeamCity config templates
+    Grab all the environment variables here to avoid setting params all over
 
     Returns:
         object: generic object used for cleanly passing options through the test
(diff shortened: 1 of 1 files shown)


mellenburg:
@cmaloney, @spahl  this is ready for review

spahl:
please split into 3 commits.

mellenburg:
@spahl done

test_util/test_installer_ccm.py
```diff
@@ -214,15 +211,10 @@ def integration_test(
         region: string indicating AWS region in which cluster is running
         agent_list: string of comma separated agent addresses
         registry_host: string for address where marathon can pull test app
-        test_minuteman: if set to True then test for minuteman service
         test_dns_search: if set to True, test for deployed mesos DNS app
         ci_flags: optional additional string to be passed to test
 
     """
-    marker_args = '-m "not minuteman"'
```
cmaloney: should also remove the marker from integration_test.py

test_util/test_installer_ccm.py
```diff
@@ -339,6 +343,13 @@ def check_environment():
     options.ci_flags = os.getenv('CI_FLAGS', '')
     options.aws_access_key_id = os.environ.get('AWS_ACCESS_KEY_ID', '')
     options.aws_secret_access_key = os.environ.get('AWS_SECRET_ACCESS_KEY', '')
+    options.num_masters = int(os.environ.get('NUM_MASTERS', '1'))
+    options.num_agents = int(os.environ.get('NUM_AGENTS', '2'))
+
+    if options.host_list:
+        total_hosts = len(options.host_list)
+        assert total_hosts == 1 + options.num_masters + options.num_agents
```
cmaloney: comment string on this assertion?

test_util/test_installer_ccm.py
```diff
@@ -350,7 +361,9 @@ def main():
     vpc = None  # Set if the test owns the VPC
 
     if options.host_list is None:
-        vpc = make_vpc(use_bare_os=options.test_install_prereqs)
+        vpc = make_vpc(
+            count=1+options.num_masters+options.num_agents,
```
cmaloney: Note we'll have to revisit this when we add public agents

test_util/test_installer_ccm.py
```diff
@@ -257,13 +257,16 @@ def prep_hosts(ssh_runner, registry):
...
-def make_vpc(use_bare_os=False):
+def make_vpc(count, use_bare_os=False):
     """uses CCM to provision a test VPC of minimal size (3).
 
     Args:
+        count: number of instances to spin up
         use_bare_os: if True, vanilla AMI is used. If False, custom AMI is used
             with much faster prereq satisfaction time
     """
+    assert count >= 4, ('Cluster must be at least 4 hosts: '
```
cmaloney: nit: This logic is a bit spread all over the place. Might make sense to centralize the notion of "I need a block of hosts" and have that class own initializing from either environment variables or building a new CCM VPC.

test_util/test_installer_ccm.py
```diff
@@ -381,8 +394,17 @@ def make_runner(host_list):
     print("VPC hosts: {}".format(host_list))
     # use first node as bootstrap node, second node as master, all others as agents
     registry_host = local_ip[host_list[0]]
-    master_list = [local_ip[_] for _ in host_list[1:2]]
-    agent_list = [local_ip[_] for _ in host_list[2:]]
+    master_list = []
+    agent_list = []
+    print('Test/registry host public/private IP: ' + host_list[0] + '/' + registry_host)
+    for i in range(1, (options.num_masters+1)):
```
cmaloney: master_list = host_list[1:options.num_masters+1]
agent_list = host_list[options.num_masters+1:options.num_agents+1]

print("Master IPs:", master_list)
print("Agent IPs:", agent_list)

mellenburg: The goal was to print out the hosts in such a way that one could easily parse the local and public IP for a single node

cmaloney: I don't see how that new formatting is any more difficult than the old one to parse, it's still arbitrary / hand written. If we want machine parsable, then should do something like `print(json.dumps({'master_list': master_list, 'agent_list': agent_list})`

test_util/test_installer_ccm.py
mellenburg: I dont see any usage of mark in integration_test.py currently?

test_util/test_installer_ccm.py
mellenburg: In that case, the code will no longer be cleanup, so I'll take it out of here for now and put it in another PR so we can get the other two basic commits in

test_util/test_installer_ccm.py
cmaloney: Apparently when minuteman was made non-conditional, that was removed but setting the marker wasn't. So yea, nothing to do there, my bad

mellenburg:
@cmaloney this has been stripped down so it only touches the minuteman and aws values cleanup. Can it ship?

cmaloney:
:ship:. I'll add to the next train.

PR #81, opened 2016-05-19
Expand private agent ports

sargun:


sargun:
@spahl

cmaloney:
NAK from landing. Any software launched on DC/OS non-root running, if it's not very aware about the fact it could get "ports < 1024" will fail to launch when randomly assigned ports. There are also arbitrary on-host user software which could be listening to these ports, and multiple DC/OS Frameworks / Services which just claim these without port reservations.

sargun:
@cmaloney Marathon has:
```
--local_port_max (Optional. Default: 20000): Max port number to use when dynamically assigning globally unique service ports to apps. If you assign your service port statically in your app definition, it does not have to be in this range.
--local_port_min (Optional. Default: 10000): Min port number to use when dynamically assigning globally unique service ports to apps. If you assign your service port statically in your app definition, it does not have to be in this range.
```

The infinity frameworks all have the user choose a port, except Kafka, and they've offered to fix that.

sargun:
I would prefer to merge early, and merge often. It's way better than our mega merges at the end of the release. It makes it impossible to track down the bug that caused the issue. 
CC: @spahl

cmaloney:
This has technical reasons  it's not ready to ship, and I don't think should land in this release at all.

sargun:
@cmaloney elucidate.

PR #85, opened 2016-05-19
Move bootstrap to be cached in the cache/ directory

cmaloney:
Results in cleaner build directories, less .gitignore being necessary.


spahl:
Looks good

PR #86, opened 2016-05-19
[WIP] Test --modules_dir flag

karya0:


packages/logrotate/build
```diff
@@ -10,6 +10,12 @@ make
 make install
 
 
+# Create the module file
+module="$PKG_PATH/dcos.target.wants_slave/logrotate.json"
```
cmaloney: I'm guessing you want this to be `$PKG_PATH/etc/mesos-slave-modules/` so it comes out where you want

karya0: Yeah, I was thinking about that but then got confused when I saw several `wants_` things in other build files and thought maybe there is some pkgpanda magic going on here :). I'll fix it and rerun the test suite. (There is another mesos build issue that was just resolved in master, so will update my branch.)

cmaloney: dcos.target.wants is for systemd units which should be started. `etc` for config.

The `_master`, `_slave` and friends are pkgpanda stuff, what that means is "only use the files in this folder if the machine has that particular role". Primarily used to select a set of services to only start on masters, or only start on slaves, rather than to start on every machine.

cmaloney:
Will come back with Mesos 1.0 PRs (the patches landed in mesos master)

PR #87, opened 2016-05-20
MOD enabled http_callback

jstabenow:
http_callback enabled for using the EventBus on Marathon-LB


cmaloney:
Any particular reason not to just use `sse` mode of marathon-lb?

jstabenow:
damn. you are right. sorry! ;-) I always forget the event-stream (realy the better option). I close this. nice sunday

PR #91, opened 2016-05-24
Add Pkgpanda HTTP API

branden:
This adds an HTTP API for [managing DC/OS components](https://github.com/dcos/dcos/tree/master/pkgpanda) that runs on every node in a cluster.

TODO
- [x]  Systemd unit for API server (gunicorn)
- [x]  Test package activation manually
- [x]  Better testing in general. Error conditions, etc.
- [x] PUT /pkgpanda/active/ without killing pkgpanda-api before the upgrade is complete


cmaloney:
Can we pull the first two commits out of this, one at a time, into separate PRs to look at and land? They seem fairly independent and reasonable to do.

pkgpanda/http/config.py
```diff
@@ -0,0 +1,12 @@
+DCOS_ROOT = '/opt/mesosphere/'
```
cmaloney: any way we can share getting / choosing these defaults sanely with the cli code?

branden: This makes the `pkgpanda.cli` docstring less useful, but it's the best I can come up with: https://github.com/dcos/dcos/pull/91/commits/f7d7a475604178448096812acd5be85d1c23467a

branden:
@spahl @cmaloney I still need to add some unit tests that cover error conditions, but this is otherwise ready for review.

branden:
This PR relies on the agent adminrouter from #292.

branden:
Adminrouter PR: https://github.com/dcos/adminrouter/pull/10

packages/pkgpanda-api/build
```diff
@@ -0,0 +1,21 @@
...
+source /opt/mesosphere/environment.export
+
+export WORK_DIR=/run/dcos/pkgpanda-api/
+export STOP_TIMEOUT=30  # seconds
+
+pip install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/gunicorn
+
+mkdir -p $PKG_PATH/{etc,bin}
+envsubst '$WORK_DIR' < /pkg/extra/pkgpanda-api.conf.py > $PKG_PATH/etc/pkgpanda-api.conf.py
+envsubst '$WORK_DIR $STOP_TIMEOUT' < /pkg/extra/pkgpanda-api > $PKG_PATH/bin/pkgpanda-api
```
cmaloney: This seems like a lot of work to make this substitutable... Can we just put it in the two places?

packages/pkgpanda-api/extra/dcos-pkgpanda-api.service
```diff
@@ -0,0 +1,21 @@
+[Unit]
+Description=Pkgpanda API: Package Management Service
+After=dcos-gen-resolvconf.service
```
cmaloney: Before / After don't do anything useful.

packages/pkgpanda-api/extra/dcos-pkgpanda-api.service
```diff
@@ -0,0 +1,21 @@
...
+
+
+[Service]
+Type=simple
+StandardOutput=journal
+StandardError=journal
+Restart=always
```
cmaloney: This needs restart delay stuff, see the other systemd services

packages/pkgpanda-api/extra/dcos-pkgpanda-api.service
```diff
@@ -0,0 +1,21 @@
...
+Environment=PKGPANDA_HTTP_CONFIG=${PKG_PATH}/etc/pkgpanda-api.conf.py
+EnvironmentFile=/opt/mesosphere/environment
+ExecStart=${PKG_PATH}/bin/pkgpanda-api
+
+# Allow pkgpanda-api to upgrade itself gracefully.
+# When a pkgpanda-api worker stops pkgpanda-api, the master process receives
+# SIGTERM and waits for the worker to finish processing the upgrade request
+# before exiting.
+KillMode=process
```
cmaloney: This should definitely be cgroup or mixed. We don't want lingering processes around. `KillMode=mixed` looks like it has exactly the behavior you want, without the KillSignal stuff.

packages/pkgpanda-api/extra/dcos-pkgpanda-api.socket
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Pkgpanda API socket: Package Management Service Socket
+
+[Socket]
+ListenStream=/run/dcos/pkgpanda-api.sock
+
+[Install]
+WantedBy=dcos-pkgpanda-api.service
```
cmaloney: This WantedBy shouldn't be here / isn't right.

packages/pkgpanda-api/extra/pkgpanda-api
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+mkdir -p $WORK_DIR
+
+gunicorn --worker-class=sync \
```
cmaloney: Can we please just put this in the systemd file as a bash command?

Doing wrapper scripts like this messes up some process tracking (Esp. since we aren't doing an exec here).

packages/pkgpanda-api/build
```diff
@@ -0,0 +1,21 @@
...
+
+source /opt/mesosphere/environment.export
+
+export WORK_DIR=/run/dcos/pkgpanda-api/
+export STOP_TIMEOUT=30  # seconds
+
+pip install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/gunicorn
+
+mkdir -p $PKG_PATH/{etc,bin}
+envsubst '$WORK_DIR' < /pkg/extra/pkgpanda-api.conf.py > $PKG_PATH/etc/pkgpanda-api.conf.py
```
cmaloney: Can we just use a HEREDOC for this one?

Practically it feels like we should just be able to drop this entirely / have it be the default config value.

pkgpanda/cli.py
```diff
@@ -124,7 +124,14 @@ def run_checks(checks, install, repository):
...
 
 def main():
-    arguments = docopt(__doc__, version="Pkpganda Package Manager")
+    arguments = docopt(
+        __doc__.format(
+            default_config_dir=constants.config_dir,
+            default_root=constants.install_root,
+            default_repository=constants.repository_base,
+        ),
+        version="Pkpganda Package Manager",
```
cmaloney: `version=0.0.1` or left out. The old version string was a bug.

pkgpanda/docs/http-swagger.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+
+definitions:
+
+  PackageId:
+    description: A unique pkgpanda package ID. Composed of a name and a version separated by `--`.
+    type: string
+    example: 'mesos--abcdef'
+    # Pattern adapted from pkgpanda/__init__.py.
+    pattern: '/^[a-zA-Z0-9@_+]([a-zA-Z0-9@._+\-]*[a-zA-Z0-9@._+])?--[a-zA-Z0-9@_+:.]+$/'
```
cmaloney: There is a much tighter pattern this is required to follow. See: https://github.com/dcos/dcos/blob/master/pkgpanda/__init__.py#L83

pkgpanda/docs/http-swagger.yaml
```diff
@@ -0,0 +1,116 @@
...
+        - application/json
+      responses:
+        '200':
+          description: A list of the packages on this node.
+          schema:
+            '$ref': '#/definitions/PackageIdArray'
+
+  /{package-id}:
+    get:
+      summary: Get a package in the node's pkgpanda repository.
```
cmaloney: "Get basic information about a package in the node's package repository"

pkgpanda/docs/http-swagger.yaml
```diff
@@ -0,0 +1,116 @@
...
+      parameters:
+        - '$ref': '#/parameters/PackageId'
+      responses:
+        '204':
+          description: The package is present in this node's pkgpanda repository.
+        '404':
+          description: The package is not present in this node's pkgpanda repository.
+    post:
+      summary: Fetch a package from the bootstrap URL configured at /etc/mesosphere/setup-flags/repository-url on this node.
```
cmaloney: "fetch" should have a configurable url which is postable to it. The setup flags is only used by `pkgpanda setup` nothing else should use it.

cmaloney: The fetch could (in theory) take a while for the network, is there a danger of the http connection timing out before there is a response?

pkgpanda/docs/http-swagger.yaml
```diff
@@ -0,0 +1,116 @@
...
+          required: true
+          schema:
+            '$ref': '#/definitions/PackageIdArray'
+      responses:
+        '204':
+          description: The packages in the request body have been activated. (This does not necessarily mean that their services started successfully.)
+        '409':
+          description: One or more packages in the request body are not present in this node's pkgpanda repository.
+
+  /active/{package-id}:
```
cmaloney: `is_active`? Active is sort of ambiguous what it is going to do.

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+from pkgpanda import Install, Repository, actions
+
+empty_response = ('', http.client.NO_CONTENT)
+
+
+def package_index(packages, package_id=None):
+    if package_id is None:
+        return json.dumps(sorted(packages))
+    elif package_id in packages:
+        return empty_response
```
cmaloney: Would it be better to return `{"found": "true"}`?

pkgpanda/http/config.py
```diff
@@ -0,0 +1,15 @@
+from pkgpanda import constants
+
+
+DCOS_ROOT = constants.install_root
+DCOS_CONFIG_DIR = constants.config_dir
+DCOS_REPO_DIR = constants.repository_base
+DCOS_ROOTED_SYSTEMD = False
+
+try:
+    with open('/etc/mesosphere/setup-flags/repository-url') as f:
```
cmaloney: setup-flags shouldn't be used here, it's only for `pkgpanda setup` / not guaranteed to be what we want in time. (In fact, I see it highly likely we want a different url when actually installing).

Practically we're going to want to move the packages into a in-cluster repository, then probably use the mesos-dns name to fetch them from there to individual hosts.

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+    if not isinstance(request.json, list):
+        abort(http.client.BAD_REQUEST)
+    # All packages in the request body must be present in the repository.
+    if not (set(request.json) <= set(current_app.repository.list())):
+        abort(http.client.CONFLICT)
+
+    # This will stop all DC/OS services, including this app. Use a web server
+    # that supports graceful shutdown to ensure that activation is completed
+    # and a response is returned.
+    actions.activate_packages(
```
cmaloney: This can return various errors / throw various exceptions, should make sure to catch and return them as proper HTTP status codes if we're running under an HTTP server that is graceful.

pkgpanda/http/__init__.py
branden: Isn't that what the HTTP status is supposed to indicate?

pkgpanda/http/__init__.py
branden: For now I'm relying on the server to catch unhandled exceptions and serve 500 responses. We can add status codes for specific errors as the client needs them. Right now I have no plans to have this API's client behave differently under different types of activation failure.

pkgpanda/docs/http-swagger.yaml
branden: This endpoint responds with a 204 No Content on success, so that's not accurate.

pkgpanda/docs/http-swagger.yaml
branden: This pattern reflects the code you linked.

pkgpanda/docs/http-swagger.yaml
cmaloney: Right now the summary is really hard to parse / digest as someone who knows the internals of pkgpanda and what it's supposed to be doing... Needs some rewording. Maybe talk to the docs team.

pkgpanda/http/__init__.py
cmaloney: Depends on what model you are going after. Some people see the HTTP status codes as a "container" for the internal messages. So that the service processed your request and can return properly gets you a 200, then we're explicit in the body about what it's about.

Others say we're going to promote all internal errors to whatever is the most useful external error.

In this case, I think it's good to be explicit with the extra message / data inside the body in addition to the "200" response. Doesn't break anything, makes it so that if we ever decide to expand / change the endpoint in the future we have a good upgrade path.

pkgpanda/http/__init__.py
cmaloney: In the upgrade service we're going to want to be able to see why a node failed to upgrade. If we aren't catching the exceptions, and returning them as JSON blobs, then we're going to have a bad time debugging. (Oh it failed, need to ssh to the host to get absolutely any idea why).

branden: We sure will!

packages/pkgpanda-api/extra/dcos-pkgpanda-api.service
branden: Added just below.

branden:
@cmaloney This is ready for another look.

packages/python-gunicorn/buildinfo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "requires": ["python"],
+  "docker": "dcos-builder",
+  "single_source": {
+    "kind": "url_extract",
+    "url": "https://pypi.python.org/packages/source/g/gunicorn/gunicorn-19.4.1.tar.gz",
```
jgehrcke: Let's please use 19.4.5. or 19.6.0.

packages/pkgpanda-api/buildinfo.json
```diff
@@ -0,0 +1,7 @@
+{
+  "requires": [
+      "python",
+      "python-gunicorn"
+  ],
+  "docker": "dcos-builder"
```
cmaloney: nit: dcos-builder is the default, specifying it doesn't do anything.

packages/pkgpanda-api/build
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+mkdir -p $PKG_PATH/etc/
+
+cat <<EOF > $PKG_PATH/etc/pkgpanda-api.conf.py
```
cmaloney: nit: What value is having this config file and passing it through giving us currently? Seems like a lot of work which could just be replaced by a constant for now.

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    return jsonify({'error': message})
+
+
+app = Flask(__name__)
+app.config.from_object('pkgpanda.http.config')
+app.config.from_envvar('PKGPANDA_HTTP_CONFIG', silent=True)
+
+
+@app.errorhandler(Exception)
+def uncaught_exception_handler(exc):
```
cmaloney: The CLI has handling for ValidationError, PackageError to top level error messages as they're relatively expected during the normal course of operations. Can we make the flask app return them as proper JSON blob error codes?

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+        repository_url,
+        package_id,
+        current_app.config['WORK_DIR'])
+
+    return empty_response
+
+
+@app.route('/<package_id>', methods=['DELETE'])
+def remove_package(package_id):
+    if not current_app.repository.has_package(package_id):
```
cmaloney: Why isn't just calling actions.remove_package and handling the exceptions enough here? Why do we need to do this manual (in theory redundant) checking?

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    if not isinstance(request.json, list):
+        return (
+            error_response(
+                'Request body must be a json array of package IDs.'
+            ),
+            http.client.BAD_REQUEST,
+        )
+    if not (set(request.json) <= set(current_app.repository.list())):
+        return (
+            error_response('Not all packages in the request are present.'),
```
cmaloney: Can we include some info about the two sets of packages here? This will be very hard to debug without it (just including both lists would be enough).

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    current_app.repository = Repository(
+        current_app.config['DCOS_REPO_DIR'])
+
+
+@app.route('/', methods=['GET'])
+def get_package_list():
+    return package_listing_response(current_app.repository.list())
+
+
+@app.route('/<package_id>', methods=['GET'])
```
cmaloney: Can we have this and the POST / fetch_package be `/packages/<package_id>` rather than this?

As this stands, it makes it hard to know whether a request is going to go into this route, or things like `/active`. Effectively we've merged those key spaces and made it possible for people to do something ambiguous. I'd much rather keep them separate.

pkgpanda/integration-tests/test_http.py
```diff
@@ -0,0 +1,153 @@
...
+
+    response = client.get('/mesos--0.22.0')
+    assert response.status_code == 200
+    assert json.loads(response.data.decode('utf-8')) == {
+        'id': 'mesos--0.22.0',
+        'name': 'mesos',
+        'version': '0.22.0',
+    }
+
+    response = client.get('/nonexistent-package--fakeversion')
```
cmaloney: Can we test a invalid package name / version number? (`client.get('/!@#*')`, `client.get('/fooASDF')`

packages/pkgpanda-api/build
branden: It lets `pkgpanda.http.app.config['WORK_DIR']` default to a location under `/tmp`, which is a little friendlier for local hacking than `/run`.

pkgpanda/http/__init__.py
branden: `remove_package()` silently succeeds if the package doesn't exist 😢 https://github.com/dcos/dcos/blob/db8b5678caf6a958b1926a50f65ce9a37adf96b6/pkgpanda/__init__.py#L387 I added this check because I didn't want to change CLI behavior as part of this PR. I guess I could add the exception, remove this check, and have the CLI catch and ignore it?

cmaloney: Succeeding doesn't sound like a bad thing in the remove non-existent case. The remove did succeed, in that the end state (package is not on host) has been reached. That's where the original CLI design came from.

In the context of users of the API, having remove be successful if it removes or the package already exists, makes it into a bit of a CRDT / completely repeatable in parallel without error operation. If two processes in parallel remove the same package, they'll both always succeed (rather than one getting a NOT_FOUND and having to decide that's just fine). Same goes for a co-ordination node.

Technically not found is slightly more information, although I'd rather just expose it as data inside of success (so people can operate on it if they want / be strict), but making it another error case just means people have to write more code to deal with it without a lot of value.

branden: Whether a 404 is acceptable in response to a DELETE seems like a reasonable decision for a client to make. One of my goals with this API has been to expose packages just as a resource and leave as much application logic as possible up to the client.

spahl:
Is this ready to go?

branden:
@cmaloney @spahl Yeah this is ready for another look.

pkgpanda/cli.py
```diff
@@ -123,7 +123,13 @@ def run_checks(checks, install, repository):
 
 
 def main():
-    arguments = docopt(__doc__, version="Pkgpanda Package Manager")
+    arguments = docopt(
+        __doc__.format(
+            default_config_dir=constants.config_dir,
+            default_root=constants.install_root,
+            default_repository=constants.repository_base,
+        ),
```
orsenthil: You could add a version string. I would suggest a semver versioning of this cli.

pkgpanda/cli.py
```diff
@@ -185,7 +191,10 @@ def main():
 
         if arguments['remove']:
             for package_id in arguments['<id>']:
-                actions.remove_package(install, repository, package_id)
+                try:
+                    actions.remove_package(install, repository, package_id)
+                except PackageNotFound:
+                    pass
```
orsenthil: `log.info("package {id} not found".format(id=package_id)` ?

pkgpanda/cli.py
branden: Prior discussion here: https://github.com/dcos/dcos/pull/91#discussion_r71184369

I went ahead and removed it since I didn't want to get into a discussion over what the version string ought to be in this PR.

orsenthil: ACK.

pkgpanda/__init__.py
```diff
@@ -355,6 +359,9 @@ def integrity_check(self):
     # If the package is already in the repository does a no-op and returns false.
     # Returns true otherwise.
     def add(self, fetcher, id, warn_added=True):
+        # Validate the package id.
+        PackageId(id)
```
orsenthil: Just sending it through the constructor only to validate (even though supported by the comment) does not seem ideal.

* Better would be an introduction of a static method, that `PackageId.isvalid(id)` and use is here and in line 330.

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+            error_response('Package {} not found.'.format(package_id)),
+            http.client.NOT_FOUND,
+        )
+    except PackageError:
+        error_message = 'Unable to load package {}.'.format(package_id)
+        logging.exception(error_message)
+        response = (
+            error_response(error_message),
+            http.client.INTERNAL_SERVER_ERROR,
+        )
```
orsenthil: These errors `ValidationError, PackageNotFound and PackageError` could be made to accept a an argument, package_id, which and the error message could be contained within the class.  

So the exception could be like

```
except (ValidationError, PackageNotFound, PackageError) as e:
      response = (e.msg, e.http_response)
```

The same pattern could be applied to other places where different types of Errors are getting caught. 

Furthermore, I would suggest the `package_response` can return a `namedtuple` so that parts of it can referenced like `response.msg, response.http_code` .

This way, line 176 below would read like `response.http_code != http.client.OK`.

pkgpanda/http/__init__.py
```diff
@@ -0,0 +1,229 @@
...
+@app.route('/active/', methods=['GET'])
+def get_active_package_list():
+    return package_listing_response(current_app.install.get_active())
+
+
+@app.route('/active/<package_id>', methods=['GET'])
+def get_active_package(package_id):
+    response = package_response(package_id, current_app.repository)
+
+    # Return early if there was an error loading the package.
```
orsenthil: The comments at like 175 and 179 seem redundant to me. You could consider killing it.

pkgpanda/http/__init__.py
branden: I think adding HTTP error responses to exception classes would be a needless coupling. API error responses are relevant only to this module, so they probably ought to be defined here, not as part of exception classes that aren't otherwise HTTP-aware. And we can't assume a 1:1 relationship between exception classes and error responses -- the error response may be different based on context that's only available to the caller. For example there are a couple cases here where we send a different error in response to `ValidationError` and `PackageError` based on context.

👍  on namedtuple responses, I'll add that.

orsenthil: Sure. Agree with your response. My intention with the first review comment was to avoid the repetition of multiple except clauses pattern that are repeated twice in this module. Introducing a new API based exception subclassing the parent could be done but I am with as it is if you feel that status quo is best.

branden: If you're referring to the duplicated 404 in response to `PackageNotFound` here and in `remove_package()`, I can DRY that up by generating the response from a `package_404()` function. We'd still want the view to decide which error to send to the client, since different views may send different errors in response to that exception, or possibly (certainly in the case of the CLI) no error at all.

branden: I'm gonna use this instead of `namedtuple`: http://flask.pocoo.org/docs/0.11/api/#flask.Flask.make_response Saves a bunch of code and allows us to refer to the response status code by a readable name.

orsenthil: I was referring to the pattern at Line 25 and Line 145. But your call is alright.
And +1 to Flask.make_response. thank you.

branden: https://github.com/dcos/dcos/pull/91/commits/4180f9740fe077533d670170b362845b8b1a869b
https://github.com/dcos/dcos/pull/91/commits/b6114753f6ea8a05cd1b193c44de29f65dce8984

orsenthil:
Changes look good to me. ship-it from me for the python code and http api.

spahl:
:ship:

PR #92, opened 2016-05-24
WIP Secure registry for integration tests

mellenburg:


mellenburg:
I'm going with the alternative to this approach that is included in https://github.com/dcos/dcos/pull/74

I.E. use localhost as registry for all agents and then just forward their local host registry port to the designated registry with ssh tunnel

PR #93, opened 2016-05-24
Fix cached bootstrap download location

cmaloney:


cmaloney:
Tested this fixes the bug we saw in CI

PR #96, opened 2016-05-24
add prepare status json

darkonie:
Add a function to run before any command from a chain start executing. This
allows to prepare status json. For JsonDelegate we want to list all nodes
from a config right away. The status for a node will be changed as installer
proceeds.

### DIFF
--- a/ext/dcos-installer/dcos_installer/cli.py
+++ b/ext/dcos-installer/dcos_installer/cli.py
@@ -21,6 +21,9 @@ class CliDelegate(AbstractSSHLibDelegate):
     def on_done(self, name, result, host_status=None):
         print_header('STAGE {}'.format(name))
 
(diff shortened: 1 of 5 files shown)


darkonie:
@cmaloney this is ready for a code review

cmaloney:
Generally looks reasonable for what it's doing. Doesn't feel particularly clean, but I think to get more cleanup we'd have to do a much deeper refactoring / rework.

I'm not a big fan of calling it "waiting" for the state. I'd prefer a name like "unstarted" which gives a clearer indication that we're not waiting for some condition, rather it just hasn't been done yet.

darkonie:
@cmaloney status changed from `waiting` to `unstarted`

cmaloney:
:ship: 

I'll merge once it's checks pass.

PR #98, opened 2016-05-24
Add a build_local script to help doing dev work

cmaloney:
Sets up a python virtual env and uses sane commands to do a full dcos build

### DIFF
--- a/build_local.sh
+++ b/build_local.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Simple helper script to do a full local  build
+
+set -x
+set -o errexit -o pipefail
+
+# Fail quickly if docker isn't working / up
+docker ps
+
+# Cleanup from previous build
+rm -rf build/venv
+
+# Force Python stdout/err to be unbuffered.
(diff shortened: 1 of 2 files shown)


build_local.sh
```diff
@@ -0,0 +1,32 @@
...
+# Cleanup from previos build
+rm -rf build/venv
+
+# Force Python stdout/err to be unbuffered.
+export PYTHONUNBUFFERED="notemtpy"
+
+pyvenv build/env
+. build/env/bin/activate
+
+cat <<EOF > dcos-release.config.yaml
```
cmaloney: Should test if this file exists early, and if so not do this.

build_local.sh
```diff
@@ -0,0 +1,32 @@
...
+# Simple helper script to do a full local  build
+
+set -x
+set -o errexit -o pipefail
+
+# Fail quickly if docker isn't working / up
+docker ps
+
+
+# Cleanup from previos build
```
cmaloney: extra newline. + `previous`

PR #101, opened 2016-05-25
WIP / Prototype: Add agents to masters

cmaloney:
This isn't expected to land soon, but we can use it to make a concrete list of reasons we can't land it, and track landing those.

TODO / known missing items:
 - [ ] UI to hide them when not in use (otherwise you can't reach full utilization)
 - [ ] 3dt
 - [ ] Documentation on how to launch tasks on them (requires another mesos framework currently)

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -126,8 +126,8 @@ class Cluster:
             return False
         data = r.json()
         num_slaves = len([x['hostname'] for x in data['slaves']])
-        # For single node setup there is only one slave node:
-        min_slaves = min(len(self.slaves), 2)
+        # There is one slave on every master, as well as slaves on agents.
+        min_slaves = len(self.slaves) + len(self.masters)
         if num_slaves >= min_slaves:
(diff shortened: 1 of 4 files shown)


test_util/integration_test.py
```diff
@@ -126,8 +126,8 @@ def _wait_for_slaves_to_join(self):
             return False
         data = r.json()
         num_slaves = len([x['hostname'] for x in data['slaves']])
-        # For single node setup there is only one slave node:
-        min_slaves = min(len(self.slaves), 2)
+        # There is one slave on every master, as well as slaves on agents.
+        min_slaves = len(self.slaves) + len(self.masters)
```
vespian: min(len(self.slaves)+len(self.slaves), 2+len(self.masters)) ?

cmaloney: The single node master + agent case we've gotten rid of, so there isn't the odd extra case needed. Might be worthwhile to globally fail early if not give at least two slaves, but that's a different condition

spahl:
Closing right now, likely to come back albeit maybe differently

PR #104, opened 2016-05-25
Pkgpanda: Separate package management actions from CLI

branden:
See #91

### DIFF
--- a/pkgpanda/actions.py
+++ b/pkgpanda/actions.py
@@ -28,7 +28,8 @@ def activate_packages(install, repository, package_ids, systemd, block_systemd):
     assert isinstance(package_ids, collections.Sequence)
     try:
         install.activate(repository.load_packages(package_ids))
-        _start_dcos_target(systemd, block_systemd)
+        if systemd:
+            _start_dcos_target(block_systemd)
     except ValidationError as ex:
         print("Validation Error: {0}".format(ex))
         sys.exit(1)
@@ -173,7 +174,7 @@ def setup(install, repository):
         # Enable dcos.target only after we have populated it to prevent starting
         # up stuff inside of it before we activate the new set of packages.
         if install.manage_systemd:
-            _start_dcos_target(True, block_systemd=True)
+            _start_dcos_target(block_systemd=True)
         os.remove(bootstrap_path)
 
     # Check for /opt/mesosphere/install_progress. If found, recover the partial
@@ -184,12 +185,11 @@ def setup(install, repository):
             print("No recovery performed: {}".format(msg))
 
 
-def _start_dcos_target(systemd, block_systemd):
-    if systemd:
-        no_block = [] if block_systemd else ["--no-block"]
-        check_call(["systemctl", "daemon-reload"])
-        check_call(["systemctl", "enable", "dcos.target", '--no-reload'])
-        check_call(["systemctl", "start", "dcos.target"] + no_block)
+def _start_dcos_target(block_systemd):
+    no_block = [] if block_systemd else ["--no-block"]
+    check_call(["systemctl", "daemon-reload"])
+    check_call(["systemctl", "enable", "dcos.target", '--no-reload'])
+    check_call(["systemctl", "start", "dcos.target"] + no_block)
 
 
 def _do_bootstrap(install, repository):
--- a/pkgpanda/actions.py
+++ b/pkgpanda/actions.py
@@ -15,20 +15,20 @@ WantedBy=multi-user.target
 """
 
 
-def activate_packages(install, repository, package_ids, no_systemd=False, noblock_systemd=True):
+def activate_packages(install, repository, package_ids, systemd=True, block_systemd=False):
     """Replace the active package set with package_ids.
 
     install: pkgpanda.Install
     repository: pkgpanda.Repository
     package_ids: sequence of package IDs to activate
-    no_systemd: don't start/stop systemd services (default: False)
-    noblock_systemd: if not no_systemd, don't block waiting for systemd services to come up (default: True)
+    systemd: start/stop systemd services (default: True)
+    block_systemd: if systemd, block waiting for systemd services to come up (default: False)
 
     """
     assert isinstance(package_ids, collections.Sequence)
     try:
         install.activate(repository.load_packages(package_ids))
-        _start_dcos_target(no_systemd, noblock_systemd)
+        _start_dcos_target(systemd, block_systemd)
     except ValidationError as ex:
         print("Validation Error: {0}".format(ex))
         sys.exit(1)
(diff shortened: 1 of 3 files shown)


pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    # Check for /opt/mesosphere/install_progress. If found, recover the partial
+    # update.
+    if os.path.exists("/opt/mesosphere/install_progress"):
+        took_action, msg = install.recover_swap_active()
+        if not took_action:
+            print("No recovery performed: {}".format(msg))
+
+
+def _start_dcos_target(block_systemd):
+    no_block = [] if block_systemd else ["--no-block"]
```
cmaloney: nit: The change from noblock_systemd -> block_systemd is a semantic change which is mixed in with a whole lot of code motion here. Looks like all the callsites are all updated properly, but would be nice to have first a pure code moving commit. Followed by a second commit that fixes up code issues (like `start_dcos_target(False, ['--no-block'])` which is definitely wrong code / a bug I'd committed)

pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    """Replace the active package set with package_ids.
+
+    install: pkgpanda.Install
+    repository: pkgpanda.Repository
+    package_ids: sequence of package IDs to activate
+    systemd: start/stop systemd services (default: True)
+    block_systemd: if systemd, block waiting for systemd services to come up (default: False)
+
+    """
+    assert isinstance(package_ids, collections.Sequence)
```
cmaloney: why are we dropping from the list requirement to a sequence here? It doesn't look like the code paths this ends up hitting particularly care (they just iterate over the package_ids given). The one caller of this function only ever passes a list though, so if we're specifying tightly, might as well specify as tightly as possible.

pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    if new_id.name not in packages_by_name:
+        print("ERROR: No package with name {} currently active to swap with.".format(new_id.name))
+
+    packages_by_name[new_id.name] = new_id
+    new_active = list(map(str, packages_by_name.values()))
+    # Activate with the new package name
+    activate_packages(install, repository, new_active, systemd, block_systemd)
+
+
+def fetch_packages(repository, repository_url, package_ids, work_dir=None):
```
cmaloney: This intentionally use to only do a single package. If you do a list of packages, there may be a failure somewhere in the middle, and whatever is controlling the API doesn't have a good notion of what is / isn't on the host.

Could build the full list logic for the co-ordination layer and this. Or just make it do one package at a time via the API, in which case it takes more calls but we know what the state is (And can optimize if we find that to be a bottleneck)

pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    """Fetch package_ids from repository_url into repository.
+
+    repository: pkgpanda.Repository
+    repository_url: URL for remote package repository
+    package_ids: sequence of package IDs to fetch
+    work_dir: location for temporary files (default: current working dir)
+
+    """
+    assert isinstance(package_ids, collections.Sequence)
+    work_dir = work_dir or os.getcwd()
```
cmaloney: If this is going to be a library, we can't / don't want to assume os.getcwd(). work_dir needs to be passed in as a parameter.

cmaloney: note the work_dir is used iff people supply relative `file://` URIs. Absolute file URIs begin with a `file:///`

pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    # TODO(cmaloney): I guarantee there is a better way to write this and
+    # I've written the same logic before...
+    packages_by_name = dict()
+    for id_str in active:
+        pkg_id = PackageId(id_str)
+        packages_by_name[pkg_id.name] = pkg_id
+
+    new_id = PackageId(package_id)
+    if new_id.name not in packages_by_name:
+        print("ERROR: No package with name {} currently active to swap with.".format(new_id.name))
```
cmaloney: not your bug, but there should be a sys.exit(1) here.

pkgpanda/actions.py
```diff
@@ -0,0 +1,293 @@
...
+    # Validate the package id
+    PackageId(pkg_id)
+
+    def fetch(_, target):
+        extract_tarball(package_filename, target)
+
+    repository.add(fetch, pkg_id)
+
+
+def remove_packages(install, repository, package_ids):
```
cmaloney: Same thing applies for this as fetch. This intentionally only took a single package ID string.

pkgpanda/actions.py
branden: >  It doesn't look like the code paths this ends up hitting particularly care (they just iterate over the package_ids given).

That was my reasoning: the code doesn't actually need a list, it just needs an iterable sequence. I saw the list assertion as over-specifying, but I can put it back if you feel otherwise.

pkgpanda/actions.py
cmaloney: slight aside: Looks like at some point I did make it so `fetch` could take a list.  (Note the ... in   `pkgpanda fetch --repository-url=<url> <id>... [options]`[0] ). That said, we should make the API which is used for remote orchestration definitely only have "atomic" operations.

[0] https://github.com/dcos/dcos/blob/master/pkgpanda/cli.py#L7

pkgpanda/actions.py
branden: This is now in its own commit: https://github.com/dcos/dcos/pull/104/commits/22908f3ed1a8f4d1153bc9df30e5a05d3e5abfdd

pkgpanda/actions.py
branden: https://github.com/dcos/dcos/pull/104/commits/c4242a9a51e2e629ff0ede25497ecd47854c5f2c

pkgpanda/actions.py
```diff
@@ -15,28 +15,28 @@
...
+        _start_dcos_target(systemd, block_systemd)
     except ValidationError as ex:
         print("Validation Error: {0}".format(ex))
         sys.exit(1)
     except PackageError as ex:
         print("Package Error: {0}".format(ex))
 
 
-def swap_active_package(install, repository, package_id, no_systemd=False, noblock_systemd=True):
+def swap_active_package(install, repository, package_id, systemd=False, block_systemd=False):
```
cmaloney: the default for `systemd` swapped here. Should now be `systemd=True`

cmaloney: Looking at the code, we never use these defaults actually. Why not just drop them?

cmaloney:
Just the one comment (https://github.com/dcos/dcos/pull/104/commits/22908f3ed1a8f4d1153bc9df30e5a05d3e5abfdd#r64825279) then we can :ship:

pkgpanda/actions.py
branden: Whoops, good catch. Yeah I'll just remove the defaults.

PR #105, opened 2016-05-25
Wait for all slaves to join, rather than just some slaves

cmaloney:
Hopefully fixes some of the CI test flaky failures which appeared recently.

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -125,16 +125,16 @@ class Cluster:
             logging.info(msg.format(r.status_code))
             return False
         data = r.json()
-        num_slaves = len([x['hostname'] for x in data['slaves']])
-        # For single node setup there is only one slave node:
-        min_slaves = min(len(self.slaves), 2)
-        if num_slaves >= min_slaves:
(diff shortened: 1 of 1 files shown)


test_util/integration_test.py
```diff
@@ -125,16 +125,16 @@ def _wait_for_slaves_to_join(self):
...
-        # For single node setup there is only one slave node:
-        min_slaves = min(len(self.slaves), 2)
-        if num_slaves >= min_slaves:
+        # Check that there are all the slaves the test knows about. They are all
+        # needed to pass the test.
+        num_slaves = len(data['slaves'])
+        if num_slaves >= len(self.all_slaves):
             msg = "Sufficient ({} >= {}) number of slaves have joined the cluster"
-            logging.info(msg.format(num_slaves, min_slaves))
+            logging.info(msg.format(num_slaves, self.all_slaves))
```
cmaloney: note: this changes the format here from int >= int -> int >= comma separated list of IPs. I think that is actually better for debugging, but it is definitely a change.

mellenburg:
very much :+1:

PR #106, opened 2016-05-26
Work on making it so pkgpanda packages can have their software run as non-root

cmaloney:
- By default we'll use systemd-sysusers to add users to the host (You can opt to manually add users instead). RPM, Deb, and Arch packaging all automatically have packages add users
 - All DC/OS software users begin with `dcos_` to make them easy to find / distinguish.
 

# TODO
 - [x] Test by hand
 - [x] Write unittest for username validation
 - [x] Move `mkpanda --recursive` into a separate PR

# Followup
 - [ ] Move dcos-history-service changes to another PR (And get 3dt updated)
 - [ ] Write an integration test for software running non-root
 - [ ] Add tooling to make it so that we can enforce that all systemd services except a couple exceptions have `User=` lines

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -15,11 +15,12 @@ envrionment variables from the package.
 import json
 import os
 import os.path
+import pwd
 import re
 import shutil
 import tempfile
 from itertools import chain
-from subprocess import CalledProcessError, check_call
+from subprocess import CalledProcessError, check_call, check_output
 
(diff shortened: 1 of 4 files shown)


cmaloney:
Launched a cluster. Manually verified the state dir stuff works properly, as well as user creation, file ownership, and systemd running the service as the dcos user.

cmaloney:
Deferring the integration test until we can run arbitrary commands on masters, which is coming soon. Breaking the history service changes into another PR.

PR #108, opened 2016-05-26
Make number of masters configurable in CF template

thefotios:
This is just an idea, so it's not fully fleshed out yet. Mostly for sparking a conversation, but I'm willing to work on it further or hand it off.

The number of masters shouldn't need to be hard coded in various templates if we propagate it properly throughout the CF stack.

### DIFF
--- a/gen/aws/templates/advanced/advanced-master.json
+++ b/gen/aws/templates/advanced/advanced-master.json
@@ -46,6 +46,12 @@
       "Type": "String",
       "Default": "m3.xlarge",
       "Description" : "\nRegion-specific instance type. E.g. m3.xlarge"
+    },
+    "NumMasters" : {
+      "Type": "Number",
+      "Default": 1,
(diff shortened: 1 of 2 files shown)


gen/aws/templates/advanced/advanced-master.json
```diff
@@ -46,6 +46,12 @@
       "Type": "String",
       "Default": "m3.xlarge",
       "Description" : "\nRegion-specific instance type. E.g. m3.xlarge"
+    },
+    "NumMasters" : {
```
cmaloney: A bunch of internal code code which is emitted into the cloud-init for the masters hard-sets values derived from the number of masters, so you can't do one advanced master template that has a variable number of masters.

That said, having all the different advanced-master templates, and having zen be able to select which to use based on a number is feasible.

cmaloney:
No response / comment back in 4 weeks, closing

PR #109, opened 2016-05-26
Output route table ID's from infra.json

thefotios:
It would be nice to have the route table IDs output from the CF template for use in VPN configurations.

Use Case:
 - I have a VPC configured with all of the necessary VPNGateway/CustomerGateway/etc
 - After creating DC/OS subnets and route tables, they are outputted from the CF template
 - I then attach my VPNGateway to these route tables for VPN Route propagation


kartick:
@thefotios - Are you planning on using the Zen template, or will be directly accessing the individual stack outputs from CF? If you are planning to drive using Zen template, then you might want to bubble up these output Subnet values to the Zen template (into the `"Outputs"` map in `zen.json`).

If you are not doing that and accessing the outputs of the Infra template directly, then this change looks good. 🚢

PR #110, opened 2016-05-26
AdminLocation should be user configurable

thefotios:
AdminLocation is a parameter to the underlying infra template, but not configurable in the base template. This passes the value down through the CF stack.


kartick:
LGTM 🚢

PR #111, opened 2016-05-26
Fix Azure to use it's ip-detect script rather than AWS'

cmaloney:
cc: @lingmann who found the azure broken recently.


lingmann:
👍

PR #114, opened 2016-05-26
Azure test integration

mellenburg:


test_util/azure_test_driver.py
```diff
@@ -126,6 +131,34 @@ def poll_on_dcos_ui_up():
...
+        with closing(SSHTunnel('core', 'ssh_key', master_lb, port=2200)) as t:
+            integration_test(
+                    tunnel=t,
+                    test_dir='/home/core',
+                    dcos_dns=master_lb,
+                    master_list=ip_buckets['masterNodeNic'],
+                    agent_list=ip_buckets['slavePrivateNic'],
+                    public_agent_list=ip_buckets['slavePublicNic'],
+                    provider='azure',
+                    test_dns_search=False,
```
lingmann: What is `test_dns_search` controlling? If this controls tests against mesos-dns... then we definitely want it enabled on Azure. If it is testing whether or not host-level DNS lookups are working (e.g. `host $(hostname)`) then this test probably shouldn't exist... since DC/OS does not guarantee that host DNS lookups work.

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -124,7 +124,7 @@ def registry_cluster(cluster, request):
             'healthChecks': [{
                 'protocol': 'COMMAND',
                 'command': {'value': 'curl -fsSlv https://{}/v2/test_server/manifests/latest'.format(cluster.registry)},
-                'gracePeriodSeconds': 20,
+                'gracePeriodSeconds': 300,
                 'intervalSeconds': 20,
                 'timeoutSeconds': 10,
                 'maxConsecutiveFailures': 3}],
```
lingmann: The above 5 lines are all set to the Marathon healthCheck defaults... while it is nice to be explicit, it's also nice to keep the code small and readable... consider removing.

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1078,7 +1078,7 @@ def test_if_minuteman_routes_to_vip(cluster, timeout=125):
         'healthChecks': [{
             'protocol': 'COMMAND',
             'command': {
-                'value': 'test "$(curl -o /dev/null --max-time 5 -4 -w \'%{http_code}\' -s http://localhost:${PORT0}/|cut -f1 -d" ")" == 200'  # noqa
+                'value': 'curl -f -X GET http://$HOST:$PORT0/'
```
lingmann: I like the simplified health check, but why is this changing? Would be nice to have some more details in the commit message.

mellenburg: This is a good question. It seemed like this test (test_if_minuteman_routes_to_vip) was failing consistently on azure (whereas it is just slightly flaky everywhere else) and I asked @sargun if he knew what was up. His response then was trying to change the health check to this. I got three fails in a row on this case before adding this check in, so it seems like it might be needed here...

test_util/azure_test_driver.py
mellenburg: Here is the test that the switch flips:
```
def test_if_search_is_working(registry_cluster):
    """Test if custom set search is working.

    Verifies that a marathon app running on the cluster can resolve names using
    searching the "search" the cluster was launched with (if any). It also tests
    that absolute searches still work, and search + things that aren't
    subdomains fails properly.

    The application being deployed is a simple http server written in python.
    Please check test/dockers/test_server for more details.
    """
    cluster = registry_cluster
    # Launch the app
    app_definition, test_uuid = cluster.get_base_testapp_definition()
    service_points = cluster.deploy_marathon_app(app_definition)

    # Get the status
    r = requests.get('http://{}:{}/dns_search'.format(service_points[0].host,
                                                      service_points[0].port))
    if r.status_code != 200:
        msg = "Test server replied with non-200 reply: '{0} {1}. "
        pytest.fail(msg.format(r.status_code, r.reason, r.text))

    r_data = r.json()

    # Make sure we hit the app we expected
    assert r_data['test_uuid'] == test_uuid

    expected_error = {'error': '[Errno -2] Name or service not known'}

    # Check that result matches expectations for this cluster
    if cluster.dns_search_set:
        assert r_data['search_hit_leader'] in cluster.masters
        assert r_data['always_hit_leader'] in cluster.masters
        assert r_data['always_miss'] == expected_error
    else:  # No dns search, search hit should miss.
        assert r_data['search_hit_leader'] == expected_error
        assert r_data['always_hit_leader'] in cluster.masters
        assert r_data['always_miss'] == expected_error

    cluster.destroy_marathon_app(app_definition['id'])
```

lingmann: Ok, this *should* work on Azure, and if it's not... we need to investigate and figure out why the test is failing.

packages/dcos-integration-test/extra/integration_test.py
lingmann: Ok, fine as is. More concerned about getting the DNS tests working on Azure.

test_util/azure_test_driver.py
lingmann: Ok, this makes sense now... when `test_dns_search` is set to `False` that is translated into the `dns_search_set` proxy class configuration here: https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/integration_test.py#L330 ... and False is what we want for the Azure environment where the DNS search domain is not yet configurable.

lingmann:
My concerns around Azure DNS testing have been addressed, thanks @mellenburg !
🚢

mellenburg:
@spahl @cmaloney This is ready for review
Note: https://github.com/dcos/dcos/pull/388 and https://github.com/dcos/dcos/pull/389 were pulled out of here for simplicity

mellenburg:
@cmaloney @spahl This has been rebased on merged dependencies in master and is ready for reivew

spahl:
:ship:

PR #115, opened 2016-05-26
Pkgpanda: Remove exit() from actions

branden:
See #91.

/cc @cmaloney

### DIFF
--- a/pkgpanda/actions.py
+++ b/pkgpanda/actions.py
@@ -5,7 +5,7 @@ from functools import partial
 from subprocess import check_call
 
 from pkgpanda import PackageId, requests_fetcher
-from pkgpanda.exceptions import FetchError, PackageError, ValidationError
+from pkgpanda.exceptions import FetchError, ValidationError
 from pkgpanda.util import (extract_tarball, if_exists, load_json, load_string,
                            write_string)
 
@@ -26,15 +26,9 @@ def activate_packages(install, repository, package_ids, systemd, block_systemd):
 
     """
     assert isinstance(package_ids, collections.Sequence)
-    try:
-        install.activate(repository.load_packages(package_ids))
-        if systemd:
-            _start_dcos_target(block_systemd)
-    except ValidationError as ex:
-        print("Validation Error: {0}".format(ex))
-        sys.exit(1)
-    except PackageError as ex:
-        print("Package Error: {0}".format(ex))
+    install.activate(repository.load_packages(package_ids))
+    if systemd:
+        _start_dcos_target(block_systemd)
 
 
 def swap_active_package(install, repository, package_id, systemd, block_systemd):
@@ -60,8 +54,7 @@ def swap_active_package(install, repository, package_id, systemd, block_systemd)
 
     new_id = PackageId(package_id)
     if new_id.name not in packages_by_name:
-        print("ERROR: No package with name {} currently active to swap with.".format(new_id.name))
-        sys.exit(1)
+        raise ValidationError("No package with name {} currently active to swap with.".format(new_id.name))
 
     packages_by_name[new_id.name] = new_id
     new_active = list(map(str, packages_by_name.values()))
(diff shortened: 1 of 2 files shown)


pkgpanda/actions.py
```diff
@@ -60,8 +54,7 @@ def swap_active_package(install, repository, package_id, systemd, block_systemd)
 
     new_id = PackageId(package_id)
     if new_id.name not in packages_by_name:
-        print("ERROR: No package with name {} currently active to swap with.".format(new_id.name))
-        sys.exit(1)
+        raise Exception("ERROR: No package with name {} currently active to swap with.".format(new_id.name))
```
cmaloney: ValidationError

pkgpanda/actions.py
```diff
@@ -88,10 +81,12 @@ def fetcher(id_, target):
     try:
         repository.add(fetcher, package_id)
     except FetchError as ex:
-        print("\nUnable to fetch package {0}: {1}".format(package_id, ex))
-        sys.exit(1)
-    sys.stdout.write("\rFetched: {0}\n".format(package_id))
-    sys.stdout.flush()
+        raise Exception("Unable to fetch package {0}: {1}".format(package_id, ex))
```
cmaloney: ValidationError here as well most likely.

pkgpanda/actions.py
```diff
@@ -131,8 +126,7 @@ def remove_package(install, repository, package_id):
 
     """
     if package_id in install.get_active():
-        print("Refusing to remove active package {0}".format(package_id))
-        sys.exit(1)
+        raise Exception("Refusing to remove active package {0}".format(package_id))
```
cmaloney: ValidationError

pkgpanda/actions.py
```diff
@@ -202,8 +196,7 @@ def _do_bootstrap(install, repository):
     # If the config can't be grabbed from any of them, fail.
     def fetcher(id, target):
         if repository_url is None:
-            print("ERROR: Non-local package {} but no repository url given.".format(repository_url))
-            sys.exit(1)
+            raise Exception("ERROR: Non-local package {} but no repository url given.".format(repository_url))
```
cmaloney: ValidationError

pkgpanda/actions.py
```diff
@@ -216,13 +209,12 @@ def fetcher(id, target):
         for pkg_id_str in os.listdir(setup_pkg_dir):
             print("Installing setup package: {}".format(pkg_id_str))
             if not PackageId.is_id(pkg_id_str):
-                print("Invalid package id in setup package: {}".format(pkg_id_str))
-                sys.exit(1)
+                raise Exception("Invalid package id in setup package: {}".format(pkg_id_str))
```
cmaloney: ValidationError

pkgpanda/actions.py
```diff
@@ -216,13 +209,12 @@ def fetcher(id, target):
...
             pkg_id = PackageId(pkg_id_str)
             if pkg_id.version != "setup":
-                print("Setup packages (those in `{0}`) must have the version setup. Bad package: {1}"
-                      .format(setup_pkg_dir, pkg_id_str))
-                sys.exit(1)
+                raise Exception(
```
cmaloney: ValidationError

pkgpanda/cli.py
```diff
@@ -141,6 +141,9 @@ def main():
         except ValidationError as ex:
             print("Validation Error: {0}".format(ex))
             sys.exit(1)
+        except Exception as ex:
```
cmaloney: All of these have the same exception handling for ValidationError, PackageError, and Exception. (Exception adds `ERROR:` to the start of the line). Should move the exception handling to surround the argument dispatching code rather than duplicate it in each dispatch

pkgpanda/actions.py
branden: Looks like `FetchError` could have all sorts of causes: https://github.com/dcos/dcos/blob/383c868ce55ca7b1b51f0e23525fa8bb0f767207/pkgpanda/util.py#L73 I'm not confident that we can assume it's a validation-related exception.

pkgpanda/actions.py
```diff
@@ -141,14 +135,14 @@ def remove_package(install, repository, package_id):
         PackageId(package_id)
         repository.remove(package_id)
     except ValidationError:
-        print("\nInvalid package id {0}".format(package_id))
-        sys.exit(1)
+        raise ValidationError("Invalid package id {0}".format(package_id))
```
cmaloney: Should be a `raise ValidationError() from ex` so we keep the exception history (same goes with the other re-throw code which rewrites the exception being thrown)

cmaloney:
Just need to do a `raise from` for exceptions which are based on other exceptions, then we can :ship:

cmaloney:
:ship:

PR #120, opened 2016-05-27
Add a NEWS file

cmaloney:
It is intended to document breaking changes / things which should be taken into account upgrading from the previous release. Actual release Errata should live in the DC/OS Documentation, this is for keeping track of things as we dev, then use it when we release to write up end user release notes

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,0 +1,15 @@
+Documentation of changes which may break some install environments
+
+
+1.8-dev
+ - Make the admin load balancer in all templates do TCP, not HTTP load balancing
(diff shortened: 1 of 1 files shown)


spahl:
Some of these things will go in a 1.7.1-open but I can add that at the time of doing it.

NEWS
```diff
@@ -0,0 +1,16 @@
+Documentation of changes which may break some install environments
+
+
+1.8-dev
+ -
```
cmaloney: extra `-` here

PR #132, opened 2016-05-31
AWS Advanced Template support for EL7

lingmann:
Support for EL7 and custom AMI's with AWS Advanced Templates

### DIFF
--- a/gen/__init__.py
+++ b/gen/__init__.py
@@ -48,8 +48,48 @@ def add_roles(cloudconfig, roles):
     return cloudconfig
 
 
-def add_units(cloudconfig, services):
-    cloudconfig['coreos']['units'] += services
+def add_units(cloudconfig, services, cloud_init_implementation='coreos'):
+    '''
+    Takes a services dict in the format of CoreOS cloud-init 'units' and
+    injects into cloudconfig a transformed version appropriate for the
+    cloud_init_implementation.  See:
+    https://coreos.com/os/docs/latest/cloud-config.html for the CoreOS 'units'
+    specification. See: https://cloudinit.readthedocs.io/en/latest/index.html
+    for the Canonical implementation.
+
+    Parameters:
+    * cloudconfig is a dict
+    * services is a list of dict's
+    * cloud_init_implementation is a string: 'coreos' or 'canonical'
+    '''
+    if cloud_init_implementation == 'canonical':
+        for unit in services:
+            unit_name = unit['name']
+            if 'content' in unit:
+                write_files_entry = {'path': '/etc/systemd/system/{}'.format(unit_name),
+                                     'content': unit['content'],
+                                     'permissions': '0644'}
+                cloudconfig.setdefault('write_files', []).append(write_files_entry)
+            if 'enable' in unit and unit['enable']:
+                runcmd_entry = ['systemctl', 'enable', unit_name]
+                cloudconfig.setdefault('runcmd', []).append(runcmd_entry)
+            if 'command' in unit:
+                opts = []
+                if 'no_block' in unit and unit['no_block']:
+                    opts.append('--no-block')
+                if unit['command'] in ['start', 'stop', 'reload', 'restart', 'try-restart', 'reload-or-restart',
+                                       'reload-or-try-restart']:
+                    runcmd_entry = ['systemctl'] + opts + [unit['command'], unit_name]
+                else:
+                    raise Exception("Unsupported unit command: {}".format(unit['command']))
+                cloudconfig.setdefault('runcmd', []).append(runcmd_entry)
+    elif cloud_init_implementation == 'coreos':
+        cloudconfig.setdefault('coreos', {}).setdefault('units', [])
+        cloudconfig['coreos']['units'] += services
+    else:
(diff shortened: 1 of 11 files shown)


lingmann:
@spahl or @cmaloney could I please get a review?

gen/__init__.py
```diff
@@ -48,8 +48,48 @@ def add_roles(cloudconfig, roles):
...
+    * cloud_init_implementation is a string: 'coreos' or 'canonical'
+    '''
+    if cloud_init_implementation == 'canonical':
+        for unit in services:
+            unit_name = unit['name']
+            if 'content' in unit:
+                write_files_entry = {'path': '/etc/systemd/system/{}'.format(unit_name),
+                                     'content': unit['content'],
+                                     'permissions': '0644'}
+                cloudconfig.setdefault('write_files', []).append(write_files_entry)
```
cmaloney: Why not just always default this right under `cloud_init_implementation == 'canonical'`? Seems like ti would be easier to read (And I would hope an empty write_files would be fine (Not to mention the odds of us ever having no units with content are very slim))

gen/aws/templates/advanced/advanced-master.json
```diff
@@ -46,39 +46,69 @@
...
+          ]
+        }
+      ]
     }
   },
   "Mappings": {
     "RegionToAmi": {
       "ap-northeast-1": {
-        "stable": "ami-84e0c7ea"
+        "coreos": "ami-84e0c7ea",
```
cmaloney: A little odd that we inject all the AMIs into both the templates (Even though just the el7 or just the stable will work). That said, probably good enough.

gen/azure/cloud-config.yaml
```diff
@@ -110,9 +69,6 @@ runcmd:
     - [ rm, -f, /etc/resolv.conf ]
     - [ cp, -p, /tmp/resolv.conf, /etc/resolv.conf ]
     - [ systemctl, start, dcos-docker-install.service ]
-    - [ systemctl, start, dcos-link-env.service ]
```
cmaloney: Just to confirm: moving all these after dcos-config-writer is fine?

gen/installer/aws.py
```diff
@@ -56,11 +56,10 @@
 
 late_services = """- name: dcos-cfn-signal.service
   command: start
+  no_block: true
   content: |
     [Unit]
     Description=AWS Setup: Signal CloudFormation Success
-    After=dcos.target
```
cmaloney: :+1:

gen/installer/aws.py
```diff
@@ -219,13 +218,17 @@ def gen_supporting_template():
...
-extra_templates = ['aws/dcos-config.yaml', 'coreos-aws/cloud-config.yaml', 'coreos/cloud-config.yaml']
-
-
 def make_advanced_bunch(variant_args, template_name, cc_params):
+    extra_templates = ['aws/dcos-config.yaml', 'aws/templates/advanced/{}'.format(template_name)]
+    if cc_params['os_type'] == 'coreos':
+        extra_templates += ['coreos-aws/cloud-config.yaml', 'coreos/cloud-config.yaml']
+        cloud_init_implementation = 'coreos'
+    else:
+        cloud_init_implementation = 'canonical'
```
cmaloney: can we have an assert here that `cc_params['os_type']` is equal to what we expect here? (I worry someone goes to add another os_type, doesn't realize they should touch this code, things sort of work for mysterious reasons until one day it all explodes with an unrelated change)

cmaloney:
Generally looks pretty good to me! Definitely like the "Use custom AMI to extend arbitrarily"

gen/aws/templates/advanced/advanced-master.json
lingmann: Agreed, for future improvements would like to templatize the whole AMI map, but wanted to keep this simple for now and not introduce potential template rendering issues.

gen/azure/cloud-config.yaml
lingmann: Yup

gen/calc.py
```diff
@@ -234,6 +234,11 @@ def calculate_adminrouter_auth_enabled(oauth_enabled):
     return oauth_enabled
 
 
+def validate_os_type(os_type):
+    can_be = ['coreos', 'el7']
+    assert os_type in can_be, 'Must be one of {}. Got {}'.format(can_be, os_type)
```
lingmann: @cmaloney should I include a reference to this under `entry['validate']` in `gen/calc.py` or `gen/aws/calc.py`?

gen/__init__.py
lingmann: Thanks that seems like a good improvement... done.

gen/installer/aws.py
lingmann: Yeah, I had a thought about that as well... this tells me we should do it. Adjustments made.

gen/calc.py
cmaloney: For now we can drop in tin `gen/aws/calc.py` since azure doesn't follow it. We should eventually move it up another level.

PR #135, opened 2016-05-31
WIP: Volume Mount integration tests

lingmann:
Not yet ready for final review.

Original PR and code review can be found in #58

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -1,7 +1,10 @@
 import collections
+import contextlib
 import json
 import logging
 import os
+import re
+import shlex
 import urllib.parse
 import uuid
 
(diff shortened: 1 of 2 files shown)


test_util/integration_test.py
```diff
@@ -15,13 +18,19 @@
 import requests
 import retrying
 
+from ssh.ssh_tunnel import SSHTunnel, run_ssh_cmd
```
kartick: This import might break, depending on how @mellenburg is "installing" the libraries via the Dockerfile. The last I recall from my convo with Michael is that the import should be `import ssh_tunnel` rather than `from ssh import ssh_tunnel`; by extension, `from ssh.ssh_tunnel import foo` may not work. This will work locally, but not as the tests are intended to be run.

lingmann: Hmmm thanks. I changed it because `import ssh_tunnel` was also not working and complaining about import failures... @mellenburg any suggestions?

mellenburg:
Note: I am taking ownership of this work so this has been migrated to: https://github.com/dcos/dcos/pull/251

lingmann:
Cool, thanks @mellenburg

PR #136, opened 2016-05-31
Teach aws.html to link to variant-specific templates

cmaloney:


cmaloney:
A little ugly (the list says `variant. Single Master`), but good enough / the template page works now. And the Open DC/OS doesn't really change.

PR #137, opened 2016-05-31
Add new systemd rules

cmaloney:


mellenburg:
:+1:  and while we are on the topic of systemd style, what about ExecPre checks? I.E:
ExecStartPre=/bin/bash -c '[ -f /var/lib/dcos/cluster-id ] || exit 1'
vs
ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id

cmaloney:
Ideally we replace all those with other logic / a waiter.py so we don't
have to do the spinning on startup

PR #138, opened 2016-06-01
Move extra module configuration into a separate EnvironmentFile

nickbp:
This removes some assumptions relating to custom hook and isolator modules, which aren't currently being used in Open DC/OS. With this change, a separate config file* is designated for use by downstream builds. This allows keeping unused configuration logic out of Open DC/OS while making downstream changes a bit more flexible.

* `/opt/mesosphere/etc/mesos-slave-common-extras`

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -72,12 +72,6 @@ def calculate_gen_resolvconf_search(dns_search):
         return ""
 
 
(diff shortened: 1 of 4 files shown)


cmaloney:
Why?

nickbp:
(added more info in PR desc)

PR #142, opened 2016-06-01
Move dcos-oauth complicated ExecStartPre into a script

cmaloney:


cmaloney:
Launched a manual test cluster

PR #145, opened 2016-06-01
Add agent adminrouter

branden:


packages/adminrouter/extra/dcos-adminrouter-slave.service
```diff
@@ -0,0 +1,24 @@
...
+ExecStartPre=/bin/ping -c1 ready.spartan
+ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id
+ExecStart=$PKG_PATH/nginx/sbin/nginx -c $PKG_PATH/nginx/conf/nginx.agent.conf
+ExecReload=/usr/bin/kill -HUP \$MAINPID
+KillSignal=SIGQUIT
+KillMode=mixed
+Environment=SECRET_KEY_FILE_PATH=/var/lib/dcos/auth-token-secret
+Environment=AUTH_ERROR_PAGE_DIR_PATH=${PKG_PATH}/nginx/conf/errorpages
+EnvironmentFile=/opt/mesosphere/etc/adminrouter.env
```
branden: Notably, I did not include the wait for bouncer to come up, since it's not running on agents:

```
# Wait for auth backend to seed the secret key
ExecStartPre=/usr/bin/curl --fail -sS -o /dev/null localhost:8101/acs/api/v1/groups
```

But the agent adminrouter will be using the secret key to verify JWT tokens. @seb @cody @alberts is this an issue?

malnick: Thanks for clarifying this

packages/adminrouter/build
```diff
@@ -27,86 +27,34 @@ make install
...
-ExecStartPre=/bin/ping -c1 ready.spartan
-ExecStartPre=/bin/ping -c1 marathon.mesos
-ExecStartPre=/bin/ping -c1 leader.mesos
-ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id
-ExecStart=$PKG_PATH/nginx/sbin/nginx
-ExecReload=/usr/bin/kill -HUP \$MAINPID
-KillSignal=SIGQUIT
-KillMode=mixed
-EOF
+systemd_slave="$PKG_PATH"/dcos.target.wants_slave/dcos-adminrouter-slave.service
```
malnick: Not to beat a dead horse, but should we be using `agent` instead of `slave` terminology here? @spahl ?

packages/adminrouter/extra/dcos-adminrouter-slave.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Admin Router Agent: A high performance web server and a reverse proxy server
```
malnick: If we use `agent` here we should use it everywhere

branden: I followed the example of the mesos units, which use `slave` in filenames and unit names but `agent` in descriptions. https://github.com/dcos/dcos/blob/master/packages/mesos/extra/dcos-mesos-slave.service#L2

packages/adminrouter/extra/dcos-adminrouter-slave.service
malnick: I think we might want to wait for Bouncer, since requests to this service will fail if Bouncer is not available (in sec-enabled environments). However, it needs to work if sec isn't enabled (no JWT passing). We might be able to infer this from a env var or perhaps a build-time configuration during gen.

branden: > I think we might want to wait for Bouncer
There's no bouncer on the agent to wait for. If we wait until bouncer on any/all masters is up, we've created a new dependency between agents and masters.

malnick:
I have one annoying nit with interchanging `slave` and `agent` and I'm not sure what the best way forward is to deal with signaling this this cluster is security-enabled or not. If we're in a secure env, we should wait for bouncer to come up before starting this router. Or perhaps the simplest answer is to not make bouncer a dep at all, and if JWT's are present in an `Authorization` header, we talk to bouncer, and if not, we skip talking to bouncer. That way, we don't depend on bouncer for starting, but we will 500 if it's unavailable and a JWT is present which isn't really a "graceful" downgrade at all.

packages/adminrouter/extra/dcos-adminrouter-slave.service
malnick: That sounds sane. But it was my understanding that moving forward, terminology should refer to `agents` instead of `slaves`.

packages/adminrouter/build
alberts: agent please

packages/adminrouter/extra/dcos-adminrouter-slave.service
```diff
@@ -0,0 +1,24 @@
...
+StandardOutput=journal
+StandardError=journal
+ExecStartPre=/bin/ping -c1 ready.spartan
+ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id
+ExecStart=$PKG_PATH/nginx/sbin/nginx -c $PKG_PATH/nginx/conf/nginx.agent.conf
+ExecReload=/usr/bin/kill -HUP \$MAINPID
+KillSignal=SIGQUIT
+KillMode=mixed
+Environment=SECRET_KEY_FILE_PATH=/var/lib/dcos/auth-token-secret
```
alberts: this file won't exist on agents.

malnick:
@spahl @cmaloney This should be marked `In Review` now, thanks.

cmaloney:
Some first pass stuff:

cmaloney:
Some first pass stuff:
 - We can't listen on 80 and 443 (https://github.com/dcos/adminrouter/pull/4/files#diff-694293a22fae760889245e10b53087c2R8) on the agents. Customers have other services they have listening there
 - The adminrouter PIDFile needs to move to `/var/lib`, and the service needs a state-directory (AdminRouter should be non-root top to bottom)
 - I'd move the  `EnvironmentFile=/opt/mesosphere/etc/adminrouter.env` line higher in the service, just abouve the "ExecStartPre" lines so it's obvious it applies
 - EnvironmentFile=/opt/mesosphere/etc/dcos-oauth.env I suspect will conflict with downstream auth methods, along with the +ExecStartPre=/usr/bin/curl --fail -sS -o /dev/null localhost:8101/acs/api/v1/groups
  - Rather than using shell / `*` expansion in adminrouter/build to copy the config files, can we just `cp -r /pkg/src/adminrouter $nginx-config_path/`?

PR #146, opened 2016-06-01
Allow to delete Exhibitor S3 bucket after the fact

mlafeldt:
By adding its resource ID to the outputs of the CloudFormation stack.

The bucket is created with a "Retain" deletion policy, which means that destroying the stack won't delete the bucket. This change makes it possible to still delete the bucket in an automated way (figuring out the bucket name without the stack output [is much more complicated](https://github.com/dcos-labs/dcos-bootstrap/blob/1ec418ca52590199576a7d5999fad6e3d0545747/destroy.yml#L5-L9)).

### DIFF
--- a/gen/aws/templates/cloudformation.json
+++ b/gen/aws/templates/cloudformation.json
@@ -988,6 +988,10 @@
     "PublicSlaveDnsAddress" : {
       "Description" : "Public slaves",
       "Value" : { "Fn::GetAtt" : [ "PublicSlaveLoadBalancer", "DNSName" ]}
+    },
+    "ExhibitorS3Bucket" : {
+      "Description" : "Exhibitor S3 bucket name",
+      "Value" : { "Ref" : "ExhibitorS3Bucket" }
     }
   }
 }


mlafeldt:
Alternatively, we might make the deletion policy configurable. To be honest, I'm not sure how the bucket is supposed to be used after deleting the CloudFormation stack? In particular, I don't think the data can be reused when setting up another DC/OS stack later on.

cmaloney:
The reason why we don't autodelete the bucket is that you can't (or at least couldn't when we made the templates) have a bucket with a single object in it be auto-deleted by cloudformation templates. It will refuse to delete and the stack deletion will fail, whereas leaving around the bucket is fairly cheap as an expense.

mlafeldt:
@cmaloney I see. According to http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html it's still not possible to empty a bucket with CloudFormation.

Based on this, I think my change makes sense.

cmaloney:
Pinging a product guy, but overall looks  seems reasonable to me. We used to output a whole bunch of additional things in the CF templates, but cut way back because people got confused, when mostly they just wanted / needed the master load balancer address.

mlafeldt:
Thanks for merging! 🎉

PR #148, opened 2016-06-02
Bump CoreOS

sargun:


cmaloney:
SEe https://dcosjira.atlassian.net/browse/DCOS-38

sargun:
@cmaloney Does this seem reasonable to merge? Or do we want to wait?

cmaloney:
See the JIRA I linked. It is why we haven't upgraded previously. We also
have customers who by the 1.8 timeframe will not make it to Docker 1.10, so
we need to make sure we have testing of older docker still

sargun:
These images are 1053.2.0. They should have a new enough SystemD to avoid these bugs.

cmaloney:
We only ship CoreOS stable. No CoreOS stable has a new enough systemd.

iamthemuffinman:
CoreOS 766.5.0 is irrelevant now. The majority of folks aren't running Docker 1.7.1. According to [this](https://github.com/coreos/bugs/issues/462#issuecomment-208477614) the error message can be safely ignored. It's not ideal, but testing an old version of CoreOS and Docker that anyone rarely uses isn't helpful to anyone.

cmaloney:
@iamthemuffinman We have customers who will definitively not get above docker 1.8 in the DC/OS 1.8 timeline. Not all are CoreOS users. Many people can't afford to restart everything too frequently, and they have problems when switching their stuff to newer dockers.

iamthemuffinman:
@cmaloney I can understand not dropping everything every time a new Docker version comes out to update. Those left on Docker 1.7.x are few in number in comparison to the 1.8.x, 1.9.x, 1.10.x, and 1.11.x users. I know you're trying to think of your customers that are still < 1.8, but you need to think about the rest of us that aren't running super old Docker versions as well.

cmaloney:
I'm not saying that supporting new docker versions is bad. I'm saying:
1. Systemd v225 has a bug which we've considered critical and has blocked upgrading for a while. It's fixed in systemd v229. That is not available in any CoreOS stable.
2. I trust the CoreOS developers to know better when their software is stable than I do. I will not upgrade to use a version of CoreOS which is not stable.
3. DC/OS _must_ keep working for customers who are currently running it. Stranding them on an island because new DC/OS doesn't work at all with older docker isn't acceptable. Our customer support team actively is working with customers on older stuff to keep them up to date, esp. since docker and DC/OS is so young that being just a couple versions back is a big delta. Breaking people's Datacenters is bad. Making it so we can't get them to the latest DC/OS is bad. We need to keep testing the oldest that we know reasonable people are using. If things don't even have simple tests, they are almost always broken.

iamthemuffinman:
1. I don't disagree here. I wouldn't go as far as critical though.
2. I would hope that they would know better. CoreOS stable has been on systemd 225 since version 808.0.0 which was released on September 2015 when the bug was introduced. In this case, I would consider stable not stable. Beta is more stable than stable is at this point.
3. Why would they spend the time to upgrade DC/OS if they won't spend the time to upgrade Docker? DC/OS is much more difficult and time consuming to upgrade regardless if you have support. Not to mention the images on Docker Hub that are being pushed from >= 1.10; They won't be able to use them and you'll see that increasingly so as more and more people upgrade. I suppose that wouldn't be an issue if they built their own base images, but I digress. If you're not stranding them on an island, you're stranding everyone else. They can upgrade DC/OS to their hearts desire, but a _really_ old Docker version is going to handcuff them.

Anyway, that's my 2 cents on the matter.

cmaloney:
3. Different teams own DC/OS than the underlying OS. DC/OS is also upgrade-able without killing any of the containers being run. It unfortunately takes a big guide and some handholding currently. There is a simple one-button deploy in the works which will lower the barrier to entry, and it's bad that there still isn't a published guide in Open about upgrading DC/OS clusters. I long ago argued we shouldn't have released DCOS 1.0 without the upgrade button.

spahl:
@sargun lets do this let's move to coreos stable (1010.5.0)

@jgehrcke @lingmann this will bring back the sudo bug but it's better then being stuck on old versions of docker and to have the security bugs from 766.

spahl:
@jgehrcke @lingmann Thoughs about 1010.5.0? Might not be practical either.

lingmann:
I don't think we can without systemd 229... we were impacted pretty severely by the polkit bug documented [here](https://github.com/coreos/bugs/issues/462). Clusters would not always boot, `daemon-reload` would randomly fail (only way to recover is to hard reboot), etc. We rely on `daemon-reload` quite a bit for both proper cluster installation as well as development / service management. So if it locks up even a small percentage of the time, that is a pretty big problem...

There is a simple repro [here](https://github.com/coreos/bugs/issues/462#issuecomment-163460828) if someone wants to see if they can trigger with 1010.5.0. But I suspect the problem is still there and serious.

jgehrcke:
@iamthemuffinman 

> I wouldn't go as far as critical though.

Disagree. While the exact symptoms of the polkit issue seem to vary depending on the environment, I'm pretty much traumatized by the polkit bug on our platform. Regarding its severity, I concur with @lingmann. In simple words: on an affected system, you have to be afraid of touching a class of systemd commands. If you do, systemd is likely to break, and the system becomes unmanageable. That's when also the `reboot` command can fail to work (heck, it won't even return).

As expected, people [are still](https://github.com/coreos/bugs/issues/462#issuecomment-218838172) running into the problem on more recent CoreOS stable releases.

Trying out a non-stable CoreOS with systemd 229 is IMO more likely to be a good story than trying systemd 225 again. However, I also entirely understand and support Cody's conservative standpoint here to not pull in a non-stable CoreOS.

So, I do not see a way out of being stuck until we have CoreOS stable with systemd 229. I hope the CoreOS team/community prioritizes this, also thanks to inquiries like [yours](https://github.com/coreos/bugs/issues/462#issuecomment-223429759).

iamthemuffinman:
@jgehrcke Let's agree to disagree on that as I've not had any serious issues with the polkit bug. (yet)

That's something I really don't understand. systemd 225 seems to be an unholy nightmare and CoreOS is dragging their feet on getting systemd 229 into stable. It's been out for a few months. That's enough time to test any catastrophic issues.

Like I said in an earlier post, CoreOS beta is more stable than stable is.

As long as CoreOS stable has systemd 225, I don't even know why it's listed as stable if this bug is so critical.

jgehrcke:
@iamthemuffinman Arguably, our experiences with systemd 225 are from several months ago. A  credible evaluation of the severity of the polkit issue in the current context can only be done with new tests.

If you have the capacity to test on CoreOS 1010.5.0, feedback about your insights and observations would be highly appreciated!

iamthemuffinman:
@jgehrcke Definitely agree with that.

I'll give 1010.5.0 a try and let you know how it goes!

spahl:
Alright we reached a verdict:-)

We are moving to coreos alpha 1053.2.0 in order to get 229.

To be able to make sure we don't break things we will add a "minimum configuration" integration test on an older docker versions (@lingmann will be working on our AMIs).

@sargun you can put this branch back to 1053.2.0

iamthemuffinman:
@spahl I know it's not ideal to run a piece of software at the alpha level, but I believe this is the right choice. Let's just make sure we pay extra attention to integration tests, yeah? ;)

iamthemuffinman:
I gotta say, I love the responses and point of views from everyone here! Thank you!

sargun:
@spahl Done.

spahl:
:ship:

cmaloney:
Still need somethign to test older, other than me having to notice and bug engineers.

sargun:
@spahl BUMP.

cmaloney:
The advanced templates need the new AMI as well: https://github.com/dcos/dcos/tree/master/gen/aws/templates/advanced

sargun:
@cmaloney thanks. Fixed.

spahl:
Can someone create a dcosjira ticket for the minimum testing settings? I don't want to miss it.

sargun:
@cmaloney Do you know what the $CUSTOMER is running? Mind creating a ticket based on their env?

sargun:
https://dcosjira.atlassian.net/browse/DCOS-156

PR #149, opened 2016-06-02
Fix exposing spartan to outside

sargun:


spahl:
@sargun can we make it so that we can turn listening to 53 on the master off through config.yml?

sargun:
@spahl Could we add an option like master_bindall: true / false to the config.yaml?

I'm actually not entirely sure how to wire up configs...

spahl:
:ship:

PR #139, opened 2016-06-01
WIP: Make DC/OS run less software as root

cmaloney:
# TODO
 - dcos_cfn_signal / the AWS CFN Signal service (Might just opt out)
 -  packages/dcos-cluster-id/build
    - Find all references to
 - Add a test to ensure more non-root things aren't added / only the services in exceptions are allowed to be root.

# In Review

# Done
packages/dcos-oauth/extra/dcos-oauth.service
packages/exhibitor/extra/dcos-exhibitor.service
packages/marathon/build
packages/3dt/build
packages/adminrouter/build
packages/cosmos/build
packages/erlang/build
packages/keepalived/build
packages/mesos-dns/extra/dcos-mesos-dns.service

# Root Exceptions
packages/logrotate/build
  - Must rotate the logs written by mesos-slave (which is root)
packages/minuteman/build
  - Requires special privileges to use certain kernel APIs

packages/mesos/build
 - Mesos Slave
 - Logrotate for mesos slave
 - Mesos master with navstar module?

packages/spartan/build:
  - Needs special privileges for binding among other things.

PR #151, opened 2016-06-02
Make 3dt not run as root

cmaloney:


cmaloney:
Needs #156

malnick:
Something to consider: we are going to be proxying back to this service on agents and masters via a agent side adminrouter. That will done over a unix socket rather than a traditional TCP socket. That socket will be secured by file perms, and I'm not sure we have resolved what system user that will be yet.

PR #152, opened 2016-06-02
Make erlang non-root

cmaloney:


packages/erlang/buildinfo.json
```diff
@@ -6,5 +6,7 @@
       "url": "https://downloads.mesosphere.com/pkgpanda-artifact-cache/otp_src_18.3.tar.gz",
       "sha1": "efc229b75e0b0614a187efe858faf33027bcf3d1"
     }
-  }
+  },
+  "state_directory": true,
```
cmaloney: shouldn't have a state_directory

PR #154, opened 2016-06-02
Make dcos-oauth run as non-root

cmaloney:


cmaloney:
Needs #156

cmaloney:
https://github.com/dcos/dcos-oauth/blob/5780c204b87bfb5a23e209b3856d45d1f4624f4b/dcos-oauth/flags.go#L24 -- have to change that setting

PR #155, opened 2016-06-02
Make cosmos run as non-root

cmaloney:


packages/cosmos/buildinfo.json
```diff
@@ -7,5 +7,7 @@
     "kind": "url",
     "url": "https://downloads.dcos.io/cosmos/0.1.5-3/cosmos-server-0.1.5-3-one-jar.jar",
     "sha1": "8345b8dd2a7ea143c73fc2288ced4d0226d52854"
-  }
+  },
+  "username": "dcos_cosmos",
+  "state_dir": true
```
cmaloney: should be state_directory

PR #156, opened 2016-06-02
pkgpanda: Only if a package has a state directory should we chown the…

cmaloney:
… state directory.

Found in testing #152

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -690,10 +690,9 @@ class Install:
                 if package.state_directory:
                     check_call(['mkdir', '-p', state_dir_path])
 
-                if package.username:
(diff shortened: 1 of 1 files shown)


pkgpanda/__init__.py
```diff
@@ -690,10 +690,10 @@ def symlink_all(src, dest):
                 if package.state_directory:
                     check_call(['mkdir', '-p', state_dir_path])
 
-                if package.username:
-                    state_dir_path = '/var/lib/dcos/{}'.format(package.name)
-                    uid = sysusers.get_uid(package.username)
-                    check_call(['chown', '-R', str(uid), state_dir_path])
+                    if package.username:
+                        state_dir_path = '/var/lib/dcos/{}'.format(package.name)
```
cmaloney: nit: This exact calculation occurs a couple lines above

PR #161, opened 2016-06-03
Make mesos-dns non-root

cmaloney:

PR #162, opened 2016-06-03
Train 14

cmaloney:
#102, #151, #152, #153, #155, #161


cmaloney:
Launched a cluster, logged in, and verified that all the new non-root services are indeed non-root.

PR #163, opened 2016-06-03
Make mesos master run as non-root

cmaloney:
TODO:
 - [x] Move logrotate to /var/lib/dcos/mesos/logs/
 - [x] Move master work dir to /var/lib/dcos/mesos/master

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -80,13 +80,13 @@ package:
       ].
   - path: /etc/mesos-master
     content: |
-      MESOS_LOG_DIR=/var/log/mesos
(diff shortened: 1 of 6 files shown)


cmaloney:
Do we need to add another logrotate, which is root, for mesos slaves?

cmaloney:
After talking with Harpreet, do need the special case for agents. Added that, although haven't updated the 3dt test, (so integration should end up failing).

surdy:
Did some minor sanity check on the build looks good. :+1:
Just a note to be mindful of the fact that users upgrading to a build with this change will have older log files in `/var/log/mesos` and `/var/log/mesos/archive`

PR #164, opened 2016-06-03
Add more DependsOn to things that commonly fail to delete and Ref thi…

cmaloney:
…ngs that fail to delete with them.

### DIFF
--- a/gen/aws/templates/cloudformation.json
+++ b/gen/aws/templates/cloudformation.json
@@ -158,7 +158,8 @@
           { "Key" : "Application", "Value" : { "Ref" : "AWS::StackName" } },
           { "Key" : "Network", "Value" : "Public" }
         ]
-      }
+      },
+      "DependsOn": "Vpc"
     },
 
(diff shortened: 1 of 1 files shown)


cmaloney:
cc: @lingmann

gen/aws/templates/cloudformation.json
```diff
@@ -209,7 +212,8 @@
       "Properties" : {
         "VpcId" : { "Ref" : "Vpc" },
         "InternetGatewayId" : { "Ref" : "InternetGateway" }
-      }
+      },
+      "DependsOn": "Vpc"
```
lingmann: There is already a `DependsOn` in this block, needs to be combined as a list.

lingmann:
Technically this should not be necessary because all of these resources have Vpc Ref statements, which implicitly creates a dependency. But I'm willing to give it a try... If this doesn't make a difference then we should revert this change to minimize unnecessary code.

PR #168, opened 2016-06-06
Use dedicated storage account for Exhibitor storage

lingmann:
This adds a dedicated storage account for Exhibitor coordination state, also making it possible to use Premium LRS for the Agent OS disks. Previously, when a VM with Premium LRS storage was selected, Exhibitor would also be configured to use a Premium LRS account for storing coordination state via block blobs. However, this is not supported use of the Premium LRS service (see [Azure Premium Storage](https://azure.microsoft.com/en-us/documentation/articles/storage-premium-storage/)):

> Currently, Premium Storage does not support Azure Block Blobs, Azure Append Blobs, Azure Files, Azure Tables, or Azure Queues. Any other object placed in a Premium Storage account will be a Page Blob, and it will snap to one of the the supported provisioned sizes. Heance Premium Storage account is not meant for storing tiny blobs.

And the cluster would fail to launch... With this change Exhibitor coordination state will always use Standard LRS and will also be isolated from the Master node storage.

### DIFF
--- a/gen/azure/templates/acs.json
+++ b/gen/azure/templates/acs.json
@@ -507,6 +507,7 @@
     "agentStoragePublicAccountName": "[concat(variables('storageAccountBaseName'), 'agntpub')]",
     "agentStoragePrivateAccountName": "[concat(variables('storageAccountBaseName'), 'agntpri')]",
     "masterStorageAccountName": "[concat(variables('storageAccountBaseName'), 'mstr0')]",
+    "masterStorageAccountExhibitorName": "[concat(variables('storageAccountBaseName'), 'exhb0')]",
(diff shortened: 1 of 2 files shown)


lingmann:
Before this can merge we need confirmation this works for Standard_DSx and/or Standard_GSx instance types, in addition to having the standard Azure integration tests passing.

lingmann:
@cmaloney this one is also ready. It's been tested against a 3-master cluster using Standard_DS2 instance types and the new dedicated storage account for Exhibitor. Test details can be seen in the linked Azure integration test.

cmaloney:
:ship:

PR #169, opened 2016-06-06
Signal Service 1.1.5

malnick:
- Updates to tests to POST from signal to test cache server
- Updates for security enhancements 
- Updates to reporters to do asynchronous report queries and multiple, generic reporter types
- Hard fails if cluster-id is not present

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 __pycache__
+ext/dcos-installer/packages/
 this
 .idea/
 dcos-config--setup*.tar.xz
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -1267,12 +1267,18 @@ def test_signal_service(cluster):
 
     wait_for_endpoint()
 
-    test_cache_url = "http://{}:{}/signal_test_cache".format(service_points[0].host, service_points[0].port)
+    test_host = service_points[0].host
+    test_port = service_points[0].port
+    test_cache_url = "http://{}:{}/signal_test_cache".format(test_host, test_port)
+
     cmd = """
-data=`/opt/mesosphere/bin/dcos-signal -report-host leader.mesos -test 2> /dev/null`;
-curl -H 'Content-Type: application/json' -X POST -d "$data" {};
+ID_PATH="/${PWD}/test-cluster-id"
+echo 'test-id' > $ID_PATH
+/opt/mesosphere/bin/dcos-signal \
+    -cluster-id-path $ID_PATH \
+    -test-url %s
 sleep 3600
-""".format(test_cache_url)
+""" % test_cache_url
 
     test_uuid = uuid.uuid4().hex
     signal_app_definition = {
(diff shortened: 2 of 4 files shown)


malnick:
@spahl This PR has the updates to signal and the integration test we discussed this past week. It now POSTs to the test server, which should add more stability to the testing cycle. 

This also satisfies https://mesosphere.atlassian.net/browse/DCOS-6436

CC'ing @pyronicide

malnick:
@mellenburg @cmaloney If you could review the test code updates that would be super.

malnick:
@cmaloney @spahl This is ready for review.

test_util/integration_test.py
```diff
@@ -1290,75 +1292,105 @@ def wait_for_endpoint():
 
     r = requests.get(test_cache_url)
 
+    # Handy for diagnosing strange signal service test behavior, not sure if we should
```
malnick: @cmaloney Your call here

test_util/integration_test.py
```diff
@@ -1260,10 +1260,12 @@ def wait_for_endpoint():
...
 
-    test_cache_url = "http://{}:{}/signal_test_cache".format(service_points[0].host, service_points[0].port)
+    test_host = service_points[0].host
+    test_port = service_points[0].port
+    test_cache_url = "http://{}:{}/signal_test_cache".format(test_host, test_port)
+
     cmd = """
-data=`/opt/mesosphere/bin/dcos-signal -report-host leader.mesos -test 2> /dev/null`;
-curl -H 'Content-Type: application/json' -X POST -d "$data" {};
+/opt/mesosphere/bin/dcos-signal -master-url http://leader.mesos -test-url {}  2> /dev/null;
```
mellenburg: I definitely like this more than saving stdout and posting with curl!

cmaloney: no way for us to keep stdout and stderr separate here? Be nice to be able to log what came to stderr, because the signal service could be trying to tell us important. (And if we don't dump it to /dev/null, it _should_ show up in the mesos task log.)

test_util/integration_test.py
```diff
@@ -1290,75 +1292,105 @@ def wait_for_endpoint():
...
-                'provider': 'onprem',
-                'source': 'cluster',
-                'clusterId': '',
-                'customerKey': '',
-                'environmentVersion': '',
-                'variant': 'open'}
-            }
+        'diagnostics': {
+            'event': 'health',
+            'anonymousId': 'open /var/lib/dcos/cluster-id: no such file or directory',
```
cmaloney: This doesn't seem right... Probably shouldn't be erroring...

test_util/integration_test.py
```diff
@@ -1290,75 +1292,105 @@ def wait_for_endpoint():
...
+    }
+
+    # Insert the generic property data which is the same between all signal tracks
+    for k, v in generic_properties.items():
+        exp_data['diagnostics']['properties'][k] = v
+        exp_data['cosmos']['properties'][k] = v
+        exp_data['mesos']['properties'][k] = v
 
+    # Insert all the diagnostics data programmatically
     master_units = [
```
cmaloney: Please don't change this indentation here. Makes a lot of noise in the diff (Can change whitespace, just do it separately from when you're making semantic changes).

test_util/integration_test.py
malnick: This test runs on an agent, so the cluster-id does not exist. The test overwrote the error with an empty string before.

test_util/integration_test.py
malnick: It runs the POST to stdout if you execute it with -v, I've also left the print statement in the test code so we can see the return. However, since this is executed via marathon, getting the stdout from the service being executed is rather difficult.

test_util/integration_test.py
cmaloney: definitely the API should not be returning "no such file or directory", that will lead to problems. It should return json null to indicate "no user id" most likely, and have that be distinguishable from the "empty user id" (empty string).

cmaloney: Best is probably, since we're triggering the signal service manually, pass the option to change the location of the cluster-id file to one of our choosing, so we can verify the file gets loaded up, data sent properly.

test_util/integration_test.py
cmaloney: Inside the test itself it's difficult to get, but if you go to the DC/OS UI and find the task which completed, then there is the whole stdout + stderr in the sandbox. Generally I really don't like dumping messages into a black hole

test_util/integration_test.py
malnick: We purposefully had the error propagate into the signal service like that so we can see what errored out, since there are a few type of errors we can have loading the cluster-id file, null could indicate any one of those errors and we want to surface the actual problem in segment. 

There isn't an option to change the location of the cluster-id currently, since this is a well established file, whose location does not change. However, if we did have such an option would you like another marathon app to make a fake cluster-id on the agent? What would the test look like in that case? And would that test actually test the signal service in the way we expect? 

My opinion is that the signal service should be tested on a master, and that it should return a not empty string for cluster-id as we expect, but we simply can't do that right now.

test_util/integration_test.py
malnick: I'm still not sure what you're asking here. Do you want me to not dump to dev null?

malnick:
@cmaloney This is ready for another pass. I updated signal to hard fail on cluster-id not being present and added a fake cluster-id file to test against.

.gitignore
```diff
@@ -1,4 +1,5 @@
 __pycache__
+ext/dcos_installer/packages/
```
cmaloney: why are we adding this?

At the very least seems unrelated. None of the stuff any of our tooling does ends up here.

test_util/integration_test.py
```diff
@@ -1260,11 +1260,23 @@ def wait_for_endpoint():
...
 
-    test_cache_url = "http://{}:{}/signal_test_cache".format(service_points[0].host, service_points[0].port)
+    test_host = service_points[0].host
+    test_port = service_points[0].port
+    test_cache_url = "http://{}:{}/signal_test_cache".format(test_host, test_port)
+
     cmd = """
-data=`/opt/mesosphere/bin/dcos-signal -report-host leader.mesos -test 2> /dev/null`;
-curl -H 'Content-Type: application/json' -X POST -d "$data" {};
+cleanup() {
```
cmaloney: This cleanup will never get hit. Note we're also running inside a mesos task sandbox here. I think echo `"test_id" > test-cluster-id && /opt/mesosphere/bin/dcos-sign -cluster-id-path $PATH/test-cluster-id` (and the rest) would work better.

test_util/integration_test.py
```diff
@@ -1290,75 +1302,105 @@ def wait_for_endpoint():
...
+        'provider': 'onprem',
+        'source': 'cluster',
+        'clusterId': 'test-id',
+        'customerKey': '',
+        'environmentVersion': '',
+        'variant': 'open'
+    }
+
+    # Insert the generic property data which is the same between all signal tracks
+    for k, v in generic_properties.items():
```
cmaloney: Why not:
`exp_data['diagnostics']['properties'] = copy(generic_properties)` for all of these? (The copy may not even be necessary.

If we're updating existing dictionaries, then do a `.update()` - https://docs.python.org/3/library/stdtypes.html#dict.update

test_util/integration_test.py
malnick: Line 1279?

malnick: We should probably put it before the sleep()

.gitignore
```diff
@@ -1,4 +1,5 @@
 __pycache__
+ext/dcos-installer/packages/
```
malnick: Gen makes that packages/ directory in the dcos-installer dir when I execute it locally in my dev env. @spahl has seen me accidentally push .tar.xz files that were placed in it in the past, and I almost did it again testing this PR. Would like to leave it here so we don't repeat it.

test_util/integration_test.py
```diff
@@ -1260,12 +1260,19 @@ def wait_for_endpoint():
...
-    test_cache_url = "http://{}:{}/signal_test_cache".format(service_points[0].host, service_points[0].port)
+    test_host = service_points[0].host
+    test_port = service_points[0].port
+    test_cache_url = "http://{}:{}/signal_test_cache".format(test_host, test_port)
+
     cmd = """
-data=`/opt/mesosphere/bin/dcos-signal -report-host leader.mesos -test 2> /dev/null`;
-curl -H 'Content-Type: application/json' -X POST -d "$data" {};
+ID_PATH="/${PWD}/test-cluster-id"
+echo -n 'test-id' > $ID_PATH
```
cmaloney: the dcos-cluster-id service actually puts a new line into the cluster-id file, the signal service should do a `.strip()` to remove whitespace around the cluster id

malnick:
@cmaloney `-n` removed.

malnick:
@cmaloney I've got a few more changes to make to the signal code here, please don't ship this tonight.

PR #174, opened 2016-06-07
Pluggable headers for history service

mellenburg:


packages/dcos-history-service/extra/history/server.py
```diff
@@ -9,6 +9,37 @@
...
+
+try:
+    import dcos_auth_python
+    log.info('dcos_auth_python module detected; applying settings')
+    add_headers_cb = dcos_auth_python.get_auth_headers
+except ImportError:
+    log.info('no dcos_auth_python module detected; using defaults')
+
+
+def headers_cb(add_headers_cb=add_headers_cb):
```
cmaloney: why not just do a `global add_headers_cb` rather than doing this "Capture global as default argument" thing, which is a bit of odd corner behavior

cmaloney:
One thing we should change to `global`. I also wonder if we can test it with monkeypatch to verify adding a header actually works? (monkeypatch in a headers_cb function)

mellenburg:
Can definitely add the test and switch to using global

mellenburg:
@cmaloney This is ready for another pass

packages/dcos-history-service/extra/history/server.py
```diff
@@ -21,7 +53,7 @@ def home():
 
 @app.route('/ping')
 def ping():
-    return "pong"
+    return _response_("pong")
```
cmaloney: Good catch

packages/dcos-history-service/extra/history/statebuffer.py
```diff
@@ -138,12 +138,16 @@ def add_data(self, timestamp, data):
 
 class BufferUpdater():
     """Class that fetchs and pushes that fetched update to BufferCollection
+    Args:
+        headers_cb (method): a callback method that returns a dictionary
+            of headers to be used for mesos state-summary requests
     """
-    def __init__(self, buffer_collection):
+    def __init__(self, buffer_collection, headers_cb=None):
```
cmaloney: nit: I'd rather not have this be an optional argument since we always pass it.

cmaloney:
:ship: . Pulled into train 18

PR #187, opened 2016-06-09
bump concurrent ssh sessions for DC/OS installer

darkonie:
DCOS-7596
- add optional `ssh_parallelism` option to a config file. The value must be
integer within 1..100 range.

- use default ssh_parallelism of 20 for DC/OS installer.

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/utils.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/utils.py
@@ -9,9 +9,12 @@ def get_async_runner(config, hosts, async_delegate=None):
     extra_ssh_options = config.get('extra_ssh_options', '')
     ssh_key_path = config.get('ssh_key_path', '/genconf/ssh_key')
 
+    # if ssh_parallelism is not set, use 20 concurrent ssh sessions by default.
+    parallelism = config.get('ssh_parallelism', 20)
+
     return ssh.ssh_runner.MultiRunner(hosts, ssh_user=config['ssh_user'], ssh_key_path=ssh_key_path,
                                       process_timeout=process_timeout, extra_opts=extra_ssh_options,
-                                      async_delegate=async_delegate)
+                                      async_delegate=async_delegate, parallelism=parallelism)
 
 
 def add_pre_action(chain, ssh_user):
--- a/ssh/validate.py
+++ b/ssh/validate.py
@@ -96,6 +96,11 @@ def run_validate_config_chunk(key_validate_fn_map, config, keys_required):
     return errors
 
 
+def validate_ssh_parallelism(value):
+    assert isinstance(value, int), 'ssh_parallelism must be integer'
+    assert 0 < value <= 100, 'ssh_parallelism must be within the range 1..100'
+
+
 def validate_config(config):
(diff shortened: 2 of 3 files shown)


ext/dcos-installer/dcos_installer/action_lib/utils.py
```diff
@@ -8,10 +8,11 @@ def get_async_runner(config, hosts, async_delegate=None):
     process_timeout = config.get('process_timeout', 120)
     extra_ssh_options = config.get('extra_ssh_options', '')
     ssh_key_path = config.get('ssh_key_path', '/genconf/ssh_key')
+    parallelism = config.get('ssh_parallelism')
```
cmaloney: Can we load the default here rather than dropping it into the config file?

ssh/ssh_runner.py
```diff
@@ -76,6 +76,16 @@ def __init__(self, targets, async_delegate=None, ssh_user=None, ssh_key_path=Non
         self.__targets = []
         for target in targets:
             self.__targets.append(add_host(target))
+
+        # default number of parallel ssh sessions.
+        parallelism = 10
```
cmaloney: I'd really rather that this must always be set as an argument to MultiRunner. Are there a lot of callsites that are hard with the default arg?

ext/dcos-installer/dcos_installer/config/__init__.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, overrides={}, config_path=CONFIG_PATH, write_default_config=T
 ssh_port: 22
 process_timeout: 10000
 bootstrap_url: file:///opt/dcos_install_tmp
+ssh_parallelism: 20
```
cmaloney: this shouldn't be needed anymore / the `config.get()` does the same thing, can we not duplicate?

ext/dcos-installer/dcos_installer/action_lib/utils.py
darkonie: done

ssh/ssh_runner.py
darkonie: all callers are using default `parallelism`.

darkonie:
@cmaloney your comments addressed, please review

ext/dcos-installer/dcos_installer/config/__init__.py
darkonie: sounds good. We will default to 20 concurrent sessions but we will not add `ssh_parallelism` to a newly generated config.

cmaloney:
:ship:

malnick:
https://mesosphere.atlassian.net/browse/DCOS-7596

ssh/validate.py
```diff
@@ -96,6 +96,11 @@ def run_validate_config_chunk(key_validate_fn_map, config, keys_required):
     return errors
 
 
+def validate_ssh_parallelism(value):
+    assert isinstance(value, int), 'ssh_parallelism must be integer'
+    assert 0 < value <= 100, 'ssh_parallelism must be within the range 1..100'
```
malnick: 👍

PR #189, opened 2016-06-09
make curl fail for non-200 status code in download

mellenburg:
See https://mesosphere.atlassian.net/browse/DCOS-5906


cmaloney:
Can we make this use our standard set of curl arguments: https://github.com/dcos/dcos/blob/master/gen/dcos-services.yaml#L29?

There's a lot more than just `-f` which is sometimes important.

mellenburg:
@cmaloney fixed!

cmaloney:
:ship:

PR #191, opened 2016-06-10
Updated Mesos release to 1.0.0-rc1.

karya0:
* Updated Mesos release to 1.0.0-rc1 ([release notes](https://git-wip-us.apache.org/repos/asf?p=mesos.git;a=blob_plain;f=CHANGELOG;hb=1.0.0-rc1)).
* Removed mesos-buildenv package in lieu of `--enable-install-module-dependencies` flag.
* Replaced `--modules` flag for agent with `--modules_dir` (updated gen/calc.py accordingly).
* Update logrotate module to supply a json file for module manifest and install it in `/opt/mesosphere/etc/mesos-slave-modules` directory.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,11 +10,12 @@ Documentation of changes which may break some install environments
  - Make the admin load balancer in all templates do TCP, not HTTP load balancing
  - Switch Ubuntu in Azure to 16.04 LTS final release
  - Teach spartan / DNS to only listen on a local internal address
- - Make the installer preflight actually faile if SELinux is enabled
+ - Make the installer preflight actually fail if SELinux is enabled
  - Took ownership of several sysctl settings for minuteman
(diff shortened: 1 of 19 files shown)


cmaloney:
So many great new features!

Can we also add to NEWS that mesos is bumped? (https://github.com/dcos/dcos/blob/master/NEWS)

packages/logrotate/extra/logrotate.json
```diff
@@ -0,0 +1,34 @@
+{
+  "libraries": [
```
cmaloney: Should probably actually live in the `mesos` package. The mesos build is what actually builds the logrotate module. The logrotate package just builds the logrotate binary which we do for other reasons.

karya0: Good idea. Fixed now!

PR #198, opened 2016-06-10
Forward ports to both public and private agents

mellenburg:
@cmaloney I think this should stop the spurious failures of VPC integration tests


cmaloney:
:ship:

PR #200, opened 2016-06-11
Make spartan non-root, giving it just limited capabilities

cmaloney:


cmaloney:
Can't set AmbientCapabilities until newer systemd / linux than we have on platforms we ship. Spartan needs to stay a root process for now.

PR #202, opened 2016-06-13
Allow setting specific user for integration tests

mellenburg:


malnick:
@cmaloney This is `Ready for Review`

test_util/integration_test.py
```diff
@@ -21,6 +21,11 @@
 AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID', '')
 AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY', '')
 AWS_REGION = os.environ.get('AWS_REGION', '')
+AUTH_ENABLED = os.environ.get('DCOS_AUTH_ENABLED', 'true') == 'true'
+# If auth is enabled, by default, tests use hard-coded OAuth token
+# Set these to run test against a custom configured user instead
+LOGIN_UNAME = os.environ.get('DCOS_LOGIN_UNAME', '')
```
cmaloney: Be better to not specify the second arg here, so we just get the default behavior of None

test_util/integration_test.py
```diff
@@ -250,15 +261,15 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves, reg
         # Make URI never end with /
         self.dcos_uri = dcos_uri.rstrip('/')
 
-        self.is_oauth = os.environ.get('DCOS_OAUTH', 'true') == 'true'
+        self.auth_enabled = auth_enabled
```
cmaloney: This is always set to AUTH_ENABLED the global, so what value are we getting setting it to a member anymore?

mellenburg:
@cmaloney fixed and ready for another pass

cmaloney:
🚢

PR #204, opened 2016-06-13
Allow passing arbitrary env to integration_test

mellenburg:


malnick:
@cmaloney This is `Ready for Review`

test_util/test_runner.py
```diff
@@ -100,7 +100,7 @@ def integration_test(
         tunnel, test_dir,
         dcos_dns, master_list, agent_list, public_agent_list,
         variant, test_dns_search, ci_flags, timeout=None,
-        aws_access_key_id='', aws_secret_access_key='', region=''):
+        aws_access_key_id='', aws_secret_access_key='', region='', **add_env):
```
cmaloney: add_env as given here will swallow anything / everything extra passed to this function. I'd much rather have us just pass through the add_env dictionary directly, rather than exploding it into proper parameters and going back again. The goal here is to pass through a very specific dictionary.

test_util/test_runner.py
```diff
@@ -135,9 +136,15 @@ def integration_test(
...
-        '-vv', ci_flags, '/integration_test.py']
+        '-e', 'AWS_REGION='+region]
+    for key, value in add_env.items():
+        extra_env = ['-e', key+'='+value]
+        test_env.extend(extra_env)
+    docker_opts = [
+        'docker', 'run', '-v', test_dir+'/integration_test.py:/integration_test.py',
+        '--net=host', '--name='+test_container_name]
+    test_args = ['py.test', 'py.test', '-vv', ci_flags, '/integration_test.py']
+    test_cmd = docker_opts + test_env + test_args
```
cmaloney: Rather than splitting this into three parts, why not just do a:
docker_cmd = ['...']
# add_env
# append docker opts

I don't see a good reason for pulling this into three different variables than re-composing them into the overal command.

mellenburg:

cmaloney:
🚢

PR #206, opened 2016-06-14
WIP: Simplify signal service integration test

mellenburg:
Units will be better checked by post-flight

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -1295,76 +1295,13 @@ sleep 3600
     cluster.destroy_marathon_app(signal_app_definition['id'])
     cluster.destroy_marathon_app(test_server_app_definition['id'])
 
-    exp_data = {
-            'Event': 'health',
-            'UserId': '',
-            'ClusterId': '',
-            'Properties': {
-                'provider': 'onprem',
-                'source': 'cluster',
-                'clusterId': '',
-                'customerKey': '',
-                'environmentVersion': '',
-                'variant': 'open'}
-            }
+    exp_keys = ['Event', 'UserId', 'ClusterId', 'Properties']
+    exp_properties = ['provider', 'source', 'clusterId', 'customerKey', 'environmentVersion', 'variant']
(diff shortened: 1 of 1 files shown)


mellenburg:

cmaloney:
I don't think that units will actually be better-checked by a postflight. It was also very intentional design @spahl and I went back and forth on for whether the services / units should be included here, and we concluded they should. For cases where people add more units, they should just be able to specify the additional units via some config.

I think it's pretty important we make sure we're running exactly the services we expect through the entire cluster. A _lot_ of the other tests in this repo can pass while individual pieces of software may not be working.

malnick:
I agree with Cody on this one, this test has provided us with some very specific insights to what broke a cluster in other troubleshooting I've had. We don't have the bullet proof test coverage on all services, and it's important that we verify the information we get from 3DT is correct. This test is the only way to prove that 3DT is returning all expected services in a cluster.

mellenburg:
So, how about a json or a yaml that you can input that overrides all all of the units in properties?

cmaloney:
I think in general we're getting very close to the point (if not at it), that all of the integration test config should go into one big config file, definitely happy to have things move in that direction.

Something like:
test_config.yaml
I think in general we're getting very close to the point (if not at it), that all of the integration test config should go into one big config file, definitely happy to have things move in that direction.

Something like (with all the services):
test_config.yaml
```yaml
3dt_services:
  master_units:
    - adminrouter-reload-service
    - adminrouter-reload-timer
  all_node_units:
    - ddt-service
    - epmd-service
  slave_units:
    - mesos-slave-service
  slave_public_units:
    - mesos-slave-public-service
  all_slave_units:
    - logrotate-agent-service
```

mellenburg:
Given that this would require bundling a config to the test, this: https://github.com/dcos/dcos/pull/68 would help greatly. So I'll give that a cleanup first

malnick:
@mellenburg Can we close this?

PR #210, opened 2016-06-15
Make signal service integration test variant aware

malnick:
Depends on #169

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -1309,6 +1309,8 @@ sleep 3600
     print('CACHE SERVER STATUS:\n{}'.format(r.status_code))
 
     r_data = json.loads(r.json())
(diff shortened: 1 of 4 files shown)


malnick:
@spahl @cmaloney This is ready for review, please mind the dependency on #169 one which this is rebased.

test_util/integration_test.py
```diff
@@ -1361,6 +1363,14 @@ def wait_for_endpoint():
         'mesos-dns-service',
         'mesos-master-service',
         'signal-service']
+
+    ee_master_units = [
```
spahl: This should not be in open. Variants should have a clean separation and not bleed through like this.

spahl:
This is not gonna work. The implementation of variants bleeds through. This is the open project we can't just add whatever we want just to make the enterprise version easier.

PR #212, opened 2016-06-15
Added *.pyc to .gitignore

spahl:


cmaloney:
These shouldn't happen with Python 3.4 that we use: https://docs.python.org/3/whatsnew/3.2.html#pep-3147-pyc-repository-directories. Everything goes in the `__pycache__` directories which we do .gitignore.

spahl:
Somehow @alberts saw some lying around but you are right could be anything. Not important.

PR #213, opened 2016-06-15
if CI_FLAGS is set, exit with code 0, since the CI should pick up the results and fail that way

cmaloney:
Exiting non-zero causes TeamCity to hard fail, even if the test which failed is muted.

This makes CI_FLAGS interpreted as a sign that we're running in a CI environment which understands
structured test results, and that the CI system will interpret failed test as a failed run.

So the new method is either "basic" environment where a non-zero exit indicates error, or
"everything ran successfully but some tests failed" indicated by test results being reported out and
the process exiting zero.

### DIFF
--- a/test_util/test_runner.py
+++ b/test_util/test_runner.py
@@ -150,13 +150,16 @@ def integration_test(
     try:
         with remote_port_forwarding(tunnel, agent_list+public_agent_list, join(test_dir, 'ssh_key')):
             log.info('Running integration test...')
-            tunnel.remote_cmd(test_cmd, timeout=timeout)
+            try:
+                tunnel.remote_cmd(test_cmd, timeout=timeout)
+            except CalledProcessError:
+                log.exception('Test failed!')
+                if ci_flags:
+                    return False
+                raise
         log.info('Successful test run!')
     except TimeoutExpired:
         log.error('Test failed due to timing out after {} seconds'.format(timeout))
-        raise
-    except CalledProcessError:
-        log.error('Test failed!')
         raise
     finally:
         get_logs_cmd = ['docker', 'logs', test_container_name]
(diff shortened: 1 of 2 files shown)


cmaloney:
test that TeamCity if a test fails:
https://teamcity.mesosphere.io/viewLog.html?buildId=297095&buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr&guest=1
https://teamcity.mesosphere.io/viewLog.html?buildId=297097&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2
https://teamcity.mesosphere.io/viewLog.html?buildId=297091&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp

cmaloney:
Actually, all those exited with code 1, meaning this didn't work...

Looking for alternate fixes....

test_util/test_runner.py
```diff
@@ -150,18 +150,23 @@ def integration_test(
...
-            tunnel.remote_cmd(test_cmd, timeout=timeout)
+            try:
+                tunnel.remote_cmd(test_cmd, timeout=timeout)
+            except CalledProcessError:
+                log.exception('Test failed!')
+                if ci_flags:
+                    return False
+                raise
         log.info('Successful test run!')
     except TimeoutExpired:
```
mellenburg: TimeoutExpired Error must now be indented to be in order to catch the tunnel.remote_cmd(test_cmd, timeout=timeout)

cmaloney:
So, after further investigation, The only one with non-zero exit code is the Vagrant test, which is expected (it doesn't use the test runner).

Experimenting with a downstream, this does make it so in a downstream TeamCity CI running the integration test, we can mute individual tests.

PR #219, opened 2016-06-15
python: Take control of setuptools and pip version

spahl:
- setuptools 21.0.0
- pip 0.8.2


cmaloney:
I'm still not a fan of this, we're making our python be yet another special snowflake with it, for the benefit of one single line of code in one package which breaks with the builtin pip...

That said, we've had the argument before, so a begrudging 🚢

PR #225, opened 2016-06-16
Use hash instead of r_user@host:port for ctrl path

mellenburg:
Otherwise path name will be too long for a unix socket


cmaloney:
🚢

PR #227, opened 2016-06-16
Add a version flag which prints the bootstrap ID for dc/os

malnick:


malnick:
@cmaloney This is ready for review.

CC @spahl

cmaloney:
We could make a flag `--bootstrap-id` that returned this, but it's definitely not the DC/OS release `--version`

PR #229, opened 2016-06-16
Allow setting provider in integration_test.py

mellenburg:
Required for activating AWS and Azure integration testing


cmaloney:
:ship:

PR #230, opened 2016-06-16
use latest 3dt

darkonie:
add snapshot functionality
https://github.com/dcos/3dt/pull/28

### DIFF
--- a/test_util/integration_test.py
+++ b/test_util/integration_test.py
@@ -2,8 +2,10 @@ import collections
 import json
 import logging
 import os
+import tempfile
 import urllib.parse
 import uuid
+import zipfile
 
 import boto3
 import botocore.exceptions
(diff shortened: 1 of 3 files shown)


darkonie:

test_util/integration_test.py
```diff
@@ -1422,3 +1424,159 @@ def test_mesos_agent_role_assignment(cluster):
...
+
+        # check `is_running` attribute for each host. All of them must be False
+        for _, attributes in response.items():
+            assert not attributes['is_running']
+
+    wait_for_job()
+
+    # the job should be complete at this point.
+    # check the listing for a zip file
+    snapshots = _get_snapshot_list(cluster)
```
cmaloney: Is there some way we can test that the snapshot has some of the data we want in it?

test_util/integration_test.py
```diff
@@ -1422,3 +1424,159 @@ def test_mesos_agent_role_assignment(cluster):
...
+    expected_master_files = ['dcos-mesos-master.service.gz'] + expected_common_files
+
+    # for agent host
+    expected_agent_files = ['dcos-mesos-slave.service.gz'] + expected_common_files
+
+    # for public agent host
+    expected_public_agent_files = ['dcos-mesos-slave-public.service.gz'] + expected_common_files
+
+    try:
+        tmp_dir = tempfile.mkdtemp()
```
cmaloney: Python has https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory to make it so you don't have to hand write the cleanup it in the try / except here.

test_util/integration_test.py
```diff
@@ -1422,3 +1424,159 @@ def test_mesos_agent_role_assignment(cluster):
...
+        download_base_url = '/system/health/v1/report/snapshot/serve'
+        for snapshot in snapshots:
+            snapshot_full_location = os.path.join(tmp_dir, snapshot)
+            with open(snapshot_full_location, 'wb') as f:
+                r = cluster.get(path=os.path.join(download_base_url, snapshot), stream=True)
+                for chunk in r.iter_content(1024):
+                    f.write(chunk)
+
+            # validate snapshot zip file.
+            print(snapshot_full_location)
```
cmaloney: print() is used here but logging() is used above. Should be consistent. logging is the most frequently used.

test_util/integration_test.py
darkonie: fixed.
response['extra']['snapshot_name'] contains the snapshot unique filename.

test_util/integration_test.py
darkonie: done

test_util/integration_test.py
darkonie: done

darkonie:
@cmaloney this is ready for round 2 code review. addressed you comments.

cmaloney:
Waiting for an out of band go code review from one of our go experts

PR #231, opened 2016-06-16
Run tests without reconfiguring user for sudo

mellenburg:
Changing the user permissions on a host is a pretty bad paradigm for tests

### DIFF
--- a/test_util/installer_api_test.py
+++ b/test_util/installer_api_test.py
@@ -43,6 +43,7 @@ class AbstractDcosInstaller(metaclass=abc.ABCMeta):
             def scp(src, dst):
                 return tunnel.write_to_remote(src, dst)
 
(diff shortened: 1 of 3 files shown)


mellenburg:
@cmaloney This is ready for review. This change is not *strictly* necessary for enabling AWS and Azure tests, but it does make those test implementations much cleaner (as the test will no longer be modifying the host user groups)

cmaloney:
@mellenburg I'm actually really not a fan of using sudo here to access docker. It potentially messes up permissons a whole lot more (More files which are un-deletable), and requiring passwordless sudo I think is a worse option than requiring that the user the integration test is run under has permissions to access docker.sock

cmaloney:
CI failures are real bugs: https://teamcity.mesosphere.io/viewLog.html?buildId=303245&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp&tab=buildLog#_focus=282

PR #236, opened 2016-06-17
Fix mkdir in fetch_packages.sh

branden:


cmaloney:
Is there any way we can catch this?

Long term we definitely need to work on reducing these random bash scripts (esp. here we aren't even setting the bash safety options, `set -euo pipefail; set -x`)

branden:
I could write a test that runs the script and verifies that it downloads the correct packages, but that'd require a fix for the hardcoded `/stable` path in the `bootstrap_url`. The root problem is that nothing is currently depending on this script. It'll ultimately be tested by the upcoming upgrade integration test.

PR #238, opened 2016-06-17
Added disk free check for /var/lib to check_all() function

tkrausjr:
Added DF check for /var/lib to make sure that Agents have enough space to hold Docker images and Mesos Sandbox components. Note sure what the best number was to check for so I used relatively low (20 GB) but show the user what they have available.

### DIFF
--- a/gen/installer/bash.py
+++ b/gen/installer/bash.py
@@ -266,6 +266,17 @@ function check_all() {
     check_selinux
     check_sort_capability
 
+     # Check to see that the /var/lib partition has enough space
+    varFree=$(( $( df /var/lib | tail -1 | awk '{print $4}')/1024/1024 ))
+    varFreeInt=${varFree%.*}
+    if [ $varFreeInt -ge 20 ]; then
+        varFreeStatus="PASS"
+    else
+        varFreeStatus="FAIL"
+    fi
+    echo "/var has" $varFree "GB available" 
+    echo $varFreeStatus
+
     local docker_version=$(command -v docker >/dev/null 2>&1 && docker version 2>/dev/null | awk '
         BEGIN {
             version = 0


cmaloney:
@tkrausjr Needs to be rebased on master to get the code quality check to pass.

That said, I don't think this is the right approach, and don't think it would work in a lot of cases
 - We care about three directories
   - `/var/lib/docker` or whatever the docker storage directory is set to (May be something else entirely). On our internal hosts we have installed, `/var/lib/docker` is it's own special isolated mount
   - `/var/lib/mesos` on agents, because all the task sandboxes go here
   - `/opt/mesosphere` (Although the directory can't exist until `dcos-download.service` exits...)
   - `/var/lib/dcos` Where all the state we store which isn't the mesos agent goes (Zookeeper, dcos-history-service, Mesos Master replicated log, 3dt, among others)

20GB is also actually bigger than we need (And bigger than our CF templates ship). There are also a number of customers I know of who don't have that much free space. Generally just means we have to do more agressive GC (which Mesos does, Docker not so much but internally there is DCOS-7618 to hopefully before 1.8 ships add a basic systemd timer which will do a docker gc)

tkrausjr:
Yep - I agree it is situation specific.  What I was looking for is In Enterprise space we run into issues where Mesos sandbox and Docker Pulls fill up a filesystem /var/lib as you said by default very quickly (Even with GC) and customers don't understand how these directories are used. I guess ultimately I would like to see a warning put that still allows customers to proceed with the install so a RED Fail with a warning message explaining why we are noting to exception. Any ideas how we could do that especially for Enterprise customers ? Would there be a way to just WARN or implement this in JUST the EE version ?

spahl:
Closing because not the approach we want. Please file a ticket for the outstanding comments. Thanks for the PR though:-)

PR #239, opened 2016-06-17
Don't rebuild Mesos if Exhibitor changes

alberts:


cmaloney:
:ship:

cmaloney:
Note that in the future, ideally Exhibitor and Mesos should build against / with the _same_ zookeeper (Mesos currently vendors its own).

PR #240, opened 2016-06-18
Integration Testing for AWS provided deployments

mellenburg:


cmaloney:
As a note on this, we need to keep running some integration tests outside of AWS. There is a decent chunk of AWS-specific code in some of the components of DC/OS (like rex-ray), and when those get don't work when not in AWS, we won't find it if we move all our tests to live on top of AWS instances.

mellenburg:
@spahl The last issue stopping this is that test_signal_service fails because not all the units are healthy by the time the test runs (~2 minutes after the cluster was deemed available by CCM)
How do we go about fixing this? Require that no tests run until all units are healthy? Make the test retry?

malnick:
@mellenburg Do you know if the units were not actually up when signal ran? Did you manually poke around on the master to determine if it was signal not behaving (or 3DT) or if the units were actually unhealthy?

mellenburg:
@malnick haven't checked that yet, but at the moment I would guess that these are correct reports. When bringing up this or an azure deployment, we don't really have guarantee that the units should be passing when we are told 'cluster ready', unlike the onprem installation where the installer does check units for us. Rather, we just spin up the instances and wait for the UI, capabilities, and some endpoints to come up before trying to run a test

mellenburg:
@malnick and of course this last test passes without any issue, which lends towards my point that signal and 3dt are probably doing their job just fine

cmaloney:
@mellenburg There are open JIRAs about making `dcos-cfn-signal` only say that the cluster is up after it's really "up". Right now it pings with "leader.mesos" is resolvable. Sounds like we should be using a better mechanism.

https://github.com/dcos/dcos/blob/master/gen/installer/aws.py#L57

Ideally that, the SSH / Web Installer, the Azure, and the integration tests (wait for up + agents joined stuff), all should use the same exact mechanism to determine when nodes in the cluster are really "up" / ready for work.

mellenburg:
@cmaloney Do you think this PR should wait until we have addressed that issue? I would much rather us move ahead on this and have some global visibility/reproducability on the issue rather than keep it relegated to this PR

cmaloney:
This has a couple of other issues before it could land in general. At the top level I see:
 - If we want to test the AWS CF templates, we should launch that directly using boto3, not use CCM.
 - The CCM VPC changes are entirely independent of the AWS CF changes, should be separate PRs.
 - The reservations tag stuff doesn't make sense to me. That we're using "reserved" instances shouldn't leak through into this code at all. If it needs to, that needs a separate PR with an explanation in the commit why.

malnick:
A few comments:

- We're adding new flags to do the gen and deployment to S3 of the AWS CF templates, so this code will be changing as soon as that work is complete.

- That being said, we have zero testing capabilities on our current advanced or basic templates. With that in mind. I would prefer, since this work is already done, to merge this in and see where our templates are at in terms of testing. We're going to vaporize this work anyways, so updating it to use Boto lib (which I agree, is the correct way to do this over CCM), is a waste right now - this entire test framework will be re-written completely with the new dcos_gen_config work.

I vote to merge this in now, so we have some intermediate testing on our CF templates, and I'll work with Michael in the re-write as the dcos_gen_config work progresses, ensuring we use boto lib to interact with AWS services, etc. 

CC @spahl - would like your input, thanks.

mellenburg:
@cmaloney This test has been refactored with Boto as the provider interface, however before it can land as a PR check. We need to teach cloud cleaner to erase old ssh keys that are generated (also a req for getting vpcs provisioned with boto)

So, should we just add logic to cloud cleaner that says something like:
1. get all key names that start with CF-integration-test-
2. for each of those keys, if their instance isn't alive, delete that key

What do you think?

cmaloney:
@mellenburg for now I'm fine with just doing a prefix match and delete. We should i think also be able to get the ssh key associated with every instance as part of cloudcleaner, and delete any which don't have any instances and have existed more than a certain amount of time except some specific whitelists (default). Generally want everyone to always use the default key.

I'll review the code later today

mellenburg:
BTW, https://github.com/dcos/dcos/pull/622 is dependent on this and also ready for review

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
+logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+AWS_VERIFY = True
+
+
+class AwsProvision(metaclass=ABCMeta):
```
cmaloney: what's the value of having this and CloudFormation be separate classes? They both would only work / make sense as currently written for CloudFormation stacks (things like self.stack, self.stack_name, only make sense for CloudFormation)

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+        self.region = region
+        self.session = boto3.session.Session(
+            aws_access_key_id=os.environ['AWS_ACCESS_KEY_ID'],
+            aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'])
+        self.ec2_client = self.session.client(
+            service_name='ec2',
+            region_name=self.region,
+            verify=AWS_VERIFY)
+
+    def set_expiration(self, expiration_string):
```
cmaloney: Anything created by AWS CloudFormation falls under the jurisdiction of of CCM for cleanup rather than the owner + expiration CloudCleaner does.

Need to use: https://ccm.mesosphere.com/api-docs/#!/cluster/Cluster_List_POST to set / extend the expiration.

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+                    wait_fixed=3000,
+                    retry_on_exception=lambda ex: False,
+                    retry_on_result=lambda res: res is False)
+    def wait_for_cluster_creation(self):
+        logger.info('Waiting for cluster to be created...')
+        self.stack.reload()
+        current_status = self.stack.stack_status
+        if current_status == 'CREATE_IN_PROGRESS':
+            return False
+        if current_status != 'CREATE_COMPLETE':
```
cmaloney: Should detect CREATE_FAILED here (so we can exit early when we know it's dead dead), as well as hopefully do status logging or something similar to help with debugging state over time.

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+            raise Exception("Expected stack {} to have status CREATE_COMPLETE,"
+                            "but found: {}".format(self.stack.stack_id,
+                                                   current_status))
```
cmaloney: should also make sure to long `stack_status_reason`

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+                    wait_fixed=3000,
+                    retry_on_exception=lambda ex: False,
+                    retry_on_result=lambda res: res is False)
+    def wait_for_cluster_deletion(self):
+        logging.info('Waiting for cluster to be deleted...')
+        current_status = self.stack.stack_status
+        if current_status == 'DELETE_IN_PROGRESS':
+            return False
+        try:
+            self.stack.reload()
```
cmaloney: if we use the boto3 client directly (https://boto3.readthedocs.io/en/latest/reference/services/cloudformation.html#CloudFormation.Client.describe_stacks) we can set `StackName=<stack_id>` at which point the deleted stacks will still show up and can actually look for just DELETE_COMPLETE simplifying this code a bit.

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+                logging.info("Boto cannot find stack; deletion successful")
+                return True
+            else:
+                raise ex
+        if current_status != 'DELETE_COMPLETE':
+            raise Exception('Delete finished, but cluster has unexpected state: '+current_status)
+
+
+class CloudFormation(AwsProvision):
+    def create(self, name, template_url,
```
cmaloney: I don't think this is quite how we want to structure this. I think it would be better to have a generic CloudFormation provisioner that can do stack standup / shutdown with an arbitrary template url + python format dictionary of parameters, then we use / call that for the AWS Simple, advanced, etc. stacks that way we don't have to do things like the simple python doct parameters -> complicated AWS ParameterKey + ParameterValue in multiple places.

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+                     {
+                     'ParameterKey': 'AdminLocation',
+                     'ParameterValue': admin_location},
+                     {
+                     'ParameterKey': 'PublicSlaveInstanceCount',
+                     'ParameterValue': str(public_agents)},
+                     {
+                     'ParameterKey': 'SlaveInstanceCount',
+                     'ParameterValue': str(private_agents)}]
+        if pre_1_8_cluster:
```
cmaloney: this code I don't think will ever be used for Pre-1.8 clusters, can probably just drop this.

test_util/provide_aws.py
```diff
@@ -0,0 +1,215 @@
...
+import boto3
+import botocore
+import retrying
+
+
+LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
+logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+AWS_VERIFY = True
```
cmaloney: I believe this is the default and we'd never want to set it differently. Do we need this as a parameter?

test_util/provide_aws.py
cmaloney: From an API persepctive: This seems much more like we should have an AWS Helper class, and if you say .create_stack() on it it returns an instance of a Stack() class which you can then do things like `.wait_for_up`, `.delete`, etc.

test_util/provide_aws.py
mellenburg: The intention was to split out the things that were to be used for arbitrary stacks in PR #622 and I couldn't figure out a good naming convention. How about
class CloudFormation-> DcosFormation(CloudFormation) & VpcFormation(CloudFormation) ?

test_util/provide_aws.py
mellenburg: Good point. Will do

test_util/provide_aws.py
mellenburg: Ah, thanks, I knew this way was hax on hax

test_util/provide_aws.py
mellenburg: I don't think there is a reason to keep it (added for debugging at one point)

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        if stack_id is None:
+            stack_id = self.stack.stack_id
+        stack_info = self.boto_wrapper.client('cloudformation').describe_stacks(StackName=stack_id)
+        logging.debug('Stack Description for {}: {}'.format(stack_id, stack_info))
+        return stack_info['Stacks'][0]
+
+
+class SimpleDcosStack(CFStack):
+    def create(self,  name, template_url, public_agents, private_agents, admin_location, key_pair_name):
```
cmaloney: This should be a @classmethod which returns a SimpleDCOSStack() which it constructs by name

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+                StackName=name,
+                TemplateURL=template_url,
+                DisableRollback=True,
+                TimeoutInMinutes=30,
+                Capabilities=['CAPABILITY_IAM'],
+                Parameters=cf_parameters)
+
+
+class CFStack():
+    def __init__(self, region, aws_access_key_id, aws_secret_access_key):
```
cmaloney: This should take two parts:
1. the stack name
2. a boto wrapper

Making this have all the parameters of botowrapper means that as we build / change botowrapper we have to change this vs. just "construct a botowrapper any way you can"

The constructor taking a stack name or stack id means thatthe CFStack object always can have a `self.stack`, and the constructor both works for the "new one" case as well as the "attach" case, both of which put the object into the same state.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        raise an excetion if the stack deletes and it cannot find stack ID
+        """
+        logging.info('Waiting for stack to finish: '+status)
+        stack_id = self.stack.stack_id
+        current_status, current_reason = self.get_stack_status(stack_id)
+        if(current_status == status):
+            return False
+        return current_status, current_reason
+
+    def get_stack_status(self, stack_id=None):
```
cmaloney: shouldn't take a stack_id, should always just use the class member (self.stack.stack_id).

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+    def attach(self, name):
+        self.stack = self.boto_wrapper.resource('cloudformation').Stack(name)
+
+    def wait_for_stack_creation(self):
+        self.wait_for_transition('CREATE_IN_PROGRESS', 'CREATE_COMPLETE')
+
+    def wait_for_stack_deletion(self):
+        self.wait_for_transition('DELETE_IN_PROGRESS', 'DELETE_COMPLETE')
+
+    def wait_for_transition(self, start_status, end_status):
```
cmaloney: Missing `@retry`?

Also: This should probably error if there is anything other than the expected transition.

cmaloney: I see the wait_for_status_change does the retry part of this.

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,139 @@
...
+    return options
+
+
+def main():
+    options = check_environment()
+
+    random_identifier = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
+    unique_cluster_id = "CF-integration-test-{}".format(random_identifier)
+    log.info("Spinning up AWS CloudFormation with ID: {}".format(unique_cluster_id))
+    cf = test_util.provide_aws.SimpleDcosStack(
```
cmaloney: When we get this all done, I think this would be something like:
```python
bw = BotoWrapper(aws_region=options.aws_region, aws_access_key_id=options.aws_access_key_id, aws_secret_access_key=options.aws_secret_access_key)
dcos = test_util.aws.DCOSCfSimple.create(name=unique_cluster_id, template_url=options.template_url, private_agents=options.private_agents, public_agents=options.public_agents, admin_location='0.0.0.0/0', key_pair_name='default')
dcos.cf_stack.wait_for_creation()
```

cmaloney:
```
with test_util.aws.DCOSCfSimple.create(name=unique_cluster_id, template_url=options.template_url, private_agents=options.private_agents, public_agents=options.public_agents, admin_location='0.0.0.0/0', key_pair_name='default', boto_wrapper=boto_wrapper, delete_on_exception=False) as cluster_info:
   <do_stuff>
```
could be even better potentially

mellenburg:
@cmaloney ready for another pass

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+
+import boto3
+import retrying
+
+LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
+logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+# Allows adjusting the timeout in CI
+AWS_WAITING_TIMEOUT = os.getenv('AWS_WAITING_TIMEOUT', 60)
```
cmaloney: AWS_CF_STACK_CREATION_TIMEOUT?

Reading this I think that it's something to do with the poll interval or something like that.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        return self.session.resource(service_name=name, region_name=self.region)
+
+    def create_key_pair(self, key_name):
+        """Retruns private key of newly generated pair
+        """
+        key = self.client('ec2').create_key_pair(KeyName=key_name)
+        return key['KeyMaterial']
+
+    def delete_key_pair(self, key_name):
+        self.client('ec2').delete_key_pair(KeyName=key_name)
```
cmaloney: `self.resource('ec2').KeyPair(name)`? (ala http://boto3.readthedocs.io/en/latest/reference/services/ec2.html#EC2.ServiceResource.KeyPair)

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+    def client(self, name):
+        return self.session.client(service_name=name, region_name=self.region)
+
+    def resource(self, name):
+        return self.session.resource(service_name=name, region_name=self.region)
+
+    def create_key_pair(self, key_name):
+        """Retruns private key of newly generated pair
+        """
+        key = self.client('ec2').create_key_pair(KeyName=key_name)
```
cmaloney: `self.resource('ec2').KeyPair(name)`? (ala http://boto3.readthedocs.io/en/latest/reference/services/ec2.html#EC2.ServiceResource.KeyPair)

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        self.client('ec2').delete_key_pair(KeyName=key_name)
+
+    def create_stack(self, name, template_url, user_parameters):
+        """Returns boto stack object
+        """
+        logging.info('Requesting AWS CloudFormation...')
+        cf_parameters = []
+        for k, v in user_parameters.items():
+            cf_parameters.append({'ParameterKey': k, 'ParameterValue': v})
+        return self.resource('cloudformation').create_stack(
```
cmaloney: This should return a CFStack instance.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+                StackName=name,
+                TemplateURL=template_url,
+                DisableRollback=True,
+                TimeoutInMinutes=AWS_WAITING_TIMEOUT,
```
cmaloney: nit: technically / from a separation of concerns point of view this should be an optional function parameter if we want it to be configurable.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+                Parameters=cf_parameters)
+
+
+class CFStack():
+    def __init__(self, stack_name, boto_wrapper):
+        self.stack_name = stack_name
+        self.boto_wrapper = boto_wrapper
+
+    def attach(self):
+        self.stack = self.boto_wrapper.resource('cloudformation').Stack(self.stack_name)
```
cmaloney: The getting the stack should actually be part of the CFStack() constructor / __init__. I'm pretty sure in all cases we're going to want to have a guarantee if we construct a CFStack, there is indeed a CloudFormation stack there rather than having to in every piece of code we use this class from a `stack = CFStack(...); stack.attach(); ...`. I don't see any of this classe's methods being particularly useful without having a stack attached.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        if(current_status == status):
+            return False
+        return current_status, current_reason
+
+    def get_stack_status(self):
+        """NOTE: stack_id must be externally defined for deleted stacks
+        """
+        stack_info = self.get_stack_info()
+        if 'StackStatusReason' in stack_info:
+            return stack_info['StackStatus'], stack_info['StackStatusReason']
```
cmaloney: `return self.stack.stack_status, self.stack.stack_status_reason`

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            return stack_info['StackStatus'], 'None'
+
+    def get_stack_info(self):
+        """NOTE: stack_id must be externally defined for deleted stacks
+        """
+        stack_info = self.boto_wrapper.client('cloudformation').describe_stacks(StackName=self.stack.stack_id)
```
cmaloney: `return self.stack` seems like it would be equivalent to this

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        logging.debug('Stack Description for {}: {}'.format(self.stack.stack_id, stack_info))
+        return stack_info['Stacks'][0]
+
+
+class DcosCfSimple(CFStack):
+    @classmethod
+    def create(cls, stack_name, template_url, public_agents, private_agents,
+               admin_location, key_pair_name, boto_wrapper):
+        ret = DcosCfSimple(stack_name, boto_wrapper)
```
cmaloney: So this is ordered wrong / we never want to have a DcosCfSimple object without having an underlying CFStack I don't think.

cmaloney: Also saves us from having to do all the late initialization of self.stack, as well as the slightly awkward initialization of the CFStack internal structure which really breaks separation of concerns.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+            'PublicSlaveInstanceCount': str(public_agents),
+            'SlaveInstanceCount': str(private_agents)}
+        ret.stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
+        return ret
+
+    def delete(self):
+        logger.info('Starting deletion of CF stack')
+        self.stack.delete()
+
+        bucket_id = self.stack.Resource('ExhibitorS3Bucket').physical_resource_id
```
cmaloney: Probably best to grab all of these things before doing the delete just to be safe.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def wait_for_transition(self, start_status, end_status):
+        final_status, status_reason = self.wait_for_status_change(start_status)
+        assert end_status == final_status, "Expected status {} but found {}. Reason: {}".format(
+                end_status, final_status, status_reason)
+
+    @retrying.retry(wait_fixed=5000, stop_max_delay=AWS_WAITING_TIMEOUT*60*1000,
+                    retry_on_result=lambda res: res is False,
+                    retry_on_exception=lambda ex: False)
+    def wait_for_status_change(self, status):
```
cmaloney: Would it make sense to utilize http://boto3.readthedocs.io/en/latest/reference/services/cloudformation.html#CloudFormation.Waiter.StackCreateComplete here?

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        return ret
+
+    def delete(self):
+        logger.info('Starting deletion of CF stack')
+        self.stack.delete()
+
+        bucket_id = self.stack.Resource('ExhibitorS3Bucket').physical_resource_id
+        s3 = self.boto_wrapper.resource('s3')
+        bucket = s3.Bucket(bucket_id)
+        all_buckets = s3.buckets.all()
```
cmaloney: why do we need to list all buckets? We just went and got the very particular bucket.... I don't see what double checking this here does.

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+        if bucket.name not in all_buckets:
+            logger.info('Bucket {} not found. Exiting..'.format(bucket))
+            return
+        logger.info("AWS start deleted buckets {}".format(bucket))
+        all_objects = bucket.objects.all()
```
cmaloney: This if set loose in the wrong AWS account could end very, very badly. Can we assert that there is one, and only one object inside the bucket? (That's all we ever put there with exhibitor).

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
...
+    def get_tag_instances(self, tag):
+        return self.boto_wrapper.client('ec2').describe_instances(
+            Filters=[
+                {'Name': 'tag-value', 'Values': [self.stack.stack_name]},
+                {'Name': 'tag-value', 'Values': [tag]}
+                ])['Reservations'][0]['Instances']
+
+    def get_master_ips(self):
+        instances = self.get_tag_instances('MasterServerGroup')
+        return [(i['PrivateIpAddress'], i['PublicIpAddress']) for i in instances]
```
cmaloney: namedtuple?

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+
+import test_util.provide_aws
+import test_util.test_runner
+from ssh.ssh_tunnel import SSHTunnel
+
+LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
+logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+DEFAULT_AWS_REGION = 'us-west-2'
```
cmaloney: Need to keep doing the environment variable fetching like the other script does for getting this that @alberts added the other day.

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+    Returns:
+        object: generic object used for cleanly passing options through the test
+
+    Raises:
+        AssertionError: if any environment variables or resources are missing
+            or do not conform
+    """
+    options = type('Options', (object,), {})()
+
+    options.variant = os.getenv('DCOS_VARIANT', 'default')
```
cmaloney: I think this is the model we're using some other places currently. I'd like to switch to:
```
from gen.calc import calculate_environment_variable

options.variant = calculate_environment_variable('DCOS_VARIANT')
```

Should make nicer error messages for all of these by default.

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+    assert options.aws_secret_access_key, 'AWS Credentials Required!'
+
+    add_env = {}
+    prefix = 'TEST_ADD_ENV_'
+    for k, v in os.environ.items():
+        if k.startswith(prefix):
+            add_env[k.replace(prefix, '')] = v
+    options.add_env = add_env
+    options.pytest_dir = os.getenv('DCOS_PYTEST_DIR', '/opt/mesosphere/active/dcos-integration-test')
+    options.pytest_cmd = os.getenv('DCOS_PYTEST_CMD', 'py.test -vv '+options.ci_flags)
```
cmaloney: `py.test -vv -rs` currently since #591 and #612

test_util/provide_aws.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python3
```
cmaloney: Why not `test_util/aws.py`?

Why do we need the `#!/usr/bin/env python3` here. This doesn't have a main().

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+    os.chmod(ssh_key_path, stat.S_IREAD | stat.S_IWRITE)
+
+    # Create custom SSH Runnner to help orchestrate the test
+    ssh_user = 'core'
+    remote_dir = '/home/core'
+
+    master_ips = cf.get_master_ips()
+    public_agent_ips = cf.get_public_agent_ips()
+    private_agent_ips = cf.get_private_agent_ips()
+    test_host = master_ips[0][1]
```
cmaloney: namedtuple for the IP tuple will make this all a lot nicer to use. See: https://docs.python.org/3/library/collections.html#collections.namedtuple

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+                public_agent_list=public_agent_list,
+                provider='aws',
+                test_dns_search=False,
+                aws_access_key_id=options.aws_access_key_id,
+                aws_secret_access_key=options.aws_secret_access_key,
+                add_env=options.add_env,
+                pytest_dir=options.pytest_dir,
+                pytest_cmd=options.pytest_cmd)
+    if result == 0:
+        log.info("Test successsful! Deleting CloudFormation and generated keypair...")
```
cmaloney: This talks about deleting the keypair but doesn't actually have any code to do so. Update the message to reflect the current code?

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+        stack_name=unique_cluster_id,
+        template_url=options.template_url,
+        private_agents=2,
+        public_agents=1,
+        admin_location='0.0.0.0/0',
+        key_pair_name='default',
+        boto_wrapper=bw)
+    cf.wait_for_stack_creation()
+
+    log.info("Exporting ssh_key to disk as")
```
cmaloney: This reads like there was supposed to be something after the `as`

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,140 @@
...
+    public_agent_ips = cf.get_public_agent_ips()
+    private_agent_ips = cf.get_private_agent_ips()
+    test_host = master_ips[0][1]
+    log.info('Running integration test from: '+test_host)
+    master_list = [i[0] for i in master_ips]
+    log.info('Master private IPs: '+repr(master_list))
+    agent_list = [i[0] for i in private_agent_ips]
+    log.info('Private agent private IPs: '+repr(agent_list))
+    public_agent_list = [i[0] for i in public_agent_ips]
+    log.info('Public agent private IPs: '+repr(public_agent_list))
```
cmaloney: Can we print to the screen what ssh key to use here for developers + the command (assuming a specific default key path), that a developer can use to get into the cluster?

test_util/provide_aws.py
mellenburg: Sure cleans up everything pretty well except it doesn't allow adjusting the timeout :( That being said, I was just going to use an hour timeout anyway

test_util/provide_aws.py
mellenburg: Why would this be more preferable?

cmaloney: The ec2 "resource" objects are the high level APIs that are structured for easier use / readability / management. The lower level "client" APIs are intended to be for boto3 internal implementation details as well as AWS services which don't have high level clients yet

test_util/test_aws_cf.py
mellenburg: Sounds good, except for this specific item (DCOS_VARIANT) I would like to provide a default valu

mellenburg:
@cmaloney Ready for another pass

test_util/aws.py
```diff
@@ -0,0 +1,109 @@
...
+            'KeyName': key_pair_name,
+            'AdminLocation': admin_location,
+            'PublicSlaveInstanceCount': str(public_agents),
+            'SlaveInstanceCount': str(private_agents)}
+        boto_wrapper.create_stack(stack_name, template_url, parameters)
+        return cls(stack_name, boto_wrapper)
+
+    def delete(self):
+        logger.info('Starting deletion of CF stack')
+        self.stack.delete()
```
cmaloney: Should we switch from a stack name based stack object to a stack id based one here?

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
+"""Deploys DCOS AWS CF template and then runs integration_test.py
```
cmaloney: DC/OS

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,139 @@
...
+        object: generic object used for cleanly passing options through the test
+
+    Raises:
+        AssertionError: if any environment variables or resources are missing
+            or do not conform
+    """
+    options = type('Options', (object,), {})()
+
+    # Defaults
+    options.variant = os.getenv('DCOS_VARIANT', 'default')
```
cmaloney: There is no such variant `default` in DC/OS. There is the python `None` variant, which wen stringified is the empty string, and we have a downstream which is an `ee` variant

test_util/test_aws_cf.py
```diff
@@ -0,0 +1,139 @@
...
+    options.ci_flags = os.getenv('CI_FLAGS', '')
+    options.aws_region = os.getenv('DEFAULT_AWS_REGION', 'us-west-2')
```
cmaloney: eu-central-1 please

cmaloney:
Once you get the test passing + some minor nits, I think this looks shippable as a v1, we can iterate more later if needed

test_util/aws.py
mellenburg: I'm not sure I see why. When self.stack is created, it will have a stack.stack_id set...

mellenburg:
@cmaloney once I get clarification on that last point, this is ready to go!

test_util/aws.py
cmaloney: Sounds good, I was thinking we wanted to make sure it was all based on a stack id rather than a stack name since from the docs stack_id based you can access after deletion

mellenburg: Ah, I was just thinking about how the waiter for delete would get around that issue, and forgetting a user might want to reload after deletion. Will update...

mellenburg:
@cmaloney this is ready to go
Note: i did have to add a retry to one of the the 3dt tests  as apparently AWS is a little slower to reach fully healthy units

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -283,6 +283,7 @@ def test_3dt_units(cluster):
         )
 
 
+@retrying.retry(wait_fixed=2000, stop_max_delay=LATENCY * 1000)
```
cmaloney: We really need to fix the "we don't wait for everything to come up before reporting healthy" problem. Although yes, we can leave that out of here for now.

This is causing at least:
 - https://dcosjira.atlassian.net/browse/DCOS-340
 - https://mesosphere.atlassian.net/browse/DCOS-4183
 - https://mesosphere.atlassian.net/browse/DCOS-3229
 - https://mesosphere.atlassian.net/browse/DCOS-6168

mellenburg: Should I sprinkle some #FIXME comments on those retries? Or ship as is?

test_util/aws.py
```diff
@@ -0,0 +1,113 @@
...
+import logging
+from collections import namedtuple
+
+import boto3
+
+LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
+logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+Host = namedtuple('Host', ['private_ip', 'public_ip'])
```
cmaloney: :+1: Eventually we should sink this into the general test_util / runner libraries probably (Azure, GCE have the same things)

packages/dcos-integration-test/extra/test_3dt.py
cmaloney: Ship as is, just noting what this comes from

test_util/aws.py
```diff
@@ -0,0 +1,113 @@
...
+        logging.info('Requesting AWS CloudFormation...')
+        cf_parameters = []
+        for k, v in user_parameters.items():
+            cf_parameters.append({'ParameterKey': k, 'ParameterValue': v})
+        self.resource('cloudformation').create_stack(
+            StackName=name,
+            TemplateURL=template_url,
+            DisableRollback=True,
+            TimeoutInMinutes=deploy_timeout,
+            Capabilities=['CAPABILITY_IAM'],
```
cmaloney: nit: capabilities is technically template-specific (DCOS stack will want it, but the VPC stacks don't need it), but we can move that later when the code actually needs it

test_util/aws.py
```diff
@@ -0,0 +1,113 @@
...
+class DcosCfSimple(CFStack):
+    @classmethod
+    def create(cls, stack_name, template_url, public_agents, private_agents,
+               admin_location, key_pair_name, boto_wrapper):
+        parameters = {
+            'KeyName': key_pair_name,
+            'AdminLocation': admin_location,
+            'PublicSlaveInstanceCount': str(public_agents),
+            'SlaveInstanceCount': str(private_agents)}
+        boto_wrapper.create_stack(stack_name, template_url, parameters)
```
cmaloney:
```
stack = boto_wrapper.create_stack()
return cls(stack.stack_id, boto_wrapper)
``` ?

File: test_util/aws.py (hunk 42 above)
Comment (mellenburg) [inline test_util/aws.py]: The next PR will definitely catch this

File: test_util/aws.py (hunk 43 above)
Comment (mellenburg) [inline test_util/aws.py]: Good catch

File: test_util/aws.py [hunk 44]
```diff
@@ -0,0 +1,113 @@
...
+        self.stack = self.boto_wrapper.resource('cloudformation').Stack(self.stack.stack_id)
+        self.stack.delete()
+        bucket_id = self.stack.Resource('ExhibitorS3Bucket').physical_resource_id
+        s3 = self.boto_wrapper.resource('s3')
+        bucket = s3.Bucket(bucket_id)
+        logger.info('Starting bucket {} deletion'.format(bucket))
+        all_objects = bucket.objects.all()
+        obj_count = len(list(all_objects))
+        if obj_count > 0:
+            assert obj_count == 1, 'Expected one object in Exhibitor S3 bucket but found: ' + str(obj_count)
```
cmaloney: random aside: cosmos is thinking of stashing things in this s3 bucket, we'll need to revisit if / when that happens.

test_util/aws.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def get_tag_instances(self, tag):
+        return self.boto_wrapper.client('ec2').describe_instances(
+            Filters=[
+                {'Name': 'tag-value', 'Values': [self.stack.stack_name]},
+                {'Name': 'tag-value', 'Values': [tag]}])['Reservations'][0]['Instances']
+
+    def get_master_ips(self):
+        instances = self.get_tag_instances('MasterServerGroup')
+        return [Host(i['PrivateIpAddress'], i['PublicIpAddress']) for i in instances]
```
cmaloney: instances_to_hosts function?

mellenburg: Sure, but it will be less-than-elegant function because it needs to know when there isnt a PublicIpAddress key. E.G.
```
hosts = []
for i in instances:
    host_i = Host(None, None)
    if 'PrivateIpAddress' in i:
         host.private_ip = i[..]
    if 'PublicIpAddress' in i:
        host.public_ip = i[...]
    hosts.append(hosts_i)
return hosts
```

Seemed to me that the one-liner comprehension was better than getting if's involved

cmaloney: I was thinking something like `return Host(i['PrivateIpAddress'], host['PublicIpAddress'] if 'PublicIpAddress' in i else None)`

which I don't think is too bad (I don't know of any way we get AWS instances without a private IP)

mellenburg: Oh yeah, I totally forgot you can just insert if statements like that in python. -_-

cmaloney: The more fun one is that is backwards for an inline if from most other languages (a ? b : c)

test_util/aws.py
```diff
@@ -0,0 +1,113 @@
...
+    @classmethod
+    def create(cls, stack_name, template_url, public_agents, private_agents,
+               admin_location, key_pair_name, boto_wrapper):
+        parameters = {
+            'KeyName': key_pair_name,
+            'AdminLocation': admin_location,
+            'PublicSlaveInstanceCount': str(public_agents),
+            'SlaveInstanceCount': str(private_agents)}
+        stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
+        return cls(stack.stack_id, boto_wrapper)
```
mellenburg: Already got it :)

mellenburg:
@cmaloney green and good to go. Once this is set I can open up resiliency tests and AWS advanced templates

cmaloney:
:ship: will put in the next train

PR #241, opened 2016-06-19
Fix DCOS-180 - Mesos-Dns error on ./build_local.sh - Permission denie…

jangelfdez:
…d cloning the repository

Changed to HTTPs url on mesos-dns github repository in order to avoid Public Key Denied authentication when running a DC/OS local build

Issue: https://dcosjira.atlassian.net/projects/DCOS/issues/DCOS-180?filter=allissues

PR #242, opened 2016-06-20
Fix typo for Mesos-DNS repo

surdy:
Fixing typo introduced here : https://github.com/dcos/dcos/pull/241


cmaloney:
:ship: 

Really need to add a periodic full end to end build from scratch test (This and a couple past errors would have been caught readily. This didn't cause the CI to fail since things were already checked out / cached)

PR #243, opened 2016-06-20
Fixing output format on invalid roles

dpetzel:
Set the output back to normal after the invalid role, rather than printing the literal string `"{NORMAL}"`


cmaloney:
:ship:

PR #245, opened 2016-06-20
allow master and slave roles

dpetzel:
I'm not sure this is even something you'll want to do, but figured I submit this for clarification just the same.

When I run the install script, passing both master and slave as roles (multiple roles seems to be supported based on the script) I get the following error:

`sh /root/genconf/serve/dcos_install.sh master slave -d`

```
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: symlink_all(role_dir, new)
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: File "/opt/mesosphere/lib/python3.4/site-packages/pkgpanda/__init__.py", line 522, in symlink_all
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: symlink_tree(src, dest)
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: File "/opt/mesosphere/lib/python3.4/site-packages/pkgpanda/__init__.py", line 397, in symlink_tree
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: os.symlink(src_path, dest_path)
Jun 20 13:23:04 n7-z01-0a2a2202 pkgpanda[11654]: FileExistsError: [Errno 17] File exists: '/opt/mesosphere/packages/3dt--c3970af261c1e9ed9014904a6408617dafd41e6c/dcos.target.wants_master/dcos-ddt.service' -> '/etc/systemd/system/dcos.target.wants.new/dcos-ddt.service'
```

With this change, it avoids the error, and seems to allow both roles to be run. I'm super green to dcos, so I don't know how to validate this 100% just yet, but so far seems ok on my setup.

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -413,7 +413,8 @@ def symlink_tree(src, dest):
             symlink_tree(src_path, dest_path)
         else:
             try:
-                os.symlink(src_path, dest_path)
+                if not os.path.exists(dest_path):
+                    os.symlink(src_path, dest_path)
             except FileNotFoundError as ex:
                 raise ConflictingFile(src_path, dest_path, ex) from ex
 


cmaloney:
Thanks for the PR. This is something that we had implemented in `DCOS` before it became the open source `DC/OS`. It's been moved away from for a couple reasons (below). Definitely it would be really nice to run a mesos agent on the masters, and there is even a prototype I put together a while back: https://github.com/dcos/dcos/pull/101.

Some history / background:
DC/OS used to allow a single host to have multiple roles. This enabled people to have a "single host vagrant" (Convenient for testing).

The symlink error is coming about because two different roles are both trying to put the same file into the same place. Both the `slave` role and `master` role happen to contain a role-specific `dcos-ddt.service` (`dcos-3dt.service in current master` per #223). The two services are not the same, and so just "ignoring" / avoiding the error, as this patch does, one of them gets chosen (Effectively at random) to be used, and the other ignored completely, making hard to debug runtime errors (versus refusing to start / come up. This sort of error can only be caused when developing DC/OS and doing something which would have lead to silent breakage / undefined behavior).

Generally we want a host to only have a "single" role for a couple reasons
 - It's a lot easier for engieneers to reason about a single than having to keep the whole set of role permutations straight
 - Most of the time we want something to behave differently on the different roles. For instance, the 3dt service has a "polling" behavior on the masters, whereas it just sits statically on the agents.
 - Software in two roles may conflict in unexpected ways. For instance, the `master` role causes a machine to start services which bind to ports which the "slave" role gives. The "slave" and "slave_public" roles if used together would cause mesos to trample all over itself.

I've actually had a patch in the past to remove the ability to specify multiple pkgpanda roles which are available on host at a given point in time. Unfortunately it broke a little bit I'd forgotten about on AWS / Azure and had to be reverted. Hoping to ressurect it when I get a bit more time.

dpetzel:
Thanks @cmaloney that makes sense. The "local development" setup is exactly what I was trying to create (although not through Vagrant). It would be really handy to be able to have some sort of "all in one" role.

cmaloney:
Agents on the master would give you that. Have to be a little careful to not tip it over and launching tasks is unfortunately more complicated than one would wish, but it's functional for really small.

Other options are https://github.com/dcos/dcos-docker or https://github.com/dcos/dcos-vagrant

Doing a manual install onto two local nodes / VMs of your choice (KVM, VMWare, etc.) I know some people do and works reasonably.

PR #247, opened 2016-06-20
Prep Installer for Many Packages

malnick:


cmaloney:
Copying the bootstrap is still going to be necessary. It's just going to be smaller.

malnick:
It's true, just realized this.

PR #250, opened 2016-06-21
Retry ensure routable

sargun:


cmaloney:
How are customers supposed to tell / be notified when the connection is ready? All their apps need to do something like this?

sargun:
@cmaloney Wait 5 seconds.

PR #251, opened 2016-06-21
WIP: Volume Mount Integration Tests

mellenburg:


lingmann:
@cmaloney I understand you and @mellenburg chatted about this PR and had some concerns, specifically around stopping/starting the agent. I'd like to carry on the discussion on this ticket so we can figure out next actions and get this to a mergeable state. It's important we move this along because we currently have no volume integration tests.

I believe the current approach taken with this PR is better than just testing the JSON generated, because it tests the full [interface](https://docs.mesosphere.com/1.7/administration/storage/mount-disk-resources/) we've provided to end users... it verifies that Mesos actually picks up the volume mounts *and* it is careful to leave the agent tested in a clean state when the test completes. It's also important to note that the bulk of this was originally written by Kartic and was reviewed previously, and now @mellenburg is just adapting it to work with the new testing harness.

Given the above, if you still think there are things that need to be addressed before this can land, can you please add to the discussion here?

mellenburg:
I'm closing this PR for the time being. The reasons are spelled out in https://mesosphere.atlassian.net/browse/DCOS-6050 where the conservation about this can continue.
TL;DR rebooting agents on AWS is terribly error-prone, ill-advised, and not consistent. Working through those 3 issues is more than was initially intended with the test automation request

PR #253, opened 2016-06-21
Update from Docker 1.11.0 to 1.11.2

lingmann:
* Fixes Docker issue with latest Azure Ubuntu 16.04 LTS image
  (16.04.201606100) where the Docker daemon does not successfully start
  during DC/OS boot up, causing Mesos Agents to also fail to start
* Update to Docker 1.11.2 and add new docker dependency: libltdl7
* Various Docker bug fixes, see:
  https://github.com/docker/docker/releases
* Improve install times and protect against edge case of another process
  locking the package database by adding exponential backoff to dpkg
  install wrapper
* Restart docker.socket once docker.service is successfully started (a work around for Docker bug https://github.com/docker/docker/issues/18444)

### DIFF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -10,8 +10,8 @@ root:
       Environment=DEBIAN_FRONTEND=noninteractive
       StandardOutput=journal+console
       StandardError=journal+console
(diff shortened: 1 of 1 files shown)


spahl:
:ship:

PR #257, opened 2016-06-21
Remove unused env variable

mellenburg:
Should have been caught when this variable was removed in the first place


cmaloney:
:ship:

PR #259, opened 2016-06-22
bump minuteman and navstar to use configurable ssl and auth tokens

spacejam:


cmaloney:
This seems to include a lot more commits than just that

PR #261, opened 2016-06-23
Update IAM to allow Cloudwatch stats publishing

lingmann:
Allow cloudwatch:PutMetricData from agent nodes.


cmaloney:
Re-engineerd, not needed to make what we're doing work.

PR #264, opened 2016-06-23
Deploy registry via marathon

mellenburg:
-Instead of deploying an insecure registry, a secure registry (with baked in certs) is grabbed from dockerhub mesosphere/test_registry:latest
-Registry is deployed as a marathon app and uses DCOS networking for local processes to reference it as registry.marathon.mesos.thisdcos.directory
-Building and pushing of test_server is done in a marathon app so that it can be dependent upon the registry deployment and also so that the test no longer requires any bootstrapping (e.g. starting and loading a registry, distributing certs, setting up a remote forward web, etc)


mellenburg:
@cmaloney @spahl This is still a little rough but functional, and it could use some feedback.

test_util/docker/test_registry/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry:2
+ADD client.key /client.key
+ADD client.cert /client.cert
```
malnick: @spahl Do we want to leave these in this PR?

test_util/integration_test.py
```diff
@@ -69,6 +69,74 @@ def auth_cluster(cluster):
...
+{test_server}
+EOF
+cat <<EOF > tmp/Dockerfile
+{dockerfile}
+EOF
+docker build -t {registry}/test_server tmp/
+docker push {registry}/test_server
+sleep 36000
+""".format(registry=cluster.registry, test_server=test_server, dockerfile=dockerfile)
+    docker_build_and_push_app = {
```
malnick: I'd like to see us put this in dockerhub too, and pin to a version in the test to make it easy to update the test against the hosted image in DH.

test_util/docker/test_registry/Dockerfile
malnick: disregard, I was confused, the hardcoded one is not being pulled from the web

malnick:
This looks good to me. I had one small qualm with the test_server image being built each time, but we can circle back to that at a later time.

cmaloney:
Docker Hub is incredibly unreliable and there's a reason all of this was engineered to not use it at all (It caused most failures for a while). The registry container needs to be built locally, or ideally we just scp the files to the right host, and volume mount them in.

cmaloney:
Actually we were pulling the registry before so that bit should be fine.

This is introducing new global state which is completely uncommitted though. Really would rather not have a artifact we rely on being pushed to docker hub independently of building / running the integration tests. Properly updating it without breaking all the other PRs in flight in the future is a hard problem.

If we just SCP the files to the right host and do a docker build, shouldn't that be enough?

Global state and when you get the "new" code committed here vs. the "global" thing, and people bumping the global stuff I'd really not get into unless we absolutely have to.

mellenburg:
How would updating mesosphere/test_registry break any other PRs in flight? With this approach, we are using a public CA bundle, so as long as we remember to push a new mesosphere/test_registry:latest with an updated cert in around a year when this free one expires, everything will keep on working as if nothing ever changed . No PRs will be aware of a 'new' or 'old' registry as long as the cert in mesosphere/test_registry is valid.

Also, for what its worth, the first commit in this PR shows what its like to hard-code a self-signed cert and scp everything and its downright hideous to look at. By adding this single point of failure (not updating the certs properly within the next year), we are able to get rid of all the hacks for getting around secure registry, plus, we get the even bigger benefit of not bootstrapping tests before they run

cmaloney:
@mellenburg We need to upgrade the registry docker container to v3 for some reason. Every single running test now pulls the "new" one, even though half of them don't have the other compatible changes.

It's roughly the same reason why we build test_server and the similar docker containers every time. More critical for those ones since people change them more often,.

Just doing a simple scp of three files, and then you don't even need to do a `docker build`, just pin the registry to a single host in the cluster using Marathon constraints and volume mount in the files + set env vars. Shouldn't be more than 50 lines of fairly straightforward code.

Definitely not re-building the Certs everytime or anything like that.

PR #267, opened 2016-06-24
add snapshot size

darkonie:
API should return snapshot file name along with its size. This will allow
to warn a user if snapshot file is too big.
https://github.com/dcos/3dt/pull/42


darkonie:

PR #271, opened 2016-06-25
Add missing CoreOS units

cmaloney:
These got lost in #132, making it only into the AWS Advanced.

Fixes https://dcosjira.atlassian.net/browse/DCOS-200


cmaloney:
Launched a cluster, agents how have 34GB of disk rather than 4GB.

PR #273, opened 2016-06-26
Bootstrap

spahl:
Replace exhibitor_wait, zk-value-consensus and cluster id generation with one bootstrap script.


packages/bootstrap/extra/bin/bootstrap
```diff
@@ -0,0 +1,102 @@
...
+    log.info('Clearing proxy environment variables')
+    os.environ.pop('HTTP_PROXY', None)
+    os.environ.pop('http_proxy', None)
+    os.environ.pop('HTTPS_PROXY', None)
+    os.environ.pop('https_proxy', None)
+    os.environ.pop('NO_PROXY', None)
+    os.environ.pop('no_proxy', None)
+
+    dirs = [
+        '/var/lib/dcos',
```
cmaloney: pkgpanda makes all the `/var/lib` state directories already.

spahl: true. I'll remove it and leave /run/dcos (or is that created too?)

cmaloney: pkgpanda doesn't do `/run/dcos` currently. It probably could / should learn to, but for now having it here works. Will need to `chmod` it to the service-appropriate user.

packages/bootstrap/extra/Makefile
```diff
@@ -0,0 +1,26 @@
+TOP := $(dir $(lastword $(MAKEFILE_LIST)))
+
+.PHONY: test install clean
+
+# You might like: make PYTEST_OPTS="-s -v" test
+PYTEST_OPTS := -v
+
+test: export MESOS_IP_DISCOVERY_COMMAND = $(TOP)/tests/ip-detect
```
cmaloney: I'd rather make the py.test tests run via the top level tox.ini than another new embedded thing

packages/bootstrap/extra/bin/bootstrap
```diff
@@ -0,0 +1,102 @@
...
+        log.error('bootstrap must be run as root')
+        sys.exit(1)
+
+    opts = parse_args()
+
+    logging.basicConfig(format='[%(levelname)s] %(message)s', level='INFO')
+    log.setLevel(logging.DEBUG)
+
+    log.info('Clearing proxy environment variables')
+    os.environ.pop('HTTP_PROXY', None)
```
cmaloney:
```python
for name in ['HTTP_PROXY', 'HTTPS_PROXY', 'NO_PROXY']:
  os.environ.pop(name, None)
  os.environ.pop(name.lower(), None)
```

packages/bootstrap/extra/Makefile
spahl: Yeah I was planning on adding it to tox. Just need to clean up the tests a bit first.

packages/bootstrap/extra/bin/bootstrap
spahl: :+1:

packages/bootstrap/extra/bin/bootstrap
```diff
@@ -0,0 +1,102 @@
...
+    os.environ.pop('no_proxy', None)
+
+    dirs = [
+        '/var/lib/dcos',
+        '/run/dcos',
+        '/var/lib/dcos/dcos-oauth',
+    ]
+    for d in dirs:
+        log.info('Preparing directory {}'.format(d))
+        os.makedirs(d, exist_ok=True)
```
cmaloney: The run dirs all need a `chown` or we shouldn't make them currently since we aren't using them.

packages/bootstrap/extra/bin/bootstrap
spahl: Will add it to pkgpanda in a separate pr.

packages/bootstrap/extra/bin/bootstrap
```diff
@@ -0,0 +1,102 @@
...
+    for service in opts.services:
+        if service not in bootstrappers:
+            log.error('Unknown service: {}'.format(service))
+            sys.exit(1)
+        log.debug('bootstrapping {}'.format(service))
+        bootstrappers[service](b, opts)
+
+
+def parse_args():
+    if os.path.exists('/etc/mesosphere/roles/master'):
```
cmaloney: nit: We should probably make pkgpanda contain a utility function to get the roles which were present when the last `pkgpanda activate` raun (in theory the list of roles could change)

packages/bootstrap/extra/dcos_internal_utils/bootstrap/__init__.py
```diff
@@ -0,0 +1 @@
+from .bootstrap import *  # noqa
```
cmaloney: why not just write the code in `__init__.py` if we're going to do this?

packages/bootstrap/extra/dcos_internal_utils/bootstrap/bootstrap.py
```diff
@@ -0,0 +1,101 @@
...
+                if os.path.exists(path):
+                    fileid = utils.read_file_line(path)
+                    if fileid == zkid:
+                        log.info('Cluster ID in ZooKeeper and file are the same: {}'.format(zkid))
+                        return zkid
+
+                log.info('Writing cluster ID from ZK to {} via rename'.format(path))
+
+                tmppath = path + '.tmp'
+                with open(tmppath, 'w') as f:
```
cmaloney: Should make this a helper function to do an "atomic write", then have both this and _write_file use it.

packages/bootstrap/extra/dcos_internal_utils/exhibitor/exhibitor.py
```diff
@@ -0,0 +1,43 @@
+import codecs
+import json
+import logging
+import os
+import urllib.request
```
cmaloney: I'd rather we just use requests (All the rest of our code was ported to requests only)

packages/bootstrap/extra/bin/bootstrap
spahl: we will not make them. I will add it to pkgpanda.

packages/bootstrap/extra/dcos_internal_utils/exhibitor/exhibitor.py
spahl: +1

packages/bootstrap/extra/dcos_internal_utils/bootstrap/bootstrap.py
spahl: Yeah I notice the similarity yesterday.

spahl:
@alberts PTAL

packages/bootstrap/extra/dcos_internal_utils/bootstrap/bootstrap.py
alberts: yeah, this code was written quite early on. should be able to make it use the same stuff

spahl:
@cmaloney All comments addressed and more testing added. If @alberts is happy with it I'm merging this.

alberts:
:ship: 

few small comments, but we can always do them later

packages/bootstrap/extra/dcos_internal_utils/exhibitor/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+
+    cluster_size = int(utils.read_file_line(master_count_filename))
+    log.info('Expected cluster size: {}'.format(cluster_size))
+
+    log.info('Waiting for ZooKeeper cluster to stabilize')
+    try:
+        response = requests.get(EXHIBITOR_STATUS_URL)
+    except requests.exceptions.ConnectionError as ex:
+        log.error('Could not connect to exhibitor: {}'.format(ex))
+        sys.exit(1)
```
spahl: ok. right now it's only used in one place. When we need this we'll address it.

packages/bootstrap/extra/dcos_internal_utils/bootstrap/bootstrap.py
spahl: Well it's done. _write_file is the utility function now.

spahl:
:ship:

PR #278, opened 2016-06-27
Update IAM to allow Cloudwatch stats publishing

spahl:
Allow cloudwatch:PutMetricData and autoscaling:DescribeAutoScalingGroups from agents.


cmaloney:
🚢 . Ideally we'd have something in open that would actually use / see this, but it's not too complicated and makes it easier for people to get autoscaling more right.

PR #284, opened 2016-06-27
try to list snapshots after the job is done

darkonie:
sometimes it can take up to 2-3 seconds to list the available snapshots off the
filesystem after job is finished.


spahl:
:ship:

PR #286, opened 2016-06-27
Switch bootstrap to use entry_points, cleanup packaging a little

cmaloney:
Should help with some of the dirty build problems we've been seeing

### DIFF
--- a/packages/bootstrap/extra/dcos_internal_utils/bootstrap/cli.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/bootstrap/cli.py
@@ -84,7 +84,3 @@ def parse_args():
         default='/opt/mesosphere/etc/master_count',
         help='File with number of master servers')
     return parser.parse_args()
-
-
-if __name__ == '__main__':
-    main()
--- a/tox.ini
+++ b/tox.ini
@@ -82,5 +82,5 @@ deps=
   teamcity-messages
 changedir=packages/bootstrap/extra
 commands=
-  python setup.py install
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -82,5 +82,5 @@ deps=
   teamcity-messages
 changedir=packages/bootstrap/extra
 commands=
-  python setup.py install
+  pip install .
```
orsenthil: To maintain consistency with `packages/bootstrap/build` using `pip3` might be a good idea.

spahl:
:ship:

packages/bootstrap/extra/dcos_internal_utils/bootstrap/cli.py
```diff
@@ -84,7 +84,3 @@ def parse_args():
         default='/opt/mesosphere/etc/master_count',
         help='File with number of master servers')
     return parser.parse_args()
-
-
-if __name__ == '__main__':
-    main()
```
cmaloney: nit: should remove the shebang at the top as well, but needing to land this quickly as a _lot_ of builds are borked currently.

tox.ini
cmaloney: here we should be running inside of a python 3.4 virtual environment, so _should_ be safe. That said, be a nice after this lands misc. cleanup PR (Trying to land this quick to un-break)

cmaloney: Should actually replace all `pip` usage in our codebase with `pip3` (This is far from the only one)

PR #298, opened 2016-06-29
Add metronome to the masters as the new jobs backend

spahl:


spahl:
:ship:

packages/metronome/build
```diff
@@ -0,0 +1,26 @@
...
+cp -a * "$PKG_PATH/"
+cp -f /pkg/extra/logback.xml "$PKG_PATH/conf/"
+
+metronome_service="$PKG_PATH/dcos.target.wants_master/dcos-metronome.service"
+mkdir -p $(dirname "$metronome_service")
+
+cat <<EOF > "$metronome_service"
+[Unit]
+Description=Jobs Service: DC/OS Metronome
+After=dcos-mesos-master.service
```
cmaloney: Don't use before/ after, they don't actually do what is wanted.

packages/metronome/build
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+pushd "/pkg/src/metronome"
+rm -rf bin/metronome.bat README.md RUNNING_PID share/ logs/
```
cmaloney: Metronome is new, why are things like RUNNING_PID inside the package? That seems like something we should fix in the upstream.

packages/metronome/build
```diff
@@ -0,0 +1,26 @@
...
+After=dcos-mesos-master.service
+
+[Service]
+Restart=always
+StartLimitInterval=0
+RestartSec=15
+EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=/opt/mesosphere/etc/metronome
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-metronome
+ExecStartPre=/bin/bash -c 'echo "METRONOME_LEADER_ELECTION_HOSTNAME=\$(\$MESOS_IP_DISCOVERY_COMMAND)" > /run/dcos/etc/metronome/service.env'
```
cmaloney: Why are we building these hacks into new code instead of just making it work right?

PR #300, opened 2016-06-30
Add support for bootstrap_package_list in treeinfo files

branden:
This PR allows treeinfo files to include a `bootstrap_package_list`, which declares the packages that are installed into the bootstrap tarball. This changes the semantics of `core_package_list` to mean the set of all packages that comprise a DC/OS release, of which `bootstrap_package_list` may be a subset.

If `bootstrap_package_list` isn't declared, it defaults to `core_package_list`. This means existing treeinfo files will produce the same build output.

Packages that aren't included in the bootstrap tarball are not yet included in a release, so `bootstrap_package_list` should not be used except for testing. A followup PR will modify release code to include all built packages along with the bootstrap tarball.

### DIFF
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -164,6 +164,22 @@ class PackageSet:
         )
         self.validate_package_tuples(self.all_packages, treeinfo, package_store)
 
+        if treeinfo.bootstrap_package_list:
+            self.bootstrap_packages = self.package_tuples_with_dependencies(
+                treeinfo.bootstrap_package_list,
+                treeinfo,
+                package_store
+            )
+            self.validate_package_tuples(self.bootstrap_packages, treeinfo, package_store)
+        else:
+            self.bootstrap_packages = self.all_packages
+
+        # Validate bootstrap packages are a subset of all packages.
+        for package_name, variant in self.bootstrap_packages:
+            if (package_name, variant) not in self.all_packages:
+                raise BuildError("Bootstrap package {} (variant {}) not found in set of all packages".format(
+                    package_name, pkgpanda.util.variant_name(variant)))
+
     @staticmethod
     def package_tuples_with_dependencies(package_names, treeinfo, package_store):
         package_tuples = set((name, treeinfo.variants.get(name)) for name in set(package_names))
@@ -629,7 +645,7 @@ def build_tree(package_store, mkbootstrap, tree_variant):
     def make_bootstrap(package_set):
         print("Making bootstrap variant:", pkgpanda.util.variant_name(package_set.variant))
         package_paths = list()
-        for name, pkg_variant in package_set.all_packages:
+        for name, pkg_variant in package_set.bootstrap_packages:
             package_paths.append(built_packages[name][pkg_variant])
 
         if mkbootstrap:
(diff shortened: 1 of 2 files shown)


branden:
@spahl @cmaloney This is ready for review.

pkgpanda/build/__init__.py
```diff
@@ -100,6 +100,111 @@ def get_src_fetcher(src_info, cache_dir, working_directory):
...
+
+        """
+        excludes = excludes or list()
+        package_list = treeinfo_dict.get(key, list())
+
+        # Validate package list.
+        if not isinstance(package_list, list):
+            raise BuildError("{} must be either null (meaning don't use) or a list of package names.".format(key))
+        for package_name in package_list:
+            if not isinstance(package_name, str):
```
cmaloney: nit: Could add a PackageId.validate_name() here

https://github.com/dcos/dcos/blob/master/pkgpanda/__init__.py#L101

pkgpanda/build/__init__.py
```diff
@@ -100,6 +100,111 @@ def get_src_fetcher(src_info, cache_dir, working_directory):
...
+        all_unexcluded_packages = package_store.packages_by_name.keys() - treeinfo.excludes
+
+        self.all_packages = self.package_tuples_with_dependencies(
+            treeinfo.core_package_list or all_unexcluded_packages,
+            treeinfo,
+            package_store
+        )
+        self.validate_package_tuples(self.all_packages, treeinfo, package_store)
+
+        self.bootstrap_packages = self.package_tuples_with_dependencies(
```
cmaloney: `self.bootstrap_packages = self.package_tuples_with_dependencies(treeinfo.bootstrap_pacakge_list, treeinfo, package_store) if treeinfo.bootstrap_package_list else self.all_packages` possibly (neither seems particularly better)?

pkgpanda/build/__init__.py
```diff
@@ -616,20 +606,20 @@ def visit(pkg_tuple):
...
-        tree_visit_list = list(sorted(package_store.list_trees(), key=pkgpanda.util.variant_str))
+    if tree_variant:
+        package_sets = [package_store.get_package_set(tree_variant)]
     else:
-        tree_visit_list = [tree_variant]
+        package_sets = package_store.get_all_package_sets()
 
-    for tree in tree_visit_list:
-        visit_packages_in_tree(tree)
+    # Build all required packages for all tree variants.
```
cmaloney: "Create the build ordering by walking the list of packages to build with a depth first traversal of their requires, adding packages to the build queue after all their dependencies are in the build queue"

cmaloney:
A couple little nits but looking good! 🚢

pkgpanda/build/__init__.py
branden: I'll go with that, it saves some work if the treeinfo doesn't have any bootstrap packages.

pkgpanda/build/__init__.py
branden: Added just below.

pkgpanda/build/__init__.py
branden: I pushed a comment fixup commit that documents that detail in `visit()`: https://github.com/dcos/dcos/pull/300/commits/4fa96a505fc9a272cd246352984297e5038075a7

spahl:
:ship:

PR #310, opened 2016-07-06
WIP: `dcos launch`, GCE backed integration test clusters

cmaloney:
This is _very_ in progress. 

Adds a `dcos launch` command which (in theory) can launch a isolated network / VPC with hosts in it inside GCE using Deployment Manager.

That same bit of functionality is also applied to the integration tests, which are now able to (hopefully) be launched on top of GCE (With lots of TeamCity tweaks to make everything happy).

Likely the bits that aren't `dcos launch` which are a core for running the integration tests on a GCE VPC will be pulled out and landed as one bit, enabling quite a bit faster integration tests in GCE (Instances just start a whole lot quicker).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,0 +1,4 @@
+deployment: cody-test
+project: inbound-bee-664
+num_agent: 1
+num_agent_public: 1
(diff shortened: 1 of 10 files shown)


cmaloney:
No progress for a month, closing. Will re-build on the new test refactoring stuff by @branden and @mellenburg which closes the gap this was missing.

PR #312, opened 2016-07-06
No JSON for --version

malnick:
Do not use JSON output for a utility whose primary purpose is to print to STDOUT. This makes it difficult to parse in the shell.

### DIFF
--- a/ext/dcos-installer/dcos_installer/cli_dispatcher.py
+++ b/ext/dcos-installer/dcos_installer/cli_dispatcher.py
@@ -104,11 +104,8 @@ def validate_ssh_config_or_exit():
 
 
 def do_version(args):
-    print(json.dumps(
-        {
-            'version': gen.calc.entry['must']['dcos_version'],
-            'variant': os.environ['BOOTSTRAP_VARIANT']},
-        indent=2, sort_keys=True))
+    print("version: ", gen.calc.entry['must']['dcos_version'])
+    print("variant: ", os.environ['BOOTSTRAP_VARIANT']})
     return 0
 
 
--- a/ext/dcos-installer/dcos_installer/cli_dispatcher.py
+++ b/ext/dcos-installer/dcos_installer/cli_dispatcher.py
@@ -104,8 +104,8 @@ def validate_ssh_config_or_exit():
 
 
 def do_version(args):
-    print("version: ", gen.calc.entry['must']['dcos_version'])
-    print("variant: ", "open" if os.environ['BOOTSTRAP_VARIANT'] is None else os.environ['BOOTSTRAP_VARIANT'])
+    print("Version: ", gen.calc.entry['must']['dcos_version'])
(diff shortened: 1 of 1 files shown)


malnick:
CC @karlkfi

karlkfi:
looks ok, tho I'd prefer capitalized keys, if we're going for human readability

cmaloney:
Why is a custom format better than JSON? Was the JSON not readable? This means anyone who wants to deal with this programatically has to hand-write a parser which accepts it "well enough"... Most places have really easy to get to JSON nowadays (or jq)

cmaloney:
Also: It's just as easy to grep the JSON if you don't have a JSON parser handy as it is to grep this new output format.... So I don't really see any benefit...

malnick:
The argument here isn't readability, it is your final point on parsability. It's hard to parse JSON with shell built-ins, having to install jq or another tool just so we can easily do a `cut -d: -f2` is not rational. It would make sense to output JSON if we were sending this over the wire somewhere, but since it's just dumping to STDOUT you can assume it's going to be read by a shell script.

malnick:
I'd also add that JSON is really handy if we were sending this to a logging service, but that's also not the case here.

cmaloney:
My point is that forcing everyone to always write a custom parser to parse an arbitrary non-formal format is harder and worse. With JSON, most people have some library readily available and it's obvious how the format is going to expand / contract / etc. it's well-specified in all cases, including what happens if we start having a list of things. Or a list in a list, etc.

`wget jq && ./dcos_generate_config.sh | jq .version` isn't particularly complicated / difficult from bash.

Doing the manual field grepping also isn't particularly difficult:
`dcos_generate_config.sh | grep 'version\": "' | cut -d ':' -f 1`

Making things not use a well specified language doesn't make the problem easier.

malnick:
I agree that when the output becomes more complicated, having a structured, well defined API to parse the data would be more useful than dumping strings to STDOUT. But, if our `--version` output becomes this complicated, requiring more than 2 components (canonical version and git revision), then we should probably rethink what this command is for, and breaking those extra datasets into their own argument parsers.

malnick:
My point is that `--version` has 1 job: print the version. It doesn't speak to other applications over anything by STDOUT, and will most likely be used by bash or other shell interpreters. Making it JSON is an over-complication.

karlkfi:
I tend to agree with Jeff. But let's ask a different question: 
- Is this consistent with the other inputs/outputs of this tool?
- Is this consistent with other similar tools/commands?
- Is this consistent with what users/customers expect?

If this were a REST API, I'd expect json, but as a CLI I expect human readability and shell parsability. 

If this were complex data, I'd expect a structured format, but as a flat small map of k/v pairs I expect "key: value". 

The CLI tools I enjoy using most give me an output option, defaulting to concise and human readable with a flag for json or yaml output. The latest best in class ones like docker and kubectl do that and also give you filter options so you don't have to install jq on every node in a cluster just to use it.

karlkfi:
fwiw, if you used yaml w/o the leading ---, it would look just like "key: value"

PR #316, opened 2016-07-06
make sure we can make wheels

alberts:
https://stackoverflow.com/questions/26664102/why-can-i-not-create-a-wheel-in-python


cmaloney:
None of this local install stuff uses wheels as far as I'm aware. `pip install -e` isn't wheel based. The `prep_teamcity` does do wheels, but that is also doing something fairly different than `prep_local` does.

PR #317, opened 2016-07-07
fix xztar format in local_fetcher 

alberts:


alberts:
builds on #316

cmaloney:
🚢 Although I don't see any reason why the wheel change needs to be underneath this / in the same commit chain....

PR #318, opened 2016-07-07
add group to make 3dt read journal

darkonie:
3dt needs to read the journal. In order to do that it needs to be in
systemd-journal group. The issue is that coreos uses several backends for
users and groups. One is a regular /etc/passwd, /etc/groups and the other is
located under /usr/share/baselayout. If we try to append dcos_3dt user with
flag -G this will not work. (Tested on CoreOS beta (1068.3.0))
This seems to be a long running issue with CoreOS
https://github.com/coreos/bugs/issues/312
However if we try to change a user's primary group with -g flag all works
as expected.

### DIFF
--- a/packages/dcos-integration-test/extra/integration_test.py
+++ b/packages/dcos-integration-test/extra/integration_test.py
@@ -1,4 +1,5 @@
 import collections
+import gzip
 import json
 import logging
 import os
@@ -1725,6 +1726,16 @@ def test_3dt_bundle_create(cluster):
     assert bundles[0] == response['extra']['bundle_name']
 
 
+def verify_unit_response(zip_ext_file):
+    assert isinstance(zip_ext_file, zipfile.ZipExtFile)
+    unit_output = gzip.decompress(zip_ext_file.read())
+
+    # TODO: This seems like a really fragile string to be searching for. This might need to be changed for
+    # different localizations.
+    assert 'Hint: You are currently not seeing messages from other users and the system' not in str(unit_output), (
+        '3dt does not have permission to run `journalctl`')
+
+
 def test_3dt_bundle_download_and_extract(cluster):
     """
     test bundle download and validate zip file
@@ -1769,6 +1780,10 @@ def test_3dt_bundle_download_and_extract(cluster):
                 assert 'ip' in health_report
                 assert health_report['ip'] == master_ip
 
+                # make sure systemd unit output is correct and does not contain error message
+                gzipped_unit_output = z.open(master_folder + 'dcos-mesos-master.service.gz')
+                verify_unit_response(gzipped_unit_output)
+
                 for expected_master_file in expected_master_files:
                     expected_file = master_folder + expected_master_file
                     assert expected_file in archived_items, 'expecting {} in {}'.format(expected_file, archived_items)
@@ -1782,6 +1797,10 @@ def test_3dt_bundle_download_and_extract(cluster):
                 assert 'ip' in health_report
                 assert health_report['ip'] == slave_ip
 
+                # make sure systemd unit output is correct and does not contain error message
+                gzipped_unit_output = z.open(agent_folder + 'dcos-mesos-slave.service.gz')
+                verify_unit_response(gzipped_unit_output)
+
                 for expected_agent_file in expected_agent_files:
                     expected_file = agent_folder + expected_agent_file
                     assert expected_file in archived_items, 'expecting {} in {}'.format(expected_file, archived_items)
@@ -1794,6 +1813,10 @@ def test_3dt_bundle_download_and_extract(cluster):
                 health_report = json.loads(z.read(agent_public_folder + '3dt-health.json').decode())
                 assert 'ip' in health_report
                 assert health_report['ip'] == public_slave_ip
+
+                # make sure systemd unit output is correct and does not contain error message
+                gzipped_unit_output = z.open(agent_public_folder + 'dcos-mesos-slave-public.service.gz')
+                verify_unit_response(gzipped_unit_output)
 
                 for expected_public_agent_file in expected_public_agent_files:
                     expected_file = agent_public_folder + expected_public_agent_file
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -12,6 +12,7 @@ Each package contains a pkginfo.json. That contains a list of requires as well a
 envrionment variables from the package.
 
 """
+import grp
 import json
 import os
 import os.path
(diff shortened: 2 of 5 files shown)


cmaloney:
The group name has changed a couple times and differes between CentOS 7 and CoreOS.

darkonie:
@cmaloney it looks like both CentOS7 and CoreOS use group `systemd-journal`, i just tested and the group name is right. If the group solution is not good, what would be a better solution? I need `dcos_3dt` user to have permissions to run `journalctl`

spahl:
Let's ship with this. Still better then giving it root. Anything you need to add @darkonie? A test that checks that it works would be good I think because we could break it down the line.

cmaloney:
@spahl This needs a code review before it can land

spahl:
Correct and the test

darkonie:
@spahl i will rework this PR and add a test today. Then it should be ready for review.

darkonie:
@spahl @cmaloney this is ready for review. deploy-vpc-cli test passed

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1760,6 +1761,13 @@ def test_3dt_bundle_download_and_extract(cluster):
                 assert 'ip' in health_report
                 assert health_report['ip'] == master_ip
 
+                # make sure systemd unit output is correct and does not contain error message
+                gzipped_unit_output = z.open(master_folder + 'dcos-mesos-master.service.gz')
+                unit_output = gzip.decompress(gzipped_unit_output.read())
+                assert ('Hint: You are currently not seeing messages from '
```
cmaloney: This seems like a really fragile string to be searching for. Works for now, but would be nice if there was some definite thing we could hook onto. If we start localizing / internationalizing (some discussions going around), will be necessary.

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1773,6 +1781,13 @@ def test_3dt_bundle_download_and_extract(cluster):
                 assert 'ip' in health_report
                 assert health_report['ip'] == slave_ip
 
+                # make sure systemd unit output is correct and does not contain error message
+                gzipped_unit_output = z.open(agent_folder + 'dcos-mesos-slave.service.gz')
```
cmaloney: We do this exact same thing three times, can we make it a helper function?

pkgpanda/__init__.py
```diff
@@ -463,14 +475,23 @@ def add_user(self, username):
                                   "automatic user addition is disabled".format(username))
 
         # Add the user:
+        add_user_cmd = [
```
cmaloney: Along the same thing as the "if we aren't allowed to add users, error and exit" just above this, we need to do that same check for groups. If add_user is called, and the group already exists, and the user belongs to the group, no-op. Otherwise print an error message that things aren't pre-configured properly.

cmaloney: Should also call validate_group() in here.

pkgpanda/__init__.py
```diff
@@ -442,7 +447,14 @@ def validate_username(username):
         if not re.match(username_regex, username):
             raise ValidationError("Username must begin with `dcos_` and only have a-z and underscore after that")
 
-    def add_user(self, username):
+    @staticmethod
+    def validate_group(group):
+        try:
+            grp.getgrnam(group)
```
cmaloney: None is allowed as a value for group below in add_user, should deal with that here as well probably.

pkgpanda/__init__.py
```diff
@@ -681,7 +702,7 @@ def symlink_all(src, dest):
             # to something incompatible. We survive the first upgrade because everything goes from
             # root to specific users, and root can access all user files.
             if package.username is not None:
```
cmaloney: If package.username is none, but package.group isn't, should hard error here.

pkgpanda/__init__.py
cmaloney: Also note, this is currently only being called during package building: https://github.com/dcos/dcos/pull/318/files#diff-41557cabf54d8bfdb900c14c8a85e435R868

That happens on entirely different machines than we actually run the items on (and that machine does not require a journalctl group to exist).

The validation during build should be just "does this group name seem reasonable", than at package activation time we check "does the group exist, does the user belong to the group"

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1773,6 +1781,13 @@ def test_3dt_bundle_download_and_extract(cluster):
...
+                unit_output = gzip.decompress(gzipped_unit_output.read())
```
darkonie: we should also not call .read(), potentially the logs can be very huge. Lucky we are able to read N bytes from the beginning, this should be good enough.

pkgpanda/__init__.py
cmaloney: http://unix.stackexchange.com/questions/11477/what-are-the-allowed-group-names-for-groupadd -- should follow the conservative rules there for what is allowed.

The goal with the `validate_username` stuff is that we find bugs early / before we launch a cluster giving people a much shorter cycle.

packages/dcos-integration-test/extra/integration_test.py
darkonie: done

pkgpanda/__init__.py
darkonie: done

pkgpanda/__init__.py
darkonie: done

darkonie:
@cmaloney 3rd commit should address your comments

cmaloney:
🚢

pkgpanda/__init__.py
```diff
@@ -442,15 +452,48 @@ def validate_username(username):
...
+            raise ValidationError("Group {} has invalid name, must match the following regex: {}".format(
+                group_name, linux_group_regex))
+
+    @staticmethod
+    def validate_user_group(username, groupname):
+        user = pwd.getpwnam(username)
+        if not groupname:
+            return
+
+        if user.pw_gid != grp.getgrnam(groupname).gr_gid:
```
cmaloney: Almost ship, but users can have more than one group. They just have a "primary" group (Which is what this is). Should check that the group is in the set of groups to which the user belongs (Or that the group contains this user).

pkgpanda/__init__.py
```diff
@@ -442,15 +452,54 @@ def validate_username(username):
...
+    def validate_user_group(username, group_name):
+        user = pwd.getpwnam(username)
+        if not group_name:
+            return
+
+        group = grp.getgrnam(group_name)
+        if user.pw_gid != group.gr_gid:
+
+            # check if the user is the right group, but the group is not primary.
+            if username in group.gr_mem:
```
darkonie: @cmaloney squashed the commits and added check for username in other groups here ^^

cmaloney:
Assuming the tests back once CCM gets revived, this is good to go / ship

spahl:
:ship:

PR #320, opened 2016-07-07
Print out the variant string rather than uninterpreted variant

cmaloney:
None -> empty variant

https://dcosjira.atlassian.net/browse/DCOS-210

### DIFF
--- a/gen/installer/bash.py
+++ b/gen/installer/bash.py
@@ -547,7 +547,7 @@ def make_installer_docker(variant, bootstrap_id, installer_bootstrap_id):
             'bootstrap_active_filename': bootstrap_active_filename})
 
         fill_template('installer_internal_wrapper', {
-            'variant': variant,
+            'variant': pkgpanda.util.variant_str(variant),
             'bootstrap_id': bootstrap_id,
             'dcos_image_commit': util.dcos_image_commit})
 


malnick:
This still gives incomplete information (in this case an empty string instead of None). 

Can we update the pkgpanda utility to do something like this https://github.com/malnick/dcos/blob/019ec51f543f1dcee7ab85342a2e50260e64fb39/ext/dcos-installer/dcos_installer/cli_dispatcher.py#L108 so we get the variant we expect (in the case of None, it's assumed to be Open)?

cmaloney:
"open" and "enterprise" are not DC/OS variants that we build or use anywhere. That bit of CLI code is actually inconsistent with everything else in the codebase. The pkgpanda `variant` which is where the term comes from is whatever the prefix of the file is. so `treeinfo.json` -> "", "ee.treeinfo.json" -> `ee`, `foo.buildinfo.json` -> `foo`.

We could invent another term which is defined to be something like that, but variant isn't that.

cmaloney:
For reference: https://github.com/dcos/dcos/blob/master/pkgpanda/util.py#L15. All those `variant_` translate variants between their different forms.

spahl:
:ship:

PR #325, opened 2016-07-08
Integration Tests as Packages

mellenburg:
-Rather than build and ship a complicated container for testing, we shall distribute a package
-Allows for independent test suites with complex dependencies (like static files or libraries)
-Further simplifies integration_test infrastructure to just triggering an executable with some provided environment variables
-Allows tests to directly access files packaged in the installation without remote commands


mellenburg:
@cmaloney @spahl This is ready for a review!

test_util/run-all
```diff
@@ -96,19 +91,12 @@ export TEAMCITY_VERSION="${TEAMCITY_VERSION:-}"
 export DNS_SEARCH=true
 export DCOS_AUTH_ENABLED=false
 export DCOS_PROVIDER=onprem
-py.test -vv -s ${CI_FLAGS:-} /integration_test.py
+cd /opt/mesosphere/active/dcos-integration-test && py.test -vv -s ${CI_FLAGS:-}
```
malnick: 👍

test_util/test_runner.py
```diff
@@ -64,55 +56,29 @@ def integration_test(
...
-    tunnel.write_to_remote(join(temp_dir, pytest_image_tar), join(test_dir, pytest_image_tar))
-    log.info('Loading py.test image on remote host')
-    tunnel.remote_cmd(['docker', 'load', '-i', join(test_dir, pytest_image_tar)])
+    test_wrapper = """#!/bin/bash
+{env}
+cd /opt/mesosphere/active/dcos-integration-test
+py.test -vv {ci_flags}
+"""
+    test_env_str = ''.join(['export '+e+'\n' for e in test_env])
+    write_string('test_wrapper.sh', test_wrapper.format(env=test_env_str, ci_flags=ci_flags))
```
malnick: It would be cool if we had a script in /opt/mesosphere/bin that did exactly this, so we could SSH to a box in the cluster and just run the test. Can we include the test env as part of the package? Or as part of a separate package this is this shell script?

alberts:
In the current Enterprise code we just called this package dcos-tests. Do we want to do that here and figure out a way to make the Enterprise tests add on to this package?

test_util/test_runner.py
mellenburg: I had a couple implementations where I added scripts that made it more specific, but, you dont want to mess with the pytest api too much, so by the time that script was done, it was just an arg parser that changed the working directory before running the test, so I decided it was just garbage.

As for the test_env, that is not finalized until install time, so it can't be in a package already :(

mellenburg:
@alberts The beauty of this approach is all I did was add pytest and then place the test resources in a directory. All pkgpanda and the buildinfo do is guarantee that when the test runs, the appropriate libs are available

Thus, when we want to add a new suite (like enterprise), we can just add a whole new package called dcos-integration-tests-ee and that package would just be copy/paste of the current pytest stuff @JP made plus a new build and buildinfo.json just describing how to create deps for the new tests

alberts:
@mellenburg Sounds good. How about just calling these dcos-tests? They're more system tests than integration tests anyway (bikeshed nitpick alert).

cmaloney:
@alberts dcos-tests is already the name of a different thing. Not being ambiguous is biggest requirement.

packages/dcos-integration-test/build
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+source /opt/mesosphere/environment.export
```
cmaloney: This isn't used at all, no reason to do it that I can see.

packages/dcos-integration-test/buildinfo.json
```diff
@@ -0,0 +1,7 @@
+{
+  "requires": [
+      "dnspython",
```
cmaloney: `python-requests`?

mellenburg: I was being lazy and didnt write all the shared requirements that are in dcos-image-deps. Come to think of it, the only reason dcos-image-deps was used is that I didnt want to rebuild beautifulsoup4 or add it as a second package. Should I remove dcos-image-deps, grab bs4, and link to all the dependent packages in dcos-image-deps. Or do I leave dcos-image-deps and just add the big reqs for redundancy? (e.g. boto, python-requests, python-retrying, etc...)

test_util/run-all
```diff
@@ -96,19 +91,12 @@ export TEAMCITY_VERSION="${TEAMCITY_VERSION:-}"
...
 EOF
-cp $PWD/../../integration_test.py .
-cp $PWD/../../docker/py.test/Dockerfile .
-cp $PWD/../../docker/test_server/Dockerfile test_server_Dockerfile
-cp $PWD/../../docker/test_server/test_server.py .
-docker build -t py.test .
-popd
-rm -rf tmp_build_dir_for_pytest
-
+chmod +x test_wrapper.sh
```
cmaloney: Why don't we just write this with the executable bit set?

cmaloney: nvm, heredoc, can't really be done...

packages/dcos-integration-test/buildinfo.json
cmaloney: Ideally should depend on all the things used directly (Not splitting beautifulsoup4 out of dcos-image-deps is fine). Just note when people go and modify dcos-image-deps they'll get a longer dev cycle than they expect since more things end up needing to change (although that package is relatively quick to build).

mellenburg: Ah, good point! I wasn't thinking about the issue of rebuilds

mellenburg:
Is this ready to ship? Once this is merged, we can open up PRs for:
-independent testing suites
-replacing hackish marathon one-off's with check_calls
-splitting integration_test.py into smaller logical files (e.g. conftest.py, test_cluster_visibility, test_marathon, etc)

test_util/test_runner.py
```diff
@@ -64,55 +56,29 @@ def integration_test(
...
+        wrapper_path = join(test_dir, 'test_wrapper.sh')
         log.info('Running integration test...')
-        tunnel.remote_cmd(docker_cmd, timeout=timeout)
+        tunnel.write_to_remote('test_wrapper.sh', wrapper_path)
+        dump_logs(tunnel.remote_cmd(['bash', wrapper_path], timeout=timeout))
         log.info('Successful test run!')
-    except CalledProcessError:
+    except CalledProcessError as e:
         log.exception('Test failed!')
+        dump_logs(e.output)
```
cmaloney: Can we please echo the output as it goes by?

I'm very doubtful people / things will remember to pick up the log files...

cmaloney:
I'm not sure the splitting is actually a good idea / it needs a lot of work to make it be actually better. a lot of things use common utility functions that are currently easy to get at with the "cluster" fixutre (And in other DC/OS variants that cluster fixture needs which means making it partly replacable / insertable to do it well...). 

The fact that we're dumping all the log output into one file post-hoc makes it really hard to tell what was going on, and impossible in the case where TeamCity early terminates a job because it was running to long / most likely hanging. That code shouldn't have been updated to do that in the past.

Can go once we're streaming bytes / messages out as we get them from py.test. Can happen to also save it into a log file simultaneously if really wanted, but streaming them out immediately is critical for usage. That has bitten me several times recently, and I know it's hit @spahl as well.

test_util/test_runner.py
mellenburg: The previous version also dumps to a log without printing. The reason I do not print this is that when using the teamcity-messages plugin, some weird byte encoding causes python to fail at printing it. Specifically, most of the output is fine until a couple hundred lines in and then stops interpreting newlines and all sorts of garbage appears in the TC logs.  Maybe some more investigation would bear what the result is, but previously it was just easier to always dump straight to a file and then add if [ -e integration_test.log ] ; then cat integration_test.log; fi to the builds scripts (which, as stated above, is already in place)

cmaloney: The original version logged this directly back, and https://github.com/dcos/dcos/pull/325/files/7938eb21e5756b9ca458e0156ed1cce4c0457651#diff-35acaec42b244e7ea0b6a2474e06aa05R94 also does that, and things work.

Not streaming out logs makes it really impossible to debug things on the integration test run. It shouldn't have gotten changed to not real time responding / that is something that was definitely missed in a previous review and needs to be fixed.

The teamcity messages themselves don't use any special ascii characters or escape codes: https://github.com/JetBrains/teamcity-messages/blob/master/teamcity/messages.py#L42.

Just outputs a ascii string which has the information in a format TeamCity understands.

cmaloney: It should be possible to do either of two things.
1. Read and write the data stream as bytes (best option), never decoding it just forwarding.
2. Readn and write the data stream as utf-8, and make sure to set PYTHONENCODING to `utf-8` in the environment, then write the data back out as utf-8.

mellenburg:
@cmaloney See https://github.com/dcos/dcos/pull/342 for resolving the streaming log issue

mellenburg:
nevermind on https://github.com/dcos/dcos/pull/342 @spahl recommended that it simply be integrated into this PR

mellenburg:
Re: streaming log output. I am running into the exact same issue I had before:
https://teamcity.mesosphere.io/viewLog.html?buildId=333614&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2&tab=buildLog#_focus=757
https://teamcity.mesosphere.io/viewLog.html?buildId=333669&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2&tab=buildLog#_focus=718

The first case it is printing the stream out per-byte. In the second case, it does per-line. Note: it always starts getting getting garbled at test_if_Marathon_app_can_be_deployed. In any case, I see now cat-ing still runs into this issue, but its a little less ugly in that case:
https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTest_3&buildId=333590#_state=33&focus=642

I really don't feel digging through haystacks for a needle is worthwhile in this case and am open to suggestions as to which way to go (semi-garbled stream, minimally garbled cat'd disk file. etc)

spahl:
:ship: This is good enough now. We can iterate to make it better.

cmaloney:
This seems to have caused our tests to start taking 4-7 minutes longer than they used to on Vagrant. See: https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr&tab=buildTypeStatistics&branch_DcosIo_Dcos_VagrantIntegrationTests=__all_branches__ :/

PR #330, opened 2016-07-11
Retry ensure routable

sargun:
Last time this was closed. Vagrant runs the test without waiting long enough for the backend routing to be setup. We can alternatively sleep 5 seconds.


spahl:
:ship:

cmaloney:
The same question I asked last time still exists: How / why do customers need to randomly wait 5 seconds? How can they know to wait 5 seconds for their thing to work?

There needs to be some simple signal they can wait for, that once it fires this is guaranteed to be up and running.

spahl:
@sargun ^

PR #334, opened 2016-07-11
Use a distro abstraction for distribution type instead of variant

malnick:


malnick:
Need to implement the `distro` type in a cleaner fashion.

PR #341, opened 2016-07-13
Remove legacy config-writer service

lingmann:
*** CURRENTLY BREAKS METRONOME ***

Remove legacy config-writer service

This config-writer service was originally added to work around an issue with Marathon, which prevented it from starting when `hostname` was not resolvable to an IP address. This is no longer an issue with Marathon so we can remove this workaround.

There is the potential that this could cause problem for services which are depending on `hostname` to be resolvable however. But in most cases a service trying to communicate with the local host should use the loopback address and not `hostname`.

### DIFF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -35,17 +35,6 @@ root:
       BindIPv6Only=both
       [Install]
       WantedBy=sockets.target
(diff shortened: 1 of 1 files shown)


cmaloney:
hasn't moved in 1+ month, closing. Re-open when in progress / iteration.

PR #345, opened 2016-07-14
Add complete build metadata to metadata.json

branden:
This adds `complete_dict`, `all_completes`, and `<variant>.complete.latest.json` to build metadata.

### DIFF
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -312,9 +312,6 @@ class PackageStore:
         check_call(['mkdir', '-p', directory])
         return directory
 
-    def get_treeinfo(self, variant):
-        return TreeInfo(load_config_variant(self._packages_dir, variant, 'treeinfo.json'))
-
     def list_trees(self):
         return get_variants_from_filesystem(self._packages_dir, 'treeinfo.json')
 
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -283,22 +283,25 @@ class PackageStore:
     def get_bootstrap_cache_dir(self):
         return self._packages_dir + "/cache/bootstrap"
 
+    def get_complete_cache_dir(self):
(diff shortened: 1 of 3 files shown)


branden:
@cmaloney This is ready for review.

cmaloney:
The `release.latest` name is really bad / ambiguous with the `release` tool which is in the repository (And what it means). Need to come up with something else. Will review the rest tomorrow.

cmaloney:
🚢 . Should ideally in the next PR update test_release to make sure having a complete set and bootstrap set which are not identical produces the right results (all the packages get uploaded).

Should probably also add a hard assert in https://github.com/dcos/dcos/pull/345/files#diff-f1d07a3e273cef624bac200fe7e81dd6R311 that the bootstrap set is a subset of the complete set. The logic there assumes that, and would not upload all the right packages if it wasn't true.

branden:
@cmaloney We don't include the bootstrap package set in `metadata.json`, only the bootstrap ID, so we can't do that comparison in the release script. We assert that the bootstrap package set is a subset of the complete package set during build, when we validate the treeinfo: https://github.com/dcos/dcos/blob/b7dd4d078f900e22f3b88ed8cf9d002e5038bd45/pkgpanda/build/__init__.py#L177

cmaloney:
@branden It shouldn't be that hard to get the set of packages inside of the bootstrap. The active.json has it if nothing else. Wouldn't be bad to pass back more structured information either rather than using the side channel of active.json.

I like having assertions in the individual bit of codes which would be violated if the assertion changes. If at some point we make it so the bootstrap set doesn't have to be a strict subset of all the packages, people will find and change that bit of pkgpanda, but have no idea they need to find / change the code here. They'll just get silent, hard to debug failures  / odd behavior when re-factoring things, and maybe eventually someone will remember this bit of code, or figure out that it's broken. Much nicer to have an assert here as well so they are notified they need to fix

PR #347, opened 2016-07-15
Adding Rex-Ray support to DC/OS

branden:
Taking over from #220.

### DIFF
--- a/packages/dcos-integration-test/extra/integration_test.py
+++ b/packages/dcos-integration-test/extra/integration_test.py
@@ -1,4 +1,6 @@
 import collections
+import contextlib
+import copy
 import json
 import logging
 import os
@@ -29,11 +31,6 @@ AUTH_ENABLED = os.getenv('DCOS_AUTH_ENABLED', 'true') == 'true'
 # Set these to run test against a custom configured user instead
 LOGIN_UNAME = os.getenv('DCOS_LOGIN_UNAME')
 LOGIN_PW = os.getenv('DCOS_LOGIN_PW')
-
-# AWS creds for volume control (not used currently)
-AWS_ACCESS_KEY_ID = os.getenv('AWS_ACCESS_KEY_ID')
-AWS_SECRET_ACCESS_KEY = os.getenv('AWS_SECRET_ACCESS_KEY')
-AWS_REGION = os.getenv('AWS_REGION')
 
 
 @pytest.fixture(scope='module')
@@ -150,6 +147,23 @@ def _setup_logging():
     logging.getLogger("requests").setLevel(logging.WARNING)
 
 
+@contextlib.contextmanager
+def _remove_env_vars(*env_vars):
+    environ = dict(os.environ)
+
+    for env_var in env_vars:
+        try:
+            del os.environ[env_var]
+        except KeyError:
+            pass
+
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(environ)
+
+
 def _delete_ec2_volume(name, timeout=300):
     """Delete an EC2 EBS volume by its "Name" tag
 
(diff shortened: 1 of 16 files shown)


packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1124,6 +1148,79 @@ def _ensure_works():
...
+    """
+    hosts = cluster.slaves[0], cluster.slaves[-1]
+    test_uuid = uuid.uuid4().hex
+    test_label = 'integration-test-move-external-volume-{}'.format(test_uuid)
+    base_app = {
+        'mem': 32,
+        'cpus': 0.1,
+        'instances': 1,
+        'container': {
+            'type': 'MESOS',
```
jdef: can we have a similar test that also exercises the DOCKER containerizer? or maybe revise this test so that it's MESOS on agent1 and DOCKER on agent2?

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -150,6 +146,23 @@ def _setup_logging():
     logging.getLogger("requests").setLevel(logging.WARNING)
 
 
+@contextlib.contextmanager
```
cmaloney: `monkeypatch` has support for setting / removing environment variables. Can we just use that?

http://doc.pytest.org/en/latest/monkeypatch.html#_pytest.monkeypatch.monkeypatch.setenv

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -162,11 +175,23 @@ def _delete_ec2_volume(name, timeout=300):
...
+
+    # Remove AWS environment variables to force boto to use IAM credentials.
+    with _remove_env_vars('AWS_ACCESS_KEY_ID', 'AWS_SECRET_ACCESS_KEY'):
+        volumes = list(boto3.session.Session(
+            # We assume we're running these tests from a cluster node, so we
+            # can assume the region for the instance on which we're running is
+            # the same region in which any volumes were created. If we can't
+            # find the current region, we default to reading it from the
+            # environment.
+            region_name=(_get_current_aws_region() or os.getenv('AWS_REGION')),
```
cmaloney: Why do we need this many fallbacks? 

Boto3 has a built in environment variable configuration usage: https://boto3.readthedocs.io/en/latest/guide/configuration.html#environment-variable-configuration

Can we just use that rather than making a new key here?

packages/rexray/extra/dcos-rexray.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=REX-Ray: A vendor agnostic storage orchestration engine
+
+[Service]
+StartLimitInterval=0
+RestartSec=15
+Restart=always
+EnvironmentFile=/etc/environment
```
cmaloney: `/etc/environment` doesn't exist many places. `/opt/mesosphere/environment` should be plenty.

packages/rexray/extra/dcos-rexray.service
```diff
@@ -0,0 +1,13 @@
...
+EnvironmentFile=/opt/mesosphere/environment
+Environment=REXRAY_HOME=/
+ExecStart=__PKG_PATH__/bin/rexray start -f
+ExecReload=/bin/kill -HUP $MAINPID
```
cmaloney: This is not the right path on ubuntu most likely. Do we really need it?

packages/rexray/extra/dcos-rexray.service
```diff
@@ -0,0 +1,13 @@
...
+KillMode=process
```
cmaloney: Why don't we want to kill the whole cgroup? This is likely to leave lingering processes when we're doing upgrades and the like.

packages/dcos-integration-test/extra/integration_test.py
branden: I don't think that's a good fit for a util function. We'd have to pass the `monkeypatch` object from the test just so this function could use it, or use it within the test and leak this function's implementation details. `monkeypatch` is intended for mocking things used by code under test, but here we're not testing boto, just trying to get it to use the right credentials. Coupling this function to pytest would make it harder to extract. I think the cost of reinventing this (small) wheel outweighs the benefits.

branden:
We need to move the DVDCLI tarball to downloads.dcos.io. /cc @lingmann

cmaloney:
Need to move the download to downloads.dcos.io, but once that happens this can ship

lingmann:
Working on the downloads move now...

packages/rexray/buildinfo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "docker": "golang:1.6",
+  "single_source" : {
+    "kind": "git",
+    "git": "git@github.com:dcos/rexray.git",
+    "ref": "1816ee3868550e07b90f00f57bfd7f737235cd93",
+    "ref_origin": "v0.3.3-dcos"
```
jdef: we've noticed a serious regression against this branch of rexray, whereby many volumes are created when launching a Service that uses external volumes and the docker containerizer. troubleshooting this with the velocity team now.

lingmann: @jdef should we hold off on publishing the dvdcli tarball to our downloads bucket until we know more about the regression?

jdef: if you're not overwriting anything then I'd say publish it now and we can
upload a new version later if needed

lingmann: Ok thanks. @branden dvdcli artifact is now available at: https://downloads.dcos.io/pkgpanda-artifact-cache/dvdcli-Linux-x86_64-0.2.0.tar.gz

jdef: xref https://github.com/emccode/rexray/issues/511

jdef:
FWIW jie and gilbert both mentioned today that dvdcli 0.2 was probably not
compatible with the mesos docker volume isolator.

cmaloney:
Can you add a blocker JIRA for that?

jdef:
done

PR #350, opened 2016-07-15
make get_hashed_password more robust

alberts:


cmaloney:
This definitely doesn't seem like it's making this any more reliable of an operation...

alberts:
The reason this randomly fails in some scenarios is that SSH prints other junk to stdout. Still need to debug why this fix isn't working. Guessing my check for the password string isn't right.

cmaloney:
Why is SSH printing out other stuff here / in this path through the installer ever? That sounds like a bug...

test_util/installer_api_test.py
```diff
@@ -65,7 +66,9 @@ def scp(src, dst):
     def get_hashed_password(self, password):
         p = self.ssh(["bash", self.installer_path, "--hash-password", password])
         # password hash is last line output but output ends with newline
-        passwd_hash = p.decode('utf-8').split('\n')[-2]
+        stdout = p.communicate()[0]
+        stdout = stdout.decode('ascii')
+        passwd_hash = [x for x in re.split('\s+', stdout) if x.startswith('$6$')][0]
```
vespian: regexp + assert (found == True) ?

cmaloney:
Haven't seen progress on this and haven't seen any integration tests fail because of this. Closing for now. Re-open if actually going to make progress.

PR #353, opened 2016-07-15
Pin to latest Ubuntu 16.04 LTS image

lingmann:
Per discussions with Azure, we've agreed to pin to the LTS image
version. Historically these images stick around until Ubuntu EOL's the
release and if we continue to pin to 'latest' there will be a chance
that a new Ubuntu LTS build will break ACS.


spahl:
:ship:

PR #354, opened 2016-07-15
Bump networking for dcos enterprise compatbility

sargun:


sargun:
@spahl Does this approach look okay?

sargun:
We would want to use notify unknown on upgrades. But otherwise I think it's reasonable. I want to make sure you think so before integrating.

spahl:
Looking

sargun:
@spahl Nudge? If this approach is reasonable, I'll start integrating dcos-enterprise atop this today.

spahl:
Yes this seems good to me. @alberts

packages/bootstrap/extra/dcos_internal_utils/bootstrap/cli.py
```diff
@@ -84,4 +102,9 @@ def parse_args():
         type=str,
         default='/opt/mesosphere/etc/master_count',
         help='File with number of master servers')
+    parser.add_argument(
```
alberts: why is this a thing?

PR #356, opened 2016-07-16
Add CCM test for advanced installation

mellenburg:
When vagrant/bash provider tests fail due to vagrant issues, its nice to have an alternative way to test the manual install process

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,6 +53,7 @@ setup(
         'console_scripts': [
             'release=release:main',
             'ccm-deploy-test=test_util.test_installer_ccm:main',
+            'adv-install-test=test_util.test_advanced_install:main',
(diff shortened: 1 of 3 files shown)


mellenburg:
@spahl @cmaloney This is ready for review

malnick:
@mellenburg Can you rebase and fix conflicts?

cmaloney:
This is fairly outdated with current master, needs some reworking because it has a lot of more or less duplicate code with the other integration tests.

Practically I also am strongly against only running tests in AWS. We have a _lot_ of AWS specific code which gets pulled into DC/OS, and just by running there vs. other places we won't be testing if our product doesn't work outside AWS except when people do manual installs internally.

PR #357, opened 2016-07-16
Test fixes

mellenburg:
-tests that use base_marathon_app need to be apart of registry_cluster fixture
-timeout must be settable in environment for installer API
-Suppressing error codes in test make it very hard to distinguish if the test failed or if the setup of the test failed. Rather, do the suppression  in TC (there is a specific option just for this case)

### DIFF
--- a/packages/dcos-integration-test/extra/integration_test.py
+++ b/packages/dcos-integration-test/extra/integration_test.py
@@ -1102,6 +1102,7 @@ def test_if_minuteman_routes_to_vip(cluster, timeout=125):
 def test_ip_per_container(registry_cluster):
     """Test if we are able to connect to a task with ip-per-container mode
     """
+    cluster = registry_cluster
     # Launch the test_server in ip-per-container mode
 
     app_definition, test_uuid = cluster.get_base_testapp_definition(ip_per_container=True)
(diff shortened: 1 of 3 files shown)


test_util/test_installer_ccm.py
```diff
@@ -324,8 +324,6 @@ def establish_host_connectivity():
             vpc.delete()
     else:
         log.info("Test failed! VPC will remain for debugging 1 hour from instantiation")
-    if options.ci_flags:
```
cmaloney: @mellenburg The suppressing the test return code is absolutely necessary for us to be able to mute / unmute test. Assumes you have "smart" CI which can interpret the results of the test report (Which is what TeamCity does). If people aren't in CI, they get the return code.

The build is setup to fail on non-zero error code _or_ non-passing test results. If there are non-passing test results, and that also leads to a non-zero return, then you can't mute tests in CI. See the PR which added this for more explanation.

cmaloney: https://confluence.jetbrains.com/display/TCD9/Muting+Test+Failures

mellenburg: So, why can't we disable test failure on non-zero exit code? My issue with suppressing the error code like this is that when the test **setup** fails, the test doesn't return non-zero nor does the smart CI pick up a failing test.

Thus, in the current state, if you truly mess up the test so bad that the setup can't run the check will show green.

I'm fine with breaking this out into a different PR, but the way it is now needs some refinement.

mellenburg:
@cmaloney This is ready for review again. Just simple changes now. Tabling the CI issues for another PR

cmaloney:
:ship:

PR #359, opened 2016-07-18
Bump Marathon to v1.2.0-RC5

gkleiman:


cmaloney:
@gkleiman This is consistenly giving odd status codes back to basic marathon apps we launch as part of the integration tests. I wouldn't have expected that API to change. Can you investigate?

The Deploy VPC API / CLI tests run in CCM clusters, when the integration test fails you can actually log into the cluster and debug.

gkleiman:
@cmaloney: these tests failures only confirm that we fixed the service ports validation. `registry_cluster` creates a `/build-and-push` app with a dynamic port. Marathon automatically assigns it the service port 1001. Then `test_octarine_http`  tries to explicitly request port 1001 and gets a validation error, which is expected.

If these are the only failing tests, then I'ld merge the PR and create an issue to fix the failing tests.

cmaloney:
I have https://github.com/dcos/dcos/pull/399 to fix the two tests, should hopefully be able to land that then this on top.

cmaloney:
🚢 . Test is fixed in #399

PR #362, opened 2016-07-18
mesos: SSL and logrotate

alberts:


packages/mesos/extra/dcos-mesos-master.service
```diff
@@ -9,7 +9,9 @@ PermissionsStartOnly=True
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/mesos-master
 EnvironmentFile=-/opt/mesosphere/etc/mesos-master-provider
+EnvironmentFile=-/run/dcos/etc/mesos-master
 ExecStartPre=/bin/ping -c1 ready.spartan
 ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done'
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-master
+ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done'
```
cmaloney: This looks identical to what is two lines up...

packages/mesos/extra/dcos-mesos-slave-public.service
```diff
@@ -14,7 +14,9 @@ EnvironmentFile=-/opt/mesosphere/etc/mesos-slave-common-extras
 EnvironmentFile=/opt/mesosphere/etc/mesos-slave-public
 EnvironmentFile=-/var/lib/dcos/mesos-slave-common
 EnvironmentFile=/var/lib/dcos/mesos-resources
+EnvironmentFile=-/run/dcos/etc/mesos-slave-public
 ExecStartPre=/bin/ping -c1 ready.spartan
 ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-agent-public
```
cmaloney: Please only use `mesos-slave` or `mesos-agent` for all the internal stuff. I'm fine with either, but don't want to be half way straddling the border. Need to rename the service and make it consistent top to bottom, or keep using slave for now.

packages/mesos/build
```diff
@@ -62,7 +64,6 @@ disk_resource_script="$PKG_PATH/bin/make_disk_resources.py"
 cp /pkg/extra/make_disk_resources.py "$disk_resource_script"
 chmod +x "$disk_resource_script"
 
-# Copy the json file describing logrotate module.
-module_json="$PKG_PATH/etc/mesos-slave-modules/logrotate.json"
-mkdir -p "$(dirname "$module_json")"
-cp /pkg/extra/logrotate.json "$module_json"
+# Install agent modules.
+mkdir -p "$PKG_PATH/etc/mesos-agent-modules"
```
adam-mesos: We'll need to change the directory name in downstreams as well (no big deal, unless we forget).

packages/mesos/extra/dcos-mesos-master.service
adam-mesos: Yup, now the question is if that needs to happen before or after 'bootstrap'

packages/mesos/build
adam-mesos: Also, in this repo:
gen/dcos-config.yaml:171:  - path: /etc/mesos-slave-modules/overlay_slave_modules.json
gen/dcos-config.yaml:221:      MESOS_MODULES_DIR=/opt/mesosphere/etc/mesos-slave-modules

cmaloney: can we please just not rename half the stuff in this big push? Doesn't add a lot of value, makes things more ambiguous. Should be an all or nothing change / giant sed.

cmaloney:
Assuming tests pass, and downstream works this is looking good. My only nit is the layout of `/run/dcos`. Should be `/run/dcos/<service-name>` so we can easily make even bootstrap non-root later.

adam-mesos:
LGTM. :rocket: 
Note that this only compiles in SSL support, but does not actually enable Mesos SSL. Additional environment variables (`SSL_*`) are required for that.

packages/bootstrap/extra/dcos_internal_utils/bootstrap/cli.py
```diff
@@ -71,6 +75,21 @@ def main():
 def parse_args():
     if os.path.exists('/etc/mesosphere/roles/master'):
         zk_default = '127.0.0.1:2181'
+    else:
+        if os.getenv('MASTER_SOURCE') == 'master_list':
+            # Spartan agents with static master list
+            with open('/opt/mesosphere/etc/master_list', 'r') as f:
+                master_list = json.load(f)
+            assert len(master_list) > 0
+            leader = random.choice(master_list)
```
jgehrcke: Haha, that's called "leader election"  :-).

PR #366, opened 2016-07-19
Add support for downstream extras for Mesos EnvironmentFiles.

adam-mesos:


packages/marathon/build
```diff
@@ -19,6 +19,8 @@ StartLimitInterval=0
 RestartSec=15
 PermissionsStartOnly=True
 EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=-/opt/mesosphere/etc/marathon
```
adam-mesos: Apparently Marathon doesn't even have any open-source environment file flags (everything's in command-line).

cmaloney: https://mesosphere.github.io/marathon/docs/command-line-flags.html

Note - Command Line Flags May Be Specified Using Environment Variables

Seems to suggest otherwise. We do use cli flags currently, should probably move

adam-mesos: Right. I was just saying we put everything in command-line now, but we would need `EnvironmentFile=-/opt/mesosphere/etc/marathon` if we wanted to specify those in environment variables instead. Migrating from cli flags to env vars can be done as a separate PR.

alberts: Might want to double-check this. This didn't work for all options. I could also be remembering Metronome.

adam-mesos: May not work for all options, but allowing Marathon/Metronome unit files to load EnvironmentFiles seems like a good thing even if env vars only work for some options.

adam-mesos: Note: I'm not putting any contents in `/opt/mesosphere/etc/marathon` here, but it is prefaced with a `-` so it won't error if the file doesn't exist.

adam-mesos: cc: @gkleiman for moving Marathon flags into an environment variable file..

gkleiman: Marathon will ignore the environment variables. We provide a shell script that turns the env vars into cmd line flags, but DC/OS doesn't use it.

adam-mesos: Ok, well most of the environment variables I want to set (e.g. SSL_) are ready by libmesos directly.

adam-mesos:
I removed the Marathon and Metronome extras and will handle them in separate PRs.
@cmaloney PTAL

packages/mesos/extra/dcos-mesos-master.service
```diff
@@ -8,6 +8,7 @@ LimitNOFILE=infinity
 PermissionsStartOnly=True
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/mesos-master
+EnvironmentFile=-/opt/mesosphere/etc/mesos-master-extras
```
cmaloney: I'd rather have this just before the `/run/dcos` one. That way the downstream overrides get the last say.

packages/mesos/extra/dcos-mesos-slave-public.service
```diff
@@ -12,6 +12,7 @@ EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/mesos-slave-common
 EnvironmentFile=-/opt/mesosphere/etc/mesos-slave-common-extras
 EnvironmentFile=/opt/mesosphere/etc/mesos-slave-public
+EnvironmentFile=-/opt/mesosphere/etc/mesos-slave-public-extras
```
cmaloney: this one is a good position

packages/mesos/extra/dcos-mesos-slave.service
```diff
@@ -12,6 +12,7 @@ EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/mesos-slave-common
 EnvironmentFile=-/opt/mesosphere/etc/mesos-slave-common-extras
 EnvironmentFile=/opt/mesosphere/etc/mesos-slave
+EnvironmentFile=-/opt/mesosphere/etc/mesos-slave-extras
```
cmaloney: Why do we need both a `mesos-slave-common-extras` and a `mesos-slave-extras`?

packages/mesos/extra/dcos-mesos-master.service
adam-mesos: You got it dude.

packages/mesos/extra/dcos-mesos-slave-public.service
adam-mesos: But do you want to switch `/opt/mesosphere/etc/mesos-slave-public` and `/opt/mesosphere/etc/mesos-slave-common-extras` so that "downstream gets the last say"? Ideally nobody should put the same thing in both of them.

packages/mesos/extra/dcos-mesos-slave.service
adam-mesos: `mesos-slave-common-extras` is for variables that the downstream wants to apply to both public and private slaves.
`mesos-slave-extras` is for variables the downstream only wants to apply to private slaves. Maybe we should call it `mesos-slave-private-extras`, but we haven't used "private" anywhere else before.

cmaloney: My point is I don't see any usage which we'll have of mesos-slave-extras / mesos-slave-public-extras. The mesos-slave / mesos-slave-public environment files just contain the differences in resources (default role). Modifying that downstream doesn't really make sense to me.

adam-mesos: Please consider that a downstream distro/operator may want to configure public agents to use a different principal for authenticating to the master, or use different certs/keys for encrypted communication.

adam-mesos: I guess they'd just need different file contents, not different env vars. Ok, I'll remove these two until we have a stronger need for them.

cmaloney:
🚢

PR #367, opened 2016-07-19
Add link for "the CI system"

lloesche:
README mentions the CI system but doesn't actually link to it even though it's public.


cmaloney:
🚢 . Will add to the next train which should go out later today. Might be a good idea to note that there is a "login as guest" button on the login screen.

PR #368, opened 2016-07-19
fixed minor typos in README.md 

discoposse:
Small fix to change premise to premises (see http://itsonpremises.com) and varios to various.


cmaloney:
🚢 

Will add to the next train which should land later today.

PR #370, opened 2016-07-19
do not restart docker on install [WIP]

darkonie:
Currently on prem installer is being run from a docker container. We cannot
restart the container because this will break the installer.


cmaloney:
NAK

spahl:
Don't forget to mark titles at WIP

darkonie:
@cmaloney correct, this is WIP, changed the title

PR #351, opened 2016-07-15
Update dcos-history for TLS support

mellenburg:

PR #372, opened 2016-07-20
Train 32

cmaloney:
#347 (It will likely need more later, but we need rex-ray in to find more bugs, and have been missing it wayy to long)
#351, #367, #368

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -14,6 +14,17 @@ from pkgpanda import PackageId
 from pkgpanda.build import hash_checkout
 
 
+AWS_REXRAY_CONFIG = """
+rexray:
+  loglevel: info
+  storageDrivers:
+    - ec2
+  volume:
+    unmount:
+      ignoreusedcount: true
+"""
+
+
 def calculate_bootstrap_variant():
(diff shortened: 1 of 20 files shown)


gen/calc.py
```diff
@@ -70,6 +81,14 @@ def calculate_ip_detect_public_contents(ip_detect_contents):
     return ip_detect_contents
 
 
+def calculate_rexray_config_contents(rexray_config_filename):
+    try:
+        with open(rexray_config_filename, encoding='utf-8') as f:
+            return yaml.dump(f.read())
+    except IOError as err:
+        raise Exception('REX-Ray config file {}: {}'.format(rexray_config_filename, err))
```
jgehrcke: Why do we "rewrite" that exception here with information loss, instead of just letting this blow up (which yields the more complete log)?

cmaloney: Adding in context of what the filename is, since this will show up to end users as a warning / error message.

We should actually probably be catching / re-wrapping exceptions, and do a raise from...

@branden Could you make a follow up PR to fix (I want to land this train, don't think this is blocking)

branden: https://github.com/dcos/dcos/pull/375

PR #373, opened 2016-07-20
Add named VIPs integration test

sargun:


packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1105,6 +1105,79 @@ def _ensure_routable():
...
+
+    cluster.deploy_marathon_app(app_definition)
+
+    proxy_definition = {
+        'id': "/integration-test-proxy-to-minuteman-vip-%s" % test_uuid,
+        'cpus': 0.1,
+        'mem': 128,
+        'ports': [10000],
+        'cmd': 'chmod 755 ncat && ./ncat -v --sh-exec "./ncat foo.marathon.l4lb.thisdcos.directory 5000" -l $PORT0 --keep-open',
+        'uris': ['https://s3.amazonaws.com/sargun-mesosphere/ncat'],
```
cmaloney: don't want to download random stuff. Touching the internet is unreliable.

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1105,6 +1105,79 @@ def _ensure_routable():
...
 
 
+def test_if_minuteman_routes_to_named_vip(cluster, timeout=125):
+    """Test if we are able to connect to a task with a vip using minuteman.
+    """
+    # Launch the app and proxy
+    test_uuid = uuid.uuid4().hex
+
+    app_definition = {
+        'id': "/integration-test-app-with-minuteman-vip-%s" % test_uuid,
```
cmaloney: we have a standard marathon app we launch, can we just use that + add VIP config to it?

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1105,6 +1105,79 @@ def _ensure_routable():
...
+            'maxConsecutiveFailures': 3,
+            'ignoreHttp1xx': False
+        }],
+    }
+
+    service_points = cluster.deploy_marathon_app(proxy_definition)
+
+    @retrying.retry(wait_fixed=2000,
+                    stop_max_delay=timeout*1000,
+                    retry_on_result=lambda ret: ret is False,
```
cmaloney: this never returns, so this isn't adding anything...

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1105,6 +1105,79 @@ def _ensure_routable():
     _ensure_routable()
 
 
+def test_if_minuteman_routes_to_named_vip(cluster, timeout=125):
```
cmaloney: we never will change this timeout / configure it and it's only referenced one place, just make it live in one place.

2 minutesis also wayyy longer than this should take.

sargun:
@cmaloney Yeah, this was the first go at the PR, just copying the standard Minuteman test. What I will do is take the IP per container test and actually modify it to have a field where I can add a VIP to the app.

I'll probably need a re-review EOD?

cmaloney:
I'm in hamburg so the review cycle is offset / I'm asleep SF EOD.

PR #374, opened 2016-07-20
Downgrade dvdcli 0.2.0 -> 0.1.0

branden:
The Mesos volume isolator requires dvdcli 0.1.0. See
https://github.com/apache/mesos/blob/master/docs/docker-volume.md#pre-conditions.

PR #337, opened 2016-07-13
Bump dcos-ui package. Includes binary cli instructions.

rcorral:

PR #376, opened 2016-07-20
Train 33

cmaloney:
#337, #374, #375, #377

### DIFF
--- a/packages/dcos-ui/buildinfo.json
+++ b/packages/dcos-ui/buildinfo.json
@@ -1,9 +1,9 @@
 {
-  "docker": "node:4.3",
+  "docker": "node:4.4.4",
   "single_source" : {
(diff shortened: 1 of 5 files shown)


packages/dcos-ui/buildinfo.json
```diff
@@ -1,9 +1,9 @@
 {
-  "docker": "node:4.3",
+  "docker": "node:4.4.4",
   "single_source" : {
     "kind": "git",
-    "git": "https://github.com/dcos/dcos-ui.git",
-    "ref": "1aedbc0c679ee83f8899404a91b8928ed2849c47",
+    "git": "git@github.com:dcos/dcos-ui.git",
```
cmaloney: @rcorral This should be / needs to be https. Can we please go back to that?

PR #377, opened 2016-07-20
Move mesos_isolation into a gen/calc variable.

adam-mesos:
Move mesos_isolation into a gen/calc variable so downstreams can extend it.
Also dropped the deprecated isolator name 'posix/disk' in favor of the new 'disk/du'.


cmaloney:
🚢

PR #379, opened 2016-07-20
Collect pytest first to catch setup failures

mellenburg:
Also allows pluggable pytest_dir and pytest_cmd


mellenburg:
@spahl This is ready for review. Right now tests can fail in the setup stage, and then will return zero if CI_FLAGS is set, thus giving a green check despite no tests ever running

cmaloney:
🚢

PR #381, opened 2016-07-20
dcos-ui: update to https

rcorral:


cmaloney:
Turns out there were a number of these that had slipped through over time. I fixed them all in : https://github.com/dcos/dcos/pull/380/commits/2768d1c4a754a46ebdd226f4012022b7bbab562b part of #380

PR #384, opened 2016-07-21
Remove Caching Test Server from Signal Test

malnick:
Leverages new test flag which dumps to STDOUT so we can remove the usage of caching server.

### DIFF
--- a/packages/dcos-integration-test/extra/integration_test.py
+++ b/packages/dcos-integration-test/extra/integration_test.py
@@ -1635,17 +1635,17 @@ def test_signal_service(registry_cluster):
     exp_data = {
         'diagnostics': {
             'event': 'health',
-            'anonymousId': 'test-id',
+            'anonymousId': cluster_id,
             'properties': {}
         },
         'cosmos': {
             'event': 'package_list',
-            'anonymousId': 'test-id',
+            'anonymousId': cluster_id,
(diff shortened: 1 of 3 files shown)


cmaloney:
This seems to only update ref_origin but not actually change what is built.

malnick:
@spahl @cmaloney This is ready for review. Works on Vagrant!

spahl:
:ship:

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1610,87 +1611,40 @@ def test_3dt_report(cluster):
...
-    print('CACHE SERVER STATUS:\n{}'.format(r.status_code))
+    dcos_version = os.getenv("DCOS_VERSION", "")
+    signal_config = open('/opt/mesosphere/etc/dcos-signal-config.json', 'r')
+    signal_config_data = json.loads(signal_config.read())
+    customer_key = signal_config_data.get('customer_key', '')
+    cluster_id_file = open('/var/lib/dcos/cluster-id')
+    cluster_id = cluster_id_file.read().strip()
 
-    r_data = json.loads(r.json())
+    print("Version: ", dcos_version)
```
cmaloney: we shouldn't be printing here...

PR #385, opened 2016-07-21
Bump networking dcos

sargun:


alberts:
Went in as https://github.com/dcos/dcos/pull/394

PR #386, opened 2016-07-21
Add naive test environment

mellenburg:
Helps facilitate on-prem debugging

### DIFF
--- a/packages/dcos-integration-test/extra/test_env.export
+++ b/packages/dcos-integration-test/extra/test_env.export
@@ -1,0 +1,17 @@
+#!/bin/bash
+# This file sets the missing environment variables necessary to run the py.test
+# in dcos-integration-test WARNING: this script naively populates the master and
+# agent lists with nodes that have already succesfully joined the cluster. Errors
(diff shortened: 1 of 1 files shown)


mellenburg:
@alberts here is a script that does some naive arg guessing and leave the user to invoke py.test

mellenburg:

packages/dcos-integration-test/extra/naive_test_env.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
malnick: Can we have some comments up here on what this file does?

spahl:
@mellenburg can you rename this to something else? Like run_in_cluster.sh or something?

mellenburg:
@spahl Sure, I will also add a comment as @malnick pointed out as well.

However, this script doesnt run anything, just sets the env and leaves the user to run py.test (very intentional). So how about test_env.sh?

spahl:
Oh I see, sounds good.

spahl:
:ship:

PR #389, opened 2016-07-21
Adjust cluster_registry marathon app

mellenburg:
-remove default values in health checks
-add large timeout for slow build and push
-read test_server resources off disk, not via definition


cmaloney:
🚢

PR #387, opened 2016-07-21
Marathon build/config cleanup

adam-mesos:

PR #365, opened 2016-07-19
exhibitor updates

alberts:

PR #393, opened 2016-07-21
Train 35

cmaloney:
#278, #379, #387, #391 

Revert: #365. It broke the AWS CF templates.


cmaloney:
Launched a CF stack of this PR. The reverts make it so the clusters come up again.

Mesos UI sandbox download links work properly now.

PR #395, opened 2016-07-21
Add minimum version to the local build git requirement.

kaysoky:
A local build will end up shelling out with `git -C ...`.
This option is introduced in Git 1.8.5:
https://github.com/git/git/commit/44e1e4d67d5148c245db362cc48c3cc6c2ec82ca

Note: This is relevant because the default git package on CentOS 7 is version 1.8.3.


cmaloney:
🚢 . Should actually just remove the uses of `-C` (I did this once a while back, but it's quite a bit more convenient than specifying the two flags so started adding it back in).

PR #396, opened 2016-07-21
Breakout integration_test.py

mellenburg:
Death to monoliths


cmaloney:
Why does this lose 36 lines? I'd expect it to add lines (We have to have more import statements)

mellenburg:
@cmaloney It should have more lines. I see: +1,978 −1,942. So I think you might have just read that backwards. Otherwise, the only significant change here is that more waiting needs to be added to the fixture as some tests were still relying on other tests to have run it seems, as in this case: https://teamcity.mesosphere.io/viewLog.html?buildId=340952&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp

mellenburg:
@cmaloney @spahl This is ready for review. The only thing that was not copy paste in this PR was in test_3dt.py I added some retrying to account for the latency of pulling the state. If we want to discuss making the whole cluster test fixture wait instead, I think we should do that in a different PR

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -0,0 +1,482 @@
...
+TEST_APP_NAME_FMT = '/integration-test-{}'
+# If auth is enabled, by default, tests use hard-coded OAuth token
+AUTH_ENABLED = os.getenv('DCOS_AUTH_ENABLED', 'true') == 'true'
+# Set these to run test against a custom configured user instead
+LOGIN_UNAME = os.getenv('DCOS_LOGIN_UNAME')
+LOGIN_PW = os.getenv('DCOS_LOGIN_PW')
+
+
+@pytest.fixture(scope='module')
+def cluster():
```
malnick: In our closed source we refer to this as `dcos`` and here as `cluster`. We should homogenize this naming convention.

mellenburg: That is the overall direction to head in, but in a different PR as the differences between the two are much deeper than just naming

spahl:
:ship:

PR #399, opened 2016-07-22
Move to dynamic ports for Minuteman and Octarine test apps

cmaloney:
The marathon ports API is weird, when you don't say  requirePors, the port being specified is the apps "servicePort", and the app gets launched / assigned a random port on the host. The tests ended up assigning multiple apps the same service port (Which gets caught by newer versions of marathon).

This switches all of the ports to always be dynamic from Marathon so we don't have to keep track of any interference. Another option would have been to use `requirePorts` and change the ports to be unique by hand.

### DIFF
--- a/packages/dcos-integration-test/extra/integration_test.py
+++ b/packages/dcos-integration-test/extra/integration_test.py
@@ -88,7 +88,6 @@ def registry_cluster(cluster, request):
                     "value": "curl -sSfv https://registry.marathon.mesos.thisdcos.directory:$PORT0/v2/_catalog"}
                 }],
             "ports": [0],
-            "requirePorts": True
             }
     endpoints = cluster.deploy_marathon_app(registry_app)
     cluster.registry = 'registry.marathon.mesos.thisdcos.directory:'+str(endpoints[0].port)
@@ -970,7 +969,7 @@ def test_octarine_http(cluster, timeout=30):
         'id': '/integration-test-app-octarine-http-{}'.format(test_uuid),
         'cpus': 0.1,
         'mem': 128,
-        'ports': [10001],
+        'ports': [0],
         'cmd': '/opt/mesosphere/bin/octarine marathon',
         'disk': 0,
         'instances': 1,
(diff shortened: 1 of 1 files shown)


cmaloney:
cc: @sargun, @nlsun

gkleiman:
LGTM

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -970,7 +970,8 @@ def test_octarine_http(cluster, timeout=30):
         'id': '/integration-test-app-octarine-http-{}'.format(test_uuid),
         'cpus': 0.1,
         'mem': 128,
-        'ports': [10001],
+        'ports': [0],
+        'requirePorts': True,
```
gkleiman: This looks really strange. What do you want to accomplish here?

packages/dcos-integration-test/extra/integration_test.py
```diff
@@ -1047,18 +1048,18 @@ def test_if_minuteman_routes_to_vip(cluster, timeout=125):
...
-                'port': 10000,
+                'port': 0,
                 'protocol': 'tcp',
                 'name': 'test',
                 'labels': {
                     'VIP_0': '1.2.3.4:5000'
                 }
             }
         ],
+        'requirePorts': True,
```
gkleiman: Same as before, asking for a dynamic port + `requirePorts=true` shouldn't even be allowed by Marathon.

packages/dcos-integration-test/extra/integration_test.py
cmaloney: I was thinking "why are we even bothering to hand-specify a port, we should let Marathon pick one out which is non-conflicting for us rather than maintain a list manually", and have been iterating. Was likely going to drop the requirePorts and just do the `'ports': [0]` shortly.

packages/dcos-integration-test/extra/integration_test.py
cmaloney: I definitely don't disagree, just iterating / temporary state.

PR #397, opened 2016-07-22
Add docs for Ubuntu 16.04 build dependencies

lingmann:
Additional information on how to build on Ubuntu 16.04 using Pyenv and
Python 3.4.4. FIXES: DCOS-254

PR #400, opened 2016-07-22
Train 36

cmaloney:
#359, #366, #395, #397, #399


cmaloney:
#399 is getting some additional changes, will need to be pulled in fresh.

PR #401, opened 2016-07-22
use latest 3dt

darkonie:
- update history service location
https://github.com/dcos/3dt/pull/45

- make 3dt find flapping systemd units
https://github.com/dcos/3dt/pull/44

### DIFF
--- a/packages/3dt/buildinfo.json
+++ b/packages/3dt/buildinfo.json
@@ -4,7 +4,7 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/dcos/3dt.git",
-    "ref": "210c1c86fdd7e55887ec239ec730ad458c8a7cc1",
+    "ref": "8b1b65f05e521548c40f42a034e1a1c7794cd2a1",
     "ref_origin": "master"
   },
   "username": "dcos_3dt",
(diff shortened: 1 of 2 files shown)


cmaloney:
@darkonie Is there any way we can test to make sure things like finding the history service / integration works?

darkonie:
@cmaloney i have an idea of embedding a set of self tests which can be triggered by a CLI flag. What do you think about it? The issue is that under normal circumstances it is difficult to catch that kind of bugs

malnick:
@cmaloney - @darkonie is going to implement a /info endpoint here so we can catch all the configuration 3DT has when running, we can use that to detect how it's getting data such as metrics which might be any of 3 sources.

malnick:
Can we get a WIP label on this PR?

darkonie:
@cmaloney @spahl ready to review

gen/calc.py
```diff
@@ -380,7 +380,7 @@ def validate_os_type(os_type):
                 "prefix": 26                            \
               }                                         \
             ]}',
-        'rexray_config_method': 'empty'
+        'rexray_config_method': 'disabled'
```
cmaloney: this is old / current is still 'empty' to not break backwards compatibility.

spahl:
:ship:

PR #402, opened 2016-07-22
Switch to using 9/8 for the subnet to assign agent subnets

sargun:
According to this NANOG post, it is unused: https://www.nanog.org/mailinglist/mailarchives/old_archive/2003-04/msg00762.html We're using 9/8 because with the way the overlay works, each agent  takes a fixed sized subnet. We want to be able to run at least 100 Docker container on each machine. The agent takes the aforementioned subnet and splits it in half between Docker and Mesos containes. There is a product requirement to support 100+ containers per machine using this feature. We also want to support 1000+ agents.

Therefore, a reasonable approach is to give the DC/OS cluster a /8 (9/8) that's unlikely to conflict. The middle two octects are per agent, so we can support ~2**16-Masters agents

This is the default, and those doing business in a 9/8 environment can modify the defaults for their needs.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -356,13 +356,13 @@ entry = {
         'dcos_overlay_mtu': '1420',
         'dcos_overlay_enable': "true",
         'dcos_overlay_network': '{                      \
-            "vtep_subnet": "198.18.0.0/20",             \
+            "vtep_subnet": "44.128.0.0/20",             \
             "vtep_mac_oui": "70:B3:D5:00:00:00",        \
             "overlays": [                               \
               {                                         \
                 "name": "dcos",                         \
-                "subnet": "44.128.0.0/16",              \
-                "prefix": 26                            \
+                "subnet": "9.0.0.0/8",                  \
+                "prefix": 24                            \
               }                                         \
             ]}',
         'rexray_config_method': 'empty'


gen/calc.py
```diff
@@ -356,13 +356,13 @@ def validate_os_type(os_type):
...
         'dcos_overlay_network': '{                      \
-            "vtep_subnet": "198.18.0.0/20",             \
+            "vtep_subnet": "44.128.0.0/20",             \
             "vtep_mac_oui": "70:B3:D5:00:00:00",        \
             "overlays": [                               \
               {                                         \
                 "name": "dcos",                         \
-                "subnet": "44.128.0.0/16",              \
-                "prefix": 26                            \
+                "subnet": "9.0.0.0/8",                  \
```
asridharan: Since this subnet is "owned" by IBM, could there be a backlash for using a subnet owned by another entity? I understand that IBM does not expose these subnets, so they are not routeable but still wanted to make sure that we are not called out publicly for being a bad citizen :).

sargun: Meh. These are the defaults. The number of people whose DC/OS will integrate with IBM are fewer than the numbers who will. Having an experience in which they do not need to drift from the original configuration simplifies the deployments that can handle it.

I don't know who can answer the squishy question of "bad citizenship" - Maybe @spahl

spahl: I'm good with that. Please update the PR with a description explaning why we are using 9/8 for the time being and what the limitations are (for posterity).

sargun: @spahl Yes. Added to commit notes as well.

spahl:
:ship:

gen/calc.py
cmaloney: I'm doubtful about the IBM point. IBM is a major player in enterprise software. DC/OS is being sold to enterprises...

sargun:
@spahl So we can close this?

gen/calc.py
spahl: Yes but it's only a default. We will make it very clear in the docs and release notes that if you are peering with 9/8 addresses (very unlikely) you need to change the default.

PR #408, opened 2016-07-24
Added batch write support to overlay replicated log.

asridharan:
Also, modified overlay master recovery so that the state of recovery
can be inferred from the replicated log store rather than by maintain
explicit state variables, which was error prone.


asridharan:
Ready for review !!

cmaloney:
Can we get review from someone on the mesos core / C++ team for the code changes here?

cmaloney:
🚢 . Talked offline, has been reviewed by Jie.

PR #409, opened 2016-07-25
Bump Mesos to 1.0-rc4.

adam-mesos:


spahl:
:ship:

adam-mesos:
Note that this will require any downstreams depending on 1.0-rc2 or earlier to adjust a few Mesos flags for http authentication and SSL.

PR #411, opened 2016-07-25
pin flake8 in tox

mellenburg:


cmaloney:
@mellenburg. The subject should be more of a "pin flake8 in tox". This isn't pinning tox.

spahl:
Please repush with a new commit message too. I'm gonna pull it in a train.

PR #414, opened 2016-07-25
Increase log rotation threshold for /var/log/mesos

spahl:
The default log rotation is way to small by mistake. This bumps it from 7*2000k to 7*256M

### DIFF
--- a/packages/logrotate/build
+++ b/packages/logrotate/build
@@ -43,7 +43,7 @@ nomail
 
 /var/log/mesos/* {
     olddir /var/log/mesos/archive
-    maxsize 2000k
+    maxsize 256M
     daily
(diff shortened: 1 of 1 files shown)


cmaloney:
This is really big for our default AWS templates. Those have a 5.5GB root disk, which is where /var/log/mesos lives. This will cause that to fill significantly (1.8GB) + the fact most the logs are also in the journal (Although they are compressed there).

I worry that with long running instances in AWS we'll run out of disk space very quickly.

spahl:
Yeah it increases it a bit but they are also compressed by log rotate (although with delaycompress) so 256+256 + compressed other days. No matter what it's rotated daily.

PR #415, opened 2016-07-26
script to run tests under dcos-docker

alberts:


spahl:
:ship:

cmaloney:
This dirties up the current directory and doesn't add things to gitignore and the like. Also, if we ran it in CI would make the hosts very dirty / usable only for one build then have to be wiped.

test_util/run-docker
```diff
@@ -0,0 +1,31 @@
...
+  git -C dcos-docker fetch origin -t +refs/heads/*:refs/remotes/origin/*
+fi
+
+curl -v $INSTALLER_URL > dcos-docker/dcos_generate_config.sh
+
+set +o errexit
+
+finish
+make -C dcos-docker
+make -C dcos-docker test
```
malnick: Do we need to pass env variables here to configure `MASTER_COUNT` and `AGENT_COUNT` and INSTALLER_URL` as specified in the #417?

PR #416, opened 2016-07-26
Fix REX-Ray default config

branden:
[REX-Ray will fail with an empty config file](https://teamcity.mesosphere.io/viewLog.html?buildId=341742&buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr&tab=buildLog#_focus=1615). This PR adds a new config that [disables the default Docker module](http://rexray.readthedocs.io/en/v0.3.3/user-guide/config/#disabling-modules). This config will allow REX-Ray to start, and cause dvdcli to fail with `Unable to locate plugin: rexray`.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -24,6 +24,14 @@ rexray:
       ignoreusedcount: true
 """
 
+DEFAULT_REXRAY_CONFIG = """
+rexray:
+  loglevel: info
+  modules:
+    default-docker:
+      disabled: true
+"""
+
 
 def calculate_bootstrap_variant():
     variant = os.getenv('BOOTSTRAP_VARIANT')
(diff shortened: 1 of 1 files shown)


gen/calc.py
```diff
@@ -410,8 +418,8 @@ def validate_os_type(os_type):
             'aws': {
                 'must': {'rexray_config_contents': yaml.dump(AWS_REXRAY_CONFIG)},
             },
-            'empty': {
-                'must': {'rexray_config_contents': yaml.dump('')},
+            'disabled': {
```
branden: The name of this `rexray_config_method` value isn't mentioned in [Marathon docs](https://mesosphere.github.io/marathon/docs/external-volumes.html), so this change shouldn't be disruptive.

spahl:
:ship:

gen/calc.py
cmaloney: This is in the Enterprise docs: https://docs.mesosphere.com/1.7/administration/installing/custom/configuration-parameters/

I'd vote to just leave the old name, makes the docs simpler.

spahl: Agreed. @branden ^

PR #417, opened 2016-07-26
test script to wait for cluster coming up

mellenburg:
@alberts @spahl @malnick


packages/dcos-integration-test/extra/run_basic_test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
malnick: Can we call this file `run_integration_test.sh` ?

packages/dcos-integration-test/extra/wait_for_cluster.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Will continually poll the cluster until the desired number of masters and agents is reached
+
+while true; do
+    agent_count=`dig slave.mesos +short | wc -l`
+    master_count=`dig slave.mesos +short | wc -l`
+    if [ "$agent_list" -eq "$DCOS_NUM_AGENTS" ] && [ "$master_count" -eq "$DCOS_NUM_MASTERS" ]; then
```
malnick: `agent_list` -> `agent_count`?

packages/dcos-integration-test/extra/wait_for_cluster.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Will continually poll the cluster until the desired number of masters and agents is reached
+
+while true; do
+    agent_count=`dig slave.mesos +short | wc -l`
+    master_count=`dig slave.mesos +short | wc -l`
```
malnick: `slave.mesos` -> `master.mesos`?

packages/dcos-integration-test/extra/wait_for_cluster.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Will continually poll the cluster until the desired number of masters and agents is reached
+
+while true; do
```
cmaloney: Just use the python we already have.

cmaloney:
This should not have been merged. We already have the majority of the code in python. I don't want multiple copies of the "sort of wait for up. There's also no reason for us to use bash.

alberts:
Happy to delete this when there's an alternative way of achieving the same thing.

cmaloney:
There's 3 other copies of this same sort of logic in the repository, including one in the integration tests which was always supposed to run in the integration tests first.

There's also no reason to do this in python which the average knowledge level is farrr less than python.

alberts:
The problem we tried to solve here is that you can't run the tests in any environment without setting various environment variables. We want a simple command that can be run after SSHing into the cluster to run the tests, in any environment. What is that command?

cmaloney:
We an build that command, it's not very hard. Doesn't require duplicating already existing logic for checking if the cluster is up. My problem is this is yet another copy of code we don't have time to maintain. It works a little different than the other ones.

If you want to ssh into the cluster before dcos-link-env runs (if you ssh in after that all the dcos stuff is in your path), then we can even write the super complicated bash wrapper script to source /opt/mesosphere/environment.export for you then call python.

```
#!/bin/bash

source /opt/mesosphere/bin/environment.export

# Random python command to run all the tests, which is the same command the CI should use
py.test --vv <...>
```

PR #421, opened 2016-07-26
Hard fail if master count file doesn't exist.

cmaloney:
In tests we can point it to a file which does exist. It is a parameter.


cmaloney:
cc: @spahl

cmaloney:
Needs a fix for not trying to ever exhibitor_wait on agents.

PR #423, opened 2016-07-26
Teach release to build all dockers in the `pkgpanda/docker` folder

cmaloney:
Ideally all this logic should live inside pkgpanda (and the dockers should be lazily built / pulled as needed). This is an MVP so we don't have to maintain more dockers out of band.

cc: @vespian

### DIFF
--- a/release/__init__.py
+++ b/release/__init__.py
@@ -405,10 +405,10 @@ def make_abs(path):
     return os.getcwd() + '/' + path
 
 
-def do_build_packages(cache_repository_url):
-    dockerfile = pkg_resources.resource_filename('pkgpanda', 'docker/dcos-builder/Dockerfile')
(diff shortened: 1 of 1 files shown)


lingmann:
Nice... 👍

cmaloney:
🚢

vespian:
@cmaloney Do you think we could use Dockerfiles from package repos referenced in buildinfo.json instead of keeping a copy in dcos repo ?

This way I would have a way to keep dependencies close to source and use the same Dockerfile for building using my laptop as well as pkgpanda - no need for copypasting/keeping in sync these files.

cmaloney:
I think that would take a lot of work to implement currently. Makes the ordering of steps in doing things _much_ more complicated.

Currently all the dockers are built before we ever look at any packages / what might use them.

PR #426, opened 2016-07-26
exclude nodes failed in preflight step [WIP]

darkonie:
The nodes that do not pass the preflight must be excluded from further
actions.
DCOS-5697

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -7,7 +7,7 @@ import pkgpanda
 from ssh.ssh_runner import Node
 import ssh.utils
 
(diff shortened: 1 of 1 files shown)


cmaloney:
A couple things:
 - This will need a manual review that the UI works / looks right when there's a failed preflight node and all the further stages have less
- If one of the machines that fails is a master, the deployment shouln't proceed
- Rather than maintain an exclusion list and excluding it all the time, should just remove / trim the bad hosts from the deployment. Maybe put them into another list the user has / can access.
- Ideally can actually just re-run the preflight on the failed nodes. From a UI perspective would want a "Re-run preflight on host" option, or the option to "remove it from the set of hosts". Only once all hosts are resolved could the user continue.

PR #427, opened 2016-07-26
Update marathon endpoint

mellenburg:


cmaloney:
Can we get a more descriptive commit message?

If we're removing `/marathon` we should also put a note in NEWS

alberts:
We'll merge this a bit later.

alberts:
might bring back later.

PR #428, opened 2016-07-26
Propogate URL scheme (Fix test_composition.py)

mellenburg:
use protocol consistent with dns

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -2,6 +2,7 @@ import collections
 import logging
 import os
 import uuid
+from urllib.parse import urlparse
 
 import dns.exception
 import dns.resolver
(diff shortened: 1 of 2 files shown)


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -240,6 +240,7 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
 
         # URI must include scheme
         assert dcos_uri.startswith('http')
+        self.protocol = dcos_uri.split(':')[0]
```
cmaloney: If we want to start parsing up the uri, we should do with: https://docs.python.org/3/library/urllib.parse.html

mellenburg:
@cmaloney changed to use urllib.parse

mellenburg:
@spahl @cmaloney this is ready for review/merge

spahl:
:ship:

PR #429, opened 2016-07-26
Replace registry_cluster fixture

mellenburg:
marathon apps should be calling docker CLI by default. This way is a bit more sensible and instead requires that the user running the test do the docker build

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -2,6 +2,7 @@ import collections
 import logging
 import os
 import uuid
+from subprocess import check_call
 
 import dns.exception
 import dns.resolver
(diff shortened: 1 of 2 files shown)


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -454,28 +465,8 @@ def registry_cluster(cluster, request):
...
-                }]
-            }
-    cluster.deploy_marathon_app(docker_build_and_push_app, timeout=500)
-
-    def kill_registry():
-        cluster.destroy_marathon_app(docker_build_and_push_app['id'])
-        cluster.destroy_marathon_app(registry_app['id'])
-    request.addfinalizer(kill_registry)
-    return cluster
+    check_call(['sudo', 'docker', 'build', '-t', '{}/test_server'.format(cluster.registry),
```
cmaloney: Don't assume passwordless sudo. Do assume sudoless docker.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -431,16 +432,26 @@ def _destroy_complete(deployment_id):
                         "completed in {} seconds.".format(timeout))
 
 
-@pytest.fixture(scope='module')
-def registry_cluster(cluster, request):
+@pytest.yield_fixture(scope='module')
```
jgehrcke: Did you confirm that to work? When using a `yield_fixture`, there should also be the `yield` keyword in the body.

mellenburg:
Closing this in favor of https://github.com/dcos/dcos/pull/440

PR #430, opened 2016-07-27
change scope of cluster fixture to session

mellenburg:
after splitting out the test into multiple modules,
this should be session scoped as to share apps
between tests


cmaloney:
🚢

PR #434, opened 2016-07-27
Remove dockercfg hook test push.

joerg84:


gen/calc.py
```diff
@@ -164,6 +178,12 @@ def validate_dcos_overlay_network(dcos_overlay_network):
                 "values are allowed: {}".format(ex))
 
 
+def validate_dcos_remove_dockercfg_enable(dcos_remove_dockercfg_enable):
+    can_be = ['true', 'false']
```
cmaloney: nit / not yours: Overall really need to move the "one of" / can be stuff to a common helper all of these call.

packages/mesos-modules/build
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -o nounset -o pipefail
```
cmaloney: Don't need this / pkgpanda does it for you (please remove here, don't like copy / paste tech debt)

packages/mesos-modules/build
```diff
@@ -0,0 +1,17 @@
...
+
+pushd /pkg/src/mesos-modules/
+./bootstrap
+popd
+
+export LD_LIBRARY_PATH=/opt/mesosphere/lib
```
cmaloney: Rather than do this, might be worthwhile to just `source /opt/mesosphere/environment.export`, which will cause you to pick up LD_LIBRARY_PATH among other things (/opt/mesosphere/bin added to $PATH, etc). Be good to start with that.

cmaloney:
Overall looking good. Need to remove the `set -o` stuff. 

Be nice if you could switch to using /opt/mesosphere/environment.export, but not critical for landing.

packages/mesos-modules/buildinfo.json
```diff
@@ -0,0 +1,10 @@
+{
+  "requires": ["mesos"],
+    "docker": "dcos-builder",
+    "single_source" : {
+      "kind": "git",
+      "git": "https://github.com/joerg84/mesos-remove-dockercfg-modules2.git",
```
cmaloney: just adding a note here, this needs to move to `dcos/dcos-mesos-modules` pre-merge.

cmaloney:
🚢

cmaloney:
@joerg84 Would it be possible to get an integration test which launches an app with a (perhaps not real) dockercfg then confirms it gets deleted out of the sandbox if the cleanup is enabled, doesn't if it isn't?

joerg84:
@cmaloney There is actually a test already in the module repo, but I can happily turn this into an integration test. What is the timeframe for that test?

spahl:
Please do the test in another PR. I will merge that one. Timeline would be before GA.

spahl:
:ship:

PR #437, opened 2016-07-27
Prozlach/registry stuff

vespian:


vespian:
@mellenburg Could you please have a look ?

mellenburg:
I would leave off the changes to service_discovery test. It will be irrelevant with https://github.com/dcos/dcos/pull/440

:+1: on the test_applications.py and test_composition.py changes

packages/dcos-integration-test/extra/test_service_discovery.py
```diff
@@ -7,7 +7,7 @@
 MESOS_DNS_ENTRY_UPDATE_TIMEOUT = 60  # in seconds
 
 
-def _service_discovery_test(cluster, docker_network_bridge=True):
+def _service_discovery_test(rcluster, docker_network_bridge=True):
```
alberts: why rename this?

vespian: We already have a pytest fixture with the same name and currently it works because we supply this argument when this method. Keeping it this way is error-prone and risky.

vespian:
@mellenburg I have removed changes to service_discovery tests from PR.

PR #436, opened 2016-07-27
Upgrade python to 3.4.5

cmaloney:
Lets see what breaks.

PR #439, opened 2016-07-27
Train 41

cmaloney:
#436, #438, #443


cmaloney:
looks like something in the python really broke the bash. Will investigate (knock the python from the train if needed).

PR #440, opened 2016-07-27
Remove registry_cluster fixture

mellenburg:


mellenburg:
@cmaloney @spahl This is ready for review

cmaloney:
🚢  assuming the tests all pass. Would be nice to not use python:3.4.3-slim, it's still a lot bigger than needed (we have a python on the host). That can hopefully come in a future PR though.

mellenburg:
Test failure was just caught and should be green in the next run. Agreed on a smaller container, but for this PR its less moving pieces to keep the same image.

vespian:
@cmaloney Regarding our discussion yesterday about using dockerized python vs. host python.
```
$ sudo docker images | grep 3.4.5     
docker.io/python                        3.4.5-alpine                            746afe31a380        8 days ago          67.35 MB
docker.io/python                        3.4.5-slim                              40d4469f51f2        8 days ago          217.4 MB
```

PR #441, opened 2016-07-28
Include all packages in cluster_packages.json

branden:
This expands `cluster_packages.json` to include all packages for the variant, causing them to be `pkgpanda fetch`ed from the `repository_url` and `pkgpanda add`ed during cluster setup. These packages are still included in the bootstrap tarball, so the `pkgpanda` commands are no-ops. An upcoming PR will remove non-bootstrap packages from the bootstrap tarball.

The larger `cluster_packages.json` causes the simple CloudFormation templates to grow past their 51,200 byte limit, so this PR moves template validation after artifact upload, and validates them by URL rather than inline. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html.


pytest/test_release.py
```diff
@@ -605,6 +605,34 @@ def test_make_channel_artifacts(monkeypatch):
...
+                'bootstrap': 'ee_bootstrap_id',
+                'packages': ['ee-package--version'],
+            },
+        },
+        'all_completes': {
+            None: {
+                'bootstrap': 'bootstrap_id',
+                'packages': ['package--version'],
+            },
+            'ee': {
```
spahl: ee?

pytest/test_release.py
```diff
@@ -605,6 +605,34 @@ def test_make_channel_artifacts(monkeypatch):
             'installer': 'installer_bootstrap_id',
             'ee.installer': 'ee_installer_bootstrap_id'
         },
+        'complete_dict': {
+            None: {
+                'bootstrap': 'bootstrap_id',
+                'packages': ['package--version'],
+            },
+            'ee': {
```
spahl: ee?

pytest/test_release.py
```diff
@@ -605,6 +605,34 @@ def test_make_channel_artifacts(monkeypatch):
             'installer': 'installer_bootstrap_id',
             'ee.installer': 'ee_installer_bootstrap_id'
```
spahl: Same here

branden:
@spahl @cmaloney Ready for review!

@lingmann PTAL at changes to CloudFormation template validation.

branden:
Closing this in favor of #524, which went a lot smoother than I expected.

PR #442, opened 2016-07-28
Generate AWS Advanced Templates directly with gen library

lingmann:
Initial support for creating AWS templates with `dcos_generate_config.sh` using the argument `--aws-cloudformation` and a limited subset of `config.yaml` options.

Known bugs:
 - [x] Setting `aws_template_upload: true` doesn't make aws_template_storage_access_key_id and friends mandatory / validate
 - [x] Need to tell user about `config_id` folder + path information
 - [x] bootstrap_url is set wrong (Shouldn't contain the unique config_id path)
 - [x] repository_url is set wrong (Shouldn't contain the unique config_id path)

Sample config.yaml:
```
---
aws_template_storage_bucket: cody-test-advanced
aws_template_storage_bucket_path: some/path
aws_template_storage_region_name: us-west-2
aws_template_upload: true
aws_template_storage_access_key_id: <your_access_key_id>
aws_template_storage_secret_access_key: <your_secret_access_key>
```

Includes: #656

### DIFF
--- a/ext/dcos-installer/dcos_installer/backend.py
+++ b/ext/dcos-installer/dcos_installer/backend.py
@@ -5,13 +5,27 @@ libraries to support the dcos installer.
 import logging
 import os
 
+import boto3
+import botocore.exceptions
+import yaml
+
 import gen
+import gen.calc
+import gen.installer.aws
+import release
+import release.storage.aws
+import release.storage.local
 import ssh.validate
 from dcos_installer import config_util
-from dcos_installer.config import DCOSConfig
+from dcos_installer.config import DCOSConfig, stringify_configuration
 from dcos_installer.constants import CONFIG_PATH, IP_DETECT_PATH, SSH_KEY_PATH
 
 log = logging.getLogger()
+
+
+def print_messages(messages):
+    for key, error in messages.items():
+        log.error('{}: {}'.format(key, error))
 
 
 def do_configure(config_path=CONFIG_PATH):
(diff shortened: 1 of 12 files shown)


ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -27,6 +30,23 @@ def do_configure(gen_config):
...
+def do_aws_cf_configure(gen_config):
+    subprocess.check_output(['mkdir', '-p', SERVE_DIR])
+
+    print("gen_config:")
+    pprint.pprint(gen_config)
+
+    # TODO(cmaloney): CURPOS
+    gen_out = gen.installer.aws.make_custom_aws_templates(gen_config)
+
+    # TODO(lingmann): At this point, gen_out['artifacts'] is a list of Dicts with the keys 'channel_path',
```
cmaloney: Ideally we'd upload them all to s3, and there's all the code in `release` we could steal to do that from the "artifacts" form.

For now though, definitely just iterate over them, write them to <SERVE_DIR>/<channel_path>

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -36,6 +38,25 @@ def do_configure(config_path=CONFIG_PATH):
         return 0
 
 
+def do_aws_cf_configure():
+    """Tries to generate AWS templates using a custom config.yaml"""
+
+    # TODO(lingmann): Exception handling
+    config = yaml.load(open(CONFIG_PATH, 'r'))
```
cmaloney: https://github.com/dcos/dcos/pull/443 for stringifying now.

gen/installer/aws.py
```diff
@@ -139,6 +140,10 @@ def get_test_session(config=None):
 
 
 def get_cloudformation_s3_url():
+    # TODO(lingmann): Remove this hack and figure out how to calculate properly...
```
cmaloney: So ideally we should prbably pass in get_cloudformation_s3_url.

Options for the hacky method, easiest is probably to use monkeypatching to change the implementation of this function to just return the `config['cloudformation_s3_url']` that a user passes in.

gen/installer/aws.py
```diff
@@ -311,6 +321,52 @@ def gen_advanced_template(arguments, variant_prefix, channel_commit_path, os_typ
...
+def make_custom_aws_templates(arguments):
+    # TODO(lingmann): The basic structure of this function already exists in do_create() ... dedupe.
+    extra_packages = list()
+    artifacts = list()
+
+    # Setup base arguments
+    variant_base_args = deepcopy(arguments)
+
+    # TODO(lingmann): Figure out how to properly set: variant_prefix, channel_commit_path
+    variant_prefix = pkgpanda.util.variant_prefix(None)
```
cmaloney: There's an env var DCOS_VARIANT if I recall correctly.

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -36,6 +38,25 @@ def do_configure(config_path=CONFIG_PATH):
...
+
+    config['provider'] = 'aws'
+    config['bootstrap_id'] = os.environ['BOOTSTRAP_ID']
+    # TODO(lingmann): Calculate the bootstrap_url ...
```
cmaloney: This will be the same as the cloudformation_s3_url the user needs to provide.

gen/installer/aws.py
```diff
@@ -325,9 +326,10 @@ def make_custom_aws_templates(arguments):
     # Setup base arguments
     variant_base_args = deepcopy(arguments)
 
-    # TODO(lingmann): Figure out how to properly set: variant_prefix, channel_commit_path
-    variant_prefix = pkgpanda.util.variant_prefix(None)
-    channel_commit_path = 'STUB_channel_commit_path'
+    variant_prefix = pkgpanda.util.variant_prefix(os.getenv('DCOS_VARIANT'))
+    # TODO(lingmann): Figure out what (if any) equivilant there is for channel_commit_path for self-hosted templates.
+    # For now pull it from an optional env var.
```
cmaloney: For this channel_commit_path should probably just be an empty string. It doesn't make sense for the on-prem people.

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -36,6 +39,22 @@ def do_configure(config_path=CONFIG_PATH):
         return 0
 
 
+def do_aws_cf_configure():
+    """Tries to generate AWS templates using a custom config.yaml"""
+
+    # TODO(lingmann): Exception handling
+    yaml_config = yaml.load(open(CONFIG_PATH, 'r'))
```
malnick: you can pass `do_aws_cf_configure(config_path=CONFIG_PATH)` then:

```
config = DCOSConfig(config_path=config_path)
``` 

and this will load the YAML file for you.

cmaloney: @malnick That does the wrong thing, and sets a bunch of config which is incorrect for the AWS templates.

lingmann:
@cmaloney this is ready for a review. I'm now using the same codepath used by the release library to generate the AWS artifacts. Separately, we are creating integration tests which will test the generated templates. You can generate yourself with:

```
wget https://downloads.dcos.io/dcos/testing/pull/442/dcos_generate_config.sh
chmod +x dcos_generate_config.sh
mkdir -p genconf
echo -e "---\ncloudformation_s3_url: https://s3-us-west-2.amazonaws.com/jeremy-testing/dcos" > genconf/config.yaml
./dcos_generate_config.sh --aws-cloudformation
```

The resulting directory structure looks as follows:

```
genconf/serve/aws.html
genconf/serve/bootstrap/d7552bd48f051a987e30808acb5cd58f4aef707d.active.json
genconf/serve/bootstrap/...
genconf/serve/cloudformation/el7-zen-1.json
genconf/serve/cloudformation/...
genconf/serve/packages/dcos-metadata/dcos-metadata--setup_45bd2e731989db9b3f2b4b15ab852c47ce886bb0.tar.xz
genconf/serve/packages/dcos-metadata/...
genconf/serve/packages/dcos-config/dcos-config--setup_5f3dea04dd07405651d0d815be3364b50454ac88.tar.xz
genconf/serve/packages/dcos-config/...
```

The open questions in my mind are:

1. Should we provide something analogous to `channel_commit_path` for users building template artifacts locally? The user can control the exact path with the `cloudformation_s3_url` config option by setting it to a unique path on each build. However, it would be easy to overlook this and I think there is a high probability that end-users will overwrite important template artifacts unintentionally. Any suggestions here?

2. How should we handle config option validation? For these generated templates we can only support a subset of the standard options... for example `cloudformation_s3_url` and `resolvers` can be set, but other options like `exhibitor_address`, `s3_bucket`, `s3_prefix`, etc. can not be modified currently and we should error. We have validation code in both the gen library and the installer, but it's not clear to me where the best place is to do the validation for this new configuration mode.

cc: @spahl and @malnick

malnick:
## Reuse or Not
My PR adds three new arguments, that could possibly be conflated with current gen parameters - `cloudformation_s3_bucket`, `cloudformation_s3_key`, `cloudformation_s3_region`. These are user-facing, and I'm not sure if I should re-use the current `s3_bucket` and `s3_path` parameters already present, or create these new ones for CF generation only.

## Validation
@lingmann As for validation, and this is part of another PR since it'll be a big change, we should work towards using the code-paths that on-prem generation uses: validation and calculation funcs should go in `gen/calc.py`.

`resolvers` is already in there, but when we use the AWS provider I do not think its validation function is leveraged yet.

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -44,6 +59,36 @@ def do_validate_gen_config(gen_config):
...
+        for artifact in gen_out['artifacts']:
+            channel_path = artifact['channel_path']
+            local_content = artifact['local_content']
+            assert channel_path[-1] != '/', "Channel path looks like a directory and must be a file"
+            artifact_dst = SERVE_DIR + '/' + channel_path
+            log.debug('Writing artifact: {}'.format(artifact_dst))
+            subprocess.check_output(['mkdir', '-p', dirname(artifact_dst)])
+            with open(artifact_dst, 'w') as out:
+                out.write(local_content)
+        for pkg_id in gen_out['packages']:
```
cmaloney: nit: this should be using the "copy out atomically" code we wrote for fetch_bootstrap. This could potentially leave half-written package files in the filesystem.

ext/dcos-installer/dcos_installer/cli_dispatcher.py
```diff
@@ -136,6 +136,14 @@ def do_genconf(args):
     return 0
 
 
+def do_aws_cloudformation(args):
+    print_header("EXECUTING AWS CLOUD FORMATION TEMPLATE GENERATION")
+    code = backend.do_aws_cf_configure()
```
cmaloney: why not just `return code`?

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -27,6 +29,19 @@ def do_configure(gen_config):
     pkgpanda.util.write_json('/genconf/cluster_packages.json', gen_out.cluster_packages)
 
 
+def do_aws_cf_configure(gen_config):
+    gen_out = gen.installer.aws.do_create(
+            channel_commit_path='',
+            variant_arguments={None: gen_config},
```
lingmann: Needs to get bootstrap_variant from the environment...

cmaloney: fixed

cmaloney:
I think this is now to a baseline MVP in terms of end to end code

This needs some hand testing and probably a user test before we can land it. Integration test before 1.8 ships, although after we do this as a baseline. There is a unit test included.

Need to go though / double check the code changes for some of the bits shared with release / aren't too dangerous and are correct.

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -5,32 +5,260 @@
...
+    'ap-southeast-1': 's3-ap-southeast-1.amazonaws.com',
+    'ap-southeast-2': 's3-ap-southeast-2.amazonaws.com',
+    'ap-northeast-1': 's3-ap-northeast-1.amazonaws.com',
+    'eu-central-1': 's3.eu-central-1.amazonaws.com',
+    'eu-west-1': 's3-eu-west-1.amazonaws.com',
+    'sa-east-1': 's3-sa-east-1.amazonaws.com'
+}
+
+
+def validate_aws_template_storage_region_name(aws_template_storage_region_name):
```
cmaloney: nit: All these should probably live in a separate file

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -5,32 +5,260 @@
...
+    assert aws_template_storage_region_name in region_to_endpoint, \
+        "Unsupported AWS region {}. Only {} are supported".format(
+            aws_template_storage_region_name,
+            region_to_endpoint.keys())
+
+
+def validate_aws_bucket_access(aws_template_storage_region_name,
```
cmaloney: nit: Should unit test these validation functions and the error cases.

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -5,32 +5,260 @@
...
+    # object.
+    # NOTE: the copying across, as well as validation is guaranteed to succeed because we've already
+    # done a validation run.
+    full_config = gen.calculate_config_for_targets(config_targets, gen_config)
+    gen_config['bootstrap_url'] = full_config['bootstrap_url']
+    gen_config['provider'] = full_config['provider']
+    gen_config['bootstrap_id'] = full_config['bootstrap_id']
+    gen_config['cloudformation_s3_url'] = full_config['cloudformation_s3_url']
+
+    bootstrap_variant = full_config['bootstrap_variant'] if full_config['bootstrap_variant'] else None
```
cmaloney: nit: Needs a comment (this is converting from a bootstrap variant string to a bootstrap variant None or string)

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -5,32 +5,260 @@
...
+        download_url=full_config['cloudformation_s3_url'],
+        region_name=full_config['aws_template_storage_region_name'],
+        access_key_id=full_config['aws_template_storage_access_key_id'],
+        secret_access_key=full_config['aws_template_storage_secret_access_key'])
+
+    log.warning("Uploading to AWS")
+    release.apply_storage_commands({'aws': storage_provider}, storage_commands)
+    # TODO(cmaloney): Print where the user can find the files in AWS
+    # TODO(cmaloney): Dump out a JSON with machine paths to make scripting easier.
+
```
cmaloney: `return 0`

cmaloney:
The s3 bucket used for this needs a bucket policy which allows the stacks being launched from it to be read from. Probably want to make a mandatory opt-in/out parameter for users which if opted in adds this default policy (ideally a more restrictive one that only allows things in the account, but this would be good enough for the usability need):
```
{
  "Version":"2012-10-17",
  "Statement":[
    {
      "Sid":"AddPerm",
      "Effect":"Allow",
      "Principal": "*",
      "Action":["s3:GetObject"],
      "Resource":["arn:aws:s3:::<bucket_name>/*"]
    }
  ]
}
```

cmaloney:
Managed to successfully launch a cluster using the generated templates.

Will leave the s3 bucket policy usability fix. Going to double check a downstream dcos distribution works with this, then commit later this evening/

cmaloney:
Filed JIRAs for the major outstanding bits of work.

PR #444, opened 2016-07-28
Teach gen to make a special expanded config json blob, which contains all arguments but itself

cmaloney:
TODO 
 - [x] Get it actually building
 - [x] Add integration test that checks `master_quorum` is set (or some always derived variable)

### DIFF
--- a/gen/__init__.py
+++ b/gen/__init__.py
@@ -18,6 +18,7 @@ import json
 import logging as log
 import os
 import os.path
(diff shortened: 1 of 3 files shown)


cmaloney:
Be nice to have a more exhaustive test for config.yaml having user specified things in it, but all of our different templates have a different set of parameters...

spahl:
:ship:

PR #445, opened 2016-07-28
Include localhost entry on Azure clusters

lingmann:
Fixes an issue currently impacting Azure regions in Australia and North
Europe, where the Azure DNS servers are no longer responding with an
entry for `localhost`. The startup service currently overwrites the
default /etc/hosts file so that it is an idempotent operation and can be
run multiple times if the $(detect_ip) value changes.

### DIFF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -45,7 +45,7 @@ root:
       Type=oneshot
       EnvironmentFile=/etc/environment
       EnvironmentFile=/opt/mesosphere/environment
-      ExecStart=/usr/bin/bash -c "echo $(detect_ip) $(hostname) > /etc/hosts"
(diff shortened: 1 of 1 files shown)


cmaloney:
Do we know what in DC/OS is using localhost? My grepping only shows instances in our tests and documentation (Would be nice to get bugs against things that rely on DNS)

spahl:
We don't need this here. We will use https://github.com/dcos/dcos/pull/446 to protect ourselves from broken dns responses from localhost.

@cmaloney yeah I suspect it could be something deeper in one of our components or simply an ubuntu related issue separate from the us.

spahl:
Closing

lingmann:
@BenWhitehead, you uncovered a place in Cosmos relying on `localhost`, could you open a separate bug on that issue? It's not critical but would be good to not lose track of it.

BenWhitehead:
Created https://dcosjira.atlassian.net/browse/DCOS-266

PR #454, opened 2016-07-29
Dial back rp_filter=2 from all interfaces to only necessary ones.

djosborne:
See: https://dcosjira.atlassian.net/browse/DCOS-265


sargun:
I think you need:
```
ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/vtep*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done
ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/m-*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done
ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/d-*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done
```

djosborne:
@sargun updated!

spahl:
Yes unfortunately I'm looking too late and the instances are already gone. But none of the clusters came up even after relaunching one of the tests. Something must be wrong.

sargun:
So, the problem is that the overlay interfaces aren’t created until after the agent launches and sets up the overlay. This would be too early if we’re trying to confine the changes only to the VTEPs / Bridges. That's the reason the original patch applied to all / default. 

I think you should just relax your check. The only way that the container could spoof the IP is if it has a raw socket, which requires CAP_NET_RAW or CAP_SYS_ADMIN. With these capabilities, it could probably do worse on the local system itself.

sargun:
Alternatively, I think there is some systemd magic to call scripts / actions on interface up / creation? You can try that?

sargun:
@djosborne Opinions?

djosborne:
@sargun I investigated doing systemd callbacks on interface creation but felt that the solution could quickly become too hacky. I feel the correct fix here is to make the components which create the vnic's also be responsible for configuring the rp_filter to meet there specifications on it. But the vnic creation is spread across three components, and therefore seems like a sizable task to update them.

Therefore, we're pushing forward with making an option to degrade the rp_filter check in calico's engine. Hopefully later we can implement "the correct fix", but this feels like the surest way to quickly get Calico working in DC/OS 1.8

djosborne:
@sargun I've been doing some more research on this that I wanted to share what I've found -

1. [opening a raw socket only requires one permission: CAP_NET_RAW (not CAP_SYS_ADMIN)](http://man7.org/linux/man-pages/man7/raw.7.html). 
2. In Docker, an (otherwise) unprivileged container can be launched with just `CAP_NET_RAW` by specifiying `--cap-add=NET_RAW`. 

I don't know any scenarios where users might launch a container with just `CAP_NET_RAW`, but my point is that it is not always true that a container with `CAP_NET_RAW` "could probably do worse on the local system itself." It is possible to launch a container with **just** `CAP_NET_RAW`, but no other root privileges. And if that container is compromised, it could spoof its IP if `rp_filter=loose`.

Regardless, we're still pushing forward with changes to degrade our rp_filter check, and will warn the user of the above during Calico installation. Hopefully we can still dial back DC/OS' blanket rp_filter some point soon.

PR #455, opened 2016-07-29
Remove `ec2:DeleteVolume` permission from agents

branden:
Neither Marathon nore the Mesos docker/volume isolator will delete a
volume, so this permission is unnecessary.


spahl:
:ship:

branden:
I just discovered that REX-Ray 0.3.3 allows us to [tag the volumes it creates](http://rexray.readthedocs.io/en/v0.3.3/user-guide/storage-providers/#volume-tagging-optional), so instead of removing this permission we can limit it to just those tagged volumes. I'll follow up with that in another PR.

PR #457, opened 2016-07-29
fix installer bug in awk

darkonie:
the bug is reproduced if docker version returns Client: but not Server:


malnick:
CC @surdy @pyronicide @spahl

PR #461, opened 2016-08-01
Restrict volume permissions for CloudFormation agents

branden:
* DeleteVolume, AttachVolume, DetachVolume restricted to volumes with
  tag rexraySet = AWS::StackId
* AttachVolume, DetachVolume restricted to instances with the
  tag aws:cloudformation:stack-id = AWS::StackId

This has the happy side effect of [restricting a CloudFormation cluster's REX-Ray to viewing only the volumes created by that cluster](https://github.com/dcos/rexray/blob/v0.3.3-dcos/drivers/storage/ec2/ec2.go#L678).

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -90,12 +90,16 @@ def calculate_ip_detect_public_contents(ip_detect_contents):
     return ip_detect_contents
 
 
-def calculate_rexray_config_contents(rexray_config_filename):
+def calculate_rexray_config_from_file(rexray_config_filename):
     try:
         with open(rexray_config_filename, encoding='utf-8') as f:
             return yaml.dump(f.read())
     except IOError as err:
         raise Exception('REX-Ray config file {}: {}'.format(rexray_config_filename, err)) from err
+
+
+def calculate_rexray_config_from_string(rexray_config_string):
+    return yaml.dump(rexray_config_string)
 
 
 def calculate_gen_resolvconf_search(dns_search):
@@ -421,7 +425,10 @@ entry = {
         },
         'rexray_config_method': {
             'file': {
-                'must': {'rexray_config_contents': calculate_rexray_config_contents},
+                'must': {'rexray_config_contents': calculate_rexray_config_from_file},
+            },
+            'string': {
+                'must': {'rexray_config_contents': calculate_rexray_config_from_string},
             },
             'aws': {
                 'must': {'rexray_config_contents': yaml.dump(AWS_REXRAY_CONFIG)},
--- a/gen/aws/calc.py
+++ b/gen/aws/calc.py
@@ -3,6 +3,20 @@ import pkg_resources
 import yaml
 
 
+# Use IAM Instance profile for auth and tag volumes with StackId.
+REXRAY_CONFIG = """
+rexray:
+  loglevel: info
+  storageDrivers:
+    - ec2
+  volume:
+    unmount:
+      ignoreusedcount: true
+aws:
+  rexrayTag: { "Ref" : "AWS::StackId" }
+"""
+
+
 def get_spot(name, spot_price):
     if spot_price:
         return '"SpotPrice": "{}",'.format(spot_price)
@@ -45,7 +59,7 @@ entry = {
         'cloud_config': '{{ cloud_config }}',
         'oauth_available': 'true',
         'oauth_enabled': '{ "Ref" : "OAuthEnabled" }',
-        # Use IAM Instance profile for auth
-        'rexray_config_method': 'aws'
+        'rexray_config_method': 'string',
+        'rexray_config_string': REXRAY_CONFIG,
     }
 }
--- a/gen/aws/templates/advanced/advanced-priv-agent.json
+++ b/gen/aws/templates/advanced/advanced-priv-agent.json
@@ -139,9 +139,6 @@
                     "ec2:CreateTags",
                     "ec2:DescribeInstances",
                     "ec2:CreateVolume",
-                    "ec2:DeleteVolume",
-                    "ec2:AttachVolume",
-                    "ec2:DetachVolume",
                     "ec2:DescribeVolumes",
                     "ec2:DescribeVolumeStatus",
                     "ec2:DescribeVolumeAttribute",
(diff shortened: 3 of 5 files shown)


branden:
@cmaloney @spahl Ready for review.

branden:
@cmaloney @spahl Ready for review.

branden:
Pinging @lingmann as well for review of CF template changes. I manually tested `cloudformation.json` via CCM, but I'm not sure how I should be testing the advanced templates.

gen/calc.py
```diff
@@ -421,7 +425,10 @@ def validate_os_type(os_type):
         },
         'rexray_config_method': {
             'file': {
-                'must': {'rexray_config_contents': calculate_rexray_config_contents},
+                'must': {'rexray_config_contents': calculate_rexray_config_from_file},
+            },
+            'string': {
+                'must': {'rexray_config_contents': calculate_rexray_config_from_string},
             },
```
cmaloney: Should we drop the aws method here? Nothing is using it anymore.

cmaloney:
Just the one comment from me. I'll leave it to @lingmann to review the AWS CF stuff.

gen/calc.py
branden: +1, my only hesitation is that enterprise docs mention it: https://docs.mesosphere.com/1.7/administration/installing/custom/configuration-parameters/

Do we care? I can follow up with a docs PR to remove it if not.

cmaloney: We should unfortunately probably keep / test it then.... Might be worthwile to ping @pyronicide to see if he has any numbers around usage / if it's safe to drop. Easy non-CF AWS makes some sense to have around.

pyronicide: What is this used for/by?

branden: @pyronicide It gives you a preset REX-Ray config that does the right thing if your DC/OS agents are ec2 instances with IAM roles configured. On master it's used in CloudFormation templates, but with this PR I'm moving CF REX-Ray to a CF-specific config.

pyronicide: That sounds fine to me. How will this effect upgrades?

branden: The impact to upgrades would be if we removed this option, users would have to replace their `rexray_config_method: aws` with `file` or `string` and provide their REX-Ray config explicitly. "That sounds fine to me" == keep (and add a test for) the `aws` option?

pyronicide: Let's remove and make sure the upgrade instructions are documented.

lingmann:
Looked over this PR and I have two concerns:

Regards upgrades, the [AWS::IAM::Role](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html) Policies can be updated on a running stack with no interruption... which means that if someone does a live upgrade of a cluster with existing RexRay EBS volumes attached, those instances will likely lose access until the EBS volumes are manually tagged.

The other item I'm concerned about is stacks with multiple agent pools. Each agent pool will have a unique stack-id, which (I believe) means that EBS volumes will not be able to float between agent pools within the same cluster.

@branden I'd be happy to help explore / test this a bit further with you to better understand the impact here. I'm trying to simulate a stack upgrade now from 1.7 to the templates from this PR and will post the results here.

lingmann:
A few more details on what will happen to an advanced template deployment can be seen below. Here is a 1.7 private agent stack upgrade to 1.8 templates, as seen through the cloudformation [changesets API](https://docs.aws.amazon.com/console/cloudformation/changeset-details):

```
        {
            "Type": "Resource",
            "ResourceChange": {
                "Details": [
                    {
                        "ChangeSource": "ResourceAttribute",
                        "CausingEntity": "Infrastructure.Outputs.PrivateAgentSecurityGroupId",
                        "Evaluation": "Dynamic",
                        "Target": {
                            "Attribute": "Properties",
                            "RequiresRecreation": "Never",
                            "Name": "Parameters"
                        }
                    },
                    {
                        "ChangeSource": "ResourceAttribute",
                        "CausingEntity": "MasterStack.Outputs.InternalMasterLoadBalancerDnsName",
                        "Evaluation": "Dynamic",
                        "Target": {
                            "Attribute": "Properties",
                            "RequiresRecreation": "Never",
                            "Name": "Parameters"
                        }
                    },
                    {
                        "ChangeSource": "DirectModification",
                        "Evaluation": "Dynamic",
                        "Target": {
                            "Attribute": "Properties",
                            "RequiresRecreation": "Never",
                            "Name": "Parameters"
                        }
                    },
                    {
                        "ChangeSource": "ParameterReference",
                        "CausingEntity": "CustomAMI",
                        "Evaluation": "Static",
                        "Target": {
                            "Attribute": "Properties",
                            "RequiresRecreation": "Never",
                            "Name": "Parameters"
                        }
                    },
                    {
                        "ChangeSource": "DirectModification",
                        "Evaluation": "Static",
                        "Target": {
                            "Attribute": "Properties",
                            "RequiresRecreation": "Never",
                            "Name": "TemplateURL"
                        }
                    }
                ],
                "Scope": [
                    "Properties"
                ],
                "PhysicalResourceId": "arn:aws:cloudformation:us-west-2:273854932432:stack/jeremy-20160803181241-PrivateAgentStack-BCOH2JVGM4EC/3abaa5e0-59a7-11e6-9e0a-50d5ca789e4a",
                "LogicalResourceId": "PrivateAgentStack",
                "ResourceType": "AWS::CloudFormation::Stack",
                "Action": "Modify",
                "Replacement": "False"
            }
        }
```

lingmann:
@branden I think next step we should work together to try and simulate an upgrade on a cluster with RexRay volumes. There is probably some work here that could also be leveraged to add AWS template upgrades to our integration tests.

branden:
@lingmann

> Regards upgrades, the AWS::IAM::Role Policies can be updated on a running stack with no interruption... which means that if someone does a live upgrade of a cluster with existing RexRay EBS volumes attached, those instances will likely lose access until the EBS volumes are manually tagged.

Tags are our only option for scoping these permissions, so we'll need to document how to tag volumes prior to an upgrade so that the cluster doesn't lose access to them. This means the tag value needs to be available before the upgrade happens -- I'm not sure if that's possible with the StackId.

> The other item I'm concerned about is stacks with multiple agent pools. Each agent pool will have a unique stack-id, which (I believe) means that EBS volumes will not be able to float between agent pools within the same cluster.

Good catch, that is indeed a problem. We would need to tag volumes with a cluster-unique value that's shared by the agent pools. I'll read through the advanced CF templates and see if I can come up with something.

> @branden I think next step we should work together to try and simulate an upgrade on a cluster with RexRay volumes. There is probably some work here that could also be leveraged to add AWS template upgrades to our integration tests.

👍  I'll start working on automated upgrade tests pretty soon, so this would dovetail nicely.

lingmann:
@branden 

It is possible to get the stackid(s) of a running stack prior to upgrade and they do not change, so just documenting the requirement sounds good to me. With a Zen Advanced stack deploy, there are actually 5 stack id's per cluster... one for each instantiated template (zen, infra, master, public agent, private agent). So if the tag used with EBS volumes is the stackid of the Master stack, that would be available, the user would just have to inspect the running Master stack prior to the upgrade.

As for finding a common identifier to use within the scope of private or public agent templates, we'll need to either add a parameter to Zen, Master, and Agent templates or use one of the existing variables which are passed from the Master stack to the Agent stacks. You can see what parameters are currently passed between templates in this [PrivateAgentStack](https://github.com/dcos/dcos/blob/master/gen/aws/templates/advanced/zen.json#L171) section of the zen template.

It might even make sense to add a new input parameter to the templates for this which is completely managed by the end-user, e.g. `RexRayEbsTag`. That way multiple clusters could share EBS volumes if they really wanted to. However it will add to the complexity of the template input.

cmaloney:
@lingmann  / @branden what is the status fo this? Hasn't had any progress in a month.

branden:
This PR has been on ice while I've been focusing on upgrade stuff. I'll close this out and resubmit later once I have time to rebase and respond to the review here.

PR #462, opened 2016-08-01
installer: use unix socket

darkonie:
- add `docker.socket` unit. The unit will create a unix socket used by docker.

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -340,10 +340,15 @@ sudo mkdir -p /etc/systemd/system/docker.service.d
 sudo tee /etc/systemd/system/docker.service.d/override.conf <<- EOF
 [Service]
 ExecStart=
-ExecStart=/usr/bin/docker daemon --storage-driver=overlay -H fd://
+ExecStart=/usr/bin/docker daemon --storage-driver=overlay -H unix:///var/run/docker.sock
 EOF
 
-sudo yum install -y docker-engine-1.11.2
+sudo systemctl daemon-reload
+# try to stop the older docker version, but do not hard fail if it does not
+# exist.
+sudo systemctl stop docker || true
+
+sudo yum install -y docker-engine
 sudo systemctl start docker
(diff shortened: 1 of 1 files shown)


darkonie:
@cmaloney @spahl this is ready for review

spahl:
The docker package for 1.12 does not provide the systemd socket anymore. This is why this is added.

spahl:
:ship:

cmaloney:
Docker intentionally removed the docker.socket. that the `After=` wasn't removed as well: https://github.com/docker/docker/pull/25094

As such I'd rather not have us manually add it in. Docker 1.12 is definitely not fully baked yet.

cmaloney:
https://github.com/docker/docker/commit/04104c3a1e6cad30cb41b762e8832215466c0e95

spahl:
Sounds reasonable, let's move to the default socket then.

spahl:
:ship:

spahl:
Before we ship this we need to check if docker 1.12 does not break much more.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -340,10 +340,15 @@ def _add_prereqs_script(chain):
...
 ExecStart=
-ExecStart=/usr/bin/docker daemon --storage-driver=overlay -H fd://
+ExecStart=/usr/bin/docker daemon --storage-driver=overlay -H unix:///var/run/docker.sock
 EOF
 
-sudo yum install -y docker-engine-1.11.2
+sudo systemctl daemon-reload
+# try to stop the older docker version, but do not hard fail if it does not
+# exist.
+sudo systemctl stop docker || true
```
alberts: If we're trying to get rid of old dockers, we should probably try to stop docker.socket and docker.service?

malnick: Why true?

cmaloney:
Hasn't had progress in a month, closing. Will try picking back up so hopefully DC/OS 1.9 is compatible with Docker 1.12

PR #469, opened 2016-08-02
Configure REX-Ray within config.yaml

branden:


branden:
@cmaloney This is ready for review, though I'll need to rebase it on #554 due to conflicts.

gen/calc.py
```diff
@@ -95,12 +71,16 @@ def calculate_ip_detect_public_contents(ip_detect_contents):
...
 
 
-def calculate_rexray_config_contents(rexray_config_filename):
-    try:
-        with open(rexray_config_filename, encoding='utf-8') as f:
-            return yaml.dump(f.read())
-    except IOError as err:
-        raise Exception('REX-Ray config file {}: {}'.format(rexray_config_filename, err)) from err
+def calculate_rexray_config_contents(rexray_config):
+    assert isinstance(rexray_config, str)
```
cmaloney: this will always be strings (hard guarantee from the gen library. It asserts everything which is returned from these functions or passed as an argument to one is a string).

cmaloney:
:ship:

PR #470, opened 2016-08-02
Add a pkgpanda package that will say what roles are active

cmaloney:
This makes it so we have a state of what is currently running, instead of the state of what might be running.

cc: @alberts @orsenthil @lingmann @spahl


orsenthil:
LGTM.

PR #472, opened 2016-08-02
[Mesos] Limit the default number of libprocess worker threads for executors to 8.

jmlvanre:
This limitation improves performance and memory consumption on hosts with large number of cores.
Schedulers running custom executors can still override this value by providing it in their executor definition.


cmaloney:
Test failure was flaky tests.

Would it make sense to set this on the mesos agent as a whole at the very least, as well as potentially marathon and metronome? Seems like they'd all have this same thing, not a great deal of benefit to running with a ton of libprocess worker threads (Even the master seems like it wouldn't actually gain much more parallelism for the most part).

cmaloney:
🚢 . Discussed offline with @jmlvanre some. We'll look into setting it more places in the next release

PR #473, opened 2016-08-03
Run Marathon/Metronome tasks as 'root' by default.

adam-mesos:
Marathon/Metronome are now run by the dcos_marathon and dcos_metronome linux users. Their schedulers default to setting FrameworkInfo.user to the user that launched the scheduler process. Since this used to be 'root', many users may be accustomed to just having access to things. Let's set the default FrameworkInfo.user to 'root' for now, since that's more sane than 'dcos_marathon', which is supposed to be just for root Marathon itself. In the future, I'd recommend we default this to 'nobody' or a generic 'dcos' user.
Note that this user can always be overridden on a per-task basis.


adam-mesos:
@spahl Updated. PTAL

cmaloney:
🚢

kensipe:
integration test: https://github.com/mesosphere/marathon/pull/4208

PR #477, opened 2016-08-03
Excluding docker-engine from yum update in preflight

jcortejoso:
We have faced a problem when retrying the preflight phase. The problem was that if during the first attempt docker has been installed (or it already installed in the machines), the command `sudo yum -y update` will update docker to the lastest version (currently 1.12). The PR excludes package docker-engine in the `sudo yum -y update` command, but it could be resolved by other ways (for example blacklisting the update of docker-engine package)

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib/__init__.py
+++ b/ext/dcos-installer/dcos_installer/action_lib/__init__.py
@@ -334,7 +334,7 @@ gpgcheck=1
 gpgkey=https://yum.dockerproject.org/gpg
 EOF
 
-sudo yum -y update
+sudo yum -y update --exclude=docker-engine
 
 sudo mkdir -p /etc/systemd/system/docker.service.d
 sudo tee /etc/systemd/system/docker.service.d/override.conf <<- EOF


cmaloney:
So we've already landed: https://github.com/dcos/dcos/pull/458 to prevent docker from being updated / pin it to 1.11, does that work for you?

jcortejoso:
@cmaloney that's perfect, but the problem I faced that commit was already present. The problem is not that preflight script doesn't install docker version 1.11.2. The problem is that if you run preflight script more than once (because some node failed and you try using retrying option), when the preflight script is run this second time, because docker repository is present and docker-engine is installed, running `yum update -y` ([this line](https://github.com/darkonie/dcos/blob/5287de3dd4d319d19d8c989585a63d6bb660c2b8/ext/dcos-installer/dcos_installer/action_lib/__init__.py#L337)) will update all the packages installed in the system, and that includes docker-engine. 
By the way I don't know if you want to perform a complete system update when running the preflight phase, but maybe it's more secure to do not run it and avoid the updating of any conflictive package. (yum will automatically perform a repo list update when installing any package if the cache is not updated).

cmaloney:
So there's actually two different steps that are "combined" when you use the web installer:
`--install-prerequisites`: Installing all the things DC/OS requires
`--preflight`: Make sure the host has all the dependencies of DC/OS and don't install if it doesn't.

When you use the SSH installer, or install by hand, you can skip the `--install-prerequisites` phase. The `--install-prerequisites` phase is not repeatable (whereas the actual `--preflight` is not). This is sounding like currently, the web installer re-runs the `--install-prerequisites` phase in some circumstances, when it shouldn't.

jcortejoso:
@cmaloney That sounds good. As you said, may be the problem is related to running the --install-prerequisites twice. As you pointed, the installation mode I followed when the problem arises was the web installer. To be more clear, the problem I suffered is that after the "preflight" phase on the web installer some of my nodes had Docker 1.12 installed (those ones the nodes that the first time the preflight worked correctly), and the nodes that failed the first preflight, after fixing the problem and retrying the preflight, it had Docker version 1.11.2.

If you find there is a better way to avoid this or you'd like I can try to check in which conditions the --install-prerequisites are run more than once, or we can provide a patch to blacklist the Docker package from a unwanted upgrade running yum update.

cmaloney:
@jcortejoso I'm going to think about it some, get back to you in 24 hours or less. See if there's something that's small / simple which will work for 1.8 timeframe.

cmaloney:
Hey, @jcortejoso How does #482 look to you? Basically the installer will never run install prerequisites twice on a host.

jcortejoso:
@cmaloney #482 seems good. It should fix the problem describer in this PR so I think we can close it.

Thanks!

PR #484, opened 2016-08-03
Fix AWS template validation

lingmann:
AWS templates were not being validated even when the release
configuration contains the necessary credentials. This was because the
`get_session` function was being called with the top-level config object,
but it should be receiving the `['testing']['aws']` values.


lingmann:
Depends on #485 for the newly activated integration tests to pass.

lingmann:
@cmaloney could you please review and advise on how to fix the code-quality failures introduced by this change? I investigated briefly, but it was not clear to me how we might configure pytest skipping to facilitate.

config/dcos-release.config.yaml
```diff
@@ -13,6 +13,15 @@ storage:
     account_key: $AZURE_STORAGE_ACCESS_KEY
     container: dcos
     download_url: https://dcosio.azureedge.net/dcos/
+testing:
+  aws:
+    kind: aws_s3
+    bucket: downloads.dcos.io
+    object_prefix: dcos
```
cmaloney: This really needs to not be the same as the production, it will create and delete a bunch of random things in s3, and if the code is wrong / has a bug (part of what this tests for), then we could end up deleting downloads.dcos.io

cmaloney:
Looks good to me, and scanning the builds looks like it triggered properly.

spahl:
:ship:

PR #485, opened 2016-08-04
Only render aws_dnsnames for master templates

lingmann:
Fix issue introduced with 75985bf554754babd17ba11987d6d3b1f6794f1c  The
InternalMasterLoadBalancer and ElasticLoadBalancer attributes are not
available on the private and public agent templates. Resolve by only
rendering aws_dnsnames on the advanced master templates.

### DIFF
--- a/gen/installer/aws.py
+++ b/gen/installer/aws.py
@@ -228,15 +228,19 @@ def make_advanced_bunch(variant_args, template_name, cc_params):
     else:
         raise RuntimeError('Unsupported os_type: {}'.format(cc_params['os_type']))
 
-    results = gen.generate(
-        arguments=variant_args,
-        extra_templates=extra_templates,
-        cc_package_files=[
+    cc_package_files = [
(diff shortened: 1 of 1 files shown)


lingmann:
@cmaloney or @spahl can I get a review? This will fix the:

> Template error: instance of Fn::GetAtt references undefined resource

failures we are seeing with advanced agent templates.

cc: @alberts

lingmann:
Also, if there is a way to accomplish this with a switch statement in `dcos-config.yaml` I believe that would better... since it would mean the `aws_dnsnames` config would only be preset on master nodes, regardless of whether the simple or advanced templates are used.

cmaloney:
Every host gets all the same files / config, so pushing it into dcos-config.yaml or something similar unfortunately wouldn't really help. This looks reasonable to me. That we're going to be leaving the odd AWS Ref thing in random files on the agents and the public agents is unfortunate.

spahl:
:ship:

PR #486, opened 2016-08-04
Use marathon 1.2.0-RC7

unterstein:


unterstein:
will do an other PR for RC9

PR #498, opened 2016-08-05
exclude docker-engine* from upgrade

darkonie:

PR #500, opened 2016-08-06
Train 47

spahl:
#484
#497 
#498


spahl:
Remove #494 due to problems on top of master

PR #501, opened 2016-08-06
Introduce Legacy Resources Mode, allowing for more ports when disabled

sargun:
```
commit 496ce70b4be91fcfd1074d34be288af7ab29365b
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Sat Aug 6 15:26:20 2016 -0700

    gen: Add true_false_enum_validator helper for config validation, Introduce legacy_agent_resources_mode

    This commit adds a true_false_enum_validator to replace a pattern
    that was being repeated throughout gen/calc.py.

    legacy_agent_resources_mode makes it so that by default (if set to
    false), it will make nearly all ports 1-32000 available in offers,
    versus traditional mode, which restricts access to ports 1-1024.
```

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -34,6 +34,19 @@ rexray:
 """
 
 
+def enum_validator(enums):
+
+    def validator(enum):
+        assert enum in enums, 'Must be one of {}. Got {}.'.format(enums, enum)
+    return validator
+
+true_false_enum_validator = enum_validator(['true', 'false'])
+
+
+def validate_legacy_agent_resources_mode(legacy_agent_resources_mode):
+    true_false_enum_validator(legacy_agent_resources_mode)
+
+
 def calculate_bootstrap_variant():
     variant = os.getenv('BOOTSTRAP_VARIANT')
     assert variant is not None, "BOOTSTRAP_VARIANT must be set"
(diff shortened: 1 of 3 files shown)


sargun:
@spahl

cmaloney:
This has the same issues as last time it was porosed. People already have a hard time with "why doesn't my thing launch". By doing this we make that an even harder question for our support engineers to answer.

spahl:
I believe the benefits outweigh the costs. And we don't have to make it the default right away with this method. Lets chat all three next week.

sargun:
I propose that we flip the flag the other way, and it gives people the ability to opt-in if they read the manual. I think that sufficiently rules out the "I don't understand what happened?" case.

cmaloney:
Does not solve the "support can't tell what is going on or why things do / don't launch" problem.

sargun:
Won't the case be pretty obvious? If they are trying to launch a task on the !public_slave && port < 1025, then make sure the flag is flipped. By default (today), people can't launch these tasks, but on fresh installs, we can allow people to flip this flag. Also, state.json (which comes with the support package) would indicate what's going on pretty quickly.

cmaloney:
Talk to support about debugging "why won't my app launch". it's one of our biggest problems. Every week there is a couple people who need help with it.

sargun:
What's the alternative? I don't see how this is makes the average case any worse? In fact, it gives support an out, which is "Oh, you didn't enable the flag, feel free to do that, and it'll work"

sargun:
CC: @artemharutyunyan

cmaloney:
I'm pretty sure most customers won't accept "enable this flag and restart every task in your cluster" which is what this requires.

PR #507, opened 2016-08-09
exclude service check from preflight on agent node

darkonie:
if the preflight gets executed on agent or public agent node, do not check
for master services.

https://dcosjira.atlassian.net/browse/DCOS-278
https://mesosphere.atlassian.net/browse/DCOS-9045

### DIFF
--- a/gen/installer/bash.py
+++ b/gen/installer/bash.py
@@ -261,6 +261,13 @@ function check_all() {
     # error exit codes
     set +e
     echo -e "${BOLD}Running preflight checks${NORMAL}"
+    AGENT_ONLY=0
+    for ROLE in $ROLES; do
+        if [[ $ROLE = "slave" || $ROLE = "slave_public" ]]; then
+            AGENT_ONLY=1
+            break
+        fi
+    done
 
     check_preexisting_dcos
     check_selinux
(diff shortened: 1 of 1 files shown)


cmaloney:
Please keep the independent changes separate. Checking for port 53 only on masters is a very separate change from continuing to try installing things.

malnick:
This might be a good opportunity to pull utility funcs out of the action_lib init file and into a more clear location, such as action_utils.py. This can also be tossed into the bin of other big refactors on this library, and be another PR if not this one.

darkonie:
@cmaloney ready for review

gen/installer/bash.py
```diff
@@ -261,6 +261,10 @@
     # error exit codes
     set +e
     echo -e "${BOLD}Running preflight checks${NORMAL}"
+    AGENT_ONLY=0
+    if [[ $ROLES = "slave" || $ROLES = "slave_public" ]]; then
```
cmaloney: Roles is technically an array currently, so we need to iterate through all the roles to check (Moving it to not be an array because really we only ever want one role is in #303, still WIP).

gen/installer/bash.py
```diff
@@ -336,19 +340,22 @@
...
-      "8080 marathon" \
-      "3888 zookeeper" \
-      "8181 exhibitor" \
-      "8123 mesos-dns"
-    do
-      check_service $service
-    done
+    # Run service check on master node only
+    if [[ $AGENT_ONLY -eq 0 ]]; then
+        for service in \
```
cmaloney: We need to still do a service check on agents. Agents require a different set of ports, which we should gather. Not having them should result in a failure of the preflight.

cmaloney:
Some comments inline. Overall probably want to move so that running a preflight requires specifying a node role.

darkonie:
@cmaloney fixed

cmaloney:
@darkonie what about https://github.com/dcos/dcos/pull/507#discussion_r74332830? We still need to do a check on agents that a number of ports are free (5051 at the very least, as well as others).

darkonie:
@cmaloney added a check for agent ports. Do we need to check for TCP ports defined under All roles here: https://github.com/dcos/dcos/blob/master/docs/admin-ports.md ?
it looks like we are not checking those at all.

cmaloney:
@darkonie that document is unfortunately outdated / needs an update. What we should probably do at this point is spin up a 1.8 cluster, then do a `ss -tulnp` to get what all is listening, and make sure that we check all of those. Do it on a master, agent, and public agent (the latter two _should_ be identical but it's always good to check).

darkonie:
can you guys please double check TCP ports required to be free for service you worked on / know about @spahl @sargun @branden @malnick

cmaloney:
Just did a quick audit of enterprise DC/OS, that has 31 listening things on the masters that are software we ship... Going to do one of open later / soon. Probably the port list in the check is overly conservative.

cmaloney:
Looking at a current Open DC/OS testing/master cluster this is what is listening we need to check for:

# Master
53 -- spartan
80 -- adminrouter
443 -- adminrouter
1050 -- 3dt
2181 -- zookeeper
5050 -- mesos-master
7070 -- cosmos
8080 -- marathon
8101 -- dcos-oauth
8123 -- mesos-dns
8181 -- exhibitor
9000 -- metronome
9942 -- metronome
9990 -- cosmos
15055 -- dcos-history
33107 -- navstar
36771 -- marathon
41281 -- zookeeper
42819 -- spartan
43911 -- minuteman
46839 -- metronome
61053 -- mesos-dns
61053 -- mesos-dns
61420 -- epmd
61421 -- minuteman
62053 -- spartan
62080 -- navstar

System Services:
22 -- sshd
68 -- systemd-networkd
546 -- sytemd-networkd

# Agent
53 -- spartan
5051 -- mesos-agent
7979 -- rexray
34451 -- navstar
39851 -- spartan
43995 -- minuteman
61001 -- agent adminrouter
61420 -- epmd
61421 -- minuteman
62053 -- spartan
62080 -- navstar

System Services:
68 -- systemd-networkd
546 -- systemd-networkd

darkonie:
@cmaloney updated commit to check the port from your comment. As for the system services, do you want the preflight to check for those port to be available as well? or it should be some different kind of check. I assume port 22 on master might be used and we don't want to fail the preflight, right?

gen/installer/bash.py
```diff
@@ -336,19 +343,57 @@
...
+            "62080 navstar"
+        do
+            check_service $service
+        done
+    else
+        # agent / public agent node service checks
+        for service in \
+            "53 spartan" \
+            "5051 mesos-agent" \
+            "7979 rexray" \
```
cmaloney: drop this one since we're going to be moving it in #554

cmaloney:
For the system services, we can just ignore them. I was just doing a full port accounting of hosts inside a DC/OS cluster. The systemd-networkd + sshd we don't actually depend upon, so we wouldn't gain anything other than spurious failures if we checked them.

Need to address one comment then we can ship this.

gen/installer/bash.py
darkonie: done

darkonie:
@cmaloney removed `7979 rexray`

cmaloney:
:ship:

PR #508, opened 2016-08-09
Change delete EC2 volume to a metronome one-off

mellenburg:
cc @branden

### DIFF
--- a/packages/dcos-integration-test/extra/test_ec2.py
+++ b/packages/dcos-integration-test/extra/test_ec2.py
@@ -1,71 +1,8 @@
-import contextlib
 import copy
 import logging
-import os
 import uuid
 
-import boto3
-import botocore
 import pytest
-import requests
-import retrying
-
-
-@contextlib.contextmanager
-def _remove_env_vars(*env_vars):
-    environ = dict(os.environ)
-
-    for env_var in env_vars:
-        try:
-            del os.environ[env_var]
-        except KeyError:
-            pass
-
-    try:
-        yield
-    finally:
-        os.environ.clear()
-        os.environ.update(environ)
-
-
-def _delete_ec2_volume(name, timeout=300):
-    """Delete an EC2 EBS volume by its "Name" tag
-
-    Args:
-        timeout: seconds to wait for volume to become available for deletion
-
-    """
-    @retrying.retry(wait_fixed=30 * 1000, stop_max_delay=timeout * 1000,
-                    retry_on_exception=lambda exc: isinstance(exc, botocore.exceptions.ClientError))
-    def _delete_volume(volume):
-        volume.delete()  # Raises ClientError if the volume is still attached.
-
-    def _get_current_aws_region():
-        try:
-            return requests.get('http://169.254.169.254/latest/meta-data/placement/availability-zone').text.strip()[:-1]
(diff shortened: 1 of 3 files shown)


packages/dcos-integration-test/extra/test_ec2.py
```diff
@@ -145,7 +82,17 @@ def test_move_external_volume_to_new_agent(cluster):
         cluster.deploy_marathon_app(read_app, **deploy_kwargs)
         cluster.destroy_marathon_app(read_app['id'])
     finally:
-        try:
-            _delete_ec2_volume(test_label)
-        except Exception as ex:
-            raise Exception("Failed to clean up volume {}: {}".format(test_label, ex)) from ex
```
branden: Let's keep this exception handler so that there's a clear error message with the volume name in it if the test fails to clean up the volume.

packages/dcos-integration-test/extra/test_ec2.py
```diff
@@ -145,7 +82,17 @@ def test_move_external_volume_to_new_agent(cluster):
...
+        logging.info('Deleting '+test_label)
```
branden: This message might be unclear since there's nothing in `test_label` that indicates it's a volume name. `'Deleting volume ' + test_label` would be clearer.

packages/dcos-integration-test/extra/test_ec2.py
```diff
@@ -145,7 +82,17 @@ def test_move_external_volume_to_new_agent(cluster):
...
+        delete_cmd = """#!/bin/bash
+source /opt/mesosphere/environment.export
+python /opt/mesosphere/dcos-integration-test/delete_ec2_volume.py test_label
+"""
+        delete_job = {
+                'id': 'delete-volumes-'+test_uuid,
```
branden: `'delete-volume' + test_uuid` since it's deleting only one volume.

branden:
Left some comments above about clearer log output for easier failure debugging, but otherwise LGTM once checks pass.

packages/dcos-integration-test/extra/test_ec2.py
mellenburg: All this is doing is raising a generic exception and attaching test_label. Rather, I just log this to output before delete action runs and let the exception raise itself. It is functionally identical and a little easier to read IMO. Also, I remembered @jgehrcke specifically pointing out this line of code in a JIRA as lacking, so I figured I would make his day ;)

mellenburg:
@spahl @cmaloney This is ready for review and is a blocker for activating AWS integration tests on master

packages/dcos-integration-test/extra/test_ec2.py
branden: It's raising an exception with a clear explanation that might be understandable and actionable to someone who's not familiar with the test, which means potentially fewer questions for me about what happened, since in my experience people usually don't spend too much effort debugging a test if they don't own it. It's raised `from ex`, so it'll include the traceback of the exception that caused it. No detail is lost here.

spahl:
Nice this doubles as our first metronome test.

mellenburg:
@spahl can we ship this?
@branden I reverted that change to try/except

spahl:
LGTM, :ship:

PR #510, opened 2016-08-09
Pin requests to 2.10.0 for mrest compatibility

mellenburg:
Azure broke recently with an exception in internal API calls

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
         'docopt',
         'passlib',
         'pyyaml',
-        'requests',
+        'requests==2.10.0',
         'retrying',
         'keyring==9.1'],  # FIXME: pin keyring to prevent dbus dep
     entry_points={


cmaloney:
https://github.com/dcos/dcos/blob/master/packages/dcos-image-deps/buildinfo.json#L34 -- want to make sure to match that version with any pin.

mellenburg:
@cmaloney I found the real source of the issue requests 2.11 came out and it will now raise an exception if any header fields arent passed as string (2.10 simply accepts it and will cast as string without complaining) and mrest passes content-length as an int. mrest added a fix yesterday that will be in 0.4.1. Just a note for when we try to upgrade to 2.11 that stuff may break

cmaloney:
Can we make this 2.10.0 so it matches: https://github.com/dcos/dcos/blob/master/packages/python-requests/buildinfo.json#L5?

With that can ship

packages/python-requests/build
```diff
@@ -3,4 +3,5 @@ source /opt/mesosphere/environment.export
 export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.4/site-packages"
 mkdir -p "$LIB_INSTALL_DIR"
 
-pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/$PKG_NAME
+cd /pkg/src/$PKG_NAME
+PYTHONPATH=$PYTHONPATH:$LIB_INSTALL_DIR easy_install --prefix=$PKG_PATH .
```
cmaloney: Why do we need to do this? Why isn't the pypi one good enough?

This also isn't one of the standard python install scripts we have with this change. `easy_install` is also deprecated I believe. Things are supposed to always use pip + wheel.

setup.py
```diff
@@ -48,9 +48,10 @@ def get_advanced_templates():
         'docopt',
         'passlib',
         'pyyaml',
-        'requests',
+        'requests==2.10',
```
cmaloney: please just pin this to 2.10.0 directly and not random requests commit zip.

test_util/azure_test_driver.py
```diff
@@ -55,7 +55,7 @@ def run():
     # tenant_id = os.environ.get('AZURE_TENANT_ID')
     # client_id = os.environ.get('AZURE_CLIENT_ID')
     # client_secret = os.environ.get('AZURE_CLIENT_SECRET')
-    group_name = 'tesing' + ''.join(random.choice('01234567890abcdef') for n in range(10))
+    group_name = 'testing' + ''.join(random.choice('01234567890abcdef') for n in range(10))
```
cmaloney: this is an independent change, another pr is pretty simple / easy.

mellenburg:
@cmaloney I mistakenly took "so that it matches" to be 'the guaranteed same bits" as opposed to 'add .0'

cmaloney:
@mellenburg Sorry about that. In theory the "add the .0" will cause pip to install exactly that version using the same url, with a lot less work.

cmaloney:
🚢

PR #517, opened 2016-08-10
gen/template.py: add error message for unhandled switch/case value

jgehrcke:
You modified the config template and template.py throws a ValueError in your face? This adds an error message that hopefully helps you understand what went wrong.


cmaloney:
🚢

cmaloney:
three flakes first run of tests... (2 were CCM failures, 1 was a falky test...)

PR #518, opened 2016-08-10
installer: update getting a list of nodes logic

darkonie:
If installer started from CLI the nodes will always be derived from config.yaml.
For example, master nodes will be taken from `master_list` field in a config.

If installer started with --web flag a list of nodes will be derived from:
 preflight -> config.yaml
 install prereqs -> nodes that successfully executed preflight step
 deploy -> nodes that successfully executed preflight step
 postflight -> nodes that successfully executed deploy step

NOTE: if a master node fails, web installer fails as well. If an agent node
      fails, installer will exclude those nodes, but will continue.

related PR:
https://github.com/dcos/dcos/pull/426

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib.py
+++ b/ext/dcos-installer/dcos_installer/action_lib.py
@@ -43,27 +43,73 @@ class ExecuteException(Exception):
     """Raised when execution fails"""
 
 
-def nodes_count_by_type(config):
-    total_agents_count = len(config.get('agent_list', [])) + len(config.get('public_agent_list', []))
+def get_successful_nodes_from(step: str, state_json_dir: str) -> list:
+    assert step in ['preflight', 'deploy', 'postflight']
+    step += '.json'
+
+    step_path = os.path.join(state_json_dir, step)
+    step_json = _read_state_file(step_path)
+    if not step_json:
+        # TODO: Fix exception here
+        raise Exception("Could not get hosts from {}".format(step_path))
+
+    assert 'hosts' in step_json, "Field `hosts` not found, location {}".format(step_path)
+
+    nodes = []
+    for node_ip, node_params in step_json['hosts'].items():
+        assert 'host_status' in node_params, "Field `host_status` not found for node {}".format(node_ip)
+        assert 'tags' in node_params, 'Field `tags` not found for node {}'.format(node_ip)
+        if node_params['host_status'] == 'success':
+            nodes.append(Node(node_ip, tags=node_params['tags']))
+        else:
+            assert 'role' in node_params['tags'], 'tags must contain parameter role. Got {}'.format(node_params['tags'])
+            if node_params['tags']['role'] == 'master':
+                raise Exception("Master node {} has failed. Uable to continue".format(node_ip))
+            log.debug("node {} unhealthy {}".format(node_ip, node_params['host_status']))
+
+    assert nodes, "No healthy nodes found"
+    return nodes
+
+
+def nodes_count_by_type(targets):
+    assert isinstance(targets, list)
+    total_masters = 0
+    total_agents = 0
+    for target in targets:
+        assert isinstance(target, Node)
+        assert 'role' in target.tags
+        if target.tags['role'] == 'master':
+            total_masters += 1
+        else:
+            total_agents += 1
+
     return {
-        'total_masters': len(config['master_list']),
-        'total_agents': total_agents_count
+        'total_masters': total_masters,
+        'total_agents': total_agents
     }
 
 
 def get_full_nodes_list(config):
-    def add_nodes(nodes, tag):
-        return [Node(node, tag) for node in nodes]
-
-    node_role_map = {
-        'master_list': 'master',
-        'agent_list': 'agent',
-        'public_agent_list': 'public_agent'
+    role_params = {
+        'master': {
+            'tags': {'role': 'master', 'dcos_install_param': 'master'},
+            'hosts': config['master_list']
+        },
+        'agent': {
+            'tags': {'role': 'agent', 'dcos_install_param': 'slave'},
+            'hosts': config.get('agent_list', [])
+        },
+        'public_agent': {
+            'tags': {'role': 'public_agent', 'dcos_install_param': 'slave_public'},
+            'hosts': config.get('public_agent_list', [])
+        }
     }
+    ssh_port = config.get('ssh_port', 22)
+
     full_target_list = []
-    for config_field, role in node_role_map.items():
-        if config_field in config:
-            full_target_list += add_nodes(config[config_field], {'role': role})
+    for role, params in role_params.items():
+        full_target_list += [Node('{}:{}'.format(node, ssh_port), params['tags']) for node in params['hosts']]
+
     log.debug("full_target_list: {}".format(full_target_list))
     return full_target_list
 
@@ -98,8 +144,8 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     preflight_chain.add_copy(pf_script_path, REMOTE_TEMP_DIR, stage='Copying preflight script')
 
     preflight_chain.add_execute(
-        'sudo bash {} --preflight-only master'.format(
-            os.path.join(REMOTE_TEMP_DIR, os.path.basename(pf_script_path))).split(),
+        lambda node: 'sudo bash {} --preflight-only {}'.format(
+            os.path.join(REMOTE_TEMP_DIR, os.path.basename(pf_script_path)), node.tags['dcos_install_param']).split(),
         stage='Executing preflight check')
     chains.append(preflight_chain)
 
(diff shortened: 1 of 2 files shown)


darkonie:
@cmaloney ready for initial review

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -16,6 +16,35 @@ class ExecuteException(Exception):
...
+    nodes = []
+    for node_ip, node_params in step_json['hosts'].items():
+        assert 'host_status' in node_params, "Field `host_status` not found for node {}".format(node_ip)
+        if node_params['host_status'] == 'success':
+            assert 'tags' in node_params, 'Field `tags` not found for node {}'.format(node_ip)
+            nodes.append(Node(node_ip, tags=node_params['tags']))
+        else:
+            log.debug("node {} unhealthy {}".format(node_ip, node_params['host_status']))
+
+    assert nodes, "No healthy nodes found"
```
cmaloney: having no successful nodes going from one stage to the next seems like it shouldn't be an assertion error, but is rather a runtime error / we need to fail in the UI but the code isn't wrong / incorrect.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -71,8 +110,8 @@ def run_preflight(config, pf_script_path='/genconf/serve/dcos_install.sh', block
     preflight_chain.add_copy(pf_script_path, REMOTE_TEMP_DIR, stage='Copying preflight script')
 
     preflight_chain.add_execute(
-        'sudo bash {} --preflight-only master'.format(
-            os.path.join(REMOTE_TEMP_DIR, os.path.basename(pf_script_path))).split(),
+        lambda node: 'sudo bash {} --preflight-only {}'.format(
```
cmaloney: 👍

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -169,31 +208,17 @@ def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_del
...
     bootstrap_tarball = _get_bootstrap_tarball()
     log.debug("Local bootstrap found: %s", bootstrap_tarball)
 
     targets = []
     if hosts:
         targets = hosts
     else:
-        for role, params in role_params.items():
-            targets += [Node(node, params['tags']) for node in params['hosts']]
+        if state_json_dir is None:
```
cmaloney: Is this is effectively detecting whether we're running via the Web UI vs. the CLI I think? Am I reading that right?

cmaloney:
Can we add at least a unit test that nodes falling of in each stage works properly / passes through?

ext/dcos-installer/dcos_installer/action_lib/__init__.py
darkonie: yes, you are reading it right. check `state_json_dir` if the var is not None (this will happen if we use JSONDelegate in web mode) then search for nodes there. Otherwise get them from a config. This is definitely implicit, i will fix it.

darkonie:
@cmaloney basic unit test added. I am also going to manually test this change on VPC cluster.

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -16,27 +16,77 @@ class ExecuteException(Exception):
     """Raised when execution fails"""
 
 
-def nodes_count_by_type(config):
-    total_agents_count = len(config.get('agent_list', [])) + len(config.get('public_agent_list', []))
+def get_successful_nodes_from(step, state_json_dir):
+    assert isinstance(step, str)
```
cmaloney: lets use python type annotations for these rather than asserts

ext/dcos-installer/dcos_installer/action_lib/__init__.py
```diff
@@ -16,27 +16,77 @@ class ExecuteException(Exception):
...
+    assert 'hosts' in step_json, "Field `hosts` not found, location {}".format(step_path)
+
+    nodes = []
+    for node_ip, node_params in step_json['hosts'].items():
+        assert 'host_status' in node_params, "Field `host_status` not found for node {}".format(node_ip)
+        if node_params['host_status'] == 'success':
+            assert 'tags' in node_params, 'Field `tags` not found for node {}'.format(node_ip)
+            nodes.append(Node(node_ip, tags=node_params['tags']))
+        else:
+            log.debug("node {} unhealthy {}".format(node_ip, node_params['host_status']))
```
cmaloney: Should check of the node tags here, error if the node is a master

cmaloney:
This doesn't seem to be currently doing the "if a master fails don't continue / fail hard"

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -43,27 +43,73 @@ class ExecuteException(Exception):
...
+        'agent': {
+            'tags': {'role': 'agent', 'dcos_install_param': 'slave'},
+            'hosts': config.get('agent_list', [])
+        },
+        'public_agent': {
+            'tags': {'role': 'public_agent', 'dcos_install_param': 'slave_public'},
+            'hosts': config.get('public_agent_list', [])
+        }
     }
+    ssh_port = config.get('ssh_port', 22)
```
cmaloney: This looks new / wasn't there in the old code. Why are we adding it here / as part of this PR? (I kno wthere is https://dcosjira.atlassian.net/browse/DCOS-325 which this probably helps fix, but that isn't mentioned in this PR description)

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -43,27 +43,73 @@ class ExecuteException(Exception):
...
+
     full_target_list = []
-    for config_field, role in node_role_map.items():
-        if config_field in config:
-            full_target_list += add_nodes(config[config_field], {'role': role})
+    for role, params in role_params.items():
+        full_target_list += [Node('{}:{}'.format(node, ssh_port), params['tags']) for node in params['hosts']]
```
cmaloney: Seems a bit silly for us to squash these together than parse them apart then inside Node(). Could we make node take the ip and port separately?

This I think also changes what we allow here. Before, if the nodes happened to be of the form `host:port` things would work, now they don't (We always force the global ssh_port)

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -196,31 +242,20 @@ def install_dcos(config, block=False, state_json_dir=None, hosts=None, async_del
...
     if hosts:
         targets = hosts
     else:
-        for role, params in role_params.items():
-            targets += [Node(node, params['tags']) for node in params['hosts']]
+        # state_json_dir contains a path to installer states. If it is not empty, it means we have executed
+        # preflight state already and we want to take only the hosts which passed the preflight. Otherwise get a list
+        # of hosts from a config file (used in CLI where the stages do not have to be in order).
+        if state_json_dir is None:
+            targets = get_full_nodes_list(config)
```
cmaloney: If I have a config.yaml, do an install, then make / edit config.yaml and do another install, what will happen here? Will it pick up the state_json_dir from my first install and try resuming that or will it pick up my new config?

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -402,7 +441,12 @@ def _add_prereqs_script(chain):
 
 @asyncio.coroutine
 def install_prereqs(config, block=False, state_json_dir=None, async_delegate=None, options=None):
-    targets = get_full_nodes_list(config)
+    targets = []
```
cmaloney: This same set of lines keeps coming up, can we pull it out into a function?

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -43,27 +43,73 @@ class ExecuteException(Exception):
...
+    nodes = []
+    for node_ip, node_params in step_json['hosts'].items():
+        assert 'host_status' in node_params, "Field `host_status` not found for node {}".format(node_ip)
+        assert 'tags' in node_params, 'Field `tags` not found for node {}'.format(node_ip)
+        if node_params['host_status'] == 'success':
+            nodes.append(Node(node_ip, tags=node_params['tags']))
+        else:
+            assert 'role' in node_params['tags'], 'tags must contain parameter role. Got {}'.format(node_params['tags'])
+            if node_params['tags']['role'] == 'master':
+                raise Exception("Master node {} has failed. Uable to continue".format(node_ip))
```
cmaloney: Pretty sure the Web installer code won't be happy with this currently unfortunately.

adam-mesos:
Ping. Can we get a rebase, @darkonie ?

darkonie:
@adam-mesos i am pretty sure this needs to be closed

PR #520, opened 2016-08-10
python test clean up

kensipe:
1. removed python file references that no longer exist
2. changed comments which were references to older code which were refactored.

### DIFF
--- a/test_util/test_runner.py
+++ b/test_util/test_runner.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Module for running integration_test.py inside of a remote cluster
+"""Module for running integration_tests inside of a remote cluster
 Note: ssh_user must be able to use docker without sudo priveleges
 """
 import logging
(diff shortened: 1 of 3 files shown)


cmaloney:
cc: @mellenburg

mellenburg:
:+1:  LGTM thanks for picking up litter!

test_util/test_runner.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-"""Module for running integration_test.py inside of a remote cluster
+"""Module for running integration_tests inside of a remote cluster
```
branden: Was that supposed to be a space instead of an underscore?

mellenburg:
@orsenthil Yeah, that can also go. test_registry was totally removed when we started running the test_server via a docker volume mount

orsenthil:
@kensipe - opportunity in this PR. Please `git rm -r test_util/docker` and squash commits for the change.

cmaloney:
🚢 . I did the follow up commented here in #535

PR #521, opened 2016-08-10
Spartan and 3DT run on agents too. Update admin ports

surdy:
I'm sure there might be more ports/services in need to be updated here. I just edited the ones I knew for sure.

### DIFF
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -6,6 +6,8 @@ The following is a list of ports used by internal DC/OS services, and their corr
 
 ### TCP
 
+ - 53: dcos-spartan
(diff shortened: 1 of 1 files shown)


cmaloney:
@surdy Should actually be running on different ports than that on agents.

branden:
3DT listens on a unix domain socket on agents. https://github.com/dcos/adminrouter/blob/master/nginx.agent.conf#L8

surdy:
@cmaloney Ohh my bad then. I guess I updated it with based on what I saw 1.7 cluster (ofcourse outdated info)
I will close this PR out. Although this is not official doc, it might be useful to keep this updated.

PR #522, opened 2016-08-10
[WIP] Add shakedown as a package

mellenburg:
integration test should be moving towards this eventually


spahl:
This is not how I think this should be integrated. Let's not do this until we have a full story.

PR #524, opened 2016-08-11
Remove non-bootstrap packages from bootstrap tarball

branden:
This removes all packages from the bootstrap tarball except the minimum necessary to install the pkgpanda CLI. A variant's packages are now all listed in cluster_packages.json, causing pkgpanda to fetch and install them during node setup.

The larger cluster_packages.json causes the simple CloudFormation templates to grow past their 51,200 byte limit, so this PR moves template validation after artifact upload, and validates them by URL rather than inline. See https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html.


branden:
@spahl @cmaloney @lingmann PTAL

cmaloney:
We seem to be uploading all the packages many times now:
https://teamcity.mesosphere.io/viewLog.html?buildId=365269&buildTypeId=DcosIo_Dcos_BuildPr&tab=buildLog#_state=60,4861&focus=2339

branden:
@cmaloney I missed a spot where cluster_packages was repeatedly being added to artifacts. Fixed at https://github.com/dcos/dcos/pull/524/files#diff-70b329bb4c2984521f8861633d6b21bbR401.

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -74,15 +85,19 @@ def cleanup_and_exit():
...
     except subprocess.CalledProcessError as ex:
         log.error("Copy failed: %s\nOutput:\n%s", ex.cmd, ex.output)
         log.error("Removing partial artifacts")
         cleanup_and_exit()
     except KeyboardInterrupt:
         log.error("Copy out of installer interrupted. Removing partial files.")
         cleanup_and_exit()
+
+
+def fetch_packages(artifact_dir, dest_dir):
```
cmaloney: Why not just inline this? It's a single line, the arguments are always passed as constants.

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -6,34 +7,45 @@
...
-    # Get bootstrap from artifacts
-    fetch_bootstrap(gen_out.arguments['bootstrap_id'])
-    # Write some package metadata
-    pkgpanda.util.write_json('/genconf/cluster_packages.json', gen_out.cluster_packages)
+    # Generate install/upgrade prereqs.
+    subprocess.check_output(['mkdir', '-p', SERVE_DIR])
+    gen.installer.bash.generate(gen_out, SERVE_DIR)
+    fetch_bootstrap(ARTIFACT_DIR, SERVE_DIR, gen_out.arguments['bootstrap_id'])
+    fetch_packages(ARTIFACT_DIR, SERVE_DIR)
+    pkgpanda.util.write_json(os.path.join(GENCONF_DIR, 'cluster_packages.json'), gen_out.cluster_packages)
```
cmaloney: I very intentionally wasn't / don't use os.path.join. This is mainly because the behavior in some cases is never what I want:
```
os.path.join("foo", "/bar")
'/bar'
```

Rather I generally make hard assertions about how things must come in. Please switch back to not using os.path.join. Windows support for all this code is not anywhere even near the roadmap (and if it was probably would do a custom join).

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -44,20 +56,19 @@ def do_validate_gen_config(gen_config):
     return gen.validate(arguments=gen_config)
 
 
-def fetch_bootstrap(bootstrap_id):
+def fetch_bootstrap(artifact_dir, dest_dir, bootstrap_id):
```
cmaloney: That code is setup so that we "atomically" copy things out as best as possible, just in case the destination device becomes completely full or something similar people don't end up with partially extracted files (With the bootstrap tarball it did actually happen a couple of times).

We really should probably extend fetch_bootstrap to do all the cluster packages as well, they need the same "atomic" copy out guarantees.

gen/calc.py
```diff
@@ -287,14 +287,17 @@ def calculate_config_id(dcos_image_commit, user_arguments, template_filenames):
...
+        pkg_id_str = "{}--setup_{}".format(config_package_name, config_id)
         # validate the pkg_id_str generated is a valid PackageId
         return pkg_id_str
 
-    cluster_package_ids = list(sorted(map(get_package_id, json.loads(package_names))))
-    return json.dumps(cluster_package_ids)
+    return json.dumps(list(sorted(map(get_config_package_id, json.loads(config_package_names)))))
+
+
+def calculate_cluster_packages(config_package_ids, package_ids):
```
cmaloney: I don't think there's still value in doing the config package_ids stuff. Why can't we just do all the cluster packages as the single builtin?

cmaloney:
Need to teach `pkgpanda fetch` when run as part of the `pkgpanda setup` on a host (this bit of code: https://github.com/dcos/dcos/blob/master/pkgpanda/actions.py#L244) to retry on download failure now that we're going to be downloading more of the packages. Before the bootstrap was always the "one big thing" and that had a retry around it to help in failure scenarios. Now we have lots more smaller connections that could fail.

ext/dcos-installer/dcos_installer/action_lib/configure.py
```diff
@@ -6,34 +7,45 @@
...
+        log.error(message)
+        raise Exception(message) from exc
+    except ValueError as exc:
+        message = "Can't read build metadata."
+        log.error(message)
+        raise Exception(message) from exc
+
     arguments = {
         'bootstrap_id': os.environ['BOOTSTRAP_ID'],
+        'package_ids': json.dumps(package_ids),
```
cmaloney: this isn't your fault, but this isn't how we should be doing this / where we should be doing this.  Should be adding it to: https://github.com/dcos/dcos/blob/master/gen/calc.py#L407, so that it applies across providers, since in theory all of them want / need this full set of package ids.

gen/calc.py
branden: We need an easy way to access the config packages ids for rendering and generating additional artifacts to upload.

https://github.com/dcos/dcos/pull/524/files#diff-aab0f8dbae82b6ad54ff55e6f2195685R780
https://github.com/dcos/dcos/pull/524/files#diff-4dbd5bc0ea362b783a3e0273771660a9R234

I could've instead added a helper funciton that loops through the cluster packages and returns packages with versions starting with `setup_`, but that seemed less straightforward.

ext/dcos-installer/dcos_installer/action_lib/configure.py
lingmann: @cmaloney Thoughts on adding a code quality check which errors on any usage of `os.path.join`? There is a handful of places in the code currently doing this, but it seems like a recurring pattern that we could discourage early on with a build failure.

branden: I could add a custom `join()` that wraps `os.path.join()` and asserts that the first arg is an absolute path. That would prevent the behavior @cmaloney mentioned above and probably let us consolidate some path assertions. I started using `os.path.join()` in the first place because I accidentally omitted a slash somewhere and got fed up.

cmaloney:
As a note still only ~1/2 way through this PR review, will keep going later.

cmaloney:
This will very likely put us over the Azure byte limit total

cmaloney:
Bootstrap takes 36 seconds to make now (19:36:04 - 19:36:40)
Used to take 3minutes 18 seconds to make (15:15:34-15:18:52)

Inspecting one build

release/__init__.py
```diff
@@ -122,6 +122,16 @@ def load_providers():
             for name in provider_names}
 
 
+def validate_cloudformation(template_url):
```
cmaloney: This isn't something that the the "release" tool should know about (It really wants to interact with all the providers the same). We also need to be able to do cloudformation template validation of the advanced AWS templates which dcos_generate_config.sh is going to be outputting (#442)

Have two options to how to do this that I can see
1. Teach release to, after uploading artifacts, call a "validate" function for each artifact (better than nothing)
2. Try and get the templates back under the byte / complexity limit (my preference). There's probably a couple simple things we can move that will make these small enough (We really want as little as humanly possible in the cloud-config we put in the templates).

cmaloney: If at all possible I'd like to get this change out of the way / scope of the pulling apart bootstrap packages (otherwise it's just more stuff that all has to be ready for any of it to land)

cmaloney: The retry is also critical (We get rate limited)

gen/installer/util.py
```diff
@@ -43,9 +40,10 @@ def do_bundle_onprem(extra_files, gen_out, output_dir):
     for filename in extra_files:
         shutil.copy(filename, output_dir + filename)
 
-    # Copy the cluster packages
-    for name, info in gen_out.cluster_packages.items():
-        copy_makedirs(info['filename'], output_dir + info['filename'])
+    # Copy the config packages
+    for package_name in json.loads(gen_out.arguments['config_package_names']):
```
cmaloney: nit: should really make all of these atomic copy out operations like fetch_bootstrap is...

gen/test_validate.py
```diff
@@ -33,7 +35,8 @@ def test_error_during_validate():
     logger.setLevel(logging.DEBUG)
     assert gen.validate({
             'bootstrap_url': '',
-            'bootstrap_variant': ''
+            'bootstrap_variant': '',
+            'package_ids': json.dumps([])
```
cmaloney: at this point why not manually do a `"[]"`?

release/__init__.py
```diff
@@ -310,21 +320,33 @@ def add_package(package_id):
     # Add the <variant>.bootstrap.latest as a channel_path
     for name, info in sorted(all_completes.items(), key=lambda kv: pkgpanda.util.variant_str(kv[0])):
         bootstrap_filename = "{}.bootstrap.tar.xz".format(info['bootstrap'])
+        active_filename = "{}.active.json".format(info['bootstrap'])
+        active_local_path = 'packages/cache/bootstrap/' + active_filename
+        latest_filename = "{}bootstrap.latest".format(pkgpanda.util.variant_prefix(name))
```
cmaloney: nit: could pull the variant_prefix calculation out and just do it once, then stash the result.

adam-mesos:
This PR hasn't been touched in weeks, and needs a rebase. Are we still moving forward with this?

cmaloney:
@adam-mesos yes, but other things have come up as working currently. WIP PRs are allowed to sit for a bit.

cmaloney:
@branden pinging on this / making sure it's not lost in the backlog. If it's going to be a while still before we get back to it, may make sense to close and open again once progress is resuming.

branden:
👍

branden:
Re-closing while this is on the backlog.

cmaloney:
@branden can we pull a bunch of this code / commits out of this (Basically everything other than the making bootstrap tarball be smaller) and land it? Otherwise it's going to bit rot and need a lot more work to put back together.

PR #527, opened 2016-08-11
Installer CLI Argument to Upload Templates to S3 [WIP]

malnick:
## Depends
Relies on #442 

## Associated Tickets
Satisfies https://mesosphere.atlassian.net/browse/DCOS-9268

## Overview
This uses a config.yaml with a few new keys:
```
---
  cloudformation_s3_url: https://s3-us-west-2.amazonaws.com/jeremy-testing/dcos
  cloudformation_s3_bucket: jeremy
  cloudformation_s3_key: dcos
  cloudformation_s3_region: us-west-2
  zk_super_credentials: super:jeremysecret
  zk_master_credentials: dcos-master:jeremysecret1
  zk_agent_credentials: dcos-agent:jeremysecret2
  # disabled|permissive|strict                                                                                                                                                                                                      
  security: permissive
```

- `cloudformation_s3_bucket`
- `cloudformation_s3_key`
- `cloudformation_s3_region`

## TODO
- [ ] Add unit tests
- [ ] Add integration tests


malnick:
@lingmann This is ready for an initial pass by you.

ext/dcos-installer/dcos_installer/backend.py
```diff
@@ -55,6 +58,55 @@ def do_aws_cf_configure():
...
+        if e not in os.environ:
+            log.error("Must set {} to upload templates to S3".format(e))
+            return 1
+
+    cf_template_dir = '/genconf/serve/cloudformation'
+    if not os.path.exists(cf_template_dir):
+        log.error("genconf/serve/cloudformation does not exist. Try --aws-cloudformation.")
+        return 1
+
+    cf_files = glob.glob('{}/*'.format(cf_template_dir))
```
cmaloney: We know what files we want to upload exactly, lets upload those exactly. Guessing / globbing picks up random stuff in the directory.

Also note we need to upload things like all of the packages output in the entire serve directory.

cmaloney:
This needs to be reworked. The goal is that when you're generating cloudformation templates, as part of that I can say "target this path in this s3 bucket" and we'll do that. Don't want a separate independent upload path.

This needs to be rebased on master (right now it has a ton of random commits in it doing all sorts of other random things).

If just sharing between two people, just pushing to a second repo is enough that someone else can get / look at it.

malnick:
>> ` The goal is that when you're generating cloudformation templates, as part of that I can say "target this path in this s3 bucket" and we'll do that. `

I hear you saying that `--aws-cloudformation` utility should also do an upload to S3 in the same CLI argument? Is that correct?

>> `his needs to be rebased on master`

As pointed out in my PR description, this is based on #442, and requires it as a dependency. It's here for @lingmann to review, per our design meeting yesterday.

cmaloney:
Yes, in addition to `--aws-cloudformation`, when that option is given a person should be able to give the s3 bucket + path in that bucket where the templates should end up. `--s3-bucket-path=foo/bar/baz`.

That should implicitly cause things like the cloudformation s3 url to be set correctly.

malnick:
The only issue I see with that functionality is where someone may want to host their templates somewhere other than S3. In that case, making two separate action arguments makes sense. 

In line with your suggestion, we could upload only if `--s3-bucket-path` is set, however, I have those settings in the config.yaml, so we could just upload automatically if the s3 parameters are present.

malnick:
@lingmann Can you weigh in on this functionality?

cmaloney:
@malnick AWS CloudFormation does not allow launching templates which are stored in anything other than an s3 bucket.

malnick:
That's not true, you can have the templates locally and launch them with the AWS CLI.

cmaloney:
In which case you wouldn't be using the upload to s3 argument...

cmaloney:
Our advanced templates also cannot be launched from the AWS CLI like that, because they reference eachother

malnick:
Right, so back to my original suggestion: I have 3 new S3 parameters in the config.yaml to handle uploading to a specific bucket/key in a given region. If those are set, we automatically upload, if they are not, we generate and stop. This is similar to what you had with the CLI flag, and allows us to put complete action configuration in one place, though I could fall on either side of that configuration argument (flag vs. config).

cmaloney:
your code here currently adds an argument to upload them: https://github.com/dcos/dcos/pull/527/commits/536e857c1190fe15e79ca1dfa8f26582c60a86b1

That is what I was reviewing code wise.... If that's not at all what is going to be implemented ...

malnick:
Ok, so I realize now what your initial argument is: what and when to call the CLI flag that uploads to S3. 

I agree, we should use a boolean flag with the CF generation flag. I do not agree that we should configure how that upload flag operates using flags, since we require three different S3 parameters, we can simply keep the keys I've added for S3 uploads in the config.yaml.

Does that sound sane to you?

malnick:
Example: 

`--aws-cloudformation` -> generates templates 
`--aws-cloudformation --upload-s3` -> generates templates, uploads to S3 if S3 config is present.

cmaloney:
Putting the code for this into https://github.com/dcos/dcos/pull/442

PR #530, opened 2016-08-11
CLI flag to disable installer CLI telemetry

malnick:
Satisfies https://mesosphere.atlassian.net/browse/DCOS-9283

### DIFF
--- a/ext/dcos-installer/dcos_installer/cli.py
+++ b/ext/dcos-installer/dcos_installer/cli.py
@@ -49,8 +49,15 @@ def parse_args(args):
         return """
 Install Mesosophere's Data Center Operating System
 
-dcos_installer [-h] [-f LOG_FILE] [--hash-password HASH_PASSWORD] [-v]
-[--web | --genconf | --preflight | --deploy | --postflight | --uninstall | --validate-config | --version]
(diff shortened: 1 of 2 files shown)


malnick:
@spahl This is ready for review

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -50,7 +50,15 @@ def print_usage():
 Install Mesosophere's Data Center Operating System
 
 dcos_installer [-h] [-f LOG_FILE] [--hash-password HASH_PASSWORD] [-v]
```
cmaloney: `[--cli-telemetry-disabled]` should go here based on how the argument is added (it's not part of the mutually exclusive set).

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -50,7 +50,15 @@ def print_usage():
 Install Mesosophere's Data Center Operating System
```
cmaloney: Not for this PR, but why can't we just use the standard / default generated argparse help nowadays?

malnick: I think we can, we removed it initially to support overriding the PORT env from the docker wrapper (at least that's how I recall the decision to write this in manually, but don't quote me on it).

cmaloney:
🚢

PR #534, opened 2016-08-12
Fixes minuteman for named VIPs with 2 or more ports in use

sargun:
This change fixes minuteman for named VIPs with 2 or more ports in use


cmaloney:
🚢

PR #543, opened 2016-08-15
mesos-dns: set listeners to localhost

alberts:
@sargun @sargun @adam-mesos leftover from one of our internal branches... thoughts?

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -71,7 +71,8 @@ package:
         "port": 61053,
         {{ mesos_dns_resolvers_str }},
         "timeout": 5,
-        "listener": "0.0.0.0",
+        "listener": "127.0.0.1",
+        "HTTPListener": "127.0.0.1",
         "email": "root.mesos-dns.mesos",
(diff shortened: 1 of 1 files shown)


alberts:
We already have this change for our DC/OS distribution:

https://github.com/mesosphere/dcos-enterprise/blob/master/gen_extra/dcos-config.yaml#L69

spahl:
Even the normal listener should be 127.0.0.1. spartan is the dns front

cmaloney:
This is reliably failing tests, moving to work in progress.

gen/dcos-config.yaml
```diff
@@ -71,7 +71,8 @@ package:
         "port": 61053,
         {{ mesos_dns_resolvers_str }},
         "timeout": 5,
-        "listener": "0.0.0.0",
+        "listener": "127.0.0.1",
```
sargun: If you do this, how will the agents ever query Mesos-DNS?

sargun:
Spartan doesn't front Mesos-DNS requests from the agents. We can make it do that, but I think last time we had this discussion we came away against that for reliability reasons.

PR #544, opened 2016-08-15
Replace simplejson exception

mellenburg:
Currently, we raise the wrong error:
```
except json.JSONDecodeError:
    assert False, ...
AttributeError: 'module' object has no attribute 'JSONDecodeError'
```
This exception is from simplejson module. We use json

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -145,8 +145,8 @@ def validate_dcos_overlay_network(dcos_overlay_network):
     assert isinstance(dcos_overlay_network, str)
     try:
         overlay_network = json.loads(dcos_overlay_network)
-    except json.JSONDecodeError as ex:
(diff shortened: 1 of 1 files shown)


orsenthil:
LGTM.

gen/calc.py
```diff
@@ -145,8 +145,8 @@ def validate_dcos_overlay_network(dcos_overlay_network):
     assert isinstance(dcos_overlay_network, str)
     try:
         overlay_network = json.loads(dcos_overlay_network)
-    except json.JSONDecodeError as ex:
-        assert False, "Must be a valid JSON . Errors while parsing at position {}: {}".format(ex.pos, ex.msg)
+    except ValueError:
+        assert False, "Must be valid JSON"
```
cmaloney: Can we include the json here in the message? Not having that makes it really annoying to debug json decoding problem (esp. since by the time things get here, the json may be supplied by other things).

cmaloney:
🚢

PR #550, opened 2016-08-16
[WIP] Experimental GPU Support

wavesoft:
- Enabling GPU support on marathon
- Adding general purpose isolator detection to private/public slaves in mesos package (currently only detecting GPUs and adding `gpu/nvidia` isolator)

### DIFF
--- a/packages/mesos/build
+++ b/packages/mesos/build
@@ -64,6 +64,22 @@ disk_resource_script="$PKG_PATH/bin/make_disk_resources.py"
 cp /pkg/extra/make_disk_resources.py "$disk_resource_script"
 chmod +x "$disk_resource_script"
 
+# setup additional isolator detection service that is responsible for identifying additional isolators,
+# (such as GPU) and creates an EnvironmentFile that contains a MESOS_ISOLATION env variable, with
+# additional isolators APPENDED to the current environment.
+
+public_isolators_service="$PKG_PATH"/dcos.target.wants_slave_public/dcos-isolators-discovery-pub-agent.service
+mkdir -p "$(dirname ${public_isolators_service})"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-isolators-discovery-pub-agent.service > "$public_isolators_service"
+
+private_isolators_service="$PKG_PATH"/dcos.target.wants_slave/dcos-isolators-discovery-priv-agent.service
+mkdir -p "$(dirname ${private_isolators_service})"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-isolators-discovery-priv-agent.service > "$private_isolators_service"
+
+isolators_script="$PKG_PATH/bin/make_isolators.py"
+cp /pkg/extra/make_isolators.py "$isolators_script"
+chmod +x "$isolators_script"
+
 # Install slave modules.
 mkdir -p "$PKG_PATH/etc/mesos-slave-modules"
 cp /pkg/extra/slave_logrotate_module.json "$PKG_PATH/etc/mesos-slave-modules/"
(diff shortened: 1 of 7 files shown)


cmaloney:
@wavesoft any timeline on the work going to happen on making this landable as previously discussed?

wavesoft:
Hey @cmaloney ! I will have a look now.

packages/mesos/extra/make_isolators.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    nvidia_smi_binary = '/bin/nvidia-smi'
+
+    # Make sure file is there and with correct permissions
+    if not os.path.isfile(nvidia_smi_binary):
+        return False
+    if not os.access(nvidia_smi_binary, os.X_OK):
+        return False
+
+    # Try to run it
```
orsenthil: [readability review] - please kill all the comments in this file. In practice, I have seen comments that explain "why" (when needed) are helpful.  Comments that explain "what" should be avoided.

packages/mesos/build
```diff
@@ -64,6 +64,22 @@ disk_resource_script="$PKG_PATH/bin/make_disk_resources.py"
...
+# (such as GPU) and creates an EnvironmentFile that contains a MESOS_ISOLATION env variable, with
+# additional isolators APPENDED to the current environment.
+
+public_isolators_service="$PKG_PATH"/dcos.target.wants_slave_public/dcos-isolators-discovery-pub-agent.service
+mkdir -p "$(dirname ${public_isolators_service})"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-isolators-discovery-pub-agent.service > "$public_isolators_service"
+
+private_isolators_service="$PKG_PATH"/dcos.target.wants_slave/dcos-isolators-discovery-priv-agent.service
+mkdir -p "$(dirname ${private_isolators_service})"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-isolators-discovery-priv-agent.service > "$private_isolators_service"
```
wavesoft: Replying to @kaysoky 's comment in a previous PR:
> Do you really need a new service for this?
>
> Seems more appropriate to add a ExecStartPre=... to the two agent services.

Correct me if I am wrong, I don't have a linux installation to quickly test this, but aren't the `EnvironmentFile`s loaded before the `ExecStartPre` / `ExecStart` ? If true, the `ExecStartPre` won't be able to generate the environment variables on time...

packages/mesos/extra/make_isolators.py
wavesoft: Will do!

cmaloney:
To note some of the more core problems here
1. We're writing a giant program to directly reverse-engineer / undo a bunch of logic mesos is doing internally. Duplicating code to get the behavior we want effectively. Need to unwind it so mesos has a reasonable option we can use. That could be something like "when there are no gpus but gpu isolator is specified it gets skipped".
2. We don't want to run another separate discovery service. These make it incredibly hard to manage hosts and change resources / attributes / etc. Things have to go inline in the mesos slave service, and we definitely don't want to add more python scripts here.

General thing: If to land something in the product we have to write a bunch of code to undo engineering decisions made by engineers employed by mesosphere and make things work reasonably for our use case, there's a problem which needs to be fixed. The baseline components need to be as easy to use as possible. If we have to build this much extra tooling around mesos to use a feature, all our customers probably have to do it as well.

mgummelt:
@wavesoft I'd like to build this cluster to test Spark GPU support.  Do you have time to make it so the build passes so I can do that?

wavesoft:
@mgummelt will work on it later today!

wavesoft:
Hi all, I based this PR over `dcos/1.8.1` branch and the errors listed does not seem to come from my additions. Any suggestions? Shall I try to rebase on `master` ?

cmaloney:
We're adding short term experimental gpu support via #766 (all the cluster has GPUs or none of them)

PR #552, opened 2016-08-17
Do a rename to move our resolv.conf into place to wipe out symlinks

cmaloney:
Should resolve https://dcosjira.atlassian.net/browse/DCOS-305

Basically: Systemd by default makes `/etc/resolv.conf` a symlink to `/run/systemd/resolve/resolv.conf` which is owned by systemd-resolved / systemd-networkd. Just writing directly to that file, there ends up being some "contested" / racy ownership.

By writing to a temp file next to that, and moving into place we do two things
1. The symlink gets replaced by a file so there isn't a race anymore
2. Changing resolv.conf is "atomic" / can't have partial writes and the like show up in the file things use.

### DIFF
--- a/packages/spartan/extra/gen_resolvconf.py
+++ b/packages/spartan/extra/gen_resolvconf.py
@@ -76,8 +76,17 @@ else:
 
 # Generate the resolv.conf config
 print('Updating {}'.format(resolvconf_path))
-with open(resolvconf_path, 'w') as f:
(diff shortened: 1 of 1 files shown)


orsenthil:
Pass by review comment: `resolvconf_path + ".tmp"` at line 79 and line 90 could be extracted to a single constant.

PR #554, opened 2016-08-18
Move REX-Ray out of the agent port range

branden:
TODO:

- [x] Manually test REX-Ray CLI


branden:
REX-Ray doesn't support unix domain sockets for this option 😢 I'll just pick a port for it.

cmaloney:
There is a block over 65k that is outside of those we give to mesos and outside of the Linux ephemeral port range. If you look at the or to change agent port checking, should be able to pick it out

PR #555, opened 2016-08-22
Add wait for metronome in conftest

mellenburg:


mellenburg:

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -177,6 +177,15 @@ def _wait_for_srouter_slaves_endpoints(self):
             assert "id" in data
             assert data["id"] == slave_id
 
+    @retrying.retry(wait_fixed=2000,
+                    retry_on_result=lambda r: r is False,
+                    retry_on_exception=lambda _: False)
+    def _wait_for_metronome(self):
+        r = self.get('/service/metronome/v1/jobs')
+        if r.status_code == 500 or r.status_code == 504:
```
cmaloney: Can we get a comment as to why 500 and 504 we should retry on, but on all other error codes we shouldn't (I'm a fan of only waiting for the codes we know happen from adminrouter before it comes up, but this code doesn't give any explanation currently why these ones)

mellenburg: My intention was to make this more specific than what is used in the fixture elsewhere:
```
# resp_code >= 500 -> backend is still down probably
if r.status_code <= 500:
            logging.info("service is probably up")
            return True
        else:
            # retry
```
Would you prefer to use this less specific check and confine the metronome specificity to an actual metronome test?

cmaloney: I prefer the being specific like this. I'm just saying a comment / documentation in the code would be nice other than "We seem to be checking two random ports" when someone reads this code.

mellenburg: Got it. Done!

cmaloney:
:ship:

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -177,6 +177,18 @@ def _wait_for_srouter_slaves_endpoints(self):
             assert "id" in data
             assert data["id"] == slave_id
 
+    @retrying.retry(wait_fixed=2000,
+                    retry_on_result=lambda r: r is False,
```
orsenthil: [Readability comment] - It's better to have a description variable name here in `lambda r` - Does `r` denote `result`?

branden: The `retry_on_result` kwarg communicates to me that `r` is a result.

orsenthil:
@branden - realized when commenting. LGTM.

PR #560, opened 2016-08-22
Switch to running mesos-native containers for the most part, small docker when required.

cmaloney:
We're using the host python anyways. Should hopefully help avoid dockerhub pull issues a bit.

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -1,4 +1,5 @@
 import collections
+import copy
 import logging
 import os
 import uuid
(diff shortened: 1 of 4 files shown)


cmaloney:
note: Ideally we drop being inside a docker container for the vast majority of tests and just use the mesos containerizer, but the networking integration tests are too enterwined currently (Mesos containers are much faster to deploy).

sargun:
LGTM.

cmaloney:
For reference, this cut ~3 minutes off our integration test time (And should make the docker pull tests more reliable)

PR #551, opened 2016-08-16
3dt: add check for `ExecMainStatus`

darkonie:
make sure the main process returned 0 status code
https://github.com/dcos/3dt/pull/57
DCOS-9323

PR #561, opened 2016-08-22
Split dcos-adminrouter-reload.service between masters and agents

cmaloney:
Previously it was failing every run because the `nginx reload` pointed to a
non-existent configuration file. This makes it point to the master configuration
on masters, and agent configuration on agents.

Should unblock #551

cc: @branden @jgehrcke @darkonie


cmaloney:
failed integration tests are either flaky tests or AWS having issues currently :/

cmaloney:
:ship: . If you click through the status checks they're all green.

jgehrcke:
:+1:

PR #566, opened 2016-08-23
Train 55

cmaloney:
To land needs
 - [x] Confirmation rexray-cli works on #554 
 - [x] SSH into an agent + public agent and verify the adminrouter reload is working now.
 - [x] Ack from @rcorral that the Version API in the installer is good enough


cmaloney:
Added #568

PR #567, opened 2016-08-23
Up the systemd-journald burst rate

cmaloney:
Changing this setting globally, since most customers don't seem to touch it, and we have some issues currently in customer logs with things getting dropped fairly frequently.

cc: @bernadinm


cmaloney:
The 3dt failure is very consistent... Needs more investigation

cmaloney:
#647 should hopefully unblock this.

PR #569, opened 2016-08-23
Tweak flake8

cmaloney:
- Move flake8 excludes to the config line rather than command line arg (Easier to read / deal with)
 - Add pep8-naming check

On my roadmap is enabling McCabe complexity checking, but that is more involved.

cc: @orsenthil @mellenburg 

The biggest thing this changes is a bunch of names in tests.


orsenthil:
LGTM.

PR #570, opened 2016-08-23
Enable Marathon health checks for overlay network integration tests

gkleiman:


sargun:
It looks like this is the right change, and tests are passing. :+1:

sargun:
(Please merge)

gkleiman:
cc/ @spahl @cmaloney

PR #571, opened 2016-08-23
Make metronome, marathon run as non-root

cmaloney:
Fixes DCOS-317

TODO
 - [ ] Manually verify on a cluster metronome and marathon are indeed non-root.


adam-mesos:
LGTM!

PR #572, opened 2016-08-23
Greatly increase retention of logs.

kaysoky:
Changes:
* [x] Move logrotate service configs into `dcos-config.yaml`.
* [x] Fix master logrotate config to use the master's config (not the agent's)
* [x] Use `MESOS_EXTERNAL_LOG_FILE` instead of `MESOS_LOG_DIR` to bypass GLOG's own logrotate.
* [x] Add parameter for tweaking master/agent log retention.
* [x] Add an environment variable to the `LogrotateContainerLogger` which will give frameworks a knob to tweak task log retention
  * Mesos review: https://reviews.apache.org/r/51565/
  * Cherry-picked the commit onto Mesos 1.0.
* [x] Remove wildcards from logrotate service.
* [x] Update logrotate service's `postrotate` to strictly limit the number of files.

Testing:
* [x] Verify that the `tee_stderr` wrapper doesn't muck with systemd process tracking.
* [x] Verify that logrotation still works (it's quite easy to spam logs).
* [x] Verify that task log rotation can be modified to do custom things (there is a unit test in Mesos).
* [x] Verify that the updated `postrotate` deletes the oldest files, when there are too many.


packages/mesos/extra/dcos-mesos-slave-public.service
```diff
@@ -21,4 +21,4 @@ ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-slave-public
 ExecStartPre=/opt/mesosphere/bin/make_disk_resources.py /var/lib/dcos/mesos-resources
 ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done'
-ExecStart=$PKG_PATH/bin/mesos-slave
+ExecStart=/bin/bash -c '$PKG_PATH/bin/mesos-slave 2>&1 | tee -a /var/log/mesos/mesos-agent.log'
```
cmaloney: using bin bash like this isn't allowed in our sytemd units. It makes systemd have a harder time tracking things,  and is overly complicated for an exec start line. See: https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md

cmaloney:
We have to have a hard filesize cap on logs. 2 weeks of mesos logs will fill up our AWS agents (And we've had customer issues with that before). Those have only a couple GB of root disk.

packages/logrotate/build
```diff
@@ -21,9 +21,9 @@ nomail
 
 /var/lib/dcos/mesos/log/* {
     olddir /var/lib/dcos/mesos/log/archive
-    maxsize 2000k
-    daily
-    rotate 7
+    maxsize 256M
+    hourly
+    maxage 14
```
vinodkone: what is the behavior if both maxage and maxsize are specified?

kaysoky: `maxage` determines when files should be deleted.  `hourly` and `maxsize` (and `size`) determine when files should be rotated.

packages/logrotate/build
```diff
@@ -21,9 +21,9 @@ nomail
 
 /var/lib/dcos/mesos/log/* {
     olddir /var/lib/dcos/mesos/log/archive
-    maxsize 2000k
-    daily
-    rotate 7
+    maxsize 256M
```
vinodkone: do we still need maxsize?

cmaloney: We must cap size. We run on many hosts with very small root disks (AWS instances), and have customers run out of disk. It's why we added the logrotate in the first place.

kaysoky: Keep in mind that `maxsize` is the size of a single file, not the whole directory.

How large is the total log size allowed be be?  10 GB?
As long as we get rid of the GLOG logrotate, we can keep the `rotate N` option here and have no disk leakage.

cmaloney: Our current AWS instances we launch in our CloudFormation template:
```
Filesystem      Size  Used Avail Use% Mounted on
devtmpfs        7.4G     0  7.4G   0% /dev
tmpfs           7.4G     0  7.4G   0% /dev/shm
tmpfs           7.4G  724K  7.4G   1% /run
tmpfs           7.4G     0  7.4G   0% /sys/fs/cgroup
/dev/xvda9      5.5G  2.0G  3.2G  39% /
/dev/xvda3      985M  651M  283M  70% /usr
/dev/xvda1      128M   38M   90M  30% /boot
tmpfs           7.4G  288K  7.4G   1% /tmp
tmpfs           7.4G     0  7.4G   0% /media
/dev/xvda6      108M   64K   99M   1% /usr/share/oem
/dev/xvdb        37G  105M   35G   1% /var/lib
tmpfs           1.5G     0  1.5G   0% /run/user/500
```

cmaloney: Mesos, Marathon, Zookeeper among other things all share that `/var/lib. The max size can do for this is probably 1-2GB (2GB being about the max). Note 2GB of logs is a lot of lines of text compressed.

vinodkone: I wish logrotate had a "rotate <size>" option in addition to the "rotate <count>" option.

packages/mesos/extra/slave_logrotate_module.json
```diff
@@ -12,19 +12,19 @@
             },
             {
               "key": "max_stdout_size",
-              "value": "2MB"
+              "value": "10MB"
             },
             {
               "key": "logrotate_stdout_options",
-              "value": "rotate 9"
+              "value": "maxage 14"
```
kaysoky: The other question is task logs.  For particularly busy services, 20 MB (the existing setting) is usually not enough.  At the same time, we aren't enforcing disk quota in Mesos, so we won't out-of-disk-kill tasks.

Technically, this change has potential for unbounded disk usage too.

packages/logrotate/build
cmaloney: One of the advantages of the systemd-journald approach is they actually will cap at a percentage of the underlying disk backing the journal, so that this sort of thing is automatically taken care of / they guarantee that the disk won't fill up because of something the jounral wrote.

gen/calc.py
```diff
@@ -360,6 +373,8 @@ def validate_os_type(os_type):
         'master_dns_bindall': 'true',
         'mesos_dns_ip_sources': '["host", "netinfo"]',
         'mesos_container_logger': __logrotate_slave_module_name,
+        'mesos_log_retention_mb': '10240',
```
cmaloney: 10000

gen/dcos-config.yaml
```diff
@@ -211,9 +211,57 @@ package:
...
+      notifempty
+      missingok
+      nomail
+      /var/lib/dcos/mesos/log/* {
+          olddir /var/lib/dcos/mesos/log/archive
+          maxsize 256M
+          rotate {{ mesos_log_retention_count }}
+          copytruncate
+          postrotate
+              find /var/lib/dcos/mesos/log/ /var/lib/dcos/mesos/log/archive -mtime +14 -delete
```
cmaloney: It's quite possible the mesos_log_rotation count would be more than 14, in which case this needs to do whatever that is plus a couple more

packages/mesos/extra/dcos-mesos-master.service
```diff
@@ -15,4 +15,4 @@ EnvironmentFile=-/run/dcos/etc/mesos-master
 ExecStartPre=/bin/ping -c1 ready.spartan
 ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done'
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-master
-ExecStart=${PKG_PATH}/bin/mesos-master
+ExecStart=$PKG_PATH/bin/tee_stderr $PKG_PATH/bin/mesos-master /var/lib/dcos/mesos/log/mesos-master.log
```
cmaloney: Why do we need both this and the MESOS_EXTERNAL_LOG_FILE? They seem to be telling things to use the same file.

kaysoky: `MESOS_EXTERNAL_LOG_FILE` modifies the link that shows up in the (somewhat obscure) `LOG` link in the Mesos web UI.  When `MESOS_LOG_DIR` is set, the link refers to the glog file.

cmaloney:
re-marked WIP / RFC until the TODO list is done

gen/calc.py
```diff
@@ -77,6 +77,18 @@ def calculate_mesos_dns_resolvers_str(resolvers):
         return '"externalOn": false'
 
 
+def validate_mesos_log_retention_mb(mesos_log_retention_mb):
+    assert int(mesos_log_retention_mb) >= 512, "Must retain at least 512 MB of logs"
```
cmaloney: I'd suggest this actually be set at 1GB / 1024MB. As long as these are considered "mission critical", we want more than 512 MB probably.

packages/logrotate/build
```diff
@@ -10,84 +10,22 @@ make
...
 
-config_file="$PKG_PATH/etc/logrotate_master.conf"
-mkdir -p $(dirname $config_file)
-cat <<'EOF' > "$config_file"
-compress
-delaycompress
-notifempty
-missingok
-nomail
+export PKG_PATH=$PKG_PATH
```
cmaloney: This has never needed to be re-exported before. Why here?

gen/dcos-config.yaml
```diff
@@ -211,9 +211,57 @@ package:
...
+          endscript
+      }
+  - path: /etc_slave/logrotate_agent.config
```
cmaloney: These are identical, why not just write one to `/etc/logrotate_agent.config` and then just make the unit point to that one?

packages/mesos/extra/dcos-mesos-slave-public.service
```diff
@@ -21,4 +21,4 @@ ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-slave-public
 ExecStartPre=/opt/mesosphere/bin/make_disk_resources.py /var/lib/dcos/mesos-resources
 ExecStartPre=/bin/bash -c 'for i in /proc/sys/net/ipv4/conf/*/rp_filter; do echo 2 > $i; echo -n "$i: "; cat $i; done'
-ExecStart=$PKG_PATH/bin/mesos-agent
+ExecStart=$PKG_PATH/bin/tee_stderr $PKG_PATH/bin/mesos-agent /var/log/mesos/mesos-agent.log
```
cmaloney: This may muck with systemd's process tracking, we'll need to manually verify (if tee_stderr keeps running but mesos-agent does not, what happens?)

gen/dcos-config.yaml
kaysoky: This is supposed to delete based on when the files are last modified (anything older than 2 weeks is deleted).  We could keep it or remove it.

I can see how this might have been useful when the mesos-fetcher created a billion empty files in the agent's log directory.  But now that we're keeping a single file, outside of GLOG's control, I don't think its strictly necessary anymore.

cmaloney: This needs to stay. When you have a lot of logs in a directory, logrotate will segfault which happened at Verizon. This prevents that log pileup

kaysoky:
Launched a little cluster to do some ad-hoc validation.

This marathon app results in 10 MB stdout files which are compressed after rotation:
(The default is 2 MB, not compressed.)
```
{
  "id": "/logging-custom-task",
  "cmd": "while true; do s=$(printf \"%1024s\" \"...\"); echo \"${s// /z}\"; done",
  "env": {
    "MESOS_LOGROTATE_LOGGER_MAX_STDOUT_SIZE": "10mb",
    "MESOS_LOGROTATE_LOGGER_LOGROTATE_STDOUT_OPTIONS": "rotate 9\ncompress"
  },
  "instances": 1,
  "cpus": 0.1,
  "mem": 64,
  "disk": 64,
  "gpus": 0
}
```

kaysoky:
With regards to process tracking:

The wrapper now spawns three processes:
1) /bin/bash
2) mesos-master/mesos-agent
3) tee

When (1) or (2) are killed, systemd will restart the service.  If (3) is killed, nothing is restarted.

kaysoky:
Sanity checked that logrotate still works by running this a couple of times (to grow the log file):
```
cat mesos-master.log >> temp.log && cat temp.log >> mesos-master.log
```
The two-minute logrotate timer did what I expected (rotated the files after 256MB and put the old files into the `archive` folder).

packages/mesos/buildinfo.json
```diff
@@ -3,8 +3,8 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/mesosphere/mesos",
-    "ref": "f6581f2e8fcb07074fca9e72b3134476f6fc74ab",
-    "ref_origin" : "dcos-mesos-1.0.1"
+    "ref": "b546e7544c04a6d3e0da87cd6207ae8f4d827add",
+    "ref_origin" : "dcos-mesos-1.0.1-logrotate"
```
adam-mesos: Let's cherry-pick these onto `dcos-mesos-1.0.1` so we have a clear linear progression of patches added onto one branch off an official Apache Mesos release (1.0.1), rather than potentially losing patches between the many "feature" branches. Totally appropriate for testing, but I'm convinced that it's time to cherry-pick these onto the "official" branch now.

packages/logrotate/build
```diff
@@ -10,84 +10,20 @@ make
...
-    maxsize 2000k
-    daily
-    rotate 7
-    copytruncate
-    postrotate
-        find  /var/lib/dcos/mesos/log/ /var/lib/dcos/mesos/log/archive -mtime +14 -delete
-    endscript
-}
-EOF
+systemd_slave="$PKG_PATH"/dcos.target.wants_slave/dcos-logrotate-agent.service
```
adam-mesos: Might as well name this variable `systemd_agent` since the service is called `dcos-logrotate-agent`, and it's only relevant in this one file.

packages/logrotate/extra/dcos-logrotate-agent.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Logrotate Mesos Master: Rotate Mesos agent logs
```
adam-mesos: s/Master/Agent/

packages/logrotate/extra/dcos-logrotate-agent.service
```diff
@@ -0,0 +1,7 @@
...
+[Service]
+Type=simple
+EnvironmentFile=/opt/mesosphere/environment
+ExecStartPre=/usr/bin/mkdir -p /var/log/mesos/archive
```
adam-mesos: Should this mkdir `/var/lib/dcos/mesos/log/archive` like the Master version does? Why the difference?

packages/mesos/README.md
```diff
@@ -9,3 +9,4 @@
 <li>[e444edd6e7d45eb842fe223ac484a4bc4c1a327a] Allowed all flags load methods to specify a prefix. (landing in 1.1.0)
 <li>[3dd498d797a44593d376b4d50a33e2b8214c0dd4] Split src/openssl.hpp adding include/process/ssl/flags.hpp. (landing in 1.1.0)
 <li>[f6581f2e8fcb07074fca9e72b3134476f6fc74ab] Updated includes to follow the SSL flag split. (landing in 1.1.0)
+<li>[b546e7544c04a6d3e0da87cd6207ae8f4d827add] Added a way to set logrotate settings per executor. (landing in 1.1.0)
```
adam-mesos: #676 was just merged. Please rebase and add the SHAs for all 3 SHAs in your branch.

> b546e7544c04a6d3e0da87cd6207ae8f4d827add Added a way to set logrotate settings per executor.
1d45ba9f4f7f943d24282e0a91fb01c68815d6c3 Fixed sign comparisons in logrotate module.
3e703af02bc309a9d6653392d02627bf3d3ce930 Updated logrotation module to use `os::pagesize()`.

packages/logrotate/extra/dcos-logrotate-agent.service
kaysoky: The agent's log directory has historically been different from the master's log directory.  I'm not sure what the history behind that is.

The helper scripts some of us use to grab logs appear to have this agent directory hard-coded.  So better not to change it.

gen/calc.py
```diff
@@ -377,6 +390,8 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
         'master_dns_bindall': 'true',
         'mesos_dns_ip_sources': '["host", "netinfo"]',
         'mesos_container_logger': __logrotate_slave_module_name,
+        'mesos_log_retention_mb': '10000',
```
cmaloney: This is too big as a default

gen/dcos-config.yaml
cmaloney: This +{number} still needs to be updated

kaysoky: I don't think this needs an update.

We introduced the `postrotate` because the `mesos-fetcher` was inheriting the agent's GLOG flags and thereby creating a bunch of empty files in the log directory every time it was called.  Over time, this cluttered the log directory (and made logrotate segfault).

With the changes to logrotation, DC/OS will only ever create one log file (`mesos-master.log` or `mesos-agent.log`) and up to `mesos_log_retention_count` files in the `archive` folder.  The `mesos-fetcher` files will be no more.

After these commits, the `postrotate` will only be there to clean up unknown files, older than two weeks, in the log directory.  These files will have no relation to the Mesos logs.

cmaloney: This has nothing to do with mesos-fetcher, it has everything to do with not having a segfaulting logrotate. 

The general plan was (num log files) + 7, so that you could have some temporary user files in the log directory. Right now if I configure for lots and lots of logs, this will cap you at 14 files max or delete all the things.

kaysoky: With the existing `postrotate` you can create an infinite number of files in the log directory and they won't be deleted until 2 weeks later.  The `-mtime +14` is not a file count.

`find ... -mtime +14` selects all files with a modification of at least 15 days ago (according the man page):

> -mtime n
              File's data was last modified n*24 hours ago.  See the
              comments for -atime to understand how rounding affects the
              interpretation of file modification times.

And `-atime` says:
> -atime n
              File was last accessed n*24 hours ago.  When find figures out
              how many 24-hour periods ago the file was last accessed, any
              fractional part is ignored, so to match -atime +1, a file has
              to have been accessed at least two days ago.

cmaloney:
We should meet in person to hash out the find / delete stuff. Still it conflicts with the other setting / "number of log" based config. We don't actually care about "deleting things older than 2 weeks", care about deleting excessive random files.

gen/dcos-config.yaml
```diff
@@ -211,9 +211,41 @@ package:
...
 {% case "false" %}
 {% endswitch %}
+  - path: /etc_master/logrotate_master.config
+    content: |
+      compress
+      delaycompress
+      notifempty
+      missingok
+      nomail
+      /var/lib/dcos/mesos/log/* {
```
kaysoky: From our in-person sync: Change this to point to the leading log file.  It doesn't make sense to (potentially) rotate random files in the directory anymore.

alberts:
@kaysoky is the whole TODO list done? :)

kaysoky:
I still need to verify the last two commits (the script for removing random files in the log folder) on a test cluster.

alberts:
@kaysoky thanks!

kaysoky:
Verified on a test cluster than the `postrotate` script deletes things.  Note that logrotate only calls `postrotate` scripts if something is actually rotated.  If there are tons of temporary files, but the leading log is still under the rotation threshold, the temporary files will remain untouched.

```
# Do this a couple of times, with different random names.
touch some_random_file

# Do this a couple of times with different suffixes to generate files.
find . -type f | xargs -n1 -I % sudo cp % %.1
```

packages/logrotate/extra/delete-oldest-unmanaged-files.py
```diff
@@ -0,0 +1,52 @@
...
+    '''
+    logging.basicConfig(format='%(levelname)-4s ] %(message)s',
+                        level=logging.INFO)
+
+    # For simplicity, convert all paths to absolute paths.
+    directory = os.path.abspath(directory)
+    managed_file = os.path.abspath(managed_file)
+
+    # Build a list of all files.
+    # TODO(josephw): Do we want to delete directories too?
```
alberts: Do or no do, there is no TODO :)

packages/mesos/build
```diff
@@ -56,6 +56,18 @@ disk_resource_script="$PKG_PATH/bin/make_disk_resources.py"
 cp /pkg/extra/make_disk_resources.py "$disk_resource_script"
 chmod +x "$disk_resource_script"
 
+# Copy the log-redirection wrapper script for the master/agent.
```
alberts: guessing here, but can exec -a help us?

http://stackoverflow.com/questions/31747247/change-real-process-name-in-c-on-linux

packages/mesos/extra/tee_stderr
```diff
@@ -0,0 +1,11 @@
...
+# This file takes two arguments:
+#  <program>: The command to run.
+#  <output file>: Where to `tee` the stderr of the <program>.
+
+if [ $# -ne 2 ]; then
+    echo "usage: $0 <program> <output file>"
+    exit 1
+fi
+
+$1 2>&1 | tee -a "$2"
```
alberts: my hope is that a well-placed exec or exec -a here can help us avoid having a wrappers subdirectory under bin/

alberts:
Maybe instead of the exec tricks, we should be setting SyslogIdentifier= instead?

packages/logrotate/extra/delete-oldest-unmanaged-files.py
kaysoky: The previous `postrotate` script would include directories by virtue of not filtering them out (i.e. find . -type f`).  Perhaps there is a degenerate case where logrotate segfaults due to having millions of directories under the log directory...

adam-mesos:
@kaysoky Is this ready for another round of review, now that you've updated for SyslogIdentifier?

cmaloney:
@adam-mesos chatting with him in slack earlier it id

packages/logrotate/extra/delete-oldest-unmanaged-files.py
```diff
@@ -0,0 +1,52 @@
...
+    Finds all files under the given `directory` and checks if the number
+    of files exceeds the `max_files`.  If so, deletes files starting from
+    the oldest (except for the `managed_file`) until the `max_files` is
+    no longer exceeded.
+
+    @type directory: str, absolute path of the directory to clean up
+    @type max_files: int, maximum number of files to keep
+    @type managed_file: str, one file (i.e. leading log) to excempt from cleanup
+    '''
+    logging.basicConfig(format='%(levelname)-4s ] %(message)s',
```
cmaloney: Missing `[` I think

packages/logrotate/extra/delete-oldest-unmanaged-files.py
```diff
@@ -0,0 +1,52 @@
...
+
+    logging.info("Found {0} files in log directory (max {1})".format(len(all_files), max_files))
+    if len(all_files) <= max_files:
+        return
+
+    oldest_first = sorted(all_files, key=lambda x: os.stat(x).st_mtime)
+    to_delete = oldest_first[0:len(all_files) - max_files]
+
+    for path in to_delete:
+        logging.info("Deleting unmanaged file inside log directory: {0}".format(path))
```
cmaloney: nit: "Deleting old file inside log directory: {0}".format(path)

Not necessarily not an unmanaged file. If a user makes a lot of temporary files it could be a managed / logrotated file.

packages/logrotate/extra/delete-oldest-unmanaged-files.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Exempt the one managed file.  This is presumably the leading log file.
+    all_files.remove(managed_file)
+
+    logging.info("Found {0} files in log directory (max {1})".format(len(all_files), max_files))
+    if len(all_files) <= max_files:
+        return
+
+    oldest_first = sorted(all_files, key=lambda x: os.stat(x).st_mtime)
+    to_delete = oldest_first[0:len(all_files) - max_files]
```
cmaloney: Could we log what the time cutoff here ends up being?

Would also be good to use `st_mtime_ns` rather than `st_mtime`

packages/mesos/buildinfo.json
```diff
@@ -3,7 +3,7 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/mesosphere/mesos",
-    "ref": "a22b5dd552168a8dd925682f3b349059e10fc3d3",
+    "ref": "b546e7544c04a6d3e0da87cd6207ae8f4d827add",
```
cmaloney: we'll need to re-evaluate this vs. other things landing when finally merging this PR (the mesos we include has gained more patches since this was opened)

adam-mesos: I've been watching the Mesos SHA closely. This SHA is a proper descendant of what's in master of dcos/dcos, since we only linearly progress the `dcos-mesos-1.0.1` branch, which is official Apache Mesos `1.0.1` tag plus a series of patches. So this part of the PR would be acceptable to merge now.
However, if #768 lands first, then it supplants (and includes) these changes to the dcos-mesos branch, since it includes all of the cherry-picks from `dcos-mesos-1.0.1` at Joseph's new sha (b546e75), but rebased on top of the Apache Mesos `1.0.x` branch, which includes `1.0.1` exactly, plus cherry-picks from master that are destined to be released in `1.0.2` whenever that is cut.

gen/calc.py
```diff
@@ -385,6 +405,9 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
         'master_dns_bindall': 'true',
         'mesos_dns_ip_sources': '["host", "netinfo"]',
         'mesos_container_logger': __logrotate_slave_module_name,
+        'mesos_log_retention_mb': '4000',
+        'mesos_log_retention_count': calculate_mesos_log_retention_count,
```
kaysoky: All review comments have been addressed!

cmaloney:
:ship: . Need an issue for longer term "the mesos logging really doesn't do what we want / need" because there is a lot of very intricate infrastructure here for what is a common / shared problem.

PR #573, opened 2016-08-23
Updated mesos-overlay-modules with fix for configuration failure.

asridharan:
* The mesos-overlay-module on the agent will try to re-register with the master in case of a configuration failure, forcing it to reattempt the configuration of the overlay network. On failing more than four times (default value) it will bail.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -123,6 +123,13 @@ def validate_dcos_overlay_mtu(dcos_overlay_mtu):
     assert int(dcos_overlay_mtu) >= 552, 'Linux allows a minimum MTU of 552 bytes'
 
 
+def validate_dcos_overlay_config_attempts(dcos_overlay_config_attempts):
+    assert dcos_overlay_config_attempts.isdigit(), (
+        'dcos_overlay_config_attempts needs to be a positive integer between 0 and 10')
+    assert int(dcos_overlay_config_attempts) >= 0 and int(dcos_overlay_config_attempts) < 10, (
+        'The acceptable range of values for dcos_overlay_config_attempts is between 0 and 10')
+
+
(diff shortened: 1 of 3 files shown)


asridharan:
@sargun could you review the `docs-config.yaml` changes?

Mesas-Overlay-module PR:
https://github.com/dcos/mesos-overlay-modules/pull/7
Has been reviewied by @jieyu

gen/calc.py
```diff
@@ -371,6 +371,7 @@ def validate_os_type(os_type):
         'ui_banner_footer_content': 'null',
         'ui_banner_image_path': 'null',
         'ui_banner_dismissible': 'null',
+        'dcos_overlay_config_attempts': '4',
```
sargun: Can we add validation here?

sargun:
Looks good to me. I would add validation and ensure that the number is something reasonable, like 0-5?

gen/calc.py
asridharan: Well, we don't have a minimum here. So if the user sets it to 0, there is nothing wrong, the first configuration error would cause a failure, which should be acceptable.

I will however add a validation to check it is a valid integer. Does that make sense?

cmaloney: Validation that the provided string by the user is an integer and between 0 and 1000 I think would be good.

cmaloney:
:ship:

gen/calc.py
asridharan: I will put an integer check and range check of 0 to 10. We have exponential back off in place when the configuration fails, so if go beyond 10 tries it pretty much de-generates the fix into the original behavior which no hard stop on failed configuration.

PR #577, opened 2016-08-24
Update dcos_generate_config.sh --help, simplify code a bit

cmaloney:
argparse's default `--help` isn't much better / worse than the custom one, remove the custom one.

Also:
 - Fix adding the help strings for each argument
 - Fold simple functions only called by one place to their call sites
 - Import dcos_installer.async_server inside do_web since it always sets up a bunch of webserver stuff.


cmaloney:
For reference:
```
cos_installer --help
usage: dcos_installer [-h] [--hash-password HASH_PASSWORD] [-v] [--offline]
                      [--cli-telemetry-disabled] [--web] [--version]
                      [--genconf] [--validate-config] [--postflight]
                      [--install-prereqs] [--deploy] [--preflight]
                      [--uninstall]

optional arguments:
  -h, --help            show this help message and exit
  --hash-password HASH_PASSWORD
                        Hash a password on the CLI for use in the config.yaml.
  -v, --verbose         Verbose log output (DEBUG).
  --offline             Do not install preflight prerequisites on CentOS7,
                        RHEL7 in web mode
  --cli-telemetry-disabled
                        Disable the CLI telemetry gathering for SegmentIO
  --web                 Run the web interface
  --version             Print the DC/OS version
  --genconf             Execute the configuration generation (genconf).
  --validate-config     Validate the configuration for executing --genconf and
                        deploy arguments in config.yaml
  --postflight          Execute postflight checks on a series of nodes.
  --install-prereqs     Execute the preflight checks on a series of nodes.
  --deploy              Execute a deploy.
  --preflight           Execute the preflight checks on a series of nodes.
  --uninstall           Execute uninstall on target hosts.
```

ext/dcos-installer/dcos_installer/cli.py
```diff
@@ -30,47 +30,11 @@ def setup_logger(options):
...
-CHANNEL_NAME          ADVANCED - Set build channel name
-BOOTSTRAP_ID          ADVANCED - Set bootstrap ID for build
-
-"""
-
     """
     Parse CLI arguments and return a map of options.
     """
-    parser = argparse.ArgumentParser(usage=print_usage())
+    parser = argparse.ArgumentParser()
```
cmaloney: should add `description` here with a brief snippet

cmaloney:
Note for the webserver longer term we should move the webserver building into a function (there is nothing that has to be global currently). Out of scope for this though

PR #579, opened 2016-08-24
Misc. dcos-installer cleanup

cmaloney:
- Remove some dead code in dcos_installer.util
- Move dcos_installer.util to dcos_installer.constants since it only contains constants
- Move dcos_installer/config/__init__.py to dcos_installer/config.py to just simplify directory structure a bit.


cmaloney:
cc: @malnick

PR #589, opened 2016-08-25
Update DC/OS UI version ref

orlandohohmeier:
Bump DC/OS UI version ref to latest and greatest.


cmaloney:
As a note this will make it in 1.8.3

cmaloney:
Also: 🎆  🚀  it looks like the UI build passed on the first time for once! Are the flaky build days behind us?

PR #590, opened 2016-08-25
Add cryptography to the DC/OS upstream [WIP]

malnick:
Needed for downstream validation in gen lib.


alberts:
@malnick needs some help

PR #591, opened 2016-08-25
Update Azure integration tests to support ACS templates

lingmann:
Update Azure integration tests to support ACS

* Determine the cluster IP configuration based on the detected Azure
  resource types for the deployed cluster. This is required for testing
  ACS templates which use VM Scale Sets which depend on a different API
  for extracting IP configuration.
* Fail if required environment variables are not set correctly.
* Ensure template parameters are explicitly defined for those which we
  depend on in the tests.
* Update tests to be correct, given the initial cluster configuration
  (e.g. do not test for auth if oauth is disabled via template
  parameters, skip the telemetry test if it was not enabled)
* Do not use the `master_lb` with the integration test, since this is
  not a reachable address inside an ACS cluster.

### DIFF
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -182,22 +182,26 @@ def run():
         print("ERROR: exception {}".format(ex))
         raise
     finally:
-        # Send a delete request
-        # TODO(cmaloney): The old code had a retry around this:
-        # @retry(wait_exponential_multiplier=1000, wait_exponential_max=60*1000, stop_max_delay=(30*60*1000))
-        poller = rmc.resource_groups.delete(group_name)
+        if os.getenv('AZURE_CLEANUP') == 'false':
+            print("Cluster must be cleaned up manually")
+            print("Cluster details: {}".format(azure_cluster))
+        else:
+            # Send a delete request
+            # TODO(cmaloney): The old code had a retry around this:
+            # @retry(wait_exponential_multiplier=1000, wait_exponential_max=60*1000, stop_max_delay=(30*60*1000))
+            poller = rmc.resource_groups.delete(group_name)
 
-        # poll for the delete to complete
-        print("Deleting resource group: {} ...".format(group_name))
+            # poll for the delete to complete
+            print("Deleting resource group: {} ...".format(group_name))
 
-        @retry(wait_fixed=(5 * 1000), stop_max_delay=(60 * 60 * 1000))
-        def wait_for_delete():
-            assert poller.done(), "Timed out waiting for delete"
+            @retry(wait_fixed=(5 * 1000), stop_max_delay=(60 * 60 * 1000))
+            def wait_for_delete():
+                assert poller.done(), "Timed out waiting for delete"
 
-        print("Waiting for delete ...")
-        wait_for_delete()
+            print("Waiting for delete ...")
+            wait_for_delete()
 
-        print("Clean up successful")
+            print("Clean up successful")
 
     if test_successful:
         print("Azure test deployment succeeded")
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -2,11 +2,11 @@
 import os
 import random
 import sys
+import traceback
 import uuid
 from contextlib import closing
 
 import azure.common.credentials
-import requests
 from azure.mgmt.network import NetworkManagementClient
 from azure.mgmt.resource.resources import ResourceManagementClient
 from azure.mgmt.resource.resources.models import (DeploymentMode,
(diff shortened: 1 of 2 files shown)


lingmann:
This PR is ready and now handles Azure integration testing of the ACS templates. @cmaloney or @orsenthil can I get a quick review? There is still the spartan watchdog service check that is failing, but that is being tracked via: https://dcosjira.atlassian.net/browse/DCOS-340

cmaloney:
@lingmann Can you mute just that one test in TeamCity so the status check will be green (And comment in the JIRA that it's muted for now)

lingmann:
Test muted / checks green @cmaloney

test_util/azure_test_driver.py
```diff
@@ -55,20 +71,12 @@ def run():
...
     # client_secret = os.environ.get('AZURE_CLIENT_SECRET')
-    group_name = 'tesing' + ''.join(random.choice('01234567890abcdef') for n in range(10))
+    group_name = 'testing' + ''.join(random.choice('01234567890abcdef') for n in range(10))
     deployment_name = 'deployment{}'.format(uuid.uuid4().hex)
 
     rmc = ResourceManagementClient(credentials, subscription_id)
 
     template_parameters = get_env_params()
-    if template_parameters.get('numberOfPrivateSlaves'):
-        assert template_parameters['numberOfPrivateSlaves']['value'] >= 2, 'Test requires at least 2 private slaves!'
```
cmaloney: Where are we moving setting these defaults / asserting we have enough machines to? It looks removed here but not added anywhere else.

test_util/azure_test_driver.py
```diff
@@ -123,70 +131,77 @@ def poll_deploy():
...
-                        all_ips.append(config.private_ip_address)
-                    bucket.extend(all_ips)
-
-        with closing(SSHTunnel('core', 'ssh_key', master_lb, port=2200)) as t:
+            'master': [],
+            'private': [],
+            'public': []}
+
+        for resource in rmc.resource_groups.list_resources(
+                group_name, filter=("resourceType eq 'Microsoft.Network/networkInterfaces' or "
```
cmaloney: nit: Be nice to only check the VMSS for ACS, and only the networkInterfaces for the DCOS templates, but this is reasonable for now.

test_util/azure_test_driver.py
```diff
@@ -123,70 +131,77 @@ def poll_deploy():
...
+            else:
+                raise('Unexpected resourceType: {}'.format(resource.type))
+
+            for bucket_name in ip_buckets.keys():
+                if bucket_name in resource.name:
+                    for n in nics:
+                        for config in n.ip_configurations:
+                            ip_buckets[bucket_name].append(config.private_ip_address)
+
+        print('Detected IP configuration: {}'.format(ip_buckets))
```
cmaloney: Is there a more pytest friendly way we can print this out?

test_util/azure_test_driver.py
```diff
@@ -19,11 +19,22 @@
...
-    try:
-        return requests.get(master_url)
-    except requests.exceptions.ConnectionError:
-        pass
+def validate_env():
+    '''
+    The following environment variables must be set to ensure the template can be deployed successfully and that
+    parameters we rely on from a testing perspective are explicitly set and not relying on defaults which may change.
+    '''
+    required = ['AZURE_PARAM_linuxAdminUsername', 'AZURE_PARAM_oauthEnabled',
```
cmaloney: the linuxAdminUsername doesn't seem to exist in the azuredeploy templates. Is this something from the Azure environment in general?

test_util/test_runner.py
```diff
@@ -19,7 +19,7 @@ def integration_test(
         dcos_dns, master_list, agent_list, public_agent_list,
         test_dns_search, provider,
         aws_access_key_id='', aws_secret_access_key='', region='', add_env=None,
-        pytest_cmd='py.test -vv',
+        pytest_cmd='py.test -rs -vv',
```
cmaloney: nit: Could we make the pytest command change a separate PR?

test_util/azure_test_driver.py
cmaloney: more of a nit than anything, I see the rest of this script uses print() so this is consistent.

test_util/azure_test_driver.py
lingmann: Intentionally removed, since I want the test driver to be more generic and I don't think this logic belongs here. In general, I think it should be easy for us to create CI jobs for a variety of cluster sizes and types... and adding this logic here just limits our test options.

cmaloney: The integration tests mandate 2+ private agents for some service discovery tests, or they will hard fail in every case.

test_util/azure_test_driver.py
lingmann: The short to medium term plan is to deprecate the DCOS template and replace with a version of the ACS template which has a simplified set of input parameters, but the same backend template. When this happens we will always need both resource types, since the ACS template uses both, so I'd prefer to leave this filter as is.

test_util/azure_test_driver.py
lingmann: This parameter is specific to the ACS templates, there is no equivalent with the azuredeploy templates. Essentially, I'm updating the test driver so it works with our main template (ACS) and dropping support for testing the other template type. We could continue to try and support testing the legacy template but that seems like a waste of effort right now.

test_util/test_runner.py
lingmann: Would prefer to keep it as part of this PR, since it is in a separate commit and part of this change is to intentionally skip certain tests if the cluster was configured with telemetry disabled (new behavior). So increasing the visibility of skip test details in our logs seems valuable. However I could move it to a separate PR if you feel strongly about it.

cmaloney: I'd like to land it independently / before all of this lands, hence separate PR (can leave the patch in this PR as well and after it lands git rebase / merge will fix things up).

test_util/azure_test_driver.py
lingmann: Hmm ok, still seems like an unnecessary tight coupling to me. The requirement for two agents can be explicitly captured by the CI job configuration, and integration test dependencies on the cluster spec are going to evolve over time, so I think it will be better if the test driver is generic and focuses on validating, deploying, and destroying ARM templates.

lingmann:
I've addressed all of your comments @cmaloney and triggered a fresh run of the Azure integration test. Anything else?

test_util/azure_test_driver.py
cmaloney: For many of the things we need to test in a launch, like container to container discovery and single agent failure, engineering them to run on smaller clusters isn't really valuable.

cmaloney:
:ship:

PR #594, opened 2016-08-25
Add a validate_one_of and validate_true_false to gen/calc.py

cmaloney:
These two come up really often, nice to have some common code for them.

@jgrehrcke came up with this in a closed downstream of this, moving to upstream
because it's useful.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -1,3 +1,4 @@
+
 import ipaddress
 import json
 import os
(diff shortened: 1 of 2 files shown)


gen/calc.py
```diff
@@ -15,6 +15,21 @@
...
 
 
+def validate_one_of(val, valid_values) -> None:
+    """Test if object `val` is a member of container `valid_values`.
+    Raise a ValueError if it is not a member. The exception message contains
+    both, the representation (__repr__) of `val` as well as the representation
+    of all items in `valid_values`.
+    """
+    if val not in valid_values:
+        options_string = ', '.join("'{}'".format(v) for v in valid_values)
```
jgehrcke: Let's make sure that `__repr__()` builds the string instead of `__str__()`,  via `{0!r}` (or `%r` in classical string formatting). That gives most insightful error messages.

gen/calc.py
```diff
@@ -15,6 +15,21 @@
...
+        raise AssertionError("Must be one of {}. Got '{}'.".format(options_string, val))
```
jgehrcke: Let's raise a `ValueError` here.

cmaloney: The gen library is particular about this being an AssertionError, like how all the other things in here use "assert" to check things.

https://github.com/dcos/dcos/blob/b94e399e211da753a582c5d0a5748e73b96fecf7/gen/__init__.py#L337

gen/calc.py
cmaloney: Practically val will always be a string, and valid_values should be some iterable list of strings. I can add python type annotations there which will assert it in the future.

gen/calc.py
jgehrcke: Then we should have updated the docstring ;).

cmaloney: Fix: https://github.com/dcos/dcos/pull/608

PR #596, opened 2016-08-26
integration-test/test_server improvement

sargun:
Allow you to set environment variables when running commands

### DIFF
--- a/packages/dcos-integration-test/extra/test_server.py
+++ b/packages/dcos-integration-test/extra/test_server.py
@@ -235,8 +235,17 @@ class TestHTTPRequestHandler(BaseHTTPRequestHandler):
         Sometimes there isn't enough time to write code
         """
         length = int(self.headers['Content-Length'])
-        cmd = self.rfile.read(length).decode('utf-8')
-        (status, output) = subprocess.getstatusoutput(cmd)
+        data = self.rfile.read(length).decode('utf-8')
+        if self.headers['Content-Type'] == 'application/json':
+            decoded_data = json.loads(data)
+            cmd = decoded_data['cmd']
+            environ = decoded_data.get('environ', {})
+            environ.update(os.environ)
+            proc = subprocess.Popen(cmd, shell=True, env=environ, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+            output = proc.communicate()[0].decode('utf-8')
+            status = proc.wait()
+        else:
+            (status, output) = subprocess.getstatusoutput(data)
         data = {"status": status, "output": output}
         self._send_reply(data)
 


cmaloney:
Should never use `shell=True`
Python passes through environment variables by default

sargun:
1) I want to invoke the following command: `/opt/mesosphere/packages/dcos-integration-test--8bd5e08f2f5ac68560be714b8ee8f621265bb77d/test_erlang_ssl.escript networking_api@$(detect_ip)`
2) I want to overload the environment with: `{'ERL_FLAGS': '-ssl_dist_opt client_cacertfile /run/dcos/pki/CA/certs/ca.crt -ssl_dist_opt client_keyfile /run/dcos/pki/tls/private/minuteman.key -ssl_dist_opt client_certfile /run/dcos/pki/tls/certs/minuteman.crt -ssl_dist_opt client_verify verify_peer -ssl_dist_opt client_depth 10 -ssl_dist_opt server_cacertfile /run/dcos/pki/CA/certs/ca.crt -ssl_dist_opt server_keyfile /run/dcos/pki/tls/private/minuteman.key -ssl_dist_opt server_certfile /run/dcos/pki/tls/certs/minuteman.crt'}`

sargun:
`$(detect_ip)` wont work without shell=True

cmaloney:
1. Just use os.environ['LIBPROCESS_IP'] formatted into your proper command line passed as probably an array
2. It's reasonable to be able to pass in environment variables as a parameter, there should still be only one code path, and things should probably use subprocess.check_output.

This code is very far from the right way to do the subprocess in python currently

sargun:
1) The other command I want to interpolate is `$(dig +short leader.mesos)`. I can alternatively add endpoints to run all of the checks manually (navstar, spartan, minuteman, networking API). 

2) check_output doesn't let you pass environment. The Popen constructor is the only one that lets you at the moment. I'd also rather not have a test that modifies the underlying environment, because I need to run multiple tests (is SSL working for each service)

I realize that shell is bad, and it has flaws, but I'm curious as to what route you'd recommend going down (hard code test endpoints, etc..).

As far as combining the execution paths, that seems reasonable enough. I'll do that once I have a go-forward plan.

cmaloney:
check_output lets you pass environment and all the standard arguments.

From the docs:
https://docs.python.org/3.4/library/subprocess.html#subprocess.check_output
"The arguments shown above are merely the most common ones, described below in Frequently Used Arguments (hence the use of keyword-only notation in the abbreviated signature). The full function signature is largely the same as that of the Popen constructor - this functions passes all supplied arguments other than input and timeout directly through to that interface. In addition, stdout is not permitted as an argument, as it is used internally to collect the output from the subprocess."

sargun:
Ah, that was a mistake. I was testing with interpolation, but I forgot for a moment it's the shell that does that.

PR #598, opened 2016-08-26
Octarine package and integration test update

nlsun:
The update includes aesthetic changes (should be no effect) as well as a small change to how URLs are parsed (should also be no effect). As well as a fix to a bug in the integration test.


nlsun:
Ready for review.

cmaloney:
Changing the golang version seems like a big change for this late in a release cycle / we're supposed to be bugfix only not adding / changing features including compiler versions which can often introduce new subtle changes that cause issues.

Other than that this looks good

nlsun:
Removed the golang update. Ready for review.

cmaloney:
:ship:

Will include in next train to land

PR #600, opened 2016-08-27
Switch do_create to yield artifacts rather than having every function build a list

cmaloney:
cc: @lingmann to review if this seems actually simpler.


cmaloney:
Chatted with @lingmann, seems reasonable to him, going to land once it's all green.

PR #604, opened 2016-08-29
Fix for Docker 1.11.2 startup issue

lingmann:
On a percentage of DC/OS agents (~5%) with Docker 1.11.2, the Daemon
will fail to start up with the following error:

> Error starting daemon: Error initializing network controller: Error
> creating default "bridge" network: failed to allocate gateway
> (172.17.0.1): Address already in use

This seems to be related to a Docker bug around the network controller
initialization, where the controller has allocated an ip pool and
persisted some state but not all of it. See:

* https://github.com/docker/docker/issues/22834
* https://github.com/docker/docker/issues/23078

This fix simply removes the docker0 interface if it exists before
starting the Docker daemon. This fix will need to be re-evaluated if we
want to enable the 1.12+ containerd live-restore like Docker options as
discussed in:

* https://docs.docker.com/engine/admin/live-restore/
* https://github.com/docker/docker/issues/2658

### DIFF
--- a/ext/dcos-installer/dcos_installer/action_lib.py
+++ b/ext/dcos-installer/dcos_installer/action_lib.py
@@ -377,6 +377,10 @@ sudo yum -y update --exclude="docker-engine*"
 sudo mkdir -p /etc/systemd/system/docker.service.d
 sudo tee /etc/systemd/system/docker.service.d/override.conf <<- EOF
 [Service]
+Restart=always
+StartLimitInterval=0
+RestartSec=15
+ExecStartPre=-/sbin/ip link del docker0
 ExecStart=
 ExecStart=/usr/bin/docker daemon --storage-driver=overlay -H fd://
 EOF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -57,11 +57,8 @@ runcmd:
     - [ ln, -s, /bin/mount, /usr/bin/mount ]
     - [ ln, -s, /bin/bash, /usr/bin/bash ]
     - [ ln, -s, /usr/sbin/useradd, /usr/bin/useradd ]
-    - [ systemctl, stop, resolvconf.service ]
-    - [ systemctl, disable, resolvconf.service ]
-    - [ systemctl, stop, lxc-net.service ]
-    - [ systemctl, disable, lxc-net.service ]
-    - [ systemctl, mask, lxc-net.service ]
+    - [ systemctl, disable, --now, resolvconf.service ]
+    - [ systemctl, mask, --now, lxc-net.service ]
     - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /tmp/1.deb https://az837203.vo.msecnd.net/dcos-deps/libipset3_6.29-1_amd64.deb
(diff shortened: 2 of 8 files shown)


gen/azure/cloud-config.yaml
```diff
@@ -19,6 +19,7 @@ root:
       Restart=always
       StartLimitInterval=0
       RestartSec=15
+      ExecStartPre=-/sbin/ip link del docker0
```
cmaloney: `/usr/bin/env ip link del docker0` / do we need this for AWS hosts?

lingmann: Yup, my plan is to also land this fix on the AWS EL7 AMI's that we ship by default. That will involve changing the cloud_images AMI script as well as the AWS templates. It does not happen with Docker 1.10, so we don't see the problem on CoreOS.

ext/dcos-installer/dcos_installer/action_lib.py
```diff
@@ -377,6 +377,10 @@ def _add_prereqs_script(chain):
 sudo mkdir -p /etc/systemd/system/docker.service.d
 sudo tee /etc/systemd/system/docker.service.d/override.conf <<- EOF
 [Service]
+Restart=always
+StartLimitInterval=0
+RestartSec=15
+ExecStartPre=-/sbin/ip link del docker0
```
lingmann: @malnick or @mellenburg do we have integration test coverage for this code in the `deploy-vpc` tests?

mellenburg: Yes and no. Yes, we have an option in test_installer_ccm.py called test_install_prereqs that will cover this, but no, its not running anywhere (as it is a very slow and expensive job). But these docker issues are killer, so maybe its time to turn it on for real

lingmann: I'm pretty confident that this change will result in a more robust docker install for anyone using the web installer against EL7 hosts... I've also tested this change extensively against our AWS EL7 builds running Docker 1.11.2 and it's working great.

But it would still be nice to have CI button we can press to do a full test of the web installer. Not a priority right now, since that will likely be a flaky test (it depends on a lot of external things)... but @mellenburg can you add it to a wish list to keep track of?

lingmann:
@cmaloney I believe this is good to go. In addition to our standard integrations tests, I ran the following scale tests:

* Azure 5Mx62A cluster: (0% failure rate)
* Azure 5Mx62A cluster: (0% failure rate)
* AWS EL7 1Mx60A cluster: (0% failure rate)
* AWS EL7 3Mx80A cluster: (0% failure rate)

Note I stuck with `/sbin/ip` because it exists in this location on both platforms where we need the fix (CentOS and Ubuntu), and it uses less characters than wrapping in bash, which is unfortunately a precious cloud config resource right now.

gen/azure/cloud-config.yaml
lingmann: See the linked `cloud_images/centos7/install_prereqs.sh` changes in this PR for the corresponding AWS EL7 AMI updates.

cmaloney:
:ship:

PR #605, opened 2016-08-29
[WIP] Experimental GPU Support

klueska:


klueska:
This is not intended to be merged. It's only being built as a means to build an image to use as a demo for MesosCon EU.

PR #606, opened 2016-08-29
Support non-standard ports for upstreams

drewkerrigan:
Addresses the Gen portion of https://github.com/mesosphere/mesos-dns/issues/459

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -96,6 +96,29 @@ def validate_ip_list(json_str: str):
     validate_ipv4_addresses(nodes_list)
 
 
+def validate_ip_port_list(json_str: str):
+    nodes_list = validate_json_list(json_str)
+    check_duplicates(nodes_list)
+    # Validate azure addresses which are a bit magical late binding stuff independently of just a
+    # list of static IPv4 addresses
+    if any(map(is_azure_addr, nodes_list)):
+        assert all(map(is_azure_addr, nodes_list)), "Azure resolver list and IP based static " \
+            "resolver list cannot be mixed. Use either all Azure IP references or IPv4 addresses."
+        return
+    # Create a list of only ip addresses by spliting the port from the node. Use the resulting
+    # ip_list to validate that it is an ipv4 address. If the port was specified, validate its
+    # value is between 1 and 65535.
+    ip_list = []
+    for node in nodes_list:
+        ip, separator, port = node.rpartition(':')
+        if not separator:
+            ip = node
+        else:
+            validate_int_in_range(port, 1, 65535)
+        ip_list.append(ip)
+    validate_ipv4_addresses(ip_list)
+
+
 def calculate_environment_variable(name):
     value = os.getenv(name)
     assert value is not None, "{} must be a set environment variable".format(name)
@@ -259,7 +282,7 @@ def validate_master_list(master_list):
 
 
 def validate_resolvers(resolvers):
-    return validate_ip_list(resolvers)
+    return validate_ip_port_list(resolvers)
 
 
 def validate_mesos_dns_ip_sources(mesos_dns_ip_sources):
--- a/packages/spartan/extra/gen_resolvconf.py
+++ b/packages/spartan/extra/gen_resolvconf.py
@@ -69,7 +69,20 @@ if len(spartans_up) > 0:
 
 # If Spartan is not up, fall back, and insert the upstreams
 else:
-    fallback_servers = os.environ['RESOLVERS'].split(',')
+    fallback_servers = []
+
+    # Resolvconf does not support custom ports, skip if not default
+    for ns in os.environ['RESOLVERS'].split(','):
+        ip, separator, port = ns.rpartition(':')
+        if not separator:
+            fallback_servers.append(ns)
+            continue
+        if port == "53":
+            fallback_servers.append(ip)
+            continue
+        print('Skipping DNS server {}: non-default ports are not supported in resolvconf'.format(
+            ns), file=sys.stderr)
+
     random.shuffle(fallback_servers)
     for ns in fallback_servers[:MAX_SERVER_COUNT]:
         contents += "nameserver {}\n".format(ns)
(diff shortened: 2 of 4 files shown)


sargun:
I think you're going to want to use a separate validation function because validate_ipv4_addrs is used by validate_host_list, which in turn is used by the master list as well. In addition, calculate_mesos_dns_resolvers_str might need to be changed, I'm not sure.

cmaloney:
Yea, most the validate_host_list expect no port (and should hard assert there is no port). This will need it's own validation function which checks with port.

drewkerrigan:
@sargun `calculate_mesos_dns_resolvers_str` should be fine, it just makes a decision based on the length of the list

cmaloney:
@drewkerrigan the code in gen resolvconf sets up /etc/resolv.conf with the "upstream" DNS if spartan isn't up for whatever reason (https://github.com/dcos/dcos/blob/master/packages/spartan/extra/gen_resolvconf.py#L63). `/etc/resolv.conf` doesn't allow custom port resolvers.

Need to have some solution for that (Could maybe exclude non-default-port resolvers, although I could see that causing some problems).

drewkerrigan:
@cmaloney yes that could be a problem. I think skipping is ok as long as it is clearly noted like it is here: https://github.com/dcos/dcos/blob/master/packages/spartan/extra/gen_resolvconf.py#L31

Thoughts?

drewkerrigan:
To make sure it doesn't get lost in the mix, this functionality also relies on https://github.com/mesosphere/mesos-dns/pull/468

gen/calc.py
```diff
@@ -224,6 +224,17 @@ def validate_host_list(host_list):
         return host_list
     assert not any(azure_format_check), "Azure static master list and IP based static master list cannot be mixed. Use "
     "either all Azure IP references or IPv4 addresses."
+
```
cmaloney: this should be a separate function. Technically duplicate IPs with different ports are fine. I'm not a fan of making a function that radically changes behavior based on a parameter.

packages/spartan/extra/gen_resolvconf.py
```diff
@@ -69,7 +69,20 @@ def check_server(addr):
...
+    fallback_servers = []
+
+    # Resolvconf does not support custom ports, skip if not default
+    for ns in os.environ['RESOLVERS'].split(','):
+        ip, separator, port = ns.rpartition(':')
+        if not separator:
+            fallback_servers.append(ns)
+            continue
+        if port == "53":
+            fallback_servers.append(ip)
```
cmaloney: This doesn't seem to match the docs: https://docs.python.org/3.4/library/stdtypes.html#str.rpartition

drewkerrigan: @cmaloney Not sure what you mean:

```
>>> host, sep, port = '192.168.1.1'.rpartition(':')
>>> not sep
True
>>> host, sep, port = '192.168.1.1:53'.rpartition(':')
>>> not sep
False
```

Are you just suggesting a stricter check on separator being an empty string or populated?

gen/calc.py
drewkerrigan: The duplicate check happens on the list of hosts after json decode, so duplicate IPs with different ports wouldn't break the assertion as written.

Re: separating the function, I see now that I did break the norm by adding a default argument, will address

packages/spartan/extra/gen_resolvconf.py
cmaloney:
```
>>> "1:2".rpartition(':')
('1', ':', '2')
>>> "2".rpartition(':')
('', '', '2')
```
In the case wehere sep doesn't exist, the data ends up in port / the last chunk. From the docs:
"If the separator is not found, return a 3-tuple containing two empty strings, followed by the string itself." -- the first two end up empty.

drewkerrigan: correct, when sep isn't there, the code adds `ns` to `fallback_servers`. Otherwise it will use `ip`

cmaloney:
@drewkerrigan can you rebase on master and fix conflicts. Things like the validate_host_port_list will run into stuff now (that code was reworked in gen/calc.py)

gen/calc.py
```diff
@@ -227,6 +227,30 @@ def validate_host_list(host_list):
...
 
 
+def validate_host_port_list(host_list):
+    host_list = validate_json_list(host_list)
+    azure_format_check = []
+    ips = []
+    validate_duplicates(host_list)
+    for host in host_list:
+        assert isinstance(host, str), 'Host must be of type string, got {}'.format(type(host))
+        if host.startswith('[[[reference(') and host.endswith(').ipConfigurations[0].properties.privateIPAddress]]]'):  # noqa
```
sargun: Comment, what is this behavior?

gen/calc.py
```diff
@@ -435,7 +459,7 @@ def validate_os_type(os_type):
         'provider': {
             'onprem': {
                 'default': {
-                    'resolvers': '["8.8.8.8", "8.8.4.4"]',
+                    'resolvers': '["8.8.8.8", "8.8.4.4:53"]',
```
sargun: Why? Just a matter of example, or testing?

packages/spartan/extra/gen_resolvconf.py
```diff
@@ -69,7 +69,20 @@ def check_server(addr):
...
+            continue
+        print('Skipping DNS server {}: non-default ports are not supported in resolvconf'.format(
```
sargun: /etc/resolv.conf -- resolvconf is yet another piece of DNS configuration software

gen/calc.py
drewkerrigan: Yes that was for testing, it will be removed it the next commit.

drewkerrigan:
@cmaloney Conflicts taken care of.
@sargun I've addressed your comments, take another look when you get a chance.

cmaloney:
:ship:

cmaloney:
In hold until https://github.com/mesosphere/mesos-dns/pull/468 makes it into a branch this can point to

drewkerrigan:
@cmaloney mesos-dns version bumped.

cmaloney:
:ship:

PR #610, opened 2016-08-29
Convert test_bootstrap to pytest format

cmaloney:
cc: @spahl @alberts

### DIFF
--- a/packages/bootstrap/extra/tests/test_bootstrap.py
+++ b/packages/bootstrap/extra/tests/test_bootstrap.py
@@ -1,7 +1,7 @@
 import getpass
 import logging
-import os
 import pwd
+import pytest
 import random
 import string
 import subprocess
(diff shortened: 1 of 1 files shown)


cmaloney:
the old code used a single hardcoded tmpdir which always ended up living inside the `packages/bootstrap/extra` folder which could help lead to dirty builds.

This moves it to the same style as all the rest of our tests.

packages/bootstrap/extra/tests/test_bootstrap.py
```diff
@@ -14,60 +14,57 @@
...
+    cmd_retry_policy = KazooRetry(max_tries=3, delay=0.3, backoff=1, max_delay=1, ignore_expire=False)
+    zk = KazooClient(hosts=zk_hosts, connection_retry=conn_retry_policy, command_retry=cmd_retry_policy)
+    zk.start()
 
-        children = zk.get_children('/')
-        for child in children:
-            if child == 'zookeeper':
-                continue
-            zk.delete('/' + child, recursive=True)
+    children = zk.get_children('/')
```
cmaloney: nit: Running in a docker with no volume mounts this shouldn't be needed / should probably hard error if there is anything to delete / the server isn't empty.

packages/bootstrap/extra/tests/test_bootstrap.py
```diff
@@ -14,60 +14,57 @@
...
+        zk.delete('/' + child, recursive=True)
 
-        self.zk = zk
+    yield zk
 
-    def teardown_class(self):
-        self.zk.stop()
-        self.zk.close()
-        subprocess.check_call(['docker', 'rm', '-f', self.zk_container_name])
+    zk.stop()
```
cmaloney: nit: should probably try moving kazoo to have a context manager interface option

PR #614, opened 2016-08-30
Cleanup test_signal_service / move it to the non-caching 3dt endpoint

cmaloney:
Cuts a minute off our integration testing time.


cmaloney:
I think this needs a dcos-signal code update

PR #620, opened 2016-08-30
Multiple Parameter Validation Functions

cmaloney:
TODO:
 - [ ] Unit test
 - [ ] Come up with a good way to forward these to the UI / make them show up in the UI reasonably

cc: @malnick @jgehrcke who have both wanted this recently.

### DIFF
--- a/gen/__init__.py
+++ b/gen/__init__.py
@@ -414,7 +414,7 @@ class DFSArgumentCalculator():
             if skip:
                 continue
 
-            # Call the validatoin function, catching AssertionErrors and turning them into errors in
+            # Call the validation function, catching AssertionErrors and turning them into errors in
             # the error dictionary.
             try:
                 validate_fn(**kwargs)
--- a/gen/__init__.py
+++ b/gen/__init__.py
@@ -276,16 +276,15 @@ class DFSArgumentCalculator():
         # Re-arrange the validation functions so we can more easily access them by
         # argument name.
         self._validate_by_arg = dict()
+        self._multi_arg_validate = dict()
+
         for fn in validate_fns:
             parameters = get_function_parameters(fn)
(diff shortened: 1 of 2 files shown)


gen/__init__.py
```diff
@@ -401,6 +400,27 @@ def evaluate_var(name):
...
+            skip = False
+            for parameter in parameter_set:
+                if parameter not in self._arguments:
+                    skip = True
+                    break
+                kwargs[parameter] = self._arguments[parameter]
+            if skip:
+                continue
+
+            # Call the validatoin function, catching AssertionErrors and turning them into errors in
```
jgehrcke: mini typo: validatoin

gen/__init__.py
```diff
@@ -401,6 +400,27 @@ def evaluate_var(name):
...
+            # the error dictionary.
+            try:
+                validate_fn(**kwargs)
+            except AssertionError as ex:
+                self._errors[parameter_set] = ex.args[0]
```
jgehrcke: Looks like we lose the exception message here. Is that what we want?

cmaloney: Yes, we keep the first argument to the exception, but drop the rest. See the existing cases: https://github.com/dcos/dcos/blob/master/gen/__init__.py#L337

This always works for python's `assert` assertions and our manually raised AssertionErrors

This is basically saying "Catch all asserts" (because asserts are easier with a nicer syntax than lots of other options), and turn the message that was printed into the assert, to the message in a great big error dictionary.

jgehrcke: :+1: 

> gen/__init__.py#L337

I blindly trust code anchored at `#L337`! :)

cmaloney:
Replaced by #628. That includes the fix, a integration test by way of the ssh validation tests exercising this, and includes code to "squash" multi-argument things to error on all parameters for the validation failure to solve the UI problem for now.

PR #622, opened 2016-08-30
Replace ccm provider

mellenburg:
Builds on work from
https://github.com/dcos/dcos/pull/240


mellenburg:
@cmaloney this is ready for view
@branden this needs to be coordinated with https://github.com/dcos/dcos/pull/663

branden:
@mellenburg cool, I'll rebase #663 onto this branch.

test_util/test_aws_vpc.py
```diff
@@ -191,20 +154,39 @@ def main():
     vpc = None  # Set if the test owns the VPC
 
     if options.host_list is None:
-        log.info('CCM_VPC_HOSTS not provided, requesting new VPC from CCM...')
-        vpc = make_vpc(use_bare_os=options.test_install_prereqs)
-        host_list = vpc.hosts()
-    else:
```
branden: This `else` case was lost in the refactor.

mellenburg: Ah, that was actually very intentional as i intended on making everything to be options.host_list because above it is parsed as a list:
options.host_list = os.environ['VPC_HOSTS'].split(',')

But, I flubbed making everything else options.host_list, so good catch!

test_util/test_aws_vpc.py
```diff
@@ -191,20 +154,41 @@ def main():
...
+            instance_os=os_name,
+            instance_count=5,  # 1 bootstrap, 1 master, 2 agents, 1 public agent
+            admin_location='0.0.0.0/0',
+            key_pair_name='default',
+            boto_wrapper=bw)
+        vpc.wait_for_stack_creation()
+
+        vpc_info = vpc.get_vpc_host_ips()
+        log.info('AWS provided VPC info: ' + repr(vpc_info))
+        options.host_list = [i.public_ip for i in vpc_info]
```
branden: `check_environment()` looks like the function that's responsible for assembling the `options` object, so seeing an assignment to it outside that context is a surprise. It'd be clearer to assign to `host_list` here, and move the below `host_list = options.host_list` into an `else` block.

mellenburg: Fair point. Will change

test_util/test_aws_vpc.py
```diff
@@ -191,20 +154,41 @@ def main():
...
         host_list = options.host_list
 
-    assert os.path.exists('ssh_key'), 'Valid SSH key for hosts must be in working dir!'
-    # key must be chmod 600 for test_runner to use
-    os.chmod('ssh_key', stat.S_IREAD | stat.S_IWRITE)
-
     # Create custom SSH Runnner to help orchestrate the test
     ssh_user = 'centos'
-    ssh_key_path = 'ssh_key'
+    ssh_key_path = 'default_ssh_key'
```
branden: In the previous version, we knew the SSH key was at `ssh_key_path` because `make_vpc()` retrieved it from CCM and wrote it to that location. Where is it coming from now?

mellenburg: Right now I am having CI do it by magic (part of the job definition). This is not ideal, but we will soon switch to generated keys and then this wont be an issue, we just need to make sure we have the infra to cleanup all the keys before we do that so we don't get locked out of making more

branden: Since this code is no longer responsible for generating the key, would it make sense to pass the key's path as a config parameter rather than hardcoding it?

mellenburg: Yeah, I actually did this in the other variant of this script: https://github.com/dcos/dcos/pull/689/files
will port

test_util/aws.py
```diff
@@ -115,3 +125,116 @@ def get_public_agent_ips(self):
...
+            'AllowAccessFrom': admin_location,
+            'ClusterSize': str(instance_count),
+            'InstanceType': str(instance_type),
+            'AmiCode': ami_code}
+        stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
+        return cls(stack.stack.stack_id, boto_wrapper)
+
+    def delete(self):
+        # boto stacks become unusable after deletion (e.g. status/info checks) if name-based
+        self.stack = self.boto_wrapper.resource('cloudformation').Stack(self.stack.stack_id)
```
branden: Why create a new instance property in a `delete()` method?

mellenburg: Not a new property. self.stack is set in __init__. The point of this is that after you call delete you may want to wait for the delete or check why the delete failed, but you can't do that if the stack is still set by name, so switch to id here

branden: Gotcha.

setup.py
```diff
@@ -55,7 +55,9 @@ def get_advanced_templates():
     entry_points={
         'console_scripts': [
             'release=release:main',
-            'ccm-deploy-test=test_util.test_installer_ccm:main',
+            # Remove this once CI is updated
+            'ccm-deploy-test=test_util.test_aws_vpc:main',
```
cmaloney: can we just keep using the old name for now rather than make it so the CI doesn't work on old commits / old builds? Eventually we should move all of these to some common wrapper where we can dispatch to the individual variations with flags rather than one binary per test shape

test_util/aws.py
```diff
@@ -115,3 +125,116 @@ def get_public_agent_ips(self):
...
+    def get_vpc_host_ips(self):
+        reservations = self.boto_wrapper.client('ec2').describe_instances(Filters=[{
+            'Name': 'tag-value', 'Values': [self.stack.stack_name]}])['Reservations']
+        logging.debug('Reservations for {}: {}'.format(self.stack.stack_id, reservations))
+        instances = reservations[0]['Instances']
+        return instances_to_hosts(instances)
+
+
+OS_AMIS = {
+    'cent-os-6': {'ap-northeast-1': 'ami-82640282',
```
cmaloney: I'd vote on killing all this CentOS 6 map. We don't ship / support CentOS 6 and I'm pretty confident at this point we never will.

test_util/aws.py
```diff
@@ -115,3 +125,116 @@ def get_public_agent_ips(self):
...
+    'debian-8': {'ap-northeast-1': 'ami-fe54f3fe',
+                 'ap-southeast-1': 'ami-60989c32',
+                 'ap-southeast-2': 'ami-07e3993d',
+                 'eu-central-1': 'ami-b092aaad',
+                 'eu-west-1': 'ami-0ed89d79',
+                 'sa-east-1': 'ami-a5bd3fb8',
+                 'us-east-1': 'ami-8b9a63e0',
+                 'us-west-1': 'ami-a5d621e1',
+                 'us-west-2': 'ami-3d56520d'},
+    'rhel-6': {'ap-northeast-1': 'ami-78379d78',
```
cmaloney: I'd drop this for the same reason as the centos6

test_util/aws.py
```diff
@@ -115,3 +125,116 @@ def get_public_agent_ips(self):
...
+    'rhel-7': {'ap-northeast-1': 'ami-35556534',
+               'ap-southeast-1': 'ami-941031c6',
+               'ap-southeast-2': 'ami-83e08db9',
+               'eu-central-1': 'ami-e25e6cff',
+               'eu-west-1': 'ami-8cff51fb',
+               'sa-east-1': 'ami-595ce844',
+               'us-east-1': 'ami-a8d369c0',
+               'us-west-1': 'ami-33cdd876',
+               'us-west-2': 'ami-99bef1a9'},
+    'ubuntu-12-04': {'ap-northeast-1': 'ami-88bca789',
```
cmaloney: We won't ever support 12.04 or 14.04, so I don't see value in having them (16.04 will be supported eventually, so we could have that one).

test_util/test_aws_vpc.py
```diff
@@ -180,7 +148,7 @@ def check_environment():
     options.add_env = add_env
 
     options.pytest_dir = os.getenv('DCOS_PYTEST_DIR', '/opt/mesosphere/active/dcos-integration-test')
-    options.pytest_cmd = os.getenv('DCOS_PYTEST_CMD', 'py.test -rs -vv ' + options.ci_flags)
+    options.pytest_cmd = os.getenv('DCOS_PYTEST_CMD', 'py.test -vv ' + options.ci_flags)
```
cmaloney: `-rs` is missing

test_util/test_aws_vpc.py
```diff
@@ -191,20 +159,41 @@ def main():
     vpc = None  # Set if the test owns the VPC
 
     if options.host_list is None:
-        log.info('CCM_VPC_HOSTS not provided, requesting new VPC from CCM...')
-        vpc = make_vpc(use_bare_os=options.test_install_prereqs)
-        host_list = vpc.hosts()
+        log.info('VPC_HOSTS not provided, requesting new VPC ...')
+        random_identifier = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
```
cmaloney: nit: We have this "make a random string of length X then append to prefix" a few places, might want to pull it into a common util.

test_util/test_aws_vpc.py
```diff
@@ -191,20 +159,41 @@ def main():
...
+        else:
+            os_name = "cent-os-7-dcos-prereqs"
+        # TODO(mellenburg): Switch to using generated keys
+        bw = test_util.aws.BotoWrapper(
+            region=DEFAULT_AWS_REGION,
+            aws_access_key_id=options.aws_access_key_id,
+            aws_secret_access_key=options.aws_secret_access_key)
+        vpc = test_util.aws.VpcCfStack.create(
+            stack_name=unique_cluster_id,
+            instance_type='m4.xlarge',
```
cmaloney: Why not make this environment variable controllable?

test_util/test_aws_vpc.py
mellenburg: Agreed, very tired of writing this function over and over. Just need to add a "give me a good-enough unique ID for something that will only be around for day"

test_util/test_aws_vpc.py
mellenburg: Will do

setup.py
mellenburg: Sure, I'll update the comment and remove the new endpoint. It just feels _wrong_ to have an executable called ccm-* that doesnt touch CCM :P

test_util/aws.py
mellenburg: Was copy/pasting from CCM so that anyone could expect the same set of AMIs in this API, but yes, definitely pointless to keep around if its a DC/OS specific utility

setup.py
cmaloney: agreed, soon we shall cleanse that. Just want to do it all at once rather than piecemeal

test_util/test_aws_vpc.py
mellenburg: Actually, lets do this in a different PR. Need to fix the AWS CF one also and don't want that in this PR

mellenburg:
@cmaloney Is this shippable? it will help greatly in building out tests that do resiliency on VPC

cmaloney:
:ship: definite improvement

PR #628, opened 2016-08-31
Pull out a "Config Target" from gen.calculate, use it for ssh configuration

cmaloney:
The ConfigTarget will also get used for AWS Key validation and allow using one config.yaml which doesn't have all of it's parameters go into the cluster: AWS keys, SSH keys, etc in config.yaml shouldn't end up on every host.

# TODO
 - [x] Make backend.do_validate_gen_config() work
 - [x] Make backend.do_validate_ssh_config() wok
 - [x] Make sure an [Azure PR Test](https://teamcity.mesosphere.io/viewType.html?buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssAzureIntegrati_2) passes

This includes:
#627 (although can land independently)
Replaces
#620 (Adds fixes, integration test by way of ssh validation test)

# Needed followup
 - Make it so gen knows all parameters that are ever available and makes sure the user doesn't have random unknown parameters
 - Make ConfigTarget get the full computed arguments for that one config target and stash them on the ConfigTarget. This is critical so that with the AWS Advanced templates from dcos_generate_config.sh the AWS keys for uploading templates don't end up on every host in the cluster / in the cluster at all

PR #633, opened 2016-09-01
Enforce PEP8 E121, E123, E126, E226, E24, E704

cmaloney:
Enforce PEP8 E121,E123,E126,E226,E24,E704
This normalizes up some inconsistencies in our code style, enforces them for the future.
 - E121: Continuation line under-indented for hanging indent
 - E123: Closing bracket must match indentation of opening bracket or be after first item
 - E126: Argument lists which wrap around must only be indented once / not be overly indented
 - E226: Spaces must be used around arithmetic operators `1+1` -> `1 + 1`
 - E241: disallow multiple spaces after ','
 - E242: disallow tab after ','
 - E704: disallow multiple statements on one line (def)

The most common ones we had in our code base was E226 followed by E121/E123/E126. The rest didn't
ever show up / we already were following them

If you're on Sublime I'd recommend Anaconda for helping find / catch these as you code. For other
things, anything which works with flake8 and pycodestyle should be pretty good, see tox.ini for
flake8 settings. If you don't like live hints, just run `tox -e py34-syntax` to catch these errors.


jgehrcke:
I encourage everyone who contributes Python code to watch this talk:
https://www.youtube.com/watch?v=wf-BqAjZb8M

I am OK with this change. But let's be open to future adjustment. Let's watch out carefully and see if some of these checks get into people's way by complaining about syntax that does not actually violate PEP 8.

(If you actually read PEP 8 it leaves quite some room in many regards, and every validator implementation I have worked with so far did, in subtle details, over-interpret PEP 8 and sometimes imposed opinions I did not agree with -- this can be quite annoying).

gen/installer/azure.py
```diff
@@ -150,8 +149,7 @@ def gen_templates(user_args, arm_template):
         results.templates[arm_template + '.json'],
         variant_cloudconfig['master'],
         variant_cloudconfig['slave'],
-        variant_cloudconfig['slave_public']
-        )
+        variant_cloudconfig['slave_public'])
```
jgehrcke: This change: is it _enforced_ by one of the newly introduced checks? Then I don't want that to be introduced. It fu*** with PEP 8 in a rather invasive way. See https://www.python.org/dev/peps/pep-0008/#id17.

packages/dcos-history/extra/history/statebuffer.py
```diff
@@ -83,7 +83,7 @@ def update_and_ff(f_path, ff_end):
                 # Set timestamp to None for memory-only buffer updates
                 with open(f_path, 'r') as fh:
                     self._update_buffer(fh.read())
-                while (ff_end-self.update_period) >= self.next_update:
+                while (ff_end - self.update_period) >= self.next_update:
```
jgehrcke: This check: :+1:

pkgpanda/build/__init__.py
```diff
@@ -958,11 +958,12 @@ def cache_abs(filename):
...
-                                 "dependency {} variant {} doesn't exist. Rebuild the dependency.".format(
-                                    pkg_tar,
-                                    requires_name,
-                                    requires_variant))
+                raise BuildError(
+                    "The build tarball {} refered to by the last_build file of the dependency {} "
+                    "variant {} doesn't exist. Rebuild the dependency.".format(
+                        pkg_tar,
+                        requires_name,
+                        requires_variant))
```
jgehrcke: I'm always in favor of pulling code "to the left" as much as possible, so I very much like this change.

But: was this change now enforced by one of the new checks? I would not support that, because both of these styles are PEP 8 compliant

```
foo = long_function_name(var_one, var_two,
                         var_three, var_four)
```

```
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```

gen/installer/azure.py
cmaloney: This is one of two allowed styles currently:

```python
     # Render the arm
     arm = render_arm(
          results.templates[arm_template + '.json'],
          variant_cloudconfig['master'],
          variant_cloudconfig['slave'],
          variant_cloudconfig['slave_public'])
```
and

```python
     # Render the arm
     arm = render_arm(
          results.templates[arm_template + '.json'],
          variant_cloudconfig['master'],
          variant_cloudconfig['slave'],
          variant_cloudconfig['slave_public']
    )
```

Are both currently allowed. We can toggle between the second and:
```python
     # Render the arm
     arm = render_arm(
          results.templates[arm_template + '.json'],
          variant_cloudconfig['master'],
          variant_cloudconfig['slave'],
          variant_cloudconfig['slave_public']
        )
```

but only one of those two can be allowed at a time.

packages/dcos-history/extra/history/statebuffer.py
cmaloney: That is the one my editor kept saying and I was wondering _why_ flake8 wasn't finding it / erroring

pkgpanda/build/__init__.py
cmaloney: So both of those are accepted here. I tend to really, really hate the prior and hence moved lots of stuff to the latter. (The prior ends up with excessive wrapping for no real value)

cmaloney:
Definitely agree with the video, and definitely open for adjustment in the future. We should also add our own hacking styles as we find specific things over time / plug into flake8. These are all the things flake8 didn't enable by default because they were a bit "controversial". I'm hoping by enforcing up front we just get rid of the whole argument around "what is the pretty code wrap" (ala clang-format or go-fmt) and "if it passes then commit it and it's standardized enough"

And then we can hopefully pay attention to using the code properly / "beyond pep8" rather than the simple "is it formatted sanely"

gen/installer/azure.py
jgehrcke: >  only one of those two can be allowed at a time. 

OK. If I must choose between the last two, I would opt for the latter:
```
foo = bar(
    boo,
    far
    )
```
But this is a small dilemma:

I really tend to prefer that, because I believe it visually integrates better with Python's overall indentation concept. Using this style is my year-old habit, and if we were to enforce "closing bracket to the far left" (your 2nd code block), I would _constantly_ run against that wall.

However, what I described is just my personal style preference. I entirely get that people coming from languages with curly brackets like and tend to do what your 2nd code block shows.

Having said this, I will be "OK" with either way here.

jgehrcke:
Full ack to your last comment here. These efforts are definitely useful!

Maybe take my "dilemma" comment into consideration once again, but otherwise overall :+1: for this PR.

cmaloney:
@jgehrcke https://github.com/dcos/dcos/pull/633/commits/0c048e6e037d90ab45478546271b49bdd4d2bc8f

Landing shortly

cmaloney:
Reverting on the hanging indent part. There are 355 instances in the codebase that currently don't do the hanging indent, which is _much_ larger than the number that do.

I really don't care which style we use, but that much code motion is definitely a different PR and this matches the majority

gen/calc.py
```diff
@@ -205,7 +205,7 @@ def validate_json_list(json_list):
 
         assert type(list_data) is list, "Must be a JSON list. Got a {}".format(type(list_data))
     except ValueError:
-        assert False, "Provided input was not valid JSON: "+json_list
+        assert False, "Provided input was not valid JSON: " + json_list
```
orsenthil: Might be a good idea to consistently use `{}` instead of + at the some stmts.

gen/installer/azure.py
orsenthil: I prefer the first one style suggested above.. With the second style, if you have nested closing brackets, then it will end up indented each in one line.

orsenthil:
I had 1 minor comment that can be addressed out of band. ship-it. (merged already).

gen/calc.py
cmaloney: Definitely aiming for that.  Many, [including this one I](https://github.com/dcos/dcos/pull/628/files#diff-3cd4ddcecc89d0887dd9cb3d5fabe995L208) removed in #628 as part of other refactoring / cleanup

PR #635, opened 2016-09-01
Updating mesos-dns ref

drewkerrigan:
Changes in this update:

* Rename IAM `secret` to `private_key`: https://github.com/mesosphere/mesos-dns/pull/465
* Stability improvements: https://github.com/mesosphere/mesos-dns/pull/461
* Fix crashes related to `labels.trimCut` problems: https://github.com/mesosphere/mesos-dns/pull/460

### DIFF
--- a/packages/mesos-dns/buildinfo.json
+++ b/packages/mesos-dns/buildinfo.json
@@ -2,8 +2,8 @@
   "sources": {
     "mesos-dns": {
       "kind": "url",
-      "url": "https://734-26509179-gh.circle-artifacts.com/0//tmp/circle-artifacts.pA0dRJZ/mesos-dns-v0.5.3-rc1-16-g22146c6-linux-amd64",
-      "sha1": "04feb2e5d91dd2a51a779b900149dc312c4f1c86"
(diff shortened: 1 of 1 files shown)


drewkerrigan:
Tested a cluster with these changes, works as expected

cmaloney:
If you can test by hand, is there an integration test we can write into DC/OS for it?

drewkerrigan:
@cmaloney the change that is easily tested by hand is this: Fix crashes related to `labels.trimCut` problems: mesosphere/mesos-dns#460.

There are unit tests for it here: https://github.com/mesosphere/mesos-dns/blob/master/records/labels/labels_test.go#L146

I could add an integration test similar to this: https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_service_discovery.py but it seems like overkill to verify an edge case related to a task name like `hello--------------------------------------------------------------------------------------------------f`, thoughts?

cmaloney:
My general goal is "How do I make sure I never ship a Mesos-DNS with this bug again / regress in the integration". For this one probably good enough for now. I'm still really not a fan of the circleci artifact paths... But future things to fix.

cmaloney:
:ship:

PR #637, opened 2016-09-01
Adminrouter cloudformation variable rendering (DCOS-321)

lingmann:
Ensure that adminrouter.env is populated with the results of the
Cloudformation expression, by including it in the late-binding config
packages.


lingmann:
@alberts for https://dcosjira.atlassian.net/browse/DCOS-321

PR #638, opened 2016-09-01
calculate diagnostics job timeout

darkonie:
The tests have shown it takes about 1 second per node to fetch the logs. While
the static timeout works well for small clusters it does not work for large
ones. That being said, we shall calculate the timeout in runtime based on a
number of nodes in a cluster + 15% tolerance.


cmaloney:
I fired off an azure run for this, if that passes then we can :ship:

PR #641, opened 2016-09-01
Configuring oauth in the advanced templates

pyronicide:
FIXES: https://dcosjira.atlassian.net/browse/DCOS-321

### DIFF
--- a/gen/aws/templates/advanced/advanced-master.json
+++ b/gen/aws/templates/advanced/advanced-master.json
@@ -53,6 +53,17 @@
       "Description": "\nExisting AMI in the deploy region which has all DC/OS pre-requisites installed or 'default'",
       "AllowedPattern": "(default|ami-[a-f0-9]{8})"
     }
+{% switch oauth_available %}
+{% case "true" %}
+    ,
+    "OAuthEnabled": {
+      "Description": "\nEnable OAuth authentication",
+      "Type": "String",
+      "Default": "true",
+      "AllowedValues": ["true", "false"]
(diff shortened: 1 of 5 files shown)


lingmann:
I don't believe we strictly need the `OAuthEnabled` cloudformation snippets on the private or public agent templates... this drives the `oauth_enabled` `calc.py` parameter, which I believe is only used by adminrouter. @alberts can you confirm that this is only being used on the masters?

If it's not strictly necessary, then the question is do we remove it from the agent templates to simplify the templates / reduce line count, or do we leave it to keep configuration as similar as possible everywhere.... thoughts @cmaloney ?

Otherwise, @pyronicide this looks good to me. 👍  I've run additional tests and confirmed that advanced clusters launch, the var is being set correctly, and the /mesos is restricted.

pyronicide:
@lingmann we don't *strictly* need it. Unfortunately, the param `OAuthEnabled` is being substituted at the file level (`etc/adminrouter.env` in the cloud config). We don't currently have a method for only including some of the `cc_package_files` based off type.

I'd rather not do the right fix (allow optional `cc_package_files` based on template) in the context of this bug and hacking it will just add tech debt. Passing the var down to every template feels like the least-bad of all the possible options.

lingmann:
Yeah, that doesn't seem like a great trade-off at this point. And we will have better ways to do the late-binding variable substitutions in the future. I'm a ⛵

lingmann:
@cmaloney for the merge queue

PR #643, opened 2016-09-02
Replace isort with flake8-import-order

cmaloney:
This replaces isort which is painful to use / maintain with the
flake8-import-order plugin which means we don't have do things like manually
maintain a whitelist for isort and a blacklist for flake8. Don't have to remember
to add things to isort. Don't have to manually set known_first_party and known_third_party

Style wise the biggest difference from current the "smarkets" style which is set
now is that from `x import a, B, c` cares about sorting alphabetically without
caring about case now

Based on the code changes, this catches a bunch of basic sorting was off which says the tool config
for isort was too complicated and the new plugin is a lot easier in that respect.

I think the improvement in CI usability (They show up on the teamcity base page) and catching
programmer mistakes / harder to forget to add things to be isorted, this is worth the slightly worse
diagnostics in telling people how to fix it (isort would print a diff and auto-fix, this just says "higher or lower")

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,4 +1,0 @@
-[settings]
-known_first_party=deploy,gen,pkgpanda,release,ssh,test_util
-known_third_party=bs4,dns,docopt,kazoo,requests,retrying,pytest,yaml
-not_skip=__init__.py
--- a/ext/dcos-installer/dcos_installer/action_lib.py
+++ b/ext/dcos-installer/dcos_installer/action_lib.py
@@ -4,8 +4,8 @@ import logging
 import os
 
 import pkgpanda
-from ssh.ssh_runner import Node
 import ssh.utils
+from ssh.ssh_runner import Node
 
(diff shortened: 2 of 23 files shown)


orsenthil:
This looks better. ship-it. thank you.

jgehrcke:
Thanks, Cody! I don't know how exactly flake8-import-order works, but this looks like we just got way closer to having imports sorted into three blocks (stdlib, third party, custom/local) -- which IMO in practice is the most important (helpful) thing one can do to import statements (and it's actually recommended by PEP 8, which alphabetical sorting is not) :).

cmaloney:
I'm actually a big fan of sorting on top of the three chunks: 
1. It creates order in chaos. A lot of coding is "formalizing" things, and when I can assume things because I know it will be reliably put together that can be helpful
2. when you get a file with a reasonably long list of imports I can figure out if what I want is already imported with a quick search.

jgehrcke:
> sorting on top of the three chunks

I'm really not against it, just very much in favor of actually enforcing the three chunks! :)

PR #644, opened 2016-09-02
Change AWS Simple template AMI back to CoreOS

cmaloney:
This was inadvertently changed in 28e52bd6ffe8af5e49fcd2d12499ccc92476f2f9

That was supposed to just update the custom CentOS AMI, but caught the CoreOS
AMI as well

cc: @lingmann


cmaloney:
Confirmed by launching a single-master cluster, things come up and nodes are all CoreOS.

lingmann:
Thanks Cody this is definitely the issue. :+1:

PR #647, opened 2016-09-03
fix flaky `test_3dt_bundle_create` test

darkonie:
- latest 3dt version exposes a `job_progress_percentage`. Make the test to
  track the progress. If test detects that the diagnostics job is stale, it will
  fail right away.
  https://github.com/dcos/3dt/pull/65

- typo fix
  https://github.com/dcos/3dt/pull/64

### DIFF
--- a/packages/dcos-integration-test/extra/test_3dt.py
+++ b/packages/dcos-integration-test/extra/test_3dt.py
@@ -1,3 +1,4 @@
+import datetime
 import gzip
 import json
 import logging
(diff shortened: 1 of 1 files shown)


packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -458,28 +458,63 @@ def test_3dt_bundle_create(cluster):
...
-    # this test can be run in. Let's calculate a timeout based on how many nodes in a cluster + 15% tolerance.
-    # the timeout would looks like: (len(all_nodes) + 15%) * 1000
-    timeout = len(cluster.masters + cluster.all_slaves) * 1000
-    tolerance = int(15 * timeout / 100.0)
-    job_timeout = timeout + tolerance
-    logging.info("Using timeout {}".format(job_timeout))
-
-    @retrying.retry(stop_max_delay=job_timeout, wait_fixed=1000)
+    # lets make sure we are progressing. We need at least 2 points to compare.
+    def is_job_progressing(job_progress):
```
cmaloney: First time we run through this if 3dt is taking some time, wouldn't this hit the "not 2 data points" bit which we don't want it to since we want to fail hard if we're not making progress?

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -458,28 +458,63 @@ def test_3dt_bundle_create(cluster):
...
+    class StaleJobException(Exception):
+        """Exception should be raised if the job is not progressing"""
+
+    # calculate the `wait_for_job` timeout. Let's give 15 seconds per node. Should be good enough.
+    timeout = len(cluster.masters + cluster.all_slaves) * 15000
+
+    # if we encounter StaleJobException we should fail right away.
+    @retrying.retry(wait_fixed=2000, stop_max_delay=timeout,
+                    retry_on_exception=lambda e: not isinstance(e, StaleJobException))
     def wait_for_job():
```
cmaloney: The retry loop inside a retry loop is a bit odd for a "continue forever" loop.

Why not make the main retry loop retry every 2 seconds or so, and have a couple state variables it keeps

1. Last state change time (`datetime.now()` when it got the last state change or got no state)
2. The last state retrieved (initially none)

The retry loop then does
1. fetch progress and print it for debugging (stdout/stderr will get captured by py.test)
2. check if state is done, and if so exit success
3. check if state has changed, and if so store new state, update last timestamp, continue / iterate in retry loop again
4. check if `datetime.now() - last_state_change > timedelta(seconds=15)`, and if so, exit loop with error (no progress made in 15 seconds).

cmaloney:
The manual tuple of jog progress, as well as keeping track of job progress over time I think still gets in the way of being able to simply / easily follow the business logic in this.

I'd really like the controller of the loop to be retrying  / retry to be the driver of the "wait for complete" progress loop like we do a lot in the other code that we have.

What do you think of restructuring it to be something like: https://github.com/cmaloney/dcos/blob/pr647/packages/dcos-integration-test/extra/test_3dt.py#L450

The goal is to make it really simple / easy to follow th ecore logic, and make the code more linear ideally. Also adds some additional checking, and gets rid of a bunch of branching / ifs which I think added complexity to the version you have in the PR currently.

cmaloney:
:ship: (assuming tests pass)

cmaloney:
Holding until tomorrow per submitter request

darkonie:
@cmaloney this one is ready to go. Tested on azure with 5 masters 5 agents cluster, took ~ 11 seconds https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=395876#_focus=2644

PR #648, opened 2016-09-03
Add upgrade integration test

branden:


branden:
@cmaloney Ready for review. Successful upgrade test run: https://teamcity.mesosphere.io/viewLog.html?buildId=416466&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssUpgradeTestOnPre&tab=buildResultsDiv

branden:
The systemd unit health test is failing pretty frequently on `dcos-spartan-watchdog`. It becomes healthy eventually after a few minutes, but I haven't investigated further yet. Tests look good otherwise.

test_util/cluster.py
```diff
@@ -184,6 +186,90 @@ def check_ssh(self, hosts):
...
+                tunnel.remote_cmd([
+                    'curl', '--silent', '{}:{}/metrics/snapshot'.format(host.private_ip, port)
+                ]).decode('utf-8')
+            )
+
+    def zk_mode(self, host):
+        """Return the mode of the ZooKeeper instance on host."""
+        with self.ssher.tunnel(host) as tunnel:
+            stat_out = tunnel.remote_cmd([
+                'sudo', 'docker', 'run',
```
branden: Kazoo would be ideal, but I'm deliberately SSHing to the host and then talking to the client port with the expectation that this method will be easier to maintain as masters get more secure. ZooKeeper docs recommend netcat, they don't appear to provide a tool for sending 4-letter commands. I'll update this to use the netcat bin we already have installed.

test_util/cluster.py
```diff
@@ -278,6 +361,105 @@ def install_dcos(
...
+            # False < True
+            key=lambda host_zk_mode: host_zk_mode[1] == 'leader',
+        )]
+
+    def upgrade_host(tunnel, role, bootstrap_url):
+        # Download the install script for the new DC/OS.
+        tunnel.remote_cmd(['curl', '-O', bootstrap_url + '/dcos_install.sh'])
+
+        # Remove the old DC/OS.
+        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
```
branden: If all it does is prepend `sudo` then that seems pretty low value. And it'd need a flag for whether to pass `-i`...it doesn't seem worth the indirection.

test_util/installer_api_test.py
```diff
@@ -295,7 +296,7 @@ def genconf(
         """
         test_config = {
             'cluster_name': 'SSH Installed DC/OS',
```
branden: Because our upgrade doc says to serve bootstrap files from the bootstrap host, and this upgrade procedure needs to match that doc. I assume that's our recommendation because `scp`ing files to each host is more tedious than running an nginx container and changing the bootstrap URL.

If the installer provided a command that only distributed bootstrap files among the cluster, we could use that here. But right now that's coupled to `--deploy`.

test_util/cluster.py
```diff
@@ -184,6 +186,90 @@ def check_ssh(self, hosts):
...
+
+def run_docker_container_daemon(tunnel, container_name, image, docker_run_args=None):
+    """Run a Docker container with the given name on the host at tunnel."""
+    docker_run_args = docker_run_args or []
+    tunnel.remote_cmd(
+        ['sudo', 'docker', 'run', '--name', container_name, '--detach=true'] + docker_run_args + [image]
+    )
+
+
+def run_bootstrap_zookeeper(tunnel):
```
branden: That second example is a little different since it's running the container locally, not over SSH. I was able to at least DRY the docker args: https://github.com/dcos/dcos/pull/648/commits/320d7289949935e33a9c68403b05e624d9794bbd

branden:
@cmaloney Ready for another look.

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+            '--retry', '20',
+            '--speed-limit', '100000',
+            '--speed-time', '60',
+            '--remote-name',
+            bootstrap_url + '/dcos_install.sh',
+        ])
+
+        # Remove the old DC/OS.
+        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
+        tunnel.remote_cmd(['sudo', 'rm', '-rf', '/opt/mesosphere', '/etc/mesosphere'])
```
cmaloney: why do we need to remove `/etc/mesosphere`?

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+
+        # Install the new DC/OS.
+        tunnel.remote_cmd(['sudo', 'bash', 'dcos_install.sh', '-d', role])
```
cmaloney: why are we skipping the preflight here?

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+    def wait_for_mesos_metric(cluster, host, key, value):
+        """Return True when host's Mesos metric key is equal to value."""
+        return cluster.mesos_metrics_snapshot(host).get(key) == value
+
+    assert all(h.public_ip for h in cluster.hosts), (
+        'All cluster hosts must be externally reachable. hosts: {}'.formation(cluster.hosts)
+    )
+    assert cluster.bootstrap_host, 'Upgrade requires a bootstrap host'
+
+    bootstrap_url = 'http://' + cluster.bootstrap_host.private_ip
```
cmaloney: nit: may want to check all machines can access bootstrap_url

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+        ssh_key = key_fh.read()
+    installer = test_util.installer_api_test.DcosCliInstaller()
+    with cluster.ssher.tunnel(cluster.bootstrap_host) as bootstrap_host_tunnel:
+        installer.setup_remote(
+            tunnel=bootstrap_host_tunnel,
+            installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh',
+            download_url=installer_url,
+        )
+        installer.genconf(
+            bootstrap_url=bootstrap_url,
```
cmaloney: nit / future thoughts: It'll be really nice when we just pass more of a config.yaml here rather than hand-blowing out some parameters.

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+
+    upgrade_ordering = [
+        ('master', 'master', master_upgrade_order(cluster)),
+        ('slave', 'agent', cluster.agents),
+        ('slave_public', 'public agent', cluster.public_agents),
+    ]
+    for role, role_name, hosts in upgrade_ordering:
+        logging.info('Upgrading {} nodes: {}'.format(role_name, repr(hosts)))
+        for host in hosts:
+            logging.info('Upgrading {}: {}'.format(role_name, repr(host)))
```
cmaloney: nit: Might be handy for debugging to log whether the particular machine is the mesos master or zookeeper master before restarting it as those are more interesting events than the non-leading ones.

test_util/cluster.py
```diff
@@ -278,6 +358,118 @@ def install_dcos(
...
+            with cluster.ssher.tunnel(host) as tunnel:
+                upgrade_host(tunnel, role, bootstrap_url)
+
+            if role == 'master':
+                wait_metric = 'registrar/log/recovered'
+            else:
```
cmaloney: nit: 
```
wait_metric = {
    'master': 'registrar/log/recovered',
    'agent': 'slave/registered',
    'agent_public': 'slave/registered'
}[role]
```

I really prefer to be explicit about the expected set of cases. The re-formating I give is what is commonly used in enterprise for this sort of thing. If one is commonly used and wants a shorthand can make a local variable. Really don't want new cases to be added and silently fall through the agent code path here leading to a lot of painful debugging.

test_util/installer_api_test.py
```diff
@@ -78,6 +79,14 @@ def get_hashed_password(self, password):
         passwd_hash = p.decode('utf-8').split('\n')[-2]
         return passwd_hash
 
+    @staticmethod
+    def ip_detect_script(preset_name):
+        try:
+            with open(pkg_resources.resource_filename('gen', 'ip-detect/{}.sh'.format(preset_name))) as f:
```
cmaloney: `pkg_resources.resource_string()` to save the file open / read stuff?

test_util/test_upgrade_vpc.py
```diff
@@ -0,0 +1,62 @@
...
+log = logging.getLogger(__name__)
+
+
+def main():
+    num_masters = 3
+    num_agents = 2
+    num_public_agents = 1
+    stack_name = 'upgrade-test-' + ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(10))
+
+    stable_installer_url = os.environ['STABLE_INSTALLER_URL']
```
cmaloney: nit: could we generalize this to n step upgrades? (basically pass in a list of installer urls, first one is used to install, rest are used for upgrades?) Would help a lot for testing things like 1.7-open -> 1.8.3 -> 1.8.4 -> 1.9 upgrade path testing (Although that probably also would require a list of different config.yaml).

cmaloney: Definitely fine to make that "out of scope" for now.

test_util/test_upgrade_vpc.py
```diff
@@ -0,0 +1,62 @@
...
+    installer_url = os.environ['INSTALLER_URL']
+
+    vpc, ssh_info = test_util.aws.VpcCfStack.create(
```
cmaloney: nit: a little obnoxious this code needs to be largely replicated between this and test-vpc, but works for now. Hopefully before too much longer the bigger refactors will come in and fix this.

test_util/cluster.py
```diff
@@ -92,7 +97,7 @@ def from_hosts(
         num_agents,
         num_public_agents,
     ):
-        assert all(x > 0 for x in [num_masters, num_agents, num_public_agents]), (
+        assert all(x >= 0 for x in [num_masters, num_agents, num_public_agents]), (
```
branden: It was, and that's how I discovered this faulty assertion. But once I fixed this, I discovered my own assertion in `run_integration_tests()` that at least one public agent is required for tests. I added a public agent but figured I'd leave this fix here.

test_util/cluster.py
branden: We may not need to. Sanity/safety I suppose, since this upgrade piggybacks on an install process that assumes DC/OS isn't installed.

test_util/cluster.py
branden: IIRC it'd complain about a lingering port that hadn't been cleaned up, maybe Exhibitor.

test_util/cluster.py
```diff
@@ -184,6 +202,86 @@ def check_ssh(self, hosts):
...
+    def mesos_metrics_snapshot(self, host):
+        """Return a snapshot of the Mesos metrics for host."""
+        if host in self.masters:
+            port = 5050
+        else:
+            port = 5051
+
+        with self.ssher.tunnel(host) as tunnel:
+            return json.loads(
+                tunnel.remote_cmd(
```
branden: Not all clusters have a bootstrap host, so if it's all the same then this is method is more general.

test_util/cluster.py
```diff
@@ -184,6 +202,86 @@ def check_ssh(self, hosts):
...
+        """Return the mode of the ZooKeeper instance on host."""
+        with self.ssher.tunnel(host) as tunnel:
+            stat_out = tunnel.remote_cmd([
+                'echo', 'stat', '|', '/opt/mesosphere/bin/toybox', 'nc', 'localhost', '2181'
+            ])
+        for line in stat_out.split(b'\n'):
+            line = line.strip()
+            if line.startswith(b'Mode: '):
+                mode = line.split(b':')[1].strip()
+                if mode not in (b'leader', b'follower', b'standalone'):
```
branden: It is, but implicitly. This is a more direct assertion that'll fail earlier and with a clearer message. I wanted this function to be responsible for retrieving (and validating) the mode of a single host, and `master_zk_modes()` to be responsible for retrieving (and validating) the modes of the entire cluster. Host IPs are included in the error message in both places.

branden:
@cmaloney Comments addressed, PTAL.

test_util/cluster.py
```diff
@@ -184,6 +202,85 @@ def check_ssh(self, hosts):
...
+                continue
+            mode = message[1].strip()
+            if mode not in (b'leader', b'follower', b'standalone'):
+                raise Exception('Unexpected ZooKeeper mode {} on host {}'.format(mode, host))
+            return mode.decode('utf-8')
+        raise Exception('ZooKeeper mode not found on host {}'.format(host))
+
+    def master_zk_modes(self):
+        """Return (master, zk_mode(master)) for each master."""
+        master_modes = [(master, self.zk_mode(master)) for master in self.masters]
```
cmaloney: Comment not for fixing now

Actually discovered we have some similar code in test_endpoint (https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_endpoints.py#L58) about checking for status, as well as in bootstrap for exhibitor waiting (https://github.com/dcos/dcos/blob/master/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py#L67)

Eventually we should probably standardize / normalize those three down to one.

PR #657, opened 2016-09-05
Ensure Spartan running when running watchdog

sargun:
This change ensures that Spartan itself is actually running before trying the watchdog.

### DIFF
--- a/packages/spartan/extra/dcos-spartan-watchdog.sh
+++ b/packages/spartan/extra/dcos-spartan-watchdog.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+LOCKFILE=/opt/mesosphere/active/spartan/spartan/bin/spartan-env
+# If the file is not locked
+if ! flock -n $LOCKFILE true; then
+  if ! /opt/mesosphere/active/toybox/bin/toybox timeout -k 1m 1m /opt/mesosphere/active/toybox/bin/toybox host ready.spartan 198.51.100.1; then
+    /opt/mesosphere/active/toybox/bin/toybox pkill -l 9 -f spartan
+  fi
+else
+  echo "Not running watchdog, Spartan not running"
+fi
--- a/packages/spartan/extra/dcos-spartan.service
+++ b/packages/spartan/extra/dcos-spartan.service
@@ -19,6 +19,6 @@ ExecStartPre=/usr/bin/env ip link set spartan up
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.1/32 dev spartan
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.2/32 dev spartan
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.3/32 dev spartan
-ExecStart=/opt/mesosphere/active/spartan/spartan/bin/spartan-env foreground
+ExecStart=/usr/bin/env flock -x /opt/mesosphere/active/spartan/spartan/bin/spartan-env /opt/mesosphere/active/spartan/spartan/bin/spartan-env foreground
 Environment=HOME=/opt/mesosphere
 
--- a/packages/spartan/build
+++ b/packages/spartan/build
@@ -29,6 +29,10 @@ gen_resolvconf="$PKG_PATH/bin/gen_resolvconf.py"
 cp /pkg/extra/gen_resolvconf.py "$gen_resolvconf"
 chmod +x "$gen_resolvconf"
 
+watchdog_sh="$PKG_PATH/bin/dcos-spartan-watchdog.sh"
+cp /pkg/extra/dcos-spartan-watchdog.sh "$watchdog_sh"
+chmod +x "$watchdog_sh"
+
 ### Spartan watchdog
 watchdog_service="$PKG_PATH/dcos.target.wants/dcos-spartan-watchdog.service"
 mkdir -p "$(dirname "$watchdog_service")"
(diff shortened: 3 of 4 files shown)


sargun:
@spahl 

@lingmann This should fix the issue with the Azure failures due to the watchdog failures.

packages/spartan/extra/dcos-spartan.service
```diff
@@ -19,6 +19,6 @@ ExecStartPre=/usr/bin/env ip link set spartan up
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.1/32 dev spartan
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.2/32 dev spartan
 ExecStartPre=-/usr/bin/env ip addr add 198.51.100.3/32 dev spartan
-ExecStart=/opt/mesosphere/active/spartan/spartan/bin/spartan-env foreground
+ExecStart=/usr/bin/env flock -x /opt/mesosphere/active/spartan/spartan/bin/spartan-env /opt/mesosphere/active/spartan/spartan/bin/spartan-env foreground
```
spahl: Can we make the flock an execstartpre instead? Would be a bit easier to read

sargun: No, the flock needs to hold while Spartan is running.

sargun: The other thing (other than flock) we can do is script around systemd, and ask systemd is spartan is up before running the watchdog. 

We can actually change the dcos-spartan-watchdog.service, and just move the flock check into ExecStartPre, but not for dcos-spartan.

lingmann: @sargun can you give us a timeline on what is happening around this `spartan-env` file and how this fixes the watchdog startup issue? This flock approach implies that something is trying to write to this file, and nothing should be writing to `/opt/mesosphere/`...

What does your approach to fix this with dcos-spartan-watchdog and an ExecStarPre step look like? Fixing the issue there seems to make more sense to me intuitively.

cmaloney: I This ExecStart is definitely beyond https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md. It also makes it so logs and the like aren't going to show up under the right binary name.

packages/spartan/extra/dcos-spartan-watchdog.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
cmaloney: we have python, why can't we use that here?

sargun: I can rewrite it in Python, but the Python equivalent is significantly longer, and a bit more complicated IMHO (you have to do the DNS resolution yourself, fcntl in Python isn't super intuitive as an API)

packages/spartan/extra/dcos-spartan.service
sargun: Nothing is going to write to spartan-env. The purpose of the flock is to signal that Spartan is running. 

The difference of exec start Pre, would be instead of:
```
ExecStart=/opt/mesosphere/bin/dcos-spartan-watchdog.sh
```

it'd be:

```
ExecStartPre=/bin/bash -c "if flock -n /opt/mesosphere/active/spartan/spartan/bin/spartan-env; then exit 1; fi"
ExecStart=/opt/mesosphere/active/toybox/bin/toybox timeout -k 1m 1m /bin/bash -c "/opt/mesosphere/active/toybox/bin/toybox host ready.spartan 198.51.100.1 || /opt/mesosphere/active/toybox/bin/toybox pkill -l 9 -f spartan"		
```

packages/spartan/extra/dcos-spartan-watchdog.sh
cmaloney: We already have a dns lib that we use in https://github.com/dcos/dcos/blob/master/packages/spartan/extra/gen_resolvconf.py

Actually, this and that seem like they're just entry points to the same "check if spartan is up" code.

packages/spartan/extra/dcos-spartan.service
sargun: @cmaloney I could have a one-liner which is a wrapper for spartan-env -- would that be preferrable?

sargun:
@lingmann Since we don't have After, Requires, and Wants, we need some way to tell systemd to ONLY trigger the watchdog AFTER Spartan is running. Are there any other suggestions for signaling?

cmaloney:
What I mentioned in the JIRA, that the core of this is we aren't waiting until hosts are really "up" to report them as up. The adminrouter-reload service does the same thing.

sargun:
The other thing we can do:
```
LINES=$(systemctl list-units --state=running --no-wall --no-legend --no-pager -n 1 dcos-spartan.service|wc -l)
if [ lines -eq 0 ]; then exit 1;fi 
```

sargun:
(as a pre exec for dcos-spartan-watchdog)

cmaloney:
Or just wait for all units in dcos.target to be "healthy" before we report a machine as up. Seems simpler.

sargun:
What do you think, @lingmann? I like @cmaloney's idea.

packages/spartan/extra/dcos-spartan.service
alberts: @sargun turns out you can also set SyslogIdentifier= to fix this, fwiw

PR #661, opened 2016-09-06
Set ssh bootstrap url default

cmaloney:


cmaloney:
Can't fix until `--genconf` for ssh is different than `--genconf` for advanced install

PR #663, opened 2016-09-06
Add test_util.cluster.Cluster

branden:
This is a series of refactors and fixups in support of #648. Depends on #622.

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -9,6 +9,7 @@ logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
 Host = namedtuple('Host', ['private_ip', 'public_ip'])
+SshInfo = namedtuple('SshInfo', ['user', 'home_dir'])
 
 VPC_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-cluster-template.json'
 VPC_EBS_ONLY_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-ebs-only-cluster-template.json'
@@ -87,7 +88,7 @@ class DcosCfSimple(CfStack):
             'PublicSlaveInstanceCount': str(public_agents),
             'SlaveInstanceCount': str(private_agents)}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_id, boto_wrapper)
+        return cls(stack.stack.stack_id, boto_wrapper), SSH_INFO['coreos']
 
     def delete(self):
         logger.info('Starting deletion of CF stack')
@@ -179,7 +180,7 @@ class DcosCfAdvanced(CfStack):
             'PrivateAgentInstanceType': private_agent_type,
             'PrivateSubnet': private_subnet}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_id, boto_wrapper)
+        return cls(stack.stack.stack_id, boto_wrapper), SSH_INFO['coreos']
 
     def delete(self, delete_vpc=False):
         logger.info('Starting deletion of CF Advanced stack')
@@ -255,7 +256,7 @@ class VpcCfStack(CfStack):
             'InstanceType': str(instance_type),
             'AmiCode': ami_code}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_id, boto_wrapper)
+        return cls(stack.stack.stack_id, boto_wrapper), OS_SSH_INFO[instance_os]
 
     def delete(self):
         # boto stacks become unusable after deletion (e.g. status/info checks) if name-based
@@ -268,6 +269,40 @@ class VpcCfStack(CfStack):
         logging.debug('Reservations for {}: {}'.format(self.stack.stack_id, reservations))
         instances = reservations[0]['Instances']
         return instances_to_hosts(instances)
+
+
+SSH_INFO = {
+    'centos': SshInfo(
+        user='centos',
+        home_dir='/home/centos',
+    ),
+    'coreos': SshInfo(
+        user='core',
+        home_dir='/home/core',
+    ),
+    'debian': SshInfo(
+        user='admin',
+        home_dir='/home/admin',
+    ),
+    'rhel': SshInfo(
+        user='ec2-user',
+        home_dir='/home/ec2-user',
+    ),
+    'ubuntu': SshInfo(
+        user='ubuntu',
+        home_dir='/home/ubuntu',
+    ),
+}
+
+
+OS_SSH_INFO = {
+    'cent-os-7': SSH_INFO['centos'],
+    'cent-os-7-dcos-prereqs': SSH_INFO['centos'],
+    'coreos': SSH_INFO['coreos'],
+    'debian-8': SSH_INFO['debian'],
+    'rhel-7': SSH_INFO['rhel'],
+    'ubuntu-16-04': SSH_INFO['ubuntu'],
+}
 
 
 OS_AMIS = {
--- a/test_util/cluster.py
+++ b/test_util/cluster.py
@@ -1,0 +1,304 @@
+import itertools
+import logging
+import os
+import stat
+from contextlib import contextmanager
+
+import passlib.hash
+import pkg_resources
+from retrying import retry
+
+import gen.calc
+import test_util.installer_api_test
+import test_util.test_runner
+from ssh.ssh_tunnel import SSHTunnel, TunnelCollection
+
+
+class Ssher:
+
+    def __init__(self, user, home_dir, key_path):
+        self.user = user
+        self.home_dir = home_dir
+        self.key_path = key_path
+
+        try:
+            os.chmod(self.key_path, stat.S_IREAD | stat.S_IWRITE)
+        except FileNotFoundError as exc:
+            raise Exception('SSH key not found at path {}'.format(self.key_path)) from exc
+        except OSError as exc:
+            raise Exception('Unable to set permissions on SSH key at {}: {}'.format(
+                self.key_path,
+                exc,
+            )) from exc
+
+    def tunnel(self, host):
+        return SSHTunnel(self.user, self.key_path, host.public_ip)
+
+    @contextmanager
+    def tunnels(self, hosts):
+        hostports = [host.public_ip + ':22' for host in hosts]
+        with TunnelCollection(self.user, self.key_path, hostports) as tunnel_collection:
+            yield tunnel_collection.tunnels
+
+    def remote_cmd(self, hosts, cmd):
+        with self.tunnels(hosts) as tunnels:
+            for tunnel in tunnels:
+                tunnel.remote_cmd(cmd)
+
+
+class Cluster:
+
+    def __init__(
+        self,
+        ssh_info,
+        ssh_key_path,
+        provider,
+        masters,
+        agents,
+        public_agents,
+        bootstrap_host=None,
+    ):
+        # Assert that this is a valid cluster.
+        gen.calc.validate_num_masters(str(len(masters)))
+
+        self.ssher = Ssher(ssh_info.user, ssh_info.home_dir, ssh_key_path)
+        self.provider = provider
+        self.masters = masters
+        self.agents = agents
+        self.public_agents = public_agents
+        self.bootstrap_host = bootstrap_host
+
+        assert all(h.private_ip for h in self.hosts), (
+            'All cluster hosts require a private IP. hosts: {}'.format(repr(self.hosts))
+        )
+
+        if self.provider == 'onprem':
+            self.check_ssh(self.hosts)
+        else:
+            self.check_ssh(self.masters)
+
+        logging.info('Bootstrap host: ' + repr(self.bootstrap_host))
+        logging.info('Masters: ' + repr(self.masters))
+        logging.info('Agents: ' + repr(self.agents))
+        logging.info('Public agents: ' + repr(self.public_agents))
+
+    @classmethod
+    def from_hosts(
+        cls,
+        ssh_info,
+        ssh_key_path,
+        hosts,
+        num_masters,
+        num_agents,
+        num_public_agents,
+    ):
+        assert all(x > 0 for x in [num_masters, num_agents, num_public_agents]), (
+            'num_masters, num_agents, and num_public_agents must be 0 or greater. '
+            'num_masters: {num_masters}, num_agents: {num_agents}, num_public_agents: {num_public_agents}'.format(
+                num_masters=num_masters,
+                num_agents=num_agents,
+                num_public_agents=num_public_agents,
+            )
+        )
+        assert len(hosts) == num_masters + num_agents + num_public_agents + 1, (
+            'Number of hosts must be equal to sum of masters and agents plus the bootstrap host.'
+            'num_masters: {num_masters}, num_agents: {num_agents}, num_public_agents: {num_public_agents}, '
+            'hosts: {hosts}'.format(
+                num_masters=num_masters,
+                num_agents=num_agents,
+                num_public_agents=num_public_agents,
+                hosts=repr(hosts),
+            )
+        )
+
+        bootstrap_host, masters, agents, public_agents = (
+            cls.partition_cluster(hosts, num_masters, num_agents, num_public_agents)
+        )
+
+        return cls(
+            ssh_info=ssh_info,
+            ssh_key_path=ssh_key_path,
+            provider='onprem',
+            masters=masters,
+            agents=agents,
+            public_agents=public_agents,
+            bootstrap_host=bootstrap_host,
+        )
+
+    @classmethod
+    def from_vpc(
+        cls,
+        vpc,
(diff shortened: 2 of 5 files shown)


cmaloney:
You me and @mellenburg should probably have a meeting with this. These wrappers are in the process of being rewritten / replaced: https://github.com/dcos/dcos/pull/240

mellenburg:
:+1: This layer of abstraction just covers the essential scripting of getting the installer to work automatically and won't cause terrible conflicts. I don't see any friction between this and the provider API changes.

branden:
Great, thanks for taking a look @mellenburg. LMK of course if you run into any problems trying to merge your changes in #240 (and related) with this. You might end up flushing out some bad design decisions that I didn't consider closely enough.

@cmaloney PTAL

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+import pkg_resources
+from retrying import retry
+
+import test_util.ccm
+import test_util.installer_api_test
+import test_util.test_runner
+from ssh.ssh_tunnel import SSHTunnel, TunnelCollection
+
+
+Host = namedtuple('Host', ['public_ip', 'private_ip'])
```
cmaloney: note to self / reminder when #240 and this both land we should get rid of this being both here and there.

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
+import itertools
```
cmaloney: rename this `ccm_vpc` or something similar?

test_util/cluster.py
mellenburg: I'll be sure to include it when I update https://github.com/dcos/dcos/pull/622 after this has merged

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+        return cluster
+
+    def cleanup(self):
+        if self.vpc is not None:
+            self.vpc.delete()
+
+    @staticmethod
+    def partition_cluster(hosts, num_masters, num_agents, num_public_agents):
+        """Return (bootstrap, masters, agents, public_agents) from hosts."""
+        assert num_masters in (1, 3, 5)
```
cmaloney: `gen.calc.validate_num_masters(str(num_masters))`

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+import test_util.test_runner
+from ssh.ssh_tunnel import SSHTunnel, TunnelCollection
+
+
+Host = namedtuple('Host', ['public_ip', 'private_ip'])
+
+
+class TestCluster:
+
+    def __init__(
```
cmaloney: This interface seems very strongly bound / non-generalizable beyond the "CCM VPC" type clusters, would be nice for it to work on any list of Host() objects provided, with a Vpc object which generalizes the "delete" type interface.

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+            public_ips,
+            ssh_user,
+            ssh_key_path,
+            remote_dir,
+            num_masters,
+            num_agents,
+            num_public_agents,
+            setup=True,
+    ):
+        self.vpc = None
```
cmaloney: why not have people pass in teh vpc. The VPC object should implement a basic interface:
1. get_hosts()
2. get_ssh_info()
3. delete()

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+    def cleanup(self):
+        if self.vpc is not None:
+            self.vpc.delete()
+
+    @staticmethod
+    def partition_cluster(hosts, num_masters, num_agents, num_public_agents):
+        """Return (bootstrap, masters, agents, public_agents) from hosts."""
+        assert num_masters in (1, 3, 5)
+        assert all(x >= 0 for x in [num_agents, num_public_agents])
+        assert num_agents + num_public_agents > 0
```
cmaloney: the requirements hard coded into the tests are currently
1. num_agents >= 2
2. num_public_agents >= 1

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+        assert len(hosts) == 1 + num_masters + num_agents + num_public_agents
```
cmaloney: This is asserting the set of hosts has a bootstrap host (the `+1`), can we comment that?

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+
+        hosts_iter = iter(hosts)
+        return (
+            next(hosts_iter),
+            list(itertools.islice(hosts_iter, num_masters)),
+            list(itertools.islice(hosts_iter, num_agents)),
+            list(itertools.islice(hosts_iter, num_public_agents)),
+        )
+
+    def get_private_address(self, tunnel):
```
cmaloney: I'd very much prefer this lives on the CCM VPC type object / doesn't live in the `TestCluster` type thing to separate the responsibilities / concerns.

That said, an integration test that `ip-detect` returns the private ips we expect, and `ip_detect_public` would be nice

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+            for tunnel in tunnels.tunnels:
+                private_ip[tunnel.host] = self.get_private_address(tunnel)
+                if setup:
+                    # Make the default user priveleged to use docker
+                    tunnel.remote_cmd(['sudo', 'usermod', '-aG', 'docker', self.ssh_user])
+
+        return [Host(public_ip, private_ip[public_ip]) for public_ip in public_ips]
+
+    @contextmanager
+    def host_tunnel(self, host):
```
cmaloney: why not just update SSHTunnel to have __enter__  + __exit__ since this is the only way we ever use it?

test_util/cluster.py
```diff
@@ -0,0 +1,260 @@
...
+            if setup:
+                # only do on setup so you can rerun this test against a living installer
+                logging.info('Verifying installer password hashing')
+                test_pass = 'testpassword'
+                hash_passwd = installer.get_hashed_password(test_pass)
+                assert passlib.hash.sha512_crypt.verify(test_pass, hash_passwd), 'Hash does not match password'
+                if api:
+                    installer.start_web_server()
+
+            with open(pkg_resources.resource_filename("gen", "ip-detect/aws.sh")) as ip_detect_fh:
```
cmaloney: nit: can be a future refactor, but we're going to want to pass in the ip_detect_script + config dictionary to use when installing DC/OS

cmaloney:
Overall looking good, #240 is landing in the next train, I'd like to if we're formalizing up the "Cluster" interface vs. the "CCM VPC" interface a bit more try and solidify that (There's a VPC class which has a certain assumed API, including things like the "get Host() objects", "delete()") so that this is drawing that line more clearly. Also makes it easier to do things like #310 (that just provides an alternate Vpc implementation)

test_util/cluster.py
branden: I'm hoping to make this abstraction provider-agnostic.

test_util/cluster.py
branden: I updated the interface -- the caller is now responsible for partitioning hosts by role and passing a callback for cleanup. There are then a couple convenience methods `from_vpc()` and `from_cloudformation()` that will take either a VPC or CF object and instantiate a `TestCluster` from them.

branden:
@cmaloney I'm still ushering this through the vagrant test, but it's ready for another look.

test_util/aws.py
```diff
@@ -156,6 +163,43 @@ def get_vpc_host_ips(self):
...
+OS_SSH_INFO = {
+    'cent-os-6': SSH_INFO['centos'],
+    'cent-os-7': SSH_INFO['centos'],
+    'cent-os-7-dcos-prereqs': SSH_INFO['centos'],
+    'coreos': SSH_INFO['coreos'],
+    'debian-8': SSH_INFO['debian'],
+    'rhel-6': SSH_INFO['rhel'],
+    'rhel-7': SSH_INFO['rhel'],
+    'ubuntu-12-04': SSH_INFO['ubuntu'],
+    'ubuntu-14-04': SSH_INFO['ubuntu'],
```
branden: I'll remove the extra keys here when I rebase onto master after #622 is merged.

test_util/azure_test_driver.py
```diff
@@ -163,7 +162,7 @@ def poll_deploy():
 
         print('Detected IP configuration: {}'.format(ip_buckets))
 
-        with closing(SSHTunnel(get_value('linuxAdminUsername'), 'ssh_key', master_lb, port=2200)) as t:
+        with SSHTunnel(get_value('linuxAdminUsername'), 'ssh_key', master_lb, port=2200) as t:
```
cmaloney: can we pull the closing() -> __enter__  + __exit__ into a separate very simple quick to land PR?

test_util/test_aws_vpc.py
```diff
@@ -156,8 +134,10 @@ def check_environment():
 def main():
     options = check_environment()
 
-    host_list = None
-    vpc = None  # Set if the test owns the VPC
+    cluster = None
+    num_masters = 1
```
cmaloney: Can we make these all configurable?

test_util/aws.py
```diff
@@ -64,9 +64,12 @@ def create_stack(self, name, template_url, user_parameters, deploy_timeout=60):
 
 
 class CfStack():
-    def __init__(self, stack_name, boto_wrapper):
+    def __init__(self, stack_name, boto_wrapper, ssh_user=None, ssh_home_dir=None):
```
cmaloney: "ssh access" stuff seems like it doesn't really belong in the CfStack class / seems like a separate concern.

It also doesn't make any sense for things like the SingleMaster AWS templates to set these on the CfStack.

test_util/test_runner.py
```diff
@@ -51,10 +49,7 @@ def integration_test(
         'SLAVE_HOSTS=' + ','.join(agent_list),
         'PUBLIC_SLAVE_HOSTS=' + ','.join(public_agent_list),
         'DCOS_PROVIDER=' + provider,
-        'DNS_SEARCH=' + dns_search,
-        'AWS_ACCESS_KEY_ID=' + aws_access_key_id,
```
mellenburg: note: Resiliency tests are just going to add these right back in anyway, so it doesnt really matter if you remove them

branden: Cool, I'll pull that commit out then.

test_util/aws.py
branden: I tried to put SSH info as close as possible to the operating system, since that's what determines it. To figure out SSH details you'd have to interrogate the stack object to get its OS, so I figured it might as well answer the question directly.

cmaloney:
```
SshCredentials = namedtuple('SshCredentials', ['ssh_user', 'ssh_key_path'])
cluster, credentials = CcmVpc.create()
stack, credentials = CfStack.create(os='foobar')
```

test_util/aws.py
```diff
@@ -67,6 +68,7 @@ class CfStack():
     def __init__(self, stack_name, boto_wrapper):
         self.boto_wrapper = boto_wrapper
         self.stack = self.boto_wrapper.resource('cloudformation').Stack(stack_name)
+        self.wait_for_stack_creation()
```
cmaloney: this shouldn't implicitly wait for stack creation, we may be connecting to an already created stack

branden:
@cmaloney Ready for review.

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+    def __init__(
+        self,
+        ssh_info,
+        ssh_key_path,
+        provider,
+        masters,
+        agents,
+        public_agents,
+        bootstrap_host=None,
+        cleanup_callback=None,
```
cmaloney: I'm really not a fan of cleanup_callback / this doesn't seem like this class should be owning "destruct some arbitrary thing when you don't actually own any resources yourself"

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+    ):
+        # Assert that this is a valid cluster.
+        gen.calc.validate_num_masters(str(len(masters)))
+        assert all(len(x) >= 0 for x in [agents, public_agents])
```
cmaloney: nit: This will produce really terrible error messages when it fails unfortunately, need a comment string which has agents, public_agents lists in it.

cmaloney: if we want to match the in-test assertions, public_agents >= 1, agents >= 2

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+        assert any(len(x) > 0 for x in [agents, public_agents])
+
+        self.ssher = Ssher(ssh_info.user, ssh_info.home_dir, ssh_key_path)
+        self.provider = provider
+        self.masters = masters
+        self.agents = agents
+        self.public_agents = public_agents
+        self.bootstrap_host = bootstrap_host
+
+        assert all(h.private_ip for h in self.hosts), 'All cluster hosts require a private IP'
```
cmaloney: include the host list in the error message

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+    install_prereqs_only=False,
+):
+    if cluster.provider != 'onprem':
+        raise NotImplementedError('Install is only supported for onprem clusters')
+
+    assert cluster.bootstrap_host is not None, 'Install requires a bootstrap host'
+    assert all(h.public_ip for h in cluster.hosts), 'Install requires that all cluster hosts are externally reachable'
+
+    if api:
+        installer = test_util.installer_api_test.DcosApiInstaller()
```
cmaloney: nit / not yours as I'm passing by this. Bot the web installer and ssh installer have the same "offline". Web it's called "offline mode", ssh it's called "install_prereqs=False". We should clean that up at some point (not this pr)

test_util/cluster.py
cmaloney: Actually, since the integration tests are a separate thing, this shouldn' actually check either agents or public_aggents have anything. Having 0 public agents is fine.

test_util/test_runner.py
```diff
@@ -17,7 +17,8 @@
 def integration_test(
         tunnel, test_dir,
         dcos_dns, master_list, agent_list, public_agent_list,
-        test_dns_search, provider,
+        provider,
```
cmaloney: Certain we got this re-ordering throuhg all the call sites?

test_util/test_aws_vpc.py
```diff
@@ -172,150 +162,68 @@ def main():
...
             stack_name=unique_cluster_id,
             instance_type=options.instance_type,
             instance_os=os_name,
-            instance_count=5,  # 1 bootstrap, 1 master, 2 agents, 1 public agent
+            # An instance for each cluster node plus the bootstrap.
+            instance_count=(options.masters + options.agents + options.public_agents + 1),
             admin_location='0.0.0.0/0',
             key_pair_name='default',
             boto_wrapper=bw)
         vpc.wait_for_stack_creation()
```
cmaloney: rather than all the delete stuff, may make sense to effectively say
`with vpc` and the __enter__ does a `wait_for_stack_creation()`, the __exit__ does a delete if some special env var isn't set which prevents that.

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+                self.key_path,
+                exc,
+            )) from exc
+
+    def tunnel(self, address):
+        return SSHTunnel(self.user, self.key_path, address)
+
+    @contextmanager
+    def tunnels(self, addresses):
+        hostports = [a + ':22' for a in addresses]
```
branden: `TunnelCollection` expects it in case `SSHTunnel` wants it: https://github.com/dcos/dcos/blob/38be502429c5ea93bdd4a8bc7a1916a66f786609/ssh/ssh_tunnel.py#L115 I'm not crazy about that API, but I'm hoping to limit this PR's spread to yet another module.

test_util/test_runner.py
branden: All call sites pass all their args by keyword.

test_util/cluster.py
```diff
@@ -0,0 +1,292 @@
...
+
+        self.ssher = Ssher(ssh_info.user, ssh_info.home_dir, ssh_key_path)
+        self.provider = provider
+        self.masters = masters
+        self.agents = agents
+        self.public_agents = public_agents
+        self.bootstrap_host = bootstrap_host
+
+        assert all(h.private_ip for h in self.hosts), 'All cluster hosts require a private IP'
+        assert all(h.public_ip for h in self.masters), 'All masters must be externally reachable'
```
mellenburg: Is this really a requirement? I ask because it would be really nice to merge conftest.py/Cluster with this class eventually, that way you have one consistent interface across all the DCOS code base for providing a cluster and interacting with it. As it stands right now, when intergration tests are run, PUBLIC_MASTER_LIST == MASTER_LIST in most cases

branden: I'll change this to assert at least one master with a public IP and move it to `run_integration_tests()`, which is the code that relies on it.

test_util/cluster.py
cmaloney: SGTM

test_util/test_aws_vpc.py
branden: I went ahead and put the existing logic back. I'll defer to another PR to improve cluster resource management.

cmaloney:
🚢 . Please squash into one big commit and then we'll land it

PR #667, opened 2016-09-06
Lock down distributed Erlang port usage

sargun:
This PR limits disterl to using the ports that are mentioned in the PR doc.


sargun:
@spahl

docs/admin-ports.md
```diff
@@ -13,6 +13,11 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 62053: dcos-spartan
  - 62080: dcos-navstar
 
+#### Disterl ports:
```
cmaloney: These are already in the list above, and this is user-facing / visible, where people won't know / be interested in "disterl ports" vs. "general ports"

cmaloney:
:ship:

PR #674, opened 2016-09-07
Cleanup containers

aaronjwood:
Figured this would be good to have in place so that we don't leave 200+ exited containers lying around after each build.

### DIFF
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -2,7 +2,9 @@ import copy
 import hashlib
 import json
 import os.path
+import random
 import shutil
+import string
 import tempfile
 from os import mkdir
 from os.path import exists
@@ -37,8 +39,14 @@ class DockerCmd:
         self.environment = dict()
         self.container = str()
 
-    def run(self, cmd):
-        docker = ["docker", "run"]
+    def run(self, name, cmd):
+        docker = ["docker", "run", "--name={}".format(
+            "{}-{}".format(
+                name, ''.join(
+                    random.choice(string.ascii_lowercase) for _ in range(10)
+                )
+            )
+        )]
         for host_path, container_path in self.volumes.items():
             docker += ["-v", "{0}:{1}".format(host_path, container_path)]
 
@@ -48,6 +56,12 @@ class DockerCmd:
         docker.append(self.container)
         docker += cmd
         check_call(docker)
+        DockerCmd.clean(name)
+
+    @staticmethod
+    def clean(name):
+        """Cleans up the specified container"""
+        check_call(["docker", "rm", "-v", name])
 
 
 def get_variants_from_filesystem(directory, extension):
@@ -1053,7 +1067,7 @@ def build(package_store, name, variant, clean_after_build, recursive=False):
             package_store.get_package_cache_folder(name): "/pkg/:rw",
         }
         cmd.container = "ubuntu:14.04.4"
-        cmd.run(["rm", "-rf", "/pkg/src", "/pkg/result"])
+        cmd.run("package-cleaner", ["rm", "-rf", "/pkg/src", "/pkg/result"])
 
     clean()
 
@@ -1154,7 +1168,7 @@ def build(package_store, name, variant, clean_after_build, recursive=False):
         # TODO(cmaloney): Run a wrapper which sources
         # /opt/mesosphere/environment then runs a build. Also should fix
         # ownership of /opt/mesosphere/packages/{pkg_id} post build.
-        cmd.run([
+        cmd.run("package-builder", [
             "/bin/bash",
             "-o", "nounset",
             "-o", "pipefail",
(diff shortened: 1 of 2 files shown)


pkgpanda/build/__init__.py
```diff
@@ -38,7 +38,7 @@ def __init__(self):
         self.container = str()
 
     def run(self, cmd):
-        docker = ["docker", "run"]
+        docker = ["docker", "run", "--rm"]
```
cmaloney: I want to only clean this one up if the build was successful (It's handy for people to be able to re-enter the build container when things fail so they can experiment)

aaronjwood: Sure, just changed how it works. LMK what you think.

pkgpanda/build/__init__.py
```diff
@@ -48,6 +48,13 @@ def run(self, cmd):
         docker.append(self.container)
         docker += cmd
         check_call(docker)
+        DockerCmd.clean()
+
+    @staticmethod
+    def clean():
+        """Removes the last container that was spawned"""
+        container_id = check_output(["docker", "ps", "--latest", "-q"]).rstrip()
+        check_call(["docker", "rm", "-v", container_id])
```
aaronjwood: @cmaloney would you want to do a `-f` here as well?

cmaloney:
In the integration test we run which uses the `-d` flow it's giving:
```[12:23:59]Conflicting options: --rm and -d```

File: pkgpanda/build/__init__.py [hunk 3]
```diff
@@ -48,6 +48,13 @@ def run(self, cmd):
         docker.append(self.container)
         docker += cmd
         check_call(docker)
+        DockerCmd.clean()
+
+    @staticmethod
+    def clean():
+        """Removes the last container that was spawned"""
+        container_id = check_output(["docker", "ps", "--latest", "-q"]).rstrip()
```
cmaloney: This seems very likely to potentially accidentally pick up and kill the wrong docker container. The value of auto-cleaning here really to me isn't worth a chance of cleaning up some random unrelated docker.

aaronjwood: Another idea on how to clean this up would be to make sure all `docker run`s provide a very specific name for the container they're launching. Then we can just remove that container by it's name afterwards. What are your thoughts?

aaronjwood:
Thanks, I'll fix that issue.

pkgpanda/build/__init__.py
```diff
@@ -37,8 +37,8 @@ def __init__(self):
         self.environment = dict()
         self.container = str()
 
-    def run(self, cmd):
-        docker = ["docker", "run"]
+    def run(self, name, cmd):
+        docker = ["docker", "run", "--name={}".format(name)]
```
cmaloney: conflicting names result in non-launching containers, so after the first failure this will fail to ever launch again without manual intervention, could we add a bit of a random name to the end of this to make it so by default run w/ fail -> run just works?

aaronjwood: Good catch, just added some randomness to the end of each name.

cmaloney:
Looks good! Can we squash the commits down and then I'll land it?

Only small nit is if every time we add the "random" to the end of name by hand, seems like we should put the random bit inside the `.run()` function, but either way works fine and I see arguments for both APIs.

aaronjwood:
I like your idea of moving the random generation inside of `run()` since it reduces code duplication a bit. I'll squash these down now too.

aaronjwood:
What's the build failure I'm getting? Looks like there's something wrong with the SHA1 hash of the output from those tests...?

pkgpanda/build/__init__.py
```diff
@@ -37,8 +39,14 @@ def __init__(self):
         self.environment = dict()
         self.container = str()
 
-    def run(self, cmd):
-        docker = ["docker", "run"]
+    def run(self, name, cmd):
+        docker = ["docker", "run", "--name={}".format(
```
cmaloney: `"--name={}-{}".format(name, ''.join(random.choice(string.ascii_lowercase) for _ in range(10)` as another way to write this?

pkgpanda/build/__init__.py
```diff
@@ -48,6 +56,12 @@ def run(self, cmd):
         docker.append(self.container)
         docker += cmd
         check_call(docker)
+        DockerCmd.clean(name)
```
cmaloney: Here you're passing the original / base name, not the new name with random, which is leading to:
https://teamcity.mesosphere.io/viewLog.html?buildId=395451&buildTypeId=DcosIo_Dcos_BuildPr&tab=buildLog#_focus=635

aaronjwood: My fault, fixed.

aaronjwood:
Do you want me to do another squash?

cmaloney:
One or two commits I'm fine with, just when it gets more than that for a small semantic change PR then like squashing

cmaloney:
:ship:. I'll land it with the next train I make which will be tomorrow

PR #675, opened 2016-09-07
Add context manager for marathon apps

mellenburg:
allows cleaner deploy->checks->cleanup

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -3,6 +3,7 @@ import copy
 import logging
 import os
 import uuid
+from contextlib import contextmanager
 from urllib.parse import urlparse, urlunparse
 
 import dns.exception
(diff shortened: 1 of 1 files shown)


orsenthil:
LGTM.

mellenburg:
@cmaloney This is ready for review. Long overdue change that can be incorporated as the tests that deploy apps are touched

cmaloney:
@mellenburg https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_networking.py#L28 doesn't look updated currently

mellenburg:
@cmaloney What I was trying to communicate above is that this is to just add the method and when other PRs touch a given test, they should do this cleanup. However, if you would prefer a larg-ish cleanup that catches the cases where no one added the delete, I could do that as well

cmaloney:
We can make a separate PR from this one that does the cleanup, but adding new infrastructure without porting the (not that many) existing cases tends to result in people seeing the old tests, copy/pasting, editing slightly. In PR review that generally doesn't get caught unless the reviewer happens to remember "oh yea, there was a new infrastructure for this" Being in the state of having both needs to be minimized.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -376,31 +377,29 @@ def get_test_app_in_docker(self, ip_per_container):
...
-        r = requests.get('http://{}:{}/test_uuid'.format(service_points[0].host,
-                                                         service_points[0].port))
-        if r.status_code != 200:
-            msg = "Test server replied with non-200 reply: '{0} {1}. "
-            msg += "Detailed explanation of the problem: {2}"
-            pytest.fail(msg.format(r.status_code, r.reason, r.text))
-
-        r_data = r.json()
+        with self.marathon_deploy_and_cleanup(app) as service_points:
+            r = requests.get('http://{}:{}/test_uuid'.format(service_points[0].host,
```
cmaloney: nit / note for future cleanup: Some sort of requests wrapper where I could do a `.get('/test_uuid') would be cool for this.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -376,31 +377,29 @@ def get_test_app_in_docker(self, ip_per_container):
...
+                                                             service_points[0].port))
+            if r.status_code != 200:
+                msg = "Test server replied with non-200 reply: '{0} {1}. "
```
cmaloney: nit: Two string literals in a row get auto-concatenated in python so why not just:
```
 msg = "Test server replied with non-200 reply: '{0} {1}. " \
    "Detailed explanation of the problem: {2}"
```

Also note you can go out to 120 characters (https://github.com/dcos/dcos/blob/master/tox.ini#L2) to not have to wrap this sort of thing as much.

cmaloney:
:ship: , some nits / ideas for the future. Definitely looking for a future PR to move the other apps we launch to the new shiny infrastructure and remove the old infrastructure from being a public API.

PR #678, opened 2016-09-08
Remove `After=` lines that don't do what people want / expect.

cmaloney:
see [Systemd Rules](https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md)

When we have restarting services, Before= and After= don't really mean anything.

That these don't match expectations can be seen by the needing to add `ping` checks for the various services / DNS resolutions because the services came up before what they were `After`


cmaloney:
:ship:  in next train

PR #679, opened 2016-09-08
Make dcos-spartan-watchdog start later rather than have a random 60 second sleep

cmaloney:
cc: @sargun

### DIFF
--- a/packages/spartan/extra/dcos-spartan-watchdog.service
+++ b/packages/spartan/extra/dcos-spartan-watchdog.service
@@ -9,5 +9,4 @@ EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
 EnvironmentFile=/opt/mesosphere/etc/dns_search_config
 EnvironmentFile=-/opt/mesosphere/etc/dns_config_master
-ExecStartPre=/bin/sleep 60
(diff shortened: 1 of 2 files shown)


sargun:
No, the purpose of this is to prevent the two systemd units from starting at the same time, and on slow systems this means that you get into this awkward restart loop where both systemd units keep starting at the same time. And the watchdog resets spartan, etc..

cmaloney:
@sargun Please tell me how `sleep 60` is any different from telling the timer to start / fire 60 seconds later? I fail to see how there is any difference there....

sargun:
Will Systemd always keep them on different intervals? Systemd has a habit of starting things together, and this is the behaviour I'm trying to avoid.

cmaloney:
Having the sleep inside the unit vs. the delay on the outside doesn't seem like it makes any difference in there, esp. since the `After=` line is being removed as well so there's no direct link between the two

cmaloney:
we could change the fire / restart loop interval to make sure they aren't on the same timing cycle potentially, but I'm pretty sure `dcos-spartan` is already on a 5 second restart loop (https://github.com/dcos/dcos/blob/master/packages/spartan/extra/dcos-spartan.service#L8) while this is a 5 minute loop (https://github.com/dcos/dcos/blob/master/packages/spartan/extra/dcos-spartan-watchdog.timer#L5)

cmaloney:
Lowered fire period to 2 minutes, if spartan which is just a DNS proxy takes > 2 minutes to initialize, much of our other software probably has issues

PR #680, opened 2016-09-08
Allow changing the bootstrap download folder from `/tmp`

cmaloney:
Fixes: https://mesosphere.atlassian.net/browse/DCOS-6484 ('/tmp' is sometimes too small to contain the bootstrap tarball by allowing people to use a custom folder.


cmaloney:
:ship:  in next train

PR #681, opened 2016-09-08
Add support for configuring a proxy via config.yaml

cmaloney:
Sample config.yaml snippet:
```
use_proxy: true
http_proxy: <your_http_proxy>
https_proxy: <your_https_proxy>
```

All but .mesos domains will be proxied

Users can add more domains to the `no_proxy` list by setting `no_proxy` to a list of things to exclude in their config.yaml

TODO:
 - [x] Allow adding more to no_proxy
 - [ ] Integration test?
 - [ ] Allow arbitrary extension of the config / setting extra proxy variables?

Fixes: https://mesosphere.atlassian.net/browse/DCOS-6619
Makes progress towards: https://dcosjira.atlassian.net/browse/DCOS-34

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -296,6 +296,11 @@ def validate_cluster_packages(cluster_packages):
             raise AssertionError(str(ex)) from ex
 
 
+def calculate_no_proxy(no_proxy):
+    user_proxy_config = validate_json_list(no_proxy)
+    return ",".join(['*.mesos,127.0.0.1,localhost'] + user_proxy_config)
+
+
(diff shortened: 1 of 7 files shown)


trevorgetty:
the proxy settings work correctly on the actual DC/OS installation, but it fails to setup or configure the required Proxy information for the NIX type installation packages.
E.g. yum fails to install due to no http proxy setup, on the first m1 machine, before it ever reaches the DC/OS installation.  
Should it be using the vagrant-proxyconf plugin to setup the NIX proxy information as part of the vagrant.yaml file?

PR #682, opened 2016-09-08
Move timeouts from 5 minutes to 2 minutes

cmaloney:
Based on ~1k test runs, all of these finish in 20-40 seconds if they're going to succeed. This makes it so they fail a bit faster when they're going to.

cc: @mellenburg


mellenburg:
:+1:  Yes, I had a feeling that after a minute my app was probably doomed but this settles it

cmaloney:
:ship: in next train

PR #683, opened 2016-09-08
Add spotify/docker-gc to help cleanup left around dockers on agents

cmaloney:
To Enable/Disable in config.yaml:
```
enable_docker_gc: true
```

Disabled by default (landing late, no integration test). See https://github.com/spotify/docker-gc for documentation on configuration, the `/etc/` configuration folders mentioned should work. It is
run hourly in DC/OS.

Fixes internal bug: https://mesosphere.atlassian.net/browse/DCOS-7618

# TODO
 - [ ] Integration test triggering dcos-docker-gc.service and making sure some docker containers are cleaned up.

### DIFF
--- a/gen/aws/calc.py
+++ b/gen/aws/calc.py
@@ -17,7 +17,8 @@ entry = {
         'num_private_slaves': '5',
         'num_public_slaves': '1',
         'os_type': '',
-        'aws_masters_have_public_ip': 'true'
+        'aws_masters_have_public_ip': 'true',
+        'enable_docker_gc': 'true'
     },
     'must': {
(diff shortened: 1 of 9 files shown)


cmaloney:
Needs an update to test_compisition.test_signal_service to include dcos-docker-gc-service, dcos-docker-gc-timer

cmaloney:
This has issues because the once an hour job sometimes fails on the first run meaning that it's not healthy everywhere.

amitaekbote:
@cmaloney ready for review

cmaloney:
nit: should really add an integration test that the docker-gc is working at all, but this works for now as is / solves a pain point.

PR #687, opened 2016-09-09
Resiliency test

mellenburg:
Add first resiliency test. This test will:
1. Start 2 apps using VIPs
2. Ping the two apps
3. kill all nodes
4. wait for autoscaling group to repopulate the nodes
5. reset the agent lists
6. ping the 2 apps

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -24,6 +24,7 @@ LOGIN_PW = os.getenv('DCOS_LOGIN_PW')
 def pytest_configure(config):
     config.addinivalue_line('markers', 'first: run test before all not marked first')
     config.addinivalue_line('markers', 'last: run test after all not marked last')
+    config.addinivalue_line('markers', 'resiliency: Run tests that cause critical failures')
 
 
 def pytest_collection_modifyitems(session, config, items):
@@ -39,6 +40,31 @@ def pytest_collection_modifyitems(session, config, items):
         else:
             new_items.append(item)
     items[:] = new_items + last_items
+
+
+def pytest_addoption(parser):
+    parser.addoption('--resiliency', action='store_true')
+
+
+def pytest_runtest_setup(item):
+    if item.get_marker('resiliency'):
+        if not item.config.getoption('--resiliency'):
+            pytest.skip('Test requires --resiliency option')
+
+
+@pytest.yield_fixture
+def vip_apps(cluster):
+    vip1 = '6.6.6.1:6661'
+    test_app1, _ = cluster.get_test_app()
+    test_app1['portDefinitions'][0]['labels'] = {
+        'VIP_0': vip1}
+    test_app2, _ = cluster.get_test_app()
+    test_app2['portDefinitions'][0]['labels'] = {
+        'VIP_0': 'foobarbaz:5432'}
+    vip2 = 'foobarbaz.marathon.l4lb.thisdcos.directory:5432'
+    with cluster.marathon_deploy_and_cleanup(test_app1):
+        with cluster.marathon_deploy_and_cleanup(test_app2):
+            yield ((test_app1, vip1), (test_app2, vip2))
 
 
 @pytest.fixture(scope='session')
@@ -72,6 +98,7 @@ def _setup_logging():
     """Setup logging for the script"""
     logging.basicConfig(format='[%(asctime)s] %(levelname)s: %(message)s', level=LOG_LEVEL)
     logging.getLogger("requests").setLevel(logging.WARNING)
+    logging.getLogger("botocore").setLevel(logging.WARNING)
 
 
 class Cluster:
@@ -176,7 +203,10 @@ class Cluster:
         assert r.status_code == 200
 
         data = r.json()
-        slaves_ids = sorted(x['id'] for x in data['slaves'])
+        # only check against the slaves we expect to be in the cluster
+        # so we can check that cluster has returned after a failure
+        # in which case will will have new slaves and dead slaves
+        slaves_ids = sorted(x['id'] for x in data['slaves'] if x['hostname'] in self.all_slaves)
 
         for slave_id in slaves_ids:
             # AdminRouter's slave endpoint internally uses cached Mesos
@@ -205,7 +235,7 @@ class Cluster:
             return False
         assert r.status_code == 200
 
-    def _wait_for_dcos(self):
+    def wait_for_dcos(self):
         self._wait_for_leader_election()
         self._wait_for_adminrouter_up()
         self._authenticate()
@@ -278,7 +308,7 @@ class Cluster:
         # Make URI never end with /
         self.dcos_uri = dcos_uri.rstrip('/')
 
-        self._wait_for_dcos()
+        self.wait_for_dcos()
 
     @staticmethod
     def _marathon_req_headers():
--- a/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
+++ b/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
@@ -1,0 +1,58 @@
+import os
+
+import pytest
+
+import test_util.aws
+import test_util.helpers
+
+
+@pytest.fixture(scope='session')
+def dcos_launchpad(cluster):
+    """Interface for direct interation to dcos_launchpad hardware
+    Currently only supports AWS CF with AWS VPC coming soon
+    """
+    if cluster.provider != 'aws':
+        # TODO(mellenburg): update when advanced templates are merged
+        pytest.skip('Must be AWS CF to run test')
+    aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
+    aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
+    aws_region = os.environ['AWS_REGION']
+    stack_name = os.environ['AWS_STACK_NAME']
+    bw = test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
+    return test_util.aws.DcosCfSimple(stack_name, bw)
+
+
+@pytest.mark.last
+@pytest.mark.resiliency
+def test_agent_failure(dcos_launchpad, cluster, vip_apps):
+    # make sure the app works before starting
+    test_util.helpers.wait_for_pong(vip_apps[0][1], 120)
+    test_util.helpers.wait_for_pong(vip_apps[1][1], 10)
+
+    agents = [i['InstanceId'] for i in dcos_launchpad.
+              get_group_instances(['PublicSlaveServerGroup', 'SlaveServerGroup'], state_list=['running'])]
+
+    # Agents are in autoscaling groups, so they will automatically be replaced
+    dcos_launchpad.boto_wrapper.client('ec2').terminate_instances(InstanceIds=agents)
+    dcos_launchpad.boto_wrapper.client('ec2').get_waiter('instance_terminated').wait(InstanceIds=agents)
+
+    def get_running_agents():
+        return dcos_launchpad.get_group_instances(
+            ['PublicSlaveServerGroup', 'SlaveServerGroup'], state_list=['running'])
+
+    # Wait for replacements
+    test_util.helpers.wait_for_len(get_running_agents, len(cluster.all_slaves), 600)
+
+    # Reset the cluster to have the replacement agents
+    cluster.slaves = sorted([agent.private_ip for agent in
+                             dcos_launchpad.get_private_agent_ips(state_list=['running'])])
+    cluster.public_slaves = sorted([agent.private_ip for agent in
+                                    dcos_launchpad.get_public_agent_ips(state_list=['running'])])
+    cluster.all_slaves = sorted(cluster.slaves + cluster.public_slaves)
+
+    # verify that everything else is still working
+    cluster.wait_for_dcos()
+    # finally verify that the app is again running somewhere with its VIPs
+    # Give marathon five minutes to deploy both the apps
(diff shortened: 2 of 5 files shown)


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -176,7 +205,7 @@ def _wait_for_srouter_slaves_endpoints(self):
         assert r.status_code == 200
 
         data = r.json()
-        slaves_ids = sorted(x['id'] for x in data['slaves'])
+        slaves_ids = sorted(x['id'] for x in data['slaves'] if x['hostname'] in self.all_slaves)
```
mellenburg: Add comment explaining dead slaves

mellenburg:

sargun:
@mellenburg Mind bumping the Minuteman hash to dac990fa535c5d55e8fa373be90bb6e7c2afafa2 (current master)

mellenburg:
@sargun agents pass resiliency with minuteman bump

sargun:
👍 Can we merge?

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -41,6 +42,33 @@ def pytest_collection_modifyitems(session, config, items):
...
+def vip_apps(cluster):
+    vip1 = '6.6.6.1:6661'
+    test_app1, _ = cluster.get_test_app()
+    test_app1['portDefinitions'][0]['labels'] = {
+        'VIP_0': vip1}
+    test_app2, _ = cluster.get_test_app()
+    test_app2['portDefinitions'][0]['labels'] = {
+        'VIP_0': 'foobarbaz:5432'}
+    vip2 = 'foobarbaz.marathon.l4lb.thisdcos.directory:5432'
+    cluster.deploy_marathon_app(test_app1)
```
cmaloney: /me will be happy when our base marathon app launcher API is a context object.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -0,0 +1,75 @@
...
+import retrying
+import test_util.aws
+
+
+@pytest.fixture(scope='session')
+def provider(cluster):
+    """Interface for direct interation to provider hardware
+    Currently only supports AWS CF with AWS VPC coming soon
+    """
+    if cluster.provider != 'aws':
```
cmaloney: nit: this technically should be "launch method" or something similar we pass through probably. the provider is the same for simple + advanced AWS templates.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -0,0 +1,75 @@
+import logging
```
cmaloney: is there a way to say "this folder is resiliency tests" rather than having to individually mark any of them with badness when we forget?

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -0,0 +1,75 @@
...
+    aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
+    aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
+    aws_region = os.environ['AWS_REGION']
+    stack_name = os.environ['AWS_STACK_NAME']
+    bw = test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
+    return test_util.aws.DcosCfSimple(stack_name, bw)
+
+
+@retrying.retry(wait_fixed=3000, stop_max_delay=300 * 1000)
+def wait_for_persistent_app(url):
```
cmaloney: this seems like generic conftest / cluster stuff

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -0,0 +1,75 @@
...
+    logging.info('Attempting to ping test application')
+    r = requests.get('http://{}/ping'.format(url))
+    assert r.ok, 'Bad response from test server: ' + str(r.status_code)
+    assert r.json() == {"pong": True}, 'Unexpected response from server: ' + repr(r.json())
+
+
+@retrying.retry(wait_fixed=3000, stop_max_delay=600 * 1000,
+                retry_on_result=lambda res: res is False,
+                retry_on_exception=lambda ex: False)
+def wait_for_group_population(provider, group, target):
```
cmaloney: this seems like generic conftest / cluster stuff which should be shared by this. (_wait_for_up)

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.mark.last
+@pytest.mark.resiliency
+def test_agent_failure(provider, cluster, vip_apps):
+    # make sure the app works before starting
+    wait_for_persistent_app(vip_apps[0][1])
+    wait_for_persistent_app(vip_apps[1][1])
+
+    agents = [i['InstanceId'] for i in provider.
```
cmaloney: provider.public_agents(), provider.private_agents()

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
cmaloney: How about:
```

def wait_for_match(fetch_fn, target_count):
    items = fetch_fn()
    count = len(items)
    if count == target:
        return False

    logging.info("Waiting for {} to equal {}. Current items: {}".format(count, target, items))
```

cmaloney: I think that makes this more generic / more of a general purpose helper we can re-use.

sargun:
Can we make sure this gets merged before 1.8?

packages/dcos-integration-test/extra/conftest.py
mellenburg: I'll can this with:
with marathon_deploy_and_cleanup:
    with marathon_deploy_and_cleanup:
        yield

no different to me.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Good point.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: The only way I could think of doing that without mucking with pytest internal hooks is to drop this all in a test class. I could do that, but Im not a huge fan of test classes

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Agreed. Will pot to conftest

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: How about a DcosCfSimple method that just does kill_all_agents?

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Yup. The other fixture already made it over there anyway

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Alternatively, we could also just use a fixture that skips if some env var is set which sets something in Cluster, which doesnt seem much different, but its how we mark auth tests for skipping security disabled tests

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Actually, there is no cluster specific code in here and its not a fixture so it doesnt belong in conftest, but what I could do is make it a function returned from the fixture which has the sole purpose of checking connectivity to these two apps, but that is somewhat convoluted when we could have a simple fixture and a simple function instead, right?

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Same thing with the above point. I was forgetting that conftest only works for fixtures, so if you drop this in there you have to hack it into a fixture which makes it ugly. Thus, this is a part of the argument of moving all the function bits out of conftest and into a module in test_util (which we can do starting with this)

mellenburg:
@cmaloney small refactor, PTAL

test_util/helpers.py
```diff
@@ -0,0 +1,33 @@
...
+    """Will call fetch_fn, get len() on the result and repeat until it is
+    equal to target count or timeout (in seconds) has been reached
+    """
+    @retrying.retry(wait_fixed=3000, stop_max_delay=timeout * 1000,
+                    retry_on_result=lambda res: res is False,
+                    retry_on_exception=lambda ex: False)
+    def check_for_match():
+        items = fetch_fn()
+        count = len(items)
+        logging.info('Waiting for len({}) to return {}. Current: {}'.format(fetch_fn.__name__, target_count, count))
```
cmaloney: can we include the items? helps debug "waiting for n agents"

test_util/aws.py
```diff
@@ -108,24 +109,30 @@ def delete(self):
...
+        instances = self.get_group_instances(['PublicSlaveServerGroup'], state_list)
         return instances_to_hosts(instances)
 
-    def get_private_agent_ips(self):
-        instances = self.get_tag_instances('SlaveServerGroup')
+    def get_private_agent_ips(self, state_list=None):
+        instances = self.get_group_instances(['SlaveServerGroup'], state_list)
         return instances_to_hosts(instances)
 
+    def get_group_instances(self, group_list, state_list=['running']):
```
cmaloney: Should ideally only default state_list in the functions which will call this (get_private_agent_ips, etc). or in the function being called. It's odd to always pass it in  from all the call sites here as one value, then only in one place consistently use the "other" default (which could just be embedded in get_running_agents probably).

test_util/aws.py
```diff
@@ -87,7 +88,7 @@ def create(cls, stack_name, template_url, public_agents, private_agents,
             'PublicSlaveInstanceCount': str(public_agents),
             'SlaveInstanceCount': str(private_agents)}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_id, boto_wrapper)
+        return cls(stack.stack.stack_name, boto_wrapper)
```
cmaloney: why stack_name? stack_id seems always better

packages/minuteman/buildinfo.json
```diff
@@ -4,7 +4,7 @@
     "minuteman": {
       "kind": "git",
       "git": "https://github.com/dcos/minuteman.git",
-      "ref": "cb84f5bd37c98d39a25a55e26f0d3b8564b43b65",
+      "ref": "dac990fa535c5d55e8fa373be90bb6e7c2afafa2",
```
cmaloney: nit: would have been nice to land this in a separate PR

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -41,6 +42,31 @@ def pytest_collection_modifyitems(session, config, items):
     items[:] = new_items + last_items
 
 
+def pytest_addoption(parser):
+    parser.addoption('--resiliency', action='store_true')
```
cmaloney: /me really wishes we could do this as a test "suite" rather than doing a special marker here. Really easy to forget the marker

test_util/aws.py
mellenburg: This is from Boto being stupid: when you name a stack from stack_name, you can then call stack.stack_name and you get the StackName or you can call stack.stack_id and get the physical resource ID. If you create it from the physical resource ID, then stack.stack_name returns the ID and not the name. So, when I want to find all the stacks the most sensible filter is: 
```
Filter=[{'Name': 'tag:aws:cloudformation:stack-name', 'Values': [self.stack.stack_name]}]
```
which breaks when stack.stack_name returns an ID.

Suggestion? This was the most sensible approach I could find other than setting the stack_name by describing the stack_id and getting the tag (which seemed excessive)

cmaloney: Suggestion: Add a comment about that. That is definitely fun crazy boto behavior (although I can guess where the implementation bug is...)

test_util/helpers.py
mellenburg: :+1:

mellenburg:
@sargun Although this seemed pretty solid after the update to minuteman, there might be some flakiness...
https://teamcity.mesosphere.io/viewLog.html?buildId=402490&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsDe_2

sargun:
You appear to have pulled the Minuteman hash bump out of this PR?

sargun:
This PR needs to land before GA.

sargun:
Scratch that actually, I'm going to make a separate PR for just Minuteman.

mellenburg:
@cody can this :ship: set sail?

cmaloney:
@sargun : https://github.com/dcos/dcos/pull/719

mellenburg:
@cmaloney This has been rebased on advanced templates and is ready for another pass (included some teency cleanup changes that should have been with advanced templates)

mellenburg:
@cmaloney @spahl This is been ready for a while now without additional comment. Can we please merge before bitrot?

spahl:
:ship:

PR #688, opened 2016-09-09
Add pytest ordering plugin

mellenburg:
Allows having a test that just establishes a cluster
is up. So, if the cluster fails to come up, the user
will see waiting for the test_cluster_is_up and not
some arbitrarily chosen test

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -21,6 +21,26 @@ LOGIN_UNAME = os.getenv('DCOS_LOGIN_UNAME')
 LOGIN_PW = os.getenv('DCOS_LOGIN_PW')
 
 
+def pytest_configure(config):
+    config.addinivalue_line('markers', 'first: run test before all not marked first')
+    config.addinivalue_line('markers', 'last: run test after all not marked last')
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """Reorders test using order mark
+    """
+    new_items = []
+    last_items = []
+    for item in items:
+        if hasattr(item.obj, 'first'):
+            new_items.insert(0, item)
+        elif hasattr(item.obj, 'last'):
+            last_items.append(item)
+        else:
+            new_items.append(item)
+    items[:] = new_items + last_items
+
+
 @pytest.fixture(scope='session')
 def cluster():
     assert 'DCOS_DNS_ADDRESS' in os.environ
--- a/packages/dcos-integration-test/extra/test_composition.py
+++ b/packages/dcos-integration-test/extra/test_composition.py
@@ -8,6 +8,11 @@ import pytest
 import requests
 
 
+@pytest.mark.first
+def test_cluster_is_up(cluster):
+    pass
+
+
 def test_if_all_mesos_masters_have_registered(cluster):
     # Currently it is not possible to extract this information through Mesos'es
     # API, let's query zookeeper directly.


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -21,6 +21,32 @@
...
+    for item in items:
+        if hasattr(item.obj, 'order'):
+            order_num = getattr(item.obj, 'order').args[0]
+            if order_num > 0:
+                first.append((order_num, item))
+            elif order_num < 0:
+                last.append((order_num, item))
+        else:
+            unordered.append(item)
+            continue
```
darkonie: nit: 
```
if not hasattr(item.obj, 'order'):
   unordered.append(item)
   continue

ordered_num = getattr(...
```

mellenburg: Yes, much better. Thanks!

mellenburg:

cmaloney:
:ship:

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -21,6 +21,31 @@
...
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """Reorders test using order mark
+    """
+    first = []
+    unordered = []
+    last = []
+    for item in items:
+        if not hasattr(item.obj, 'order'):
```
cmaloney: nit / thinking of other ways to write this:
```
orderings = list()
for item in items:
    order = getattr(item.obj, 'order', 0)
    assert isinstance(order, int)
    collection.insert((order, item))
return [i[1] for i in sort(orderings, key=lambda i: i[0])]
```

cmaloney:
Manually checked / verified first test run is always the new "Test_cluster_is_up" -- https://teamcity.mesosphere.io/viewLog.html?buildId=395922&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp&tab=buildLog#_state=37&focus=502

packages/dcos-integration-test/extra/conftest.py
mellenburg: This implementation doesnt account for negatives (index from end) and it will put unordered tests first. Once you try to do those two cases, I think we end up with what is already here

cmaloney: just reversing using either reverse or making the key func negate i[0] will cause us to go [3, 2, 1, 0, -1, -2, -3]. I _think_ that is the ordering we want (Destructor order should exactly reverse constructor order based on programming language history to date).

if the `[1, 2, 3, 0, -1, -2, -3]` is wanted, that's just another key function. If we can guarantee size, we could do bucketing more simply, but I don't think that is the right order for this sort of thing.

mellenburg: > if the [1, 2, 3, 0, -1, -2, -3] is wanted
What I want is [1,2,3,-3,-2,-1] so that order(1) clearly stays the first item and order(-1) clearly stays the last item. If we go order [3,2,1,0,-1,-2,-3] and then some other test must come second my test_cluster_is_up is test must be changed to be order(2) and the new test order(1) (need to edit the entire stack of orders to guarantee timing) or I would call my new test order(0), and then we are locked into using negatives thereafter and not to mention its a little awkward. Furthermore, my order(-1) test will always be at least tied for the end. with [3,2,1,0,-1,-2,-3], I need to set something like order(-sys.maxint - 1)

cmaloney: We can do [1, 2, 3, 0, -3, -2, -1] if you want 1 to always be first. They just need to be opposite orderings, if I put an initalizer in 2, it should run before 3, and if I put a finalizer in -2 for that same thing, it should come after -3. The "three" level always comes before or always comes after.

```
assert all(lambda x: x < 1000, items)
sorted([-2, 1, 3, -1, 2, 0, -3], key= lambda x: 1000-x)
```
gives the `[3, 2, 1, 0, -1, -2, 3]` ordering

mellenburg:
@cmaloney same gist, more specific implementation, PTAL

cmaloney:
:ship: assuming the first thing comes first in all the tests we run

mellenburg:
Confirmed: all runs had test_cluster_is_up as the first test

PR #689, opened 2016-09-09
Allow setting SSH key path and cluster in CF test

mellenburg:


cmaloney:
:ship: 

Still more iterating to do here probably in cleanup, but definitely getting better.

PR #692, opened 2016-09-09
exhibitor: check existance of realm file in /opt instead of in /run

jgehrcke:
- start_exhibitor.py: check existance of realm file in /opt instead of in /run
- run ExecStartPre as root and chown realm file to user dcos_exhibitor

Note that /usr/bin/chown is a standardized location that should exist on all target platforms, and that it dereferences symlinks by default.

### DIFF
--- a/packages/exhibitor/extra/dcos-exhibitor.service
+++ b/packages/exhibitor/extra/dcos-exhibitor.service
@@ -7,11 +7,11 @@ StandardError=journal
 Restart=always
 StartLimitInterval=0
 RestartSec=5
+# Run in new mount namespace to create custom resolv.conf.
 MountFlags=private
 RuntimeDirectory=dcos_exhibitor
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
 EnvironmentFile=/opt/mesosphere/etc/exhibitor
 EnvironmentFile=-/opt/mesosphere/etc/exhibitor-extras
-# run in new mount namespace to create custom resolv.conf
 ExecStart=$PKG_PATH/usr/exhibitor/start_exhibitor.py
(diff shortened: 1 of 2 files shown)


cmaloney:
nak, the `exhibitor_realm` file should get a uname to make it group / other readable.

cmaloney:
https://github.com/dcos/dcos/blob/master/gen/__init__.py#L490 -- inside fo the `dcos-config.yaml` you can actually specify a 'permissions' key which we'll chmod the file to (defaults to 644)

jgehrcke:
We should make the exhibitor_realm file readable _only_ for the user dcos_exhibitor (it may contain a cleartext password). What is currently the best way to achieve that?

cmaloney:
Everything in gen/dcos-config.yaml becomes tarballs which any host in the cluster could download, secure credential download can't be achieved through it really... For 1.8 shipping the file on host is probably the easiest. To get proper credential store probably needs to end up in a `/run` directory

packages/exhibitor/extra/dcos-exhibitor.service
```diff
@@ -7,11 +7,14 @@ StandardError=journal
...
 MountFlags=private
+PermissionsStartOnly=true
 RuntimeDirectory=dcos_exhibitor
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
 EnvironmentFile=/opt/mesosphere/etc/exhibitor
 EnvironmentFile=-/opt/mesosphere/etc/exhibitor-extras
-# run in new mount namespace to create custom resolv.conf
+# Change owner of realm file and ignore exit code, because file my not exist.
+ExecStartPre=/bin/sh -c "/usr/bin/chown dcos_exhibitor /opt/mesosphere/etc/exhibitor_realm || true"
```
cmaloney: no, this doesn't add security.

jgehrcke: Agree. I have removed the chown and instead use more open file permissions downstream.

cmaloney:
:ship:

PR #694, opened 2016-09-10
De-duplicate test logging

mellenburg:
Right now we get double the messages:
[2016-09-09 23:56:24,681|root|INFO]: Response from marathon: {'portDefinitions': [{'port': 0, 'protocol
[2016-09-09 23:56:24,681] INFO: Response from marathon: {'portDefinitions': [{'port': 0, 'protocol': 'tcp',


cmaloney:
/me wonders if we should find a way to de-duplicate all the local logging basicConfig stuff....

That said, this looks good as is.

:ship:

PR #695, opened 2016-09-10
Replace boto waiter with simple loop

mellenburg:
When boto waiters fail they just say " botocore.exceptions.WaiterError: Waiter InstanceRunning failed: Waiter encountered a terminal failure state" which is effectively worthless. So, add a wrapper to catch that error and dump the stack event info. Regardless of which client is waiting, client('cloudformation').describe_stack_events() is probably the most relevant thing to dump upon bumping into a failure in time domain

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
 import logging
+import pprint
 from collections import namedtuple
 
 import boto3
+import retrying
 
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


mellenburg:

mellenburg:
Huh, this is a little less useful and boto waiter are a bit more enigmatic than I hoped... Better than nothing though.

test_util/aws.py
```diff
@@ -58,13 +59,22 @@ def __init__(self, stack_name, boto_wrapper):
...
+        under CloudFormation and EC2 client Waiter for common options
+
+        NOTE: Cache stack ID before starting so stacks that failed to create can be examined
+        """
+        stack_id = self.stack.stack_id
+        cf_client = self.boto_wrapper.client(client)
+        try:
+            cf_client.get_waiter(waiter).wait(**kwargs)
+        except botocore.exceptions.WaiterError:
+            stack_info = repr(self.boto_wrapper.client('cloudformation').describe_stack_events(StackName=stack_id))
```
cmaloney: doesn't this mean only `cloudformation` waiters are allowed / would work here? If so, why take the client as a parameter?

mellenburg:
Note: wrapping boto not very helpful as waiters are buggy. Rather than wrap the waiter it is more prudent to use a simple waiter that will raise exceptions we can understand. Back to WIP fo rnow

mellenburg:

test_util/aws.py
```diff
@@ -68,13 +70,53 @@ def __init__(self, stack_name, boto_wrapper):
         self.boto_wrapper = boto_wrapper
         self.stack = self.boto_wrapper.resource('cloudformation').Stack(stack_name)
 
+    def wait_for_status_change(self, state_1, state_2, timeout=3600):
+        """ Note: Do not use boto waiter class, it has very poor error handling
```
orsenthil: [nit]: No space between `"""` and `Note:`

test_util/aws.py
```diff
@@ -68,13 +70,53 @@ def __init__(self, stack_name, boto_wrapper):
...
+        @retrying.retry(wait_fixed=5000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda res: res is False,
+                        retry_on_exception=lambda ex: False)
+        def wait_loop():
+            stack_details = self.get_stack_details()
+            stack_status = stack_details['StackStatus']
+            if stack_status == state_2:
+                return True
+            if stack_status != state_1:
+                logging.exception('Stack Details:\n' + pprint.pformat(stack_details))
```
orsenthil: Might be better to use string format instead of `+` in lines 100, 101 and 102

`Stack Details:\n {details}.format(details=pprint.pformat(stack_details))`

test_util/aws.py
```diff
@@ -68,13 +70,53 @@ def __init__(self, stack_name, boto_wrapper):
...
+                logging.exception('Stack Events:\n' + pprint.pformat(self.get_stack_events()))
+                raise Exception('StackStatus changed unexpectedly to:' + stack_status)
+            return False
+
+        wait_loop()
```
orsenthil: My personal preference will be to use a private function instead of inner function.
* Like if this is refactored into

`self._wait_loop_status_change(state_1, state_2)`

* This will require `stop_max_delay` to be preset instead of sending as an argument.

It's your call on this comment.

test_util/aws.py
mellenburg: Better in what sense? I feel its a bit more readable to use ```+``` when I only have a string to append. If its a matter of style I am willing to conform to the consensus

test_util/aws.py
mellenburg: I'm also not a fan of the inner function, but I really like to be able to easily set my timeout for specific cases in test scripts. Thus, between using time.sleep and a retry inner-function, I thought the latter was a little more palatable

test_util/aws.py
mellenburg: Good eye

test_util/aws.py
orsenthil: Consistency.  For e.g. in Line 87 and Line 88, .format is used and here the concatenation.
Given the string interpolation operators available and kwargs support with  .format`, if we have to standardize on something `.format` will be more suitable.

test_util/aws.py
orsenthil: I agree.

mellenburg:
Rebased and addressed comments.

orsenthil:
LGTM.

mellenburg:
@cmaloney @spahl Can we get this in a train soon? Without this change, CI will continue to come to a crashing halt when we hit the AWS API rate limit

PR #696, opened 2016-09-11
Advanced template integration tests

mellenburg:


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -195,6 +195,9 @@ def _wait_for_dcos(self):
         self._wait_for_srouter_slaves_endpoints()
         self._wait_for_metronome()
 
+    @retrying.retry(wait_fixed=2000,
```
cmaloney: can we set an upper limit on this?

test_util/aws.py
```diff
@@ -46,7 +46,6 @@ def create_stack(self, name, template_url, user_parameters, deploy_timeout=60):
         self.resource('cloudformation').create_stack(
             StackName=name,
             TemplateURL=template_url,
-            DisableRollback=True,
```
cmaloney: don't we still want this?

mellenburg: Testing change. will revert

mellenburg:
Note to self: need to make VPC, gateway, pub subnet, and priv subnet in eu-central-1 for actual CI integration

mellenburg:
@cmaloney This is ready for review. The new test passed:
https://teamcity.mesosphere.io/viewLog.html?buildId=401673&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsAd_2&tab=buildLog

Note: I will try to squeeze in the ability to select instance OS into this PR, but I don't expect much to change from how it is now

packages/dcos-integration-test/extra/conftest.py
mellenburg: Sure, although none of the other wait functions here have an upper limit. But, it seems to be a corner case in advanced where auth isnt ready after leader election and adminrouter is.

What do you suggest? 5min? 2min?

cmaloney: I 'd say 5 minutes should be plenty. We only do this after mesos has a master and adminrouter is up, at which point things should be moving along.

mellenburg:
Note: I'm removing the "ensure cleanup" option as that was only needed when generating new networks on every run. By reusing those, we can just rely on CCM to kill off clusters from test runs that encountered exception (and hence don't need a cleanup guarantee in script)

test_util/test_aws_cf.py
```diff
@@ -55,6 +60,13 @@ def check_environment():
     # Defaults
     options.ci_flags = os.getenv('CI_FLAGS', '')
     options.aws_region = os.getenv('DEFAULT_AWS_REGION', 'eu-central-1')
+    options.ensure_dcos_cleanup = os.getenv('ENSURE_DCOS_CLEANUP', 'true') == 'true'
```
mellenburg: remove this

mellenburg:
This has now passed with both CoreOs and EL-7 (centos) templates. Just needs review

cmaloney:
note this and #687, only one can land in one train, next has to be a different pr (one or the other needs some tweaks for things to be all universally compatible).

test_util/aws.py
```diff
@@ -92,8 +92,29 @@ def create(cls, stack_name, template_url, public_agents, private_agents,
     def delete(self):
         logger.info('Starting deletion of CF stack')
         # boto stacks become unusable after deletion (e.g. status/info checks) if name-based
+        self.empty_and_delete_s3_bucket_from_stack()
```
cmaloney: doesn't the comment mean this should be after we change to an id based stack?

test_util/aws.py
```diff
@@ -108,22 +129,116 @@ def delete(self):
...
+               gateway=None, vpc=None, private_subnet=None, public_subnet=None):
+        ec2 = boto_wrapper.client('ec2')
+        if not vpc:
+            logging.info('Creating new VPC...')
+            vpc = ec2.create_vpc(CidrBlock=vpc_cidr, InstanceTenancy='default')['Vpc']['VpcId']
+            ec2.get_waiter('vpc_available').wait(VpcIds=[vpc])
+            ec2.create_tags(Resources=[vpc], Tags=[{'Key': 'Name', 'Value': stack_name}])
+        logging.info('Using VPC with ID: ' + vpc)
+
+        if not gateway:
```
cmaloney: nit: i'm not sure this is we want this API to work

cmaloney: esp. if we want the "cleanup" behaivor.

test_util/test_aws_cf.py
```diff
@@ -55,6 +60,13 @@ def check_environment():
...
+    options.advanced = os.getenv('DCOS_ADVANCED_TEMPLATE', 'false') == 'true'
```
cmaloney: can we just implicitly pick this up from the template name / url?

test_util/aws.py
mellenburg: Indeed it does

test_util/test_aws_cf.py
mellenburg: Yup. If it doesnt end in single-master.cloudformation.json or multi-master.cloudformation.json == advanced

test_util/aws.py
mellenburg: I agree that leaving this in is dubious without having a better cleanup method, but I feel leaving it in is a convenient addition (e.g. it took just a second for me to make the shared Vpc, gateway, subnet combo for eu-central-1). That said, its pretty trivial so if you insist on taking it out, I wont argue

cmaloney: my thing is just how the API is structured isn't how we'll want it when we have these things. That said, I don't have the better API right now. So nit, there's definitely a better way to do this in python and structurally as an API, not sure if it's worth it right now.

test_util/test_aws_cf.py
mellenburg: Oh, wait, nope, cant do that if we want to be able to attach to an already running advanced stack by stack name, so I guess this stays as is?

cmaloney: if <has template url> then infer otherwise require it as a separate parameter explicitly set?

test_util/aws.py
mellenburg: Yes, I would like to get this at least landed in downstream ASAP and iterate on it later

test_util/test_aws_cf.py
mellenburg: Sounds good

mellenburg:
@cmaloney I think this is ready to :ship:

test_util/aws.py
```diff
@@ -93,7 +93,28 @@ def delete(self):
         logger.info('Starting deletion of CF stack')
         # boto stacks become unusable after deletion (e.g. status/info checks) if name-based
         self.stack = self.boto_wrapper.resource('cloudformation').Stack(self.stack.stack_id)
+        self.empty_and_delete_s3_bucket_from_stack()
```
cmaloney: nit: should do this after self.stack.delete(), exhibitor could use the bucket while we're deleting it otherwise.

cmaloney:
:ship:

PR #697, opened 2016-09-11
Stabilize dcos-setup.service

mellenburg:
Many stacks fail to create in AWS because of
curl: (18) transfer closed with ___ bytes remaining to read
curl's default keepalive interval is 60s

From what I can tell, this error is generated when the server side cuts off the transfer so it would make sense to bump up keepalive-time. And it helped at least one person http://stackoverflow.com/questions/33522615/curl-transfer-closed-with-outstanding-read-data-remaining


cmaloney:
:ship:

PR #698, opened 2016-09-11
Add a `dcos-shell` program so dcos doesn't have to be added to the default `$PATH`

cmaloney:
The goal is this is the only thing that gets added to $PATH (Rather than the
python environment and everything else)

People can then use:
`dcos-shell` -- get a shell with all DC/OS bits in your environment
`dcos-shell pkgpanda --uninstall` -- run a pkgpanda command with all the right environment variables.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,7 +1,9 @@
 Documentation of changes which may break some install environments
 
+1.9-dev
+ - DC/OS binaries / config is no longer injected to user environment via /etc/profile.d/dcos.sh. New installs have just a single binary, `dcos-shell` added to $PATH. Can be used as `dcos-shell <cmd>` to run a single command or `dcos-shell` to start a new shell which has the environment variables sourced in which arbitrary commands can be run.
 
-1.8-dev
+1.8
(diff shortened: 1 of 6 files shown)


cmaloney:
holding for testing, which becomes easier once the upgrade test pr lands (can swap the sudo commands to use it instead of sudo -i).

cmaloney:
1.9 targeted

NEWS
```diff
@@ -1,7 +1,9 @@
 Documentation of changes which may break some install environments
 
+1.9-dev
+ - DC/OS binaries / config is no longer injected to user environment via /etc/profile.d/dcos.sh. New installs have just a single binary, `dcos-shell` added to $PATH. Can be used as `dcos-shell <cmd>` to run a single command or `dcos-shell` to start a new shell which has the environment variables sourced in which arbitrary commands can be run.
```
orsenthil: Minor change suggestion in the second part of the sentence.

```Users can use `dcos-shell <cmd>` to run a single command or `dcos-shell` to start a new shell which has the environment variables sourced.```

PR #701, opened 2016-09-12
Use /state instead of /slaves in tests.

alberts:
With some Mesos security configurations, /slaves is not available.

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -77,7 +77,7 @@ class Cluster:
                     retry_on_result=lambda ret: ret is False,
                     retry_on_exception=lambda x: False)
     def _wait_for_slaves_to_join(self):
-        r = self.get('/mesos/master/slaves')
+        r = self.get('/mesos/master/state')
         if r.status_code != 200:
             msg = "Mesos master returned status code {} != 200 "
             msg += "continuing to wait..."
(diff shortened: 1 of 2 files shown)


jgehrcke:
:+1:

alberts:
:ship:

cmaloney:
Why? We have several customers actively relying on slaves. It is one of the ways we shed load from /state, because querying that too often will tip over a cluster.

adam-mesos:
@cmaloney Apache Mesos does not provide the hooks (yet) for authorization/filtering on the /slaves endpoint, and it's not used by either the Mesos or DC/OS UIs. If we enable the /slaves endpoint, any authenticated user would be able to view all tasks, frameworks, etc. instead of seeing a filtered view. As such, we disable this endpoint (and a few other minor ones) in security=strict mode to prevent unauthorized access to that data. It is still enabled in disabled/permissive modes, so most customers won't be affected. In a future release, where we trust strict mode more, we can/will have filtering on /slaves too.
Let me find/file the Mesos JIRA for that.

cmaloney:
I need to ping support and see what the customer impact of this will be. Several customers have definitely read the Mesos endpoint docs and started using the various listed endpoints. Need to see what the customer impact of cutting this one (one of the more common) off.

cmaloney:
Booted out of 1.8 for now at least, can revive if we decide we need to do this.

PR #702, opened 2016-09-12
Build packages with the number of cores that the machine has

aaronjwood:
@cmaloney what do you think about applying this change to many of the other components as well?

### DIFF
--- a/packages/mesos/build
+++ b/packages/mesos/build
@@ -17,8 +17,7 @@ LIBS="-lssl -lcrypto" LDFLAGS="-L/opt/mesosphere/active/openssl/lib" "/pkg/src/m
   --with-libevent=/opt/mesosphere/active/libevent \
   --with-curl=/opt/mesosphere/active/curl \
   --sbindir="$PKG_PATH/bin"
-# TODO(cmaloney): -j8 is almost certainly wrong.
-make -j8
+make -j$NUM_CORES
 
 make install
 cp src/java/target/mesos-*.jar "$libdir"
--- a/packages/adminrouter/build
+++ b/packages/adminrouter/build
@@ -20,7 +20,7 @@ pushd "/pkg/src/nginx"
   --with-http_ssl_module \
   --with-luajit
 
-make -j8
+make -j$NUM_CORES
 make install
 popd
 
--- a/packages/curl/build
+++ b/packages/curl/build
@@ -41,6 +41,6 @@ pushd /pkg/src/curl
             --without-winidn \
             --without-winssl \
             --with-ca-bundle=/opt/mesosphere/active/python-requests/lib/python3.5/site-packages/requests/cacert.pem
-make -j8
+make -j$NUM_CORES
 make install
 popd
--- a/packages/erlang/build
+++ b/packages/erlang/build
@@ -7,7 +7,7 @@ export CXXFLAGS=CFLAGS
 
 pushd /pkg/src/erlang
 ./configure --prefix=$PKG_PATH --with-ssl=/opt/mesosphere/active/openssl --with-termcap=/opt/mesosphere/active/ncurses --enable-dirty-schedulers --disable-hipe --enable-kernel-poll
-make -j8
+make -j$NUM_CORES
 make install
 popd
 
--- a/packages/libsodium/build
+++ b/packages/libsodium/build
@@ -4,6 +4,6 @@ mkdir -p build
 pushd build
 
 cd /pkg/src/libsodium
-./configure --prefix="$PKG_PATH" 
-make -j8
+./configure --prefix="$PKG_PATH"
+make -j$NUM_CORES
 make install
--- a/packages/mesos-modules/build
+++ b/packages/mesos-modules/build
@@ -10,6 +10,6 @@ pushd build
 /pkg/src/mesos-modules/configure \
      --with-mesos=/opt/mesosphere/active/mesos \
      --prefix="$PKG_PATH"
-make -j8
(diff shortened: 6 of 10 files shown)


packages/mesos/build
```diff
@@ -17,8 +17,7 @@ LIBS="-lssl -lcrypto" LDFLAGS="-L/opt/mesosphere/active/openssl/lib" "/pkg/src/m
   --with-libevent=/opt/mesosphere/active/libevent \
   --with-curl=/opt/mesosphere/active/curl \
   --sbindir="$PKG_PATH/bin"
-# TODO(cmaloney): -j8 is almost certainly wrong.
-make -j8
```
orsenthil: Two comments.

1. `-j` without any args is supposed to utilize all the cpus.  It can be verified.
2. Utilizing all cpus is not error-free. only by benchmarking can we arrive a good value for `j`

cmaloney:
I'm thinking about this, not high priority compared to some of the other stuff I need to go over at the moment though. Definitely if we're going to do it, should do it to _all_ builds. That said, changing / manipulating make parallelism most definitely results in problems with some projects and their build systems.

packages/mesos/build
cmaloney: `-j` without arguments actually results in `-j<infinity>` which is more problematic typically (And makes things go slower than doing the "right" number of things.


From the man page:
Specifies the number of jobs (commands) to run simultaneously. If there is more than one -j option, the last one is effective. If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously.

aaronjwood: Sure, would love to open this up for more discussion. Just trying to make it so that we're not wasteful or too aggressive with resources depending on what kind of machine it's on.

aaronjwood:
I believe many of the components are built in parallel. My thought was to only change the components that are already doing parallel builds to match the number of CPUs. I'll be building this on machines with 40+ (logical) cores, at least at work. It'd be much more efficient in my case to take advantage of everything that's available :)

aaronjwood:
Here are the instances that I thought we could change:
```
packages/adminrouter/build:23:make -j8
packages/curl/build:44:make -j8
packages/erlang/build:10:make -j8
packages/libsodium/build:8:make -j8
packages/mesos/build:21:make -j8
packages/mesos-modules/build:13:make -j8
packages/mesos-overlay-modules/build:15:make -j8
packages/ncurses/build:8:make -j8
packages/python/build:11:make -j8
```

packages/mesos/build
aaronjwood: It looks like if you don't specify a number the amount of jobs is unbounded https://www.gnu.org/software/make/manual/html_node/Parallel.html

"If the ‘-j’ option is followed by an integer, this is the number of recipes to execute at once; this is called the number of job slots. If there is nothing looking like an integer after the ‘-j’ option, there is no limit on the number of job slots. The default number of job slots is one, which means serial execution (one thing at a time)."

cmaloney:
I definitely get the 40 core case, the problem is that, for instance
1. not all "small" docker containers we build in have `nproc` probably (I can imagine the go container in particular, alpine containers not having it). Although that one could be worked through case by case
2. Build systems have odd bugs with parallelism, and I want relatively strong guarantees around at least what we ship that we can reproduce it. 1, 8 tend to be "sweet" spots, 1 because it serializes things, 8 because that is the number of cores your average dev workstation has had for the last bit and so devs have typically run into the issues and fixed them.

cmaloney:
That said, this is definitely after-1.8-ships when I'll have time to review / go through it fully (so probably sometime next week at the earliest).

aaronjwood:
Good point about Alpine. Let me iterate on this a bit and change the other packages to use a dynamic number and see how the results are. I'll also look into a way of doing this without relying on `nproc`. Feel free to throw any feedback/comments my way when you have time!

cmaloney:
Rather than doing the bash magic everywhere, can we centralize it into: https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L1158

And there just add an env var "'BUILD_NUM_CORES", which equals `multiprocessing.cpu_count()`

then everywhere we can just do a `make "-j$BUILD_NUM_CORES"`

which I think is a bit simpler

aaronjwood:
Good idea, I'll make adjustments!

cmaloney:
Still thinking about / looking at this. Trying to determine if I want to make all the env vars other than $PKG_PATH + PKG_ID part of the package-id. Doing so would make it so different core count machines wouldn't share cache, but it also makes it so if there's badness at particular parallelization levels, the packages don't pollute in ways where later changing the number of cores breaks them.

If I decide to drop them into the package-id I'll take this PR and add that on top myself.

aaronjwood:
Sounds good. Since the ones that are changed here were already parallelized I'd think that we wouldn't run into any issues by just increasing the parallelism. Unless of course there's a super rare race condition type scenario that can only be triggered by using N cores but I wouldn't think that'd be the case...

Are you worried that increasing the parallelism for the packages here will cause more problems?

cmaloney:
I most definitely have broken the mesos build using 40 cores before :P, and run into other oddness at unusual core counts. Parallelism is hard for people.

cmaloney:
I think this is good to land as is, needs a rebase on current master (there are conflicts) + squash (1 commit per loc change for a single semantic change is definitely too much), then I'll add it to the next train

aaronjwood:
All set!

cmaloney:
Added to Train 85 (#819) for landing

PR #703, opened 2016-09-12
Set the Environment for the disterl port lockdown in the systemd unit

sargun:
This changes the location of where the disterl port is set from the
env file to the systemd unit. The reason for this is it should only
apply when launched from systemd, and not debug scripts, such as
remote_console.


cmaloney:
:ship:

PR #706, opened 2016-09-13
Octarine bugfix

nlsun:
Octarine tests were hitting the wrong url, this was causing tests to pass even though it was misbehaving. Also updated Octarine to fix this.

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -68,7 +68,7 @@ def test_octarine_http(cluster, timeout=30):
     octarine_id = uuid.uuid4().hex
     proxy = ('"http://127.0.0.1:$(/opt/mesosphere/bin/octarine ' +
              '--client --port {})"'.format(octarine_id))
-    check_command = 'curl --fail --proxy {} marathon.mesos'.format(proxy)
+    check_command = 'curl --fail --proxy {} marathon.mesos.mydcos.directory'.format(proxy)
 
     app_definition = {
         'id': '/integration-test-app-octarine-http-{}'.format(test_uuid),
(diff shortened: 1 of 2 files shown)


nlsun:
Ready for review. I think my privileges got reset so I can't restart that 1 failed build.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -68,7 +68,8 @@ def test_octarine_http(cluster, timeout=30):
     octarine_id = uuid.uuid4().hex
     proxy = ('"http://127.0.0.1:$(/opt/mesosphere/bin/octarine ' +
              '--client --port {})"'.format(octarine_id))
-    check_command = 'curl --fail --proxy {} marathon.mesos'.format(proxy)
+    url = "marathon.mesos.mydcos.directory"
```
cmaloney: given this is a locally used only thing, why not just embed it as a constant in the check_command string?

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -107,7 +108,8 @@ def test_octarine_srv(cluster, timeout=30):
     cmd = ('/opt/mesosphere/bin/octarine {} & '.format(octarine_id) +
            '/opt/mesosphere/bin/python -m http.server ${PORT0}')
     raw_app_id = 'integration-test-app-octarine-srv-{}'.format(test_uuid)
-    check_command = ('curl --fail --proxy {} _{}._{}._tcp.marathon.mesos')
+    check_command = ('curl --fail --proxy {} ' +
```
cmaloney: Our wrap width is 120 characters, this is under that / doesn't need to wrap

To match the rest of the codebase should probably be:
```python
    check_command = 'curl --fail --proxy {} _{}._{}._tcp.marathon.mesos.mydcos.directory'.format(
        proxy
        port_name
        raw_app_id)
```

cmaloney:
How did you find this was off / is there some more end to end integration test (like using `dcos-tunnel` or the like) which we could use for this longer term?

cmaloney:
:ship:

nlsun:
I was actually doing a full demo for documentation purposes and that's when I noticed that it wasn't working.

I have a test that would have caught this that runs when I push to the CLI module part of this, so perhaps that could be added to CI.

PR #707, opened 2016-09-13
Remove LD_LIBRARY_PATH from executor environment variables, set RPATH for mesos

cmaloney:
This was supposed to happen in DC/OS 1.3 but it didn't because frameworks broke (https://github.com/mesosphere/dcos-image/commit/17aa4adf79f7946f9a536c0b78b8bd11f7e325de).

Frameworks should now be bundling their own libmesos or using the HTTP API. Same
goes for all software on top of DC/OS. This enforces the separation of the
"system" internals from stuff on top of the cluster.

# Checklist

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -447,8 +447,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "LD_LIBRARY_PATH": "/opt/mesosphere/lib",
-        "SASL_PATH": "/opt/mesosphere/lib/sasl2",
         "LIBPROCESS_NUM_WORKER_THREADS": "8"
       }
   - path: /etc/dns_search_config
--- a/packages/mesos/build
+++ b/packages/mesos/build
@@ -9,7 +9,9 @@ mkdir -p build
 pushd build
 # TODO(cmaloney): --with-glog=/usr --with-protobuf=/usr --with-boost=/usr
 # TODO(cmaloney): DESTDIR builds so we don't have to build as root?
-LIBS="-lssl -lcrypto" LDFLAGS="-L/opt/mesosphere/active/openssl/lib" "/pkg/src/mesos/configure" \
+LIBS="-lssl -lcrypto" \
+  LDFLAGS="-L/opt/mesosphere/active/openssl/lib -Wl,-rpath=/opt/mesosphere/lib"  \
+  "/pkg/src/mesos/configure" \
   --prefix="$PKG_PATH" --enable-optimize --disable-python \
   --enable-libevent --enable-ssl \
   --enable-install-module-dependencies \


lingmann:
Cool LGTM! I'm a :shipit: @cmaloney and ideas on what frameworks might be lagging on bundling libmesos and could use a heads up on this change? I'd like to reach out to them proactively if possible.

cmaloney:
I believe the core data-agility frameworks all have been updated for this. Jenkins is only just getting updated for it I think.

The best would be to audit the universe for what does volume mounts of `/opt/mesosphere`. I believe there are also several internal sales solutions which do it currently.

cmaloney:
https://mesosphere.atlassian.net/browse/DCOS-2369
https://mesosphere.atlassian.net/browse/VELOCITY-245

are at least two of the tickets I know of tracking moving.

lingmann:
@spahl or @adam-mesos can you please review? I'd like to get this merged so that we can protect ourselves from bad/unsupported app development practices. So if there are specific framework authors that we need to notify or tests we need to run before we do so... let's figure out a plan on how we are going to move this forward.

cc: @ssk2 @triclambert @jsancio @pyronicide

ssk2:
@lingmann: we have a PR for this in review. I'm assuming this will land with 1.9? Note that this is going to break any existing Jenkins implementations that are out there, so we're going to make sure we have upgrade notes for users to upgrade their existing Jenkins deployments to use the latest version of the package (once we release it) and _then_ to upgrade DC/OS itself. 

cc @colin-msphere

adammesos:
See also https://mesosphere.atlassian.net/browse/MARATHON-1371 for MoM-EE (potentially also MoM).
See https://mesosphere.atlassian.net/browse/SPARK-406 for including the bundle in the Spark image.
Note that we now have a procedure for building a libmesos bundle that also includes the mesos-modules-private, since those MUST be built using the exact same deps as libmesos or you get nasty runtime linkage errors.
None of the Infinity frameworks have moved to v1 HTTP API yet either, but https://mesosphere.atlassian.net/browse/CORE-272 covers some of the Java shim work that will get them closer.
From what I can tell, EVERY framework we ship (Infinity/Velocity/Marathon) still uses the host libmesos, except Spark which is on a special, private base docker image (which users are complaining about). So, although we're making great strides in providing a libmesos bundle and a Java shim, we need to prioritize/schedule the frameworks actually using these so we can remove their dependencies on host libmesos. Maybe in 1.10?

gen/dcos-config.yaml
```diff
@@ -447,8 +447,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "LD_LIBRARY_PATH": "/opt/mesosphere/lib",
```
adammesos: I'm not sure we can remove this yet in 1.9 (although I wish we could), since every universe framework scheduler except Spark depends on host libmesos, and each of these schedulers runs as an executor, using these executor-environment vars.

Before we merge this, I'd like to see at least Cassandra/Kafka/HDFS, Jenkins, and MoM running (and successfully registered with Mesos) on a cluster from this PR; and an EE strict-mode cluster with the EE versions of those frameworks (including MoM-EE) successfully registered. It'd be great if this was an automated integration test
If that all still works, we can drop LD_LIBRARY_PATH from here. If not, we'll need to wait another release.
Even then, all the rest of the universe frameworks not officially supported by Mesosphere are also likely to break, so we'd need to notify the developers (Do we have a dev@?)

jsancio:
I should point out that Universe can do filtering based on the version of DC/OS. The way we have implemented is such that new packages can express that they do not work on older versions of DC/OS. The opposite is not possible. We did this because we agreed that new version of DC/OS are not allowed to break old frameworks.

If we really need to do this can we first deprecate this. Give framework developers a couple of releases to fix this. After a few releases we can remove the env variable.

gen/dcos-config.yaml
cmaloney: @adam-mesos The need to remove this has been stated to the frameworks teams repeatedly since DC/OS 1.3. That is a lot more than 6 months that they've had notice to fix it. Giving more time I don't think is going to result in a solution.

cmaloney:
@adam-mesos they don't all use libmesos. They've been steadily moving off. Please go talk to the teams. Jenkins is the last one with an active jira where the last comment on the progress was that sunil needed to review the code.

gen/dcos-config.yaml
adammesos: I understand how long we've been dragging this out far too long, but I don't think breaking all our frameworks is going to result in a solution either. The real solution will be to provide them with the tools/APIs (libmesos bundle and v1 HTTP API) so that they can move off of host libmesos with ease and confidence. And those are the efforts that @karya0 and I have been involved in recently.

adammesos:
@cmaloney I _have_ been actively talking to the Infinity and Marathon teams, and neither have moved off of host libmesos yet. The v1 HTTP API doesn't work in strict mode yet, so none of them are moving to that until we unblock them there. The Infinity team (rightfully) did not want to be in charge of building libmesos and the private modules themselves, so they deferred to the Core team to build a bundle for them that they can now start including in their packages instead of relying on host libmesos. That work was only recently completed by @karya0 and now we are working with the teams to help them switch over.
cc: @zen-dog @aquamatthias @mgummelt @gabrielhartmann 

You say that "They've been steadily moving off." but I can't find a single framework (besides Spark) which doesn't rely on the host libmesos. If you know of some, let's get automated integration tests that prove that these frameworks will still work on DC/OS after this PR. I am NOT going to let this merge without appropriate integration tests. You should be comfortable with that edict (I borrowed it from you).

Examples where frameworks still rely on /opt/mesosphere (for libmesos or modules):
Cassandra: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/C/cassandra/23/marathon.json.mustache#L67
HDFS: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/H/hdfs/7/marathon.json.mustache#L66
Jenkins: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/J/jenkins/15/marathon.json.mustache#L40
Kafka: https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/K/kafka/37/marathon.json.mustache#L17
Marathon: https://github.com/mesosphere/marathon-dcos-plugins/blob/master/marathon-ee.json#L43 (see also https://mesosphere.atlassian.net/browse/MARATHON-1371 )
Spark: bases its dockerfile on a private image, which the community dislikes, so we'll move them to the bundle too: https://github.com/mesosphere/spark-build/blob/master/docker/Dockerfile#L21
(see also https://mesosphere.atlassian.net/browse/SPARK-406 )

lingmann:
To add some additional context to this... our users are running into very serious problems due to things like the Docker daemon picking up the LD_LIBRARY_PATH and loading DC/OS libraries. When we upgrade (and thereby replace libraries in /opt/mesosphere) this breaks Docker tasks. See: https://mesosphere.atlassian.net/browse/DCOS-11624 

@adam-mesos and @spahl how can we move this forward? This will need to be a concerted effort by all teams working on DC/OS components and not something that I can drive without some additional support and prioritization.

adam-mesos:
Definitely need buy-in and prioritization from Infinity/SDK, Marathon, and Velocity teams. The Mesos team is now providing a libmesos bundle and a java shim, and we're eager to help teams use them so we can eliminate dependence on /opt/mesosphere.
Let's get this at the top of the [Provisioning Area backlog](https://mesosphere.atlassian.net/wiki/pages/viewpage.action?pageId=93552674) with help from @pyronicide or the successor PM for this area.

jgehrcke:
@mhrabovcin @benclarkwood I believe it makes sense for you to look at this again, and align it with current thinking.

rukletsov:
@adam-mesos, @lingmann I've checked recent versions of all frameworks @adam-mesos mentioned before, and it seems they have all updated to use bundled libmesos: [HDFS](https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/H/hdfs/9/marathon.json.mustache#L53), [Cassandra](https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/C/cassandra/26/marathon.json.mustache#L34), [Jenkins](https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/J/jenkins/21/marathon.json.mustache#L19), [Kafka](https://github.com/mesosphere/universe/blob/version-3.x/repo/packages/K/kafka/40/marathon.json.mustache#L7), [Marathon](https://jira.mesosphere.com/browse/MARATHON_EE-769).

@adam-mesos, what is necessary to merge this PR beyond rerunning tests?

adam-mesos:
SGTM. I think 1.11 is the right time to revisit this, now that we're properly bundling libmesos in all our supported frameworks. We should also notify the community about this pending change ASAP, since this will probably break some community universe packages that would also need to use a bundled libmesos. They don't have to wait until 1.11 to use a bundled libmesos, so they can start updating their packages now.
cc: @karya0 @judithpatudith

judithpatudith:
I [drafted a heads-up email](https://docs.google.com/document/d/12hrzsTtgRrsDf5BCH-aqCfkKixct6c-Q6IpeUmc8EGs/edit?usp=sharing) to the DC/OS community mailing list. Does someone have time to review it?

Does this change impact the Mesos community? (it doesn't look like it but I just wanted to check)

spahl:
This PR needs a rebase and an upstream bump in enterprise. Otherwise looks good.

lingmann:
@rukletsov is going to open another PR and mention this PR for context and history since this PR is coming from Cody's repo.

jgehrcke:
I think we can close this now that https://github.com/dcos/dcos/pull/1857 has landed. Is that correct @adam-mesos @rukletsov @tillt @sethp-nr? Closing under the assumption that the answer is 'yes'. Please leave a comment if that's not true!

PR #713, opened 2016-09-13
WIP: dockercfg remove integration test

joerg84:


test_util/test_runner.py
```diff
@@ -38,12 +39,15 @@ def integration_test(
             the test environment
         pytest_dir: directory of test package on test host
         pytest_cmd: string representing command for py.test
+        test_dockercfg_hook: if set to True, we enable the remove dockercfg hook
+            module
```
mellenburg: I know I said earlier that we _could_ put this in this function argument, but we are going to want to control this from ci and between variants, so instead lets propagate this with TEST_ADD_ENV_*

To do this locally, make no edits to test_util/test_runner.py and just run this as TEST_ADD_ENV_DOCKERCFG_HOOK='true' ccm-deploy-test

To do this in CI, you just add another param here:
https://teamcity.mesosphere.io/admin/editBuildParams.html?id=template:ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTest
(ill go ahead and populate that in a moment, but FYI so you can toggle it)

test_util/test_runner.py
```diff
@@ -38,12 +39,15 @@ def integration_test(
...
 
     Returns:
         exit code corresponding to test_cmd run
 
     """
     dns_search = 'true' if test_dns_search else 'false'
+    dockercfg_hook = 'true' if test_dockercfg_hook else 'false'
```
mellenburg: see above

test_util/test_runner.py
```diff
@@ -52,6 +56,7 @@ def integration_test(
         'PUBLIC_SLAVE_HOSTS=' + ','.join(public_agent_list),
         'DCOS_PROVIDER=' + provider,
         'DNS_SEARCH=' + dns_search,
+        'DOCKERCFG_HOOK=' + dockercfg_hook,
```
mellenburg: see above

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -29,6 +29,7 @@ def cluster():
     assert 'SLAVE_HOSTS' in os.environ
     assert 'PUBLIC_SLAVE_HOSTS' in os.environ
     assert 'DNS_SEARCH' in os.environ
+    assert 'DOCKERCFG_HOOK' in os.environ
```
mellenburg: Is this applicable to Azure and AWS? If not, I would prefer this to have a false default instead of assert on unset

joerg84: Fixed.

cmaloney:
@mellenburg can you please review?

adam-mesos:
@joerg84 Please rebase so we can get this merged into 1.9

packages/dcos-integration-test/extra/test_remove_dockercfg_hook.py
```diff
@@ -0,0 +1,36 @@
+import uuid
+
+import pytest
+
+
+@pytest.fixture(scope='module')
+def auth_cluster(cluster):
```
mellenburg: This fixture is not used.

packages/dcos-integration-test/extra/test_remove_dockercfg_hook.py
```diff
@@ -0,0 +1,36 @@
...
+    if not cluster.dockercfg_hook_enabled:
+        pytest.skip("Skipped because not running against cluster with remove .dockercfg hook.")
+    return cluster
+
+
+def test_remove_dockercfg_hook(cluster):
```
mellenburg: I do not understand what is going on here at all. The build procedure has been changed to always dump a .dockercfg file into the integration testing package. This test should always pass with this build step, right?

packages/dcos-integration-test/extra/test_remove_dockercfg_hook.py
```diff
@@ -0,0 +1,36 @@
...
+            'cpus': 0.1,
+            'mem': 32,
+            'disk': 0,
+            'cmd': "test -f .dockercfg",
+            'artifacts': [{
+                'uri': "file:///opt/mesosphere/active/dcos-integration-test/.dockercfg"}]}}
+    try:
+        cluster.metronome_one_off(job)
+    except Exception as ex:
+        raise Exception(".dockercfg was not removed")
```
mellenburg: How does an exception mean .dockercfg was not removed? This job will fail if the .dockercfg was not pulled. This message indicates it's failing is indicative of the exact opposite of this message i.e. the dockercf WAS removed

cmaloney:
@joerg84 / @mellenburg / @spahl  / @lingmann  the dockercfg remove feature being merged was conditional on getting a test. This test still isn't done. It needs to be done or we should remove the dockercfg remove stuff.

spahl:
@joerg84 can we get this done please?

adam-mesos:
Rebased into #1091 but it still needs some work.

packages/dcos-integration-test/extra/test_remove_dockercfg_hook.py
adam-mesos: I think what we really want here is for this test to be run in at least one configuration. Cloud, on-prem, vagrant, whatever. Or all of them, it shouldn't matter where. What's the best way to configure pytest to do that? Maybe we rip out all these `conftest.py` (now `cluster_api.py`) changes and do something different.

adam-mesos: Or maybe it only works on vagrant since the test is run from a master and the agent is trying to fetch from localhost.

packages/dcos-integration-test/extra/test_remove_dockercfg_hook.py
adam-mesos: This test is supposed to verify that the dockercfg is not exposed inside of the container, when using the fetcher to inject dockercfg's for talking to private registries. We added a special hook to remove the dockercfg from the sandbox before the container starts. Here's the intended flow of this test:
1. On setup, drop a `.dockercfg` on every node's filesystem, in particular targeting the agents. This could just as well be fetched from an http/ftp URL.
2. Metronome asks Mesos to launch a task with the dockercfg's URI as a fetcher artifact
3. Mesos fetcher copies the dockercfg from the `file://` location into the sandbox prepared for the task
4. If this task was a docker image (which it's not), the container runtime (UCR or Docker) would use the dockercfg's credentials to pull the image, e.g. from a private registry
5. Then the hook swoops in and deletes the dockercfg from the sandbox directory (although it still lives in the original location from whence it was fetched)
6. Then the container is launched and the task is run, which checks for the existence of the dockercfg file in the task's sandbox. The task is gone, so `test -f` returns a non-zero status code, which `metronome_one_off`'s `wait_for_completion` will detect as a failure, and raise an exception. I'm not sure why it doesn't. Maybe because we're never enabling the hook?
I think the job should actually `test ! -f` as was happening before, and we should set `DOCKERCFG_HOOK=true` in at least one CI installer configuration
I'll clean this up a bit and see if I can make it clearer in #1091 

Bonus points:
- Verify that a fetched file of a different name is not removed.
Double-bonus points:
- Use the .dockercfg to actually fetch an image from a private registry. Do we have any integration tests that use a private docker registry?

orsenthil:
@joerg84 , please update the PR description to reflect the state of this issue.

adam-mesos:
Closing in favor of #1091

PR #714, opened 2016-09-13
Add check time

sargun:


sargun:
@spahl How can I disable check_time only on Vagrant

cmaloney:
@sargun why do we need to disable it only on vagrant? Shouldn't it work / run everywhere?

sargun:
@cmaloney Because Vagrant doesn't seem to do NTP reliably (at least based on local testing)

sargun:
Review?

spahl:
Looks good to me. Waiting for the tests but  :ship: from an implementation perspective

gen/dcos-config.yaml
```diff
@@ -1,4 +1,11 @@
 package:
+  - path: /etc/check_time.env
```
cmaloney: nit: pkginfo.json goes first

packages/check-time/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pushd /pkg/extra
+mkdir -p $PKG_PATH/bin
+gcc -std=gnu99 -Wall -Werror -o $PKG_PATH/bin/check-time check-time.c
```
cmaloney: nit: `-std=c11`? two decade old C isn't what we need to write all code in

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
...
+#include <stdio.h>
+#include <sys/timex.h>
+#include <string.h>
+#include <errno.h>
+
+#define USEC_PER_MSEC 1000L
+#define MAX_EST_ERROR_US 100L * USEC_PER_MSEC /* 100 millisecond */
+
+int main() {
+	struct timex tx;
```
cmaloney: tx {};
rc = 0;
error = 0;

to prevent fun default not initializing (the curly braces cause the C compiler to fill all not explicitly set fields with zeros).

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
...
+	int rc;
+	long long int error;
+
+	if (getenv("IGNORE_CHECK_TIME")) {
```
cmaloney:
```
char *ignore_check_time = getenv("IGNORE_CHECK_TIME");
if (strcmp(ignore_check_time, "true") == 0) {
    ....
} else if (strcmp(ignore_check_time, "false") == 0) {
   // Continue below
} else {
  fprintf(stderr, "invalid value for IGNORE_CHECK_TIME");
  return 1
}
```
hard failing on unexpected input when we generate all input is nice.

packages/check-time/build
sargun: Yeah, we can do gnu11.

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/timex.h>
+#include <string.h>
+#include <errno.h>
+
+#define USEC_PER_MSEC 1000L
+#define MAX_EST_ERROR_US 100L * USEC_PER_MSEC /* 100 millisecond */
```
cmaloney: I hope drift of about 1 minute should be reasonably acceptable, running customer clusters definitely can get further out of sync than that.

packages/check-time/build
sargun: Although, sometimes, I feel it's a bit too modern :P.

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
...
+	long long int error;
+
+	if (getenv("IGNORE_CHECK_TIME")) {
+		fprintf(stderr, "check-time ignoring time, passing no matter what\n");
+		return 0;
+	}
+
+	memset(&tx, 0, sizeof(tx));
+	rc = adjtimex(&tx);
+	error = tx.esterror - MAX_EST_ERROR_US;
```
cmaloney: Shouldn't access tx.esterror until return code has been checked.

tx.esterror may be positive or negative, so the "are we in sync enough" should probably be
```
/* Make error value absolute, avoiding abs() which is a bit odd as a function (or macro sometimes!) */
long abs_error = tx.esterror > 0 ? tx.esterror : tx.esterror
if (abs_error > MAX_EST_ERROR_US) {
  // print error message
}
```

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
...
+
+	if (rc == TIME_BAD)
+		fprintf(stderr, "Time is marked as bad\n");
+	else if (rc == -1)
+		perror("adjtimex");
```
cmaloney: this code path should presumably `return 1`

packages/check-time/extra/check-time.c
sargun: rc doesn't need to be set to 0 nor error as they are not accessed before being set. RE: tx, makes sense. It's a noop anyways.

packages/check-time/extra/check-time.c
```diff
@@ -0,0 +1,41 @@
...
+	if (getenv("IGNORE_CHECK_TIME")) {
+		fprintf(stderr, "check-time ignoring time, passing no matter what\n");
+		return 0;
+	}
+
+	memset(&tx, 0, sizeof(tx));
+	rc = adjtimex(&tx);
+	error = tx.esterror - MAX_EST_ERROR_US;
+
+	if (rc == TIME_BAD)
```
cmaloney:
```
                   The symbolic name TIME_BAD is a synonym for TIME_ERROR,
                   provided for backward compatibility.
```

TIME_ERROR then?

packages/check-time/extra/check-time.c
sargun: Eh, I think that this is overkill. If IGNORE_CHECK_TIME is set to anything (1, foo, pants) we should ignore it.

packages/check-time/build
cmaloney: If we can, would also be nice to do `-static`

packages/metronome/buildinfo.json
```diff
@@ -1,5 +1,5 @@
 {
-  "requires": ["java", "exhibitor"],
+  "requires": ["java", "exhibitor", "check-time"],
```
cmaloney: We use requires for compile time dependencies nowadays only, no need to add check-time here or to minuteman.

packages/check-time/extra/check-time.c
cmaloney: practically we should really hard error if IGNORE_CHECK_TIME isn't set. we can make a strong guarantee, and if it isn't it's an internal consistency bug (IGN0RE_CHECK_TIME)

packages/check-time/extra/check-time.c
sargun:
```
         This variable establishes the expected error of the indicated
         time relative to the primary synchronization source in
         microseconds. For NTP, the value is determined as the root
         dispersion, which represents the best estimate of the actual
         error of the system clock based on its past behavior, together
         with observations of multiple clocks within the peer group.
         This variable is computed by the synchronization daemon and
         returned in system calls, but is otherwise not used by the
         kernel.
```

This is less about how out of sync we are and more about how poorly Shor's algorithm (or the algo of choice) is performing on the cluster.

packages/check-time/extra/check-time.c
sargun: Eh, https://www.kernel.org/doc/Documentation/CodingStyle

packages/metronome/buildinfo.json
sargun: :+1:

packages/check-time/extra/check-time.c
sargun: Also, under the hood it jmps.

cmaloney: referring to @spahl. From my perspective, next person who touches this will probably end up breaking it and not meaning to. This also has the dangling else parse problem / the else statements could re-associate in fun ways you wouldn't expect with minor changes to the code.

packages/check-time/build
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+pushd /pkg/extra
+mkdir -p $PKG_PATH/bin
+gcc -std=gnu11 -Wall -Werror -o $PKG_PATH/bin/check-time check-time.c
```
cmaloney: `-static`. No reason to leave this linking to libc, the kernel api here should be much more predictable and stable.

packages/check-time/extra/check-time.c
spahl: I'm actually good with that code. It's readable and follows a simple flow IMHO.

packages/check-time/build
spahl: +1

sargun: @cmaloney I thought you don't like static binaries?

sargun:
@cmaloney I thought you don't like static binaries?
 Add a line note

sargun:
Mergeable?

cmaloney:
I like actual static binaries, particularly when they don't touch any of the APIs (pthreads, etc) which have a tight connection to the glibc on host.

cmaloney:
:ship:

PR #715, opened 2016-09-13
test: Tweaked output for deploy_marathon_app

rukletsov:
A marathon app is considered deployed when all instances are running and healthy. To simplify debugging, we should check those cases separately and print different debug messages.

### DIFF
--- a/test_util/cluster_api.py
+++ b/test_util/cluster_api.py
@@ -410,17 +410,25 @@ class ClusterApi:
                     'Application deployment failed, reason: {}'.format(data['app']['lastTaskFailure']['message'])
                 )
 
-            if (
-                data['app']['tasksRunning'] == app_definition['instances'] and
-                (not check_health or data['app']['tasksHealthy'] == app_definition['instances'])
-            ):
-                res = [Endpoint(t['host'], t['ports'][0], t['ipAddresses'][0]['ipAddress'])
-                       for t in data['app']['tasks']]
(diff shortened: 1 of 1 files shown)


mellenburg:
LGTM

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -463,17 +463,19 @@ def _pool_for_marathon_app(app_id):
...
-            ):
+            if (data['app']['tasksRunning'] != app_definition['instances']):
+                logging.info('Waiting for application to be deployed: ' \
+                             'Not all instances are running: {}'.format(repr(data)))
+                return None
+            elif (check_health and data['app']['tasksHealthy'] != app_definition['instances']):
+                logging.info('Waiting for application to be deployed: ' \
+                             'Not all instances are healthy: {}'.format(repr(data)))
+                return None
+            else:
```
cmaloney: could we make this logic so that we only return / get here if we find things to be healthy explicitly?

I don't like having the fall through of "we didn't see anything was unhealthy" -> "everything is okay".

cmaloney:
needs the 1 comment and code quality fix (you can run `tox -e py34-syntax` to check the code quality checks locally.

rukletsov:
Updated the patch, addressed Cody's comment.

cmaloney:
Replaced with #780 which fixes the syntax errors.

PR #717, opened 2016-09-13
test: Updated Marathon app definition for Mesos containerizer

rukletsov:
If Mesos containerizer shall be used, 'ContainerInfo.mesos' field should be set instead of 'ContainerInfo.docker'.

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -46,9 +46,14 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
     app, test_uuid = cluster.get_test_app()
     app['container'] = {
         'type': 'MESOS',
-        'docker': {
+        'mesos': {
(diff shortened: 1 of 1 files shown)


cmaloney:
:ship:

Gilbert88:
LGTM! Ship it!

Note: the old ContainerInfo definition is a non-existed case, currently we do nothing on it. However, we should add validation. Alex posted a patch for it already:
https://reviews.apache.org/r/51865/

nfnt:
Sorry guys, this diff won't help. This isn't a JSONified Mesos protobuf, but a JSONified Marathon app definition. Setting `docker` is the right thing to do here. Unfortunately, Marathon's validation was loosened a bit to be able to add the universal containerizer support. As a result, setting `mesos` here will be swallowed during serialization where ideally there should be a validation error. Because of that, the Mesos containerizer will be used to isolate the task, but without provisioning a Docker image (because `docker` wasn't set). This may be the reason why it may have solved some problems, but in the end this isn't what we want to test here (= deploying an app in a Docker image using the Mesos containerizer).

PR #725, opened 2016-09-14
Teach tests to get a new auth token when the current one expires

malnick:
This PR invokes the `_login()` function before running the test on registered slaves to ensure we have a fresh token in the case that we use a short lived token TTL.

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -289,11 +289,23 @@ class Cluster:
             return self.superuser_auth_header
         return {}
 
+
+    def _try_authenticate(result):
+        """If the value of the get request is 401, we should authenticate
+        in the case that we have an expired auth token."""
+        if result.status_code == 401:
+            self._authenticate()
+            return true
+        return false
+
+
+    @retry(retry_on_result=_try_authenticate, stop_max_attempt_number=1)
     def get(self, path="", params=None, disable_suauth=False, **kwargs):
         hdrs = self._suheader(disable_suauth)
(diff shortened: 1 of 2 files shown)


malnick:
Per @cmaloney side-conversation, will be adding a durable check to cluster.get() instead.

mellenburg:
Nit: title is misleading, this isnt during the test, its during every single get request made by the cluster fixture

malnick:

jgehrcke:
The change as-is can introduce severe false-positive test results as of effectively disabling the effect of the `disable_suauth` argument.

malnick:
@jgehrcke As a end-user configuration parameter?

jgehrcke:
> @jgehrcke As a end-user configuration parameter?

No. Look at how the method works w/o this changeset. If `disable_suauth == True`, then no authentication header is set. There is no point in re-trying then upon 401 retrieval.

malnick:
Thanks for pointing that out @jgehrcke ! I've updated the logic to check for the Auth header, and to only retry if it is not present.

malnick:
Thanks @jgehrcke, I updated the logic to include a check on `disable_suauth` too. Let me know what you think.

cmaloney:
@jgehrcke can you review again?

jgehrcke:
See internal ticket DCOS-9142 for reference. There I have expressed my view on how we should test the case

> Does (mini)Bouncer pick up a non-default bouncer_expiration_auth_token_days config.yaml setting?

There is a much better way than modifying _all_ tests, which potentially has quite bad side effects on the test results (false positives absolutely conceivable).

PR #726, opened 2016-09-14
[EXPERIMENTAL] add gatewayd

darkonie:


darkonie:
@cmaloney this should be targeted to 1.10

cmaloney:
@darkonie don't actually target PRs, all that tracking lives in JIRA. PRs land when they land, unless we're explicitly backporting (like 1.8.5)

PR #732, opened 2016-09-15
Restart download if it fails

mellenburg:


cmaloney:
Unfortunately if dcos-download fails, dcos-setup will never run thanks to the wonderful systemd `Requires=` + `Restart=always` problem so hosts will never actually recover (https://github.com/dcos/dcos/blob/master/gen/dcos-services.yaml#L33 for reference). That bug is why the systemd rules (https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md) disallow requires/wants... This there wasn't another way for me to ship it in the time we wanted / needed unfortunately. The dcos-download + dcos-setup stuff probably needs to be reworked as a pair to make this work.

Migrating installed hosts is also going to be complicated (these files live outside of pkgpanda's realm of influence currently)

mellenburg:
Ah, thanks for the clarification.

PR #737, opened 2016-09-16
Fix systemd unit files to work with modern systemd

sargun:


alberts:
@sargun what did they change to break this?

sargun:
@alberts Env variable setting with spaces has to be in quotes now

cmaloney:
Can you add a line to `docs/systemd-rules.md` about that?

sargun:
Updated.

alberts:
@sargun @drewkerrigan do we want this one or https://github.com/dcos/dcos/pull/739 or both?

PR #738, opened 2016-09-16
Bug DCOS-9371: Makes named VIP subnet configurable from gen

GoelDeepak:
@sargun @cmaloney  Request you to please review this PR

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -362,6 +362,26 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
         "Must be an absolute path to a directory, although leave off the `/` at the beginning and end."
 
 
+def calculate_minuteman_min_named_ip_erltuple(minuteman_min_named_ip):
+    return ip_to_erltuple(minuteman_min_named_ip)
+
+
+def calculate_minuteman_max_named_ip_erltuple(minuteman_max_named_ip):
+    return ip_to_erltuple(minuteman_max_named_ip)
+
+
+def ip_to_erltuple(ip):
+    return '{' + ip.replace('.', ',') + '}'
+
+
+def validate_minuteman_min_named_ip(minuteman_min_named_ip):
+    validate_ipv4_addresses([minuteman_min_named_ip])
+
+
+def validate_minuteman_max_named_ip(minuteman_max_named_ip):
+    validate_ipv4_addresses([minuteman_max_named_ip])
+
+
 __logrotate_slave_module_name = 'org_apache_mesos_LogrotateContainerLogger'
 
 
@@ -387,7 +407,9 @@ entry = {
         lambda dcos_overlay_config_attempts: validate_int_in_range(dcos_overlay_config_attempts, 0, 10),
         lambda dcos_remove_dockercfg_enable: validate_true_false(dcos_remove_dockercfg_enable),
         validate_rexray_config,
-        lambda check_time: validate_true_false(check_time)],
+        lambda check_time: validate_true_false(check_time),
+        validate_minuteman_min_named_ip,
+        validate_minuteman_max_named_ip],
     'default': {
         'bootstrap_tmp_dir': 'tmp',
         'bootstrap_variant': lambda: calculate_environment_variable('BOOTSTRAP_VARIANT'),
@@ -435,6 +457,8 @@ entry = {
               }                                         \
             ]}',
         'dcos_remove_dockercfg_enable': "false",
+        'minuteman_min_named_ip': '11.0.0.0',
+        'minuteman_max_named_ip': '11.255.255.255',
         'no_proxy': '',
         'rexray_config_preset': '',
         'rexray_config': json.dumps({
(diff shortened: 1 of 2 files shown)


gen/dcos-config.yaml
```diff
@@ -87,7 +87,20 @@ package:
...
         {minuteman,
           [
-            {agent_polling_enabled, false}
+            {agent_polling_enabled, false},
+            {min_named_ip, {{ minuteman_min_named_ip_str }}},
+            {max_named_ip, {{ minuteman_max_named_ip_str }}}
+          ]
+        }
+      ].
+  - path: /etc_slave/minuteman.app.config
```
sargun: You're going to want to do this for the public slave as well

gen/dcos-config.yaml
```diff
@@ -87,7 +87,20 @@ package:
...
+    content: |
+      [
+        {minuteman,
+          [
+            {agent_polling_enabled, true},
```
sargun: No need to explicitly say true here, it's enabled by default.

gen/calc.py
```diff
@@ -412,6 +441,8 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
               }                                         \
             ]}',
         'dcos_remove_dockercfg_enable': "false",
+        'minuteman_min_named_ip': '11.0.0.0',
+        'minuteman_max_named_ip': '11.0.0.254',
```
sargun: The default should be 11.255.255.255

gen/calc.py
```diff
@@ -339,6 +339,33 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
...
+
+def validate_minuteman_min_named_ip(minuteman_min_named_ip):
+    validate_ipv4_address_format(minuteman_min_named_ip)
+
+
+def validate_minuteman_max_named_ip(minuteman_max_named_ip):
+    validate_ipv4_address_format(minuteman_max_named_ip)
+
+
+def validate_ipv4_address_format(ip):
```
sargun: Don't we already have an IPv4 address validator?

gen/calc.py
```diff
@@ -339,6 +339,33 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
...
+def calculate_ip_erlstr(ip):
+    return '{' + ip.replace('.', ',') + '}'
+
+
+def validate_minuteman_min_named_ip(minuteman_min_named_ip):
+    validate_ipv4_address_format(minuteman_min_named_ip)
+
+
+def validate_minuteman_max_named_ip(minuteman_max_named_ip):
+    validate_ipv4_address_format(minuteman_max_named_ip)
```
sargun: No need to revalidate it - gen automatically validates it once you put a validator into the right list. 

See this PR for a simple example: https://github.com/dcos/dcos/pull/714

gen/calc.py
```diff
@@ -339,6 +339,33 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
...
+def calculate_minuteman_min_named_ip_str(minuteman_min_named_ip):
+    return calculate_ip_erlstr(minuteman_min_named_ip)
+
+
+def calculate_minuteman_max_named_ip_str(minuteman_max_named_ip):
+    return calculate_ip_erlstr(minuteman_max_named_ip)
+
+
+def calculate_ip_erlstr(ip):
+    return '{' + ip.replace('.', ',') + '}'
```
sargun: Probably not the best idea to call this "calculate_", probably something more like ip_to_erltuple?

gen/dcos-config.yaml
GoelDeepak: done.

gen/dcos-config.yaml
GoelDeepak: done.

gen/calc.py
GoelDeepak: done.

gen/calc.py
GoelDeepak: didn't get you. This is a validator

gen/calc.py
GoelDeepak: I had thoughts about it. There is one which takes in a list of IP addresses. I can use that one.

gen/calc.py
GoelDeepak: 11.255.255.255 is a broadcast address for the subnet but I got your point that the range should end with 11.255.255.254.

sargun: It should be okay to use `11.255.255.255`, since this isn't on a layer 2 segment.

The way that Minuteman works is by programming kernel flows (via IPTables conntrack).

sargun:
lgtm.

sargun:
@spahl  @cmaloney Any way we can review this and get it in 1.8.5?

cmaloney:
@sargun is there a customer need or bug this is addressing which makes it have to go into 1.8.5? Trying to keep that as small a pile of backports as possible.

1.9 should be fairly close in the future.

sargun:
@cmaloney This is for a customer need. The customer that the 11/8 range conflicts with.

PR #739, opened 2016-09-16
Fix Erlang ENV Problems

drewkerrigan:
The current build of DCOS warns about certain Erlang apps like navstar, minuteman, and spartan on certain operating systems due to a systemd Enviroment config problem that results in log line like this:

```
Sep 16 10:12:12 p1.dcos systemd[1]: [/opt/mesosphere/packages/navstar--525060447dd2b1b09088865e6d0ebb0d209558c0/dcos.target.wants/dcos-navstar.service:9] Invalid environment assignment, ignoring: ERL_FLAGS=-kernel inet_dist_listen_min 62502 -kernel inet_dist_listen_max 62502
```

As a result, those services never get the correct `ERL_FLAGS` env var.

This PR fixes the problem by putting those env vars in service-specific `*.env` files on all hosts.

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -37,6 +37,9 @@ package:
   - path: /etc_master/master_count
     content: |
       {{ num_masters }}
+  - path: /etc/spartan.env
+    content: |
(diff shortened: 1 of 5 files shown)


alberts:
Superseded by https://github.com/dcos/dcos/pull/763 according to @sargun

PR #742, opened 2016-09-16
Bump to Python 3.5

cmaloney:
Things noticed so far:
 - [x] Need code quality hosts to have Python 3.5 available for running bumped integration tests (might compromise and make the req `py3` to land then bump to 3.5 once we switch to ubuntu 16.04 for build hosts. Local tests show all tests passing with python 3.5


cmaloney:
Chatted with @lingmann, going to leave the integration testing on python 3.4 for now, will upgrade the system python to 3.5. When CI boxes are updated to new ubuntu, will then change the integration tests over to python 3.5

cmaloney:
:ship:

PR #736, opened 2016-09-16
Revert "test: Updated Marathon app definition for Mesos containerizer"

cmaloney:
This reverts commit de9842a766582355079ba4dab1032f832ff76485.

See comments on #717 

cc: @rukletsov

PR #746, opened 2016-09-19
Revert "test: Updated Marathon app definition for Mesos containerizer"

rukletsov:
This reverts commit de9842a766582355079ba4dab1032f832ff76485.


alberts:
@rukletsov why? :)

cmaloney:
Duplicate of #736, this is needed because the new format introduces meaningless fields which don't get used by anything and aren't validated as invalid. The old one was doing the right thing via marathon's API. The new one doesn't.

PR #747, opened 2016-09-19
Add /usr/sbin to PATH in /opt/mesosphere/environment

branden:
Resolves https://dcosjira.atlassian.net/browse/DCOS-392.

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -33,7 +33,7 @@ reserved_env_vars = ["LD_LIBRARY_PATH", "PATH"]
 
 env_header = """# Pkgpanda provided environment variables
 LD_LIBRARY_PATH={0}/lib
-PATH={0}/bin:/usr/bin:/bin:/sbin\n\n"""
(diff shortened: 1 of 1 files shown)


orsenthil:
Change LGTM.

orsenthil:
This came up in our meeting again and I wanted to explain my thought process.

In terms of design, Iwould agree, if we should not  add **any system PATHS** like (/usr/bin:/bin:/sbin) in `$PATH` in `pkgpanda/__init__.py and keep them all separate in environment files. But, if we are mutating`$PATH` explicitly,  we have to provide all the locations where the required binaries can be looked up. In the JIRA referenced it looks like `useradd` in CoreOS7 was only present in /usr/sbin/. 

/cc @sphal @lingmann

lingmann:
We decided to close this because we don't have a clear understanding of the problem and there is no clear reproduction from the end-user. We suspect this is related to sudo environment inheritance / transitions, but there is not enough importance here to invest in digging in at this time.

PR #749, opened 2016-09-19
Handle connection error on dcos-history

amitaekbote:
When running dcos-history, if the flask app cannot bind to the port assigned it does not fail and keeps retrying. Using gunicorn ensures that the app fails and master is marked unhealthy.

https://mesosphere.atlassian.net/browse/DCOS-5334 

Unit tests passed
Syntax test passed

### DIFF
--- a/packages/dcos-history/build
+++ b/packages/dcos-history/build
@@ -3,12 +3,18 @@ source /opt/mesosphere/environment.export
 export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.4/site-packages"
 mkdir -p "$LIB_INSTALL_DIR"
 
+
 # Copy from the ro /pkg/extra to a rw folder.
 cp -r /pkg/extra /build
+
+
+# Copy the gunicorn conf file to $PKG_PATH
+cp /pkg/extra/history/dcos_history_conf.py "$PKG_PATH/dcos_history_conf.py"
+
 
 pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /build
 
 # Create the service file
 service="$PKG_PATH/dcos.target.wants_master/dcos-history.service"
 mkdir -p "$(dirname "$service")"
-cp /pkg/extra/dcos-history.service  "$service"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-history.service > "$service"
--- a/packages/dcos-history/extra/history/dcos_history_conf.py
+++ b/packages/dcos-history/extra/history/dcos_history_conf.py
@@ -1,0 +1,9 @@
+from history.server import on_starting_server
+
+
+bind = '0.0.0.0:15055'
+workers = 1
+
+
+def on_starting(server):
+    on_starting_server(server)
--- a/packages/dcos-history/extra/history/server.py
+++ b/packages/dcos-history/extra/history/server.py
@@ -2,9 +2,11 @@ import logging
 import os
 import sys
 
+
 from flask import Flask, Response
 from flask.ext.compress import Compress
 from history.statebuffer import BufferCollection, BufferUpdater
+
 
 app = Flask(__name__)
 compress = Compress()
@@ -79,7 +81,7 @@ def _response_(content):
     return Response(response=content, content_type="application/json", headers=headers_cb())
 
 
-def start():
+def on_starting_server(server):
     global state_buffer
     logging.basicConfig(format='[%(levelname)s:%(asctime)s] %(message)s', level='INFO')
 
@@ -90,4 +92,10 @@ def start():
 
     state_buffer = BufferCollection(os.environ['HISTORY_BUFFER_DIR'])
     BufferUpdater(state_buffer, headers_cb).run()
-    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', '15055')))
+
+
+def start():
+    # Used for testing only; on dc/os $PATH should have gunicorn
+    # Have to be in the same folder to run this
+    # In case of failure it will not sys.exit
+    os.system("gunicorn -c dcos_history_conf.py server:app")
--- a/packages/dcos-integration-test/extra/test_endpoints.py
+++ b/packages/dcos-integration-test/extra/test_endpoints.py
@@ -1,6 +1,7 @@
 import urllib.parse
 
 import bs4
+from retrying import retry
 
 
 def test_if_dcos_ui_is_up(cluster):
@@ -116,15 +117,20 @@ def test_if_pkgpanda_metadata_is_available(cluster):
 
 
 def test_if_dcos_history_service_is_getting_data(cluster):
-    r = cluster.get('/dcos-history-service/history/last')
-    assert r.status_code == 200
-    # Make sure some basic fields are present from state-summary which the DC/OS
-    # UI relies upon. Their exact content could vary so don't test the value.
-    json = r.json()
-    assert 'cluster' in json
-    assert 'frameworks' in json
-    assert 'slaves' in json
-    assert 'hostname' in json
+
+    @retry(stop_max_delay=20000, wait_fixed=500)
+    def check_up():
+        r = cluster.get('/dcos-history-service/history/last')
+        assert r.status_code == 200
+        # Make sure some basic fields are present from state-summary which the DC/OS
+        # UI relies upon. Their exact content could vary so don't test the value.
+        json = r.json()
+        assert 'cluster' in json
+        assert 'frameworks' in json
+        assert 'slaves' in json
+        assert 'hostname' in json
+
+    check_up()
 
 
 def test_if_we_have_capabilities(cluster):
--- a/packages/dcos-history/buildinfo.json
+++ b/packages/dcos-history/buildinfo.json
@@ -1,5 +1,5 @@
 {
-  "requires": ["flask", "python", "python-requests"],
+  "requires": ["flask", "python", "python-gunicorn", "python-requests"],
   "state_directory": true,
   "username": "dcos_history"
 }
--- a/packages/dcos-history/extra/dcos-history.service
+++ b/packages/dcos-history/extra/dcos-history.service
@@ -12,4 +12,5 @@ EnvironmentFile=-/opt/mesosphere/etc/dcos-history-extras.env
 Environment=HISTORY_BUFFER_DIR=/var/lib/dcos/dcos-history
 ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-history
-ExecStart=/opt/mesosphere/bin/dcos-history
+ExecStart=/opt/mesosphere/bin/gunicorn -c $PKG_PATH/dcos_history_conf.py \
+                history.server:app
--- a/packages/dcos-history/extra/setup.py
+++ b/packages/dcos-history/extra/setup.py
@@ -20,6 +20,7 @@ config = {
     'install_requires': [
         'Flask>=0.10.1',
         'flask-compress',
+        'gunicorn',
         'requests']
 }
 


packages/dcos-history/extra/history/server.py
```diff
@@ -78,16 +80,15 @@ def _buffer_response_(name):
 def _response_(content):
     return Response(response=content, content_type="application/json", headers=headers_cb())
 
-
-def start():
+if __name__ == '__main__':
```
mellenburg: In the systemd unit you make a change so that gunicorn executable is the main thread and calls out to the app WSGI object. If that't the way to run it, then this whole section should be removed so that we only have one way to run this service

packages/dcos-history/extra/history/server.py
```diff
@@ -78,16 +80,15 @@ def _buffer_response_(name):
...
     global state_buffer
     logging.basicConfig(format='[%(levelname)s:%(asctime)s] %(message)s', level='INFO')
-
     compress.init_app(app)
-
     if 'HISTORY_BUFFER_DIR' not in os.environ:
         sys.exit('HISTORY_BUFFER_DIR must be set!')
-
     state_buffer = BufferCollection(os.environ['HISTORY_BUFFER_DIR'])
     BufferUpdater(state_buffer, headers_cb).run()
```
mellenburg: The BufferUpdater will never be called as implemented. If we are telling gunicorn to import and run the WSGI app for us, then we need to include starting the BufferUpdater when we instantiate the WSGI app. This would be a somewhat confusing approach, so maybe consider doing it like this: http://docs.gunicorn.org/en/latest/custom.html so you can easily call out the starting of the updater thread whenever you start the server

amitaekbote:
Ready for review

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -76,10 +77,15 @@ def test_if_uiconfig_is_available(cluster):
 
 
 def test_if_dcos_history_service_is_up(cluster):
-    r = cluster.get('/dcos-history-service/ping')
 
-    assert r.status_code == 200
-    assert 'pong' == r.text
+    @retry(stop_max_delay=20000, wait_fixed=500)
```
mellenburg: this functionality is already covered when we make the cluster fixture: https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/conftest.py#L87
which does:
https://github.com/dcos/dcos/blob/master/test_util/cluster_api.py#L58
Thus, I see no reason to modify this test

packages/dcos-history/build
```diff
@@ -3,12 +3,18 @@ source /opt/mesosphere/environment.export
 export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.4/site-packages"
 mkdir -p "$LIB_INSTALL_DIR"
 
+
 # Copy from the ro /pkg/extra to a rw folder.
 cp -r /pkg/extra /build
 
+
+# Copy the gunicorn conf file to $PKG_PATH
+envsubst '$PKG_PATH' < /pkg/extra/history/dcos_history_conf.py > "$PKG_PATH/dcos_history_conf.py"
```
mellenburg: I'm confused about this usage of envsubst. What is the purpose of '$PKG_PATH'? envsubst does not take an argument other than STDIN I thought...
Furthermore, you are envsubst-ing on an python file...
(1) That python file has no shell variables to substitute
(2) If it did that would be a rather odd python file
Why not just cp the file as this is effectively doing?

packages/dcos-history/extra/history/dcos_history_conf.py
```diff
@@ -0,0 +1,9 @@
+from history.server import on_starting_server
+
+
+bind = '0.0.0.0:15055'
```
mellenburg: This was configurable by environment before

packages/dcos-history/build
```diff
@@ -3,12 +3,18 @@ source /opt/mesosphere/environment.export
...
+
+
 pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /build
 
 # Create the service file
 service="$PKG_PATH/dcos.target.wants_master/dcos-history.service"
 mkdir -p "$(dirname "$service")"
-cp /pkg/extra/dcos-history.service  "$service"
+envsubst '$PKG_PATH' < /pkg/extra/dcos-history.service > "$service"
```
mellenburg: As above, I dont  see any env to subst in dcos-history.service, so just use cp

packages/dcos-history/extra/history/server.py
```diff
@@ -90,4 +92,10 @@ def start():
...
-    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', '15055')))
+
+
+def start():
+    # Used for testing only
+    # Have to be in the same folder to run this
+    # In case of failure it will not sys.exit
+    os.system("/opt/mesosphere/bin/gunicorn -c dcos_history_conf.py server:app")
```
mellenburg: I don't think we should expect that anyone running the unit tests will have gunicorn at this file path.

packages/dcos-history/extra/history/server.py
```diff
@@ -90,4 +92,10 @@ def start():
 
     state_buffer = BufferCollection(os.environ['HISTORY_BUFFER_DIR'])
     BufferUpdater(state_buffer, headers_cb).run()
-    app.run(host='0.0.0.0', port=int(os.environ.get('PORT', '15055')))
+
+
+def start():
```
mellenburg: If we need to explicitly invoke gunicorn from outside of python to get the expected behavior, then why leave this option around? I would just remove this entirely as well as the hook in setup.py

cmaloney: For easy local development / testing it's really nice to have a simple single command, this makes it so they all run through the same single codepath. With this for development / testing can still just run `dcos-history` as a command rather than needing to enter a python shell and manually start the server.

packages/dcos-integration-test/extra/test_endpoints.py
cmaloney: So this is covering a different flakiness the switch to gunicorn found / exposed. So dcos-history-service up checks that one instance somewhere in the cluster is up. This test then sometimes hits other instances in the cluster, which sometimes are up but have no data.

mellenburg: @cmaloney I don't follow, this is test is sending the exact same GET request to the same adminrouter endpoint as wait_for_dcos_history_up. The only difference is this asserts /ping is working perfectly whereas the wait_for_up check just makes sure its _probably_ somewhere. In any case, this test is hitting /ping so it will have the exact same behavior regardless if it has data

packages/dcos-history/extra/history/server.py
mellenburg: Having a testing shortcut is fine in general, but the UX of this shortcut is somewhat awkward. Specifically, I need to be aware of which directory I call it from and have files be in the right place. Before this change, you can call dcos-history wherever and it still works. Alternatively, this could pass the config path and app module location via arguments or change the working directory to always be in the right place, but my personal (and lazy) preference would be a README line on how to run it and be done

packages/dcos-integration-test/extra/test_endpoints.py
cmaloney: The nginx / adminrouter actually forwards / load balances to all master.mesos addresses: https://github.com/dcos/adminrouter/blob/master/nginx.master.conf#L17

so by hitting adminrouter, adminrouter sends it to one instance mor or less at random

packages/dcos-history/build
cmaloney: the service has a $PKG_PATH in it. https://github.com/dcos/dcos/pull/749/files#diff-ee8c631a232fb776584c0b43f493138eR15

packages/dcos-history/extra/history/server.py
cmaloney: The "needs to be in the right path" thing is relatively easily fixable by just changing the gunicorn exec to not be an absolute path. then a `pip install -e` into a virtualenv makes it easy to edit / work interactively.

packages/dcos-history/build
mellenburg: oops, very obvious. no issue here.

packages/dcos-history/extra/history/server.py
amitaekbote: I dont expect this to be used for unit testing, I will change the comment to make that clear. This is more for testing on the cluster

packages/dcos-integration-test/extra/test_endpoints.py
amitaekbote: @mellenburg We saw this test flakiness with multiple masters where we might have to wait longer for all of them to come up.

mellenburg: I get that this _can_ hit other masters, but it does so in a very liberal way:
(1) Retries for any state of request. We would like to fail if we hit a connection error
(2) I don't like retrying in tests on something we expect to be stable once the cluster is up; if history service is not up everywhere when this tests starts, then perhaps we should make a change to the wait_for_history_service up
(3) if we really want to validate that all masters have history service running, then we should make sure that we can get cluster.len(master)*FACTOR successful GET's in a row to be reasonably sure we actually hit something else.

cmaloney: Practically if we want to this "right" then we need to hit every dcos-history-service directly. Doing that requires being on every master and doing a ping directly. Making it work across open and enterprise is complicated. That's not feasible for fixing here / now. 

Doing something like "check statistically" I don't think adds any value here. Our goal is just "make sure the history service is up and running". This should produce a log of the failures + those messages, the contents to help people debug.

The issue here is purely that sometimes the history buffer has never once pulled state when the worker gets hit and pulls the state out. We could make the checks a little more tuned down, but practically if there's a  major error it'll still show up, we're not waiting for a ton of time, and we definitely do not ever wait for "everything to actually wait for up". Making things wait for up would fix this and a bunch of other issues, but is a very out of scope project for now.

packages/dcos-history/extra/history/server.py
cmaloney: @amitaekbote would still be nice if we can just say `gunicorn`. For users doing testing inside a DC/OS cluster that should be in $PATH. It hsould also be so if people are in a virtualenv and have done a `pip install -e`.

packages/dcos-integration-test/extra/test_endpoints.py
mellenburg: @cmaloney ok, when considering scope, I agree the wait_for_up issue and 'make sure endpoints on all agents and masters' are much bigger than this PR, so I'm ok with dropping them.

However, there is still the fact this test pulls an endpoint that doesnt pull state. So, just hitting /dcos/history-service/last and verifying non-empty would cover that.

packages/dcos-history/extra/history/server.py
amitaekbote: done

packages/dcos-integration-test/extra/test_endpoints.py
amitaekbote: The purpose to modify this was to cover the flakiness in multi master scenarios. Ideally we would do what @mellenburg said and verify the service is up by checking last. We will create an issue to keep track of this and have it outside the scope of this commit.

amitaekbote: https://dcosjira.atlassian.net/browse/DCOS-406

packages/dcos-history/extra/history/server.py
mellenburg: Nit: changing to gunicorn instead of /opt/mesosphere/bin/gunicorn doesn't help the fact that dcos_history_conf.py is not necessarily in working dir, but that might just be splitting hairs

cmaloney:
I really don't think go would help here / fix the issues. The problem is purely around how people chose to structure this API in python. People made a try/catch loop that caught a little too much. I guarantee there are C++ libraries and Go libraries which have the same issue around trying to bind.

There was also a _lot_ of learning about how different pieces work in the history of building this (how dcos_generate_config.sh works / is constructed among other things).

mellenburg:
@cmaloney Fair point, but what I was getting at is that the threading as implemented has some weird side effects and makes it a bit more difficult to work with

PR #751, opened 2016-09-19
Move cluster API from test to lib

mellenburg:
By moving this into test_util, we can interact with it in integration tests, broader scripted test, and downstream test suites.


mellenburg:
@cmaloney This is ready for review. Goal here is to change virtually nothing until we get downstream hooked in

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -1,24 +1,11 @@
...
-from contextlib import contextmanager
-from urllib.parse import urlparse, urlunparse
 
-import dns.exception
-import dns.resolver
 import pytest
-import requests
-import retrying
+
+import test_util.cluster_api
```
orsenthil: [Minor]: Since test_util is a package, you can just `import test_util` or just do `from test_util import cluster_api` and use it. `import a.b` and then using it as `a.b` at places does not give any advantage.

Not applicable here, but in case, if we write mocks, then the first two ways are preferable.

PR #752, opened 2016-09-20
Octarine: Golang 1.7, godep, and parsing changes

nlsun:
Update to golang 1.7 and move parsing from general HTTP handler to just SRV records. Also updated the project to use godep, a golang dependency tracking tool.

### DIFF
--- a/packages/octarine/buildinfo.json
+++ b/packages/octarine/buildinfo.json
@@ -3,7 +3,7 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/dcos/octarine.git",
-    "ref": "426a6db3ec0ea4a86af05a6c7585722e4d636e3f",
+    "ref": "abd003681cf65e984f12168d6120cfd48c179fc6",
     "ref_origin": "master"
   }
 }
--- a/packages/octarine/buildinfo.json
+++ b/packages/octarine/buildinfo.json
@@ -1,5 +1,5 @@
 {
-  "docker": "golang:1.6",
+  "docker": "golang:1.7",
   "single_source" : {
     "kind": "git",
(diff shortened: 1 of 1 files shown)


nlsun:
Ready for review.

sargun:
@nlsun We don't use godep, we use govendor, as that's the preferred tool for 1.6+

nlsun:
Well, when I search through large/popular golang projects they all seem to use godep. godep also has vendor support now.

sargun:
We use govendor in Mesos-DNS: https://github.com/mesosphere/mesos-dns/

nlsun:
Looking through large golang projects, they all use godep or roll their own so I'd assume it's more battle tested. Maybe Mesos-DNS should be switched to godep? Also, if this is going into 1.8.5 then perhaps I should remove the commit that switches to golang 1.7 @cmaloney ?

cmaloney:
@nlsun not going into 1.8.5

sargun:
We just switched away from Godep in Mesos-DNS. Do you want to file a bug, and kick off the conversation there?

nlsun:
Is godep blocking this from getting merged?

cmaloney:
There seems to be conflict between you and sargun about whether this is ready / what to do, you two as maintainers need to come to a resolution

cmaloney:
Also, since there was the conflict and this was moved back to WIP, there hasn't been any new "ready for review" to tell me things are all resolved / I should look at it again.

nlsun:
The godep stuff was figured out. Ready for review again.

cmaloney:
to me this looks good, that said holding it off until 1.9 is `master` as we have a lot of things / critical bug fixes only for master which is 1.8.5 bound still.

PR #754, opened 2016-09-20
Update tests to expect either root or nobody users.

adam-mesos:


packages/dcos-integration-test/extra/conftest.py
```diff
@@ -418,7 +418,7 @@ def deploy_test_app_and_check(self, app, test_uuid):
                 msg += "Detailed explanation of the problem: {2}"
                 pytest.fail(msg.format(r.status_code, r.reason, r.text))
 
-            assert r.json() == {'username': 'root'}
+            assert r.json() == {'username': 'root'} or r.json() == {'username': 'nobody'}
```
alberts: maybe a comment explaining why this might happen?

adam-mesos:
Closing this in favor of a more explicit approach, without magic unexplained usernames.
See https://github.com/dcos/dcos/pull/755

PR #755, opened 2016-09-21
Tests should be explicit about app user and expected user

adam-mesos:


adam-mesos:
Passes all expected tests. Ready for review, @mellenburg @alberts

adam-mesos:
Squashed.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -16,7 +16,7 @@ def test_if_marathon_app_can_be_deployed(cluster):
     the one assigned to test - test succeds.
     """
     app, test_uuid = cluster.get_test_app()
-    cluster.deploy_test_app_and_check(app, test_uuid)
+    cluster.deploy_test_app_and_check(app, test_uuid, expected_user=app['user'])
```
cmaloney: can we default / if `expected_user` is None, then use app['user']? Then I think this diff just goes to adding that default lookup, adding the default user, and adding the one extra line to change user for the exceptional mesos containerizer case.

test_util/cluster_api.py
```diff
@@ -274,6 +274,7 @@ def get_test_app(self, custom_port=False):
             'cpus': 0.1,
             'mem': 32,
             'instances': 1,
+            'user': 'nobody',
```
cmaloney: Very explicitly do _not_ want to specify a user automatically and check what the default / no app user is for some of these cases (See https://mesosphere.atlassian.net/browse/DCOS-9040). That's why the user checking was added.

adam-mesos: I see. Well, that default user could be different depending on downstream config, so we can't necessarily hardcode "root" as the expected default user.
Maybe I'll just disable this test in our downstream and clone a version that is downstream-config-aware.

cmaloney: I'm thinking we need to grab the "user to use" out of an environment variable, than various integration test runners can set how they want / need

adam-mesos: Would involve setting a `TEST_ADD_ENV_` in enterprise `util/run_ccm_test.sh`?
How would substitute in the correct expected user value based on security-mode?
Should I just read `default_task_user` out of bootstrap-config.json if it exists, and assume `root` otherwise?

adam-mesos:
Closing in favor of whatever @mellenburg decides to do for https://mesosphere.atlassian.net/browse/DCOS-9970

PR #761, opened 2016-09-22
Modify for 3DT for downstream compatibility

mellenburg:
Minimial refactor so that 3DT is compatible
with SSL enabled clusters

See: https://github.com/mesosphere/dcos-enterprise/pull/316
for downstream results with this change

### DIFF
--- a/packages/dcos-integration-test/extra/test_3dt.py
+++ b/packages/dcos-integration-test/extra/test_3dt.py
@@ -4,47 +4,53 @@ import json
 import logging
 import os
 import tempfile
+import urllib.parse
 import zipfile
 
+import pytest
 import requests
 import retrying
 
 BASE_ENDPOINT_3DT = '/system/health/v1'
-PORT_3DT = 1050
-PORT_3DT_AGENT = 61001
 # Expected latency for all 3dt units to refresh after postflight
 LATENCY = 60
 
 
-def make_3dt_request(ip, endpoint, cluster, port=80):
+@pytest.fixture
+def make_3dt_request(cluster):
     """
-    a helper function to get info from 3dt endpoint. Default port is 80 for pulled data from agents.
-    if a destination port in 80, that means all requests should go though master (adminrouter) and we can re-use
-    cluster.get otherwise we can query 3dt agents directly to port 61001 (agent-adminrouter).
+    returns a function that performs authenticated connections to 3DT on the appropriate port
+    and appends ?cache=0 to all endpoints
     """
-    if port == 80:
+    def make_request(ip, endpoint, is_agent):
         assert endpoint.startswith('/'), 'endpoint {} must start with /'.format(endpoint)
-        logging.info('GET {}'.format(endpoint))
-        json_response = cluster.get(path=endpoint).json()
-        logging.info('Response: {}'.format(json_response))
+        endpoint = BASE_ENDPOINT_3DT + endpoint + '?cache=0'
+        scheme = urllib.parse.urlparse(cluster.dcos_uri).scheme
+        # TODO: SSL settings are not a part of dcos/dcos. Migrate settings to be in the same repo
+        port = {
+            'agent': {'http': 61001, 'https': 61002},
+            'master': {'http': 80, 'https': 443}}['agent' if is_agent else 'master'][scheme]
+        url = '{}://{}:{}/{}'.format(scheme, ip, port, endpoint.lstrip('/'))
+        logging.info('GET {}'.format(url))
+        request_headers = {}
+        if cluster.auth_enabled:
+            request_headers = cluster.superuser_auth_header
+        response = requests.get(url, headers=request_headers)
+        assert response.ok
+        try:
+            json_response = response.json()
+            logging.info('Response: {}'.format(json_response))
+        except ValueError:
+            logging.exception('Coult not deserialized json response from {}\nResponse contents:{}'.format(
+                url, response.content.decode()))
+            raise
+        assert len(json_response) > 0, 'json response is invalid from {}: {}'.format(url, json_response)
         return json_response
-
-    url = 'http://{}:{}/{}'.format(ip, port, endpoint.lstrip('/'))
-    logging.info('GET {}'.format(url))
-    request = requests.get(url)
-    assert request.ok
-    try:
-        json_response = request.json()
-        logging.info('Response: {}'.format(json_response))
-    except ValueError:
-        logging.error('Coult not deserialized json response from {}'.format(url))
-        raise
-    assert len(json_response) > 0, 'json response is invalid from {}'.format(url)
-    return json_response
+    return make_request
 
 
 @retrying.retry(wait_fixed=2000, stop_max_delay=LATENCY * 1000)
-def test_3dt_health(cluster):
+def test_3dt_health(cluster, make_3dt_request):
     """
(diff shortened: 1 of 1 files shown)


mellenburg:
@darkonie PTAL
@cmaloney @spahl This is ready for review. The refactor is by no means a great example of how this should be done, but the goal is to have ~700 lines less of code to consider when overhauling the cluster test API

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -6,45 +6,55 @@
...
+        if agent and scheme == 'http':
+            port = 61001
+        elif agent and scheme == 'https':
+            port = 61002
+        elif not agent and scheme == 'http':
+            port = 80
+        elif not agent and scheme == 'https':
+            port = 443
+        else:
+            raise Exception('Scheme is invalid: {}'.format(scheme))
```
darkonie:
```
agent_port_map = {
    'http': 61001,
    'https': 61002
}

master_port_map = {
    'http': 80,
    'https': 443
}

port = master_port_map.get(schema)
if agent:
    port = agent_port_map.get(schema)

assert port
```

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -6,45 +6,55 @@
...
+    and appends ?cache=0 to all endpoints
     """
-    if port == 80:
+    def make_request(ip, endpoint, agent):
         assert endpoint.startswith('/'), 'endpoint {} must start with /'.format(endpoint)
-        logging.info('GET {}'.format(endpoint))
-        json_response = cluster.get(path=endpoint).json()
-        logging.info('Response: {}'.format(json_response))
+        endpoint = BASE_ENDPOINT_3DT + endpoint + '?cache=0'
+        scheme = cluster.dcos_uri.split(':')[0]
```
darkonie: let's use `urllib2`

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -6,45 +6,55 @@
...
+            port = 61001
+        elif agent and scheme == 'https':
+            port = 61002
+        elif not agent and scheme == 'http':
+            port = 80
+        elif not agent and scheme == 'https':
+            port = 443
+        else:
+            raise Exception('Scheme is invalid: {}'.format(scheme))
+        url = '{}://{}:{}/{}'.format(scheme, ip, port, endpoint.lstrip('/'))
```
darkonie: quick research on python url libs did not give me a good example how to construct the URLs with lib rather then a string interpolation. In go you could use https://golang.org/pkg/net/url/#URL to build a URL, but for python i think that is ok

orsenthil: @darkonie  - `urlparse.unparse` is the function to use for url construction.  You give the parts in the arguments.  But it is not wrong to construct the url like you do here.

You can do slightly better by having semantic keywords in the format string. like
`url = '{scheme}://{host}:{port}/{path}'.format(scheme=scheme, host=ip, port=port, path=endpoint.lstrip('/'))`

mellenburg: There is a nice Url class in EE that I plan to port over to this repo: https://github.com/mesosphere/dcos-enterprise/blob/master/packages/dcos-integration-test-ee/extra/dcostests/url.py

packages/dcos-integration-test/extra/test_3dt.py
mellenburg: Done

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -4,47 +4,58 @@
...
     """
-    if port == 80:
+    def make_request(ip, endpoint, agent):
         assert endpoint.startswith('/'), 'endpoint {} must start with /'.format(endpoint)
-        logging.info('GET {}'.format(endpoint))
-        json_response = cluster.get(path=endpoint).json()
-        logging.info('Response: {}'.format(json_response))
+        endpoint = BASE_ENDPOINT_3DT + endpoint + '?cache=0'
+        scheme = urllib.parse.urlparse(cluster.dcos_uri).scheme
+        agent_port_map = {
```
cmaloney: It's also a bit odd that the https ports never used in `dcos/dcos` are here. I get it for now, but should have a TODO about that as techdebt.

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -4,47 +4,58 @@
...
+        request_headers = {}
+        if cluster.auth_enabled:
+            request_headers = cluster.superuser_auth_header
+        request = requests.get(url, headers=request_headers)
+        assert request.ok
+        try:
+            json_response = request.json()
+            logging.info('Response: {}'.format(json_response))
+        except ValueError:
+            logging.error('Coult not deserialized json response from {}'.format(url))
```
cmaloney: nit: please include the json document here. (`request.content`)

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -4,47 +4,58 @@
...
+            raise
+        assert len(json_response) > 0, 'json response is invalid from {}'.format(url)
```
cmaloney: `'json response is invalid from {}: {}'.format(url, json_response)`

packages/dcos-integration-test/extra/test_3dt.py
cmaloney: Might make sense to do the 3dt master + agent port as a mandatory test environment variable

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -161,18 +172,18 @@ def test_3dt_nodes(cluster):
...
 
-def test_3dt_nodes_node(cluster):
+def test_3dt_nodes_node(cluster, make_3dt_request):
     """
     test a specific node enpoint /system/health/v1/nodes/<node>
     """
     for master in cluster.masters:
         # get a list of nodes
-        response = make_3dt_request(master, BASE_ENDPOINT_3DT + '/nodes', cluster)
+        response = make_3dt_request(master, '/nodes', False)
```
cmaloney: nit: The last parameter being passed to these isn't really obvious what it is / means. Doing `is_agent=False` explicitly for these would help quite a bit.

Could also make the `is_agent` implicit by doing `is_agent = host in cluster.agents` in the 3dt check function

packages/dcos-integration-test/extra/test_3dt.py
mellenburg: If you don't particularly mind, I'd like to avoid setting this via env var for now, as there are a few other cases like this and I would like to address them all uniformly

cmaloney: re: env var, that's fine for now, but the comment your comment was responding to doesn't talk about. The json document that this comment is about still applies

mellenburg:
@cmaloney Comments addressed assuming you are OK with my position on setting via environment variable

cmaloney:
Responded inline to you're comment, I'm fine with not doing env vars for now, but definitely need a TODO comment then.

cmaloney:
Should be good once the code-quality is fixed

cmaloney:
:ship:

PR #764, opened 2016-09-22
Always authenticate before any test

mellenburg:
wait_for_dcos needs to be in the fixture or else
running tests in arbitrary order will not work


mellenburg:
@alberts @cmaloney Can we :ship: this small usability fix?

cmaloney:
Are we certain this happens once and only once in a test run?

I invision accidentally having every person requesting the cluster object accidentally getting a new one that quietly swallows up errors around their individual services going down briefly while they're hitting them when they shouldn't be.

mellenburg:
@cmaloney Scope='session' works fine as far as I can tell and this is only run once.

PR #765, opened 2016-09-22
[WIP] Basic automated uninstallation test

mellenburg:


test_util/test_aws_vpc.py
```diff
@@ -72,6 +73,18 @@
 DEFAULT_AWS_REGION = os.getenv('DEFAULT_AWS_REGION', 'eu-central-1')
 
 
+def get_cluster_systemd_units(cluster):
+    all_hosts = [cluster.bootstrap_host] + cluster.agents + cluster.public_agents +\
```
branden: This list is already at `cluster.hosts`.

test_util/installer_api_test.py
```diff
@@ -343,3 +343,6 @@ def deploy(self, expect_errors=False):
 
     def postflight(self, expect_errors=False):
         self.run_cli_cmd('--postflight', expect_errors=expect_errors)
+
+    def uninstall(self, expect_errors=False):
```
cmaloney: the dgc.sh --uninstall is actually something we need to kill or rework.  `pkgpanda uninstall` is the canonical we should test.

cmaloney:
Closing since it hasn't had movement in a long time, branch still exists if we want to come back. There is https://dcosjira.atlassian.net/browse/DCOS-250 to track building an uninstall that actually works for DC/OS.  Part of that includes actually testing it.

PR #770, opened 2016-09-23
Bump dcos-ui for 1.8.5

rcorral:
Fixes launching apps with networking options and disables killing tasks that can't be deleted.


cmaloney:
note for self: can't land in master / is 1.8.5 only cherry-pick

PR #771, opened 2016-09-25
Bug DCOS-9363 - Testing minuteman changes for overlay network

GoelDeepak:


packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -59,3 +59,22 @@ def test_ip_per_container(cluster):
         app_port = app_definition['container']['docker']['portMappings'][0]['containerPort']
         cmd = '/opt/mesosphere/bin/curl -s -f http://{}:{}/ping'.format(service_points[1].ip, app_port)
         ensure_routable(cmd, service_points)()
+
+
+def test_ip_per_container_with_named_vip(cluster):
```
sargun: Not sure what this test is supposed to do beyond the existing test?

gen/calc.py
```diff
@@ -339,6 +339,26 @@ def validate_bootstrap_tmp_dir(bootstrap_tmp_dir):
         "Must be an absolute path to a directory, although leave off the `/` at the beginning and end."
 
 
+def calculate_minuteman_min_named_ip_erltuple(minuteman_min_named_ip):
```
sargun: Can you separate this PR / set of changes from this patch?

packages/dcos-integration-test/extra/test_networking.py
GoelDeepak: There were three tests prior to this fourth one. Here is my understanding of those three tests:
1. minuteman routes to VIP -- tests reachability to VIP without overlay.
2. minuteman routes to named VIP -- same as 1 but with a named VIP.
3. IP per container -- reachability to a container IP without load balancing.
4. IP per container with named VIP -- reachability to a container IP with load balancing.

gen/calc.py
GoelDeepak: Yeah, I also didn't like it but then my previous PR is on the master itself (didn't create a branch for that change) and this one is on the branch cut out of the master after the previous commit. Is it fine if we just let it be this way this time? From next time, I plan to create branches for individual changes to avoid such issues.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -59,3 +59,21 @@ def test_ip_per_container(cluster):
...
+    """Test if we are able to connect to a task with ip-per-container mode
+    """
+    # Launch the test_server in ip-per-container mode
+
+    app_definition, test_uuid = cluster.get_test_app_in_docker(ip_per_container=True)
```
sargun: So, this should use cluster.get_test_app(), if possible. Avoiding the use of a Docker container is ideal here.

sargun: ip_per_container in this app def exposes the hostPort on the overlay. We need to create a app that has a port only on the overlay.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -59,3 +59,21 @@ def test_ip_per_container(cluster):
...
+
+    assert len(cluster.slaves) >= 2, "IP Per Container tests require 2 private agents to work"
```
sargun: Any reason we need this? For Load balancing testing, we shouldn't need the 2nd container to actually be on the overlay.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -59,3 +59,16 @@ def test_ip_per_container(cluster):
         app_port = app_definition['container']['docker']['portMappings'][0]['containerPort']
         cmd = '/opt/mesosphere/bin/curl -s -f http://{}:{}/ping'.format(service_points[1].ip, app_port)
         ensure_routable(cmd, service_points)()
+
+
+def test_ip_per_container_with_named_vip(cluster):
+    """Test if we are able to connect to a task with ip-per-container mode using named vip
+    """
+    app_definition, test_uuid = cluster.get_test_app_in_docker(ip_per_container=True)
+    app_definition['container']['docker']['portMappings'][0]['servicePort'] = 6000
```
sargun: We need to delete the hostPort and servicePort entry to prevent port mapping from occuring.

sargun:
Once you merge the Minuteman PR, please update this PR, and punt to @cmaloney  for review.

GoelDeepak:
@cmaloney I have updated this PR to point to latest minuteman commit. This is ready for review now.

PR #772, opened 2016-09-26
Seperate user identity from cluster api

mellenburg:


mellenburg:

test_util/cluster_api.py
```diff
@@ -160,34 +161,17 @@ def wait_for_dcos(self):
...
-            if self.username and self.password:
-                js = {'uid': self.username, 'password': self.password}
-        else:
-            # no authentication required
-            return
-
-        r = requests.post(self.dcos_uri + '/acs/api/v1/auth/login', json=js)
-        # In the case of the Advanced Templates this endpoint comes up a bit later
-        if r.status_code >= 400:
+    def authenticate_default_user(self):
```
cmaloney: these return codes are never checked by the one call site. Should the `return False` really be an exception?

cmaloney: This probably should also be an `_authenticate_default_user(self)` since others calling it modifies global internal state and needs to be more managed than other things most likely.

test_util/cluster_api.py
```diff
@@ -160,34 +161,17 @@ def wait_for_dcos(self):
...
             return False
-        self.superuser_auth_header = {'Authorization': 'token=%s' % r.json()['token']}
-        self.superuser_auth_cookie = r.cookies['dcos-acs-auth-cookie']
+        self.superuser_auth_header = self.default_user.auth_header
+        self.superuser_auth_cookie = self.default_user.auth_cookie
+        return True
 
     def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
-                 dns_search_set, provider, auth_enabled, username, password):
+                 dns_search_set, provider, auth_enabled, default_user=None):
```
cmaloney: `web_auth_default_user`? I'm worried that this is going to be very ambiguous with the "default user for marathon tasks"

cmaloney: `web_auth_credentials` could also be good

test_util/helpers.py
```diff
@@ -4,10 +4,34 @@
...
 
+import test_util
+
 Host = namedtuple('Host', ['private_ip', 'public_ip'])
 SshInfo = namedtuple('SshInfo', ['user', 'home_dir'])
 
 
+class DcosUser:
+    """A lightweight user representation."""
+    def __init__(self, auth_json):
```
cmaloney: nit: This feels a lot like a named tuple + a helper function, but I can also get doing a class.

test_util/helpers.py
```diff
@@ -4,10 +4,34 @@
...
+        self.auth_json = auth_json
+        self.auth_header = None
+        self.auth_token = None
+        self.auth_cookie = None
+
+    def authenticate(self, cluster):
+        assert isinstance(cluster, test_util.cluster_api.ClusterApi), 'Unrecognized cluster object!'
+        logging.info('Attempting authentication')
+        r = cluster.post(path='/acs/api/v1/auth/login', disable_suauth=True, payload=self.auth_json)
+        if r.status_code >= 400:
```
cmaloney: `r.raise_for_status()` (http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes) I think will give us much better error messages here. Ideally want the status_code,  the message, and the `.content`. The `.content` usually contains lots of juicy info.

test_util/helpers.py
```diff
@@ -4,10 +4,34 @@
...
+            return False
+        assert r.ok, 'Authentication failed with status_code: {}'.format(r.status_code)
+        self.auth_token = r.json()['token']
```
cmaloney: would be handy to log / print r.json() before this. If we got an empty json object, this would just break hard and be complicated to debug.

test_util/cluster_api.py
mellenburg: Note: web_auth_default_user is the choice as web_auth_credentials sounds more like passing in a JSON

mellenburg:
@cmaloney comments addressed; ready for another pass

test_util/cluster_api.py
```diff
@@ -150,44 +150,25 @@ def _wait_for_metronome(self):
...
-        self.superuser_auth_header = {'Authorization': 'token=%s' % r.json()['token']}
-        self.superuser_auth_cookie = r.cookies['dcos-acs-auth-cookie']
+    @retrying.retry(wait_fixed=2000, stop_max_delay=120 * 1000)
+    def _authenticate_default_user(self):
+        """retry default auth user because in some deployments,
+        the auth endpoint might not be routable immediately
+        after adminrouter is up. DcosUser.authenticate()
+        will raise exception if authorization fails
+        """
+        self.default_user.authenticate(self)
```
cmaloney: web_auth_default_user

cmaloney:
:ship:

PR #773, opened 2016-09-26
Consolidated request methods

mellenburg:
Builds on work from https://github.com/dcos/dcos/pull/772
Note: 0e8cbec987d89dbf3dd59b8683f2027d8b11ae96 is taking advantage of this change to have adminrouter URL construction in only one place

### DIFF
--- a/packages/dcos-integration-test/extra/test_3dt.py
+++ b/packages/dcos-integration-test/extra/test_3dt.py
@@ -472,7 +472,7 @@ def test_3dt_bundle_create(cluster):
 
     # start the diagnostics bundle job
     create_url = '/system/health/v1/report/diagnostics/create'
-    create_response = cluster.post(path=create_url, payload={"nodes": ["all"]}).json()
+    create_response = cluster.post(path=create_url, json={"nodes": ["all"]}).json()
     logging.info('POST {}, response: {}'.format(create_url, create_response))
 
     # make sure the job is done, timeout is 5 sec, wait between retying is 1 sec
--- a/packages/dcos-integration-test/extra/test_3dt.py
+++ b/packages/dcos-integration-test/extra/test_3dt.py
@@ -4,11 +4,9 @@ import json
 import logging
 import os
 import tempfile
-import urllib.parse
 import zipfile
 
 import pytest
-import requests
 import retrying
 
 BASE_ENDPOINT_3DT = '/system/health/v1'
@@ -22,29 +20,18 @@ def make_3dt_request(cluster):
     returns a function that performs authenticated connections to 3DT on the appropriate port
     and appends ?cache=0 to all endpoints
     """
-    def make_request(ip, endpoint, is_agent):
+    def make_request(ip, endpoint):
         assert endpoint.startswith('/'), 'endpoint {} must start with /'.format(endpoint)
         endpoint = BASE_ENDPOINT_3DT + endpoint + '?cache=0'
-        scheme = urllib.parse.urlparse(cluster.dcos_uri).scheme
-        # TODO: SSL settings are not a part of dcos/dcos. Migrate settings to be in the same repo
-        port = {
-            'agent': {'http': 61001, 'https': 61002},
-            'master': {'http': 80, 'https': 443}}['agent' if is_agent else 'master'][scheme]
-        url = '{}://{}:{}/{}'.format(scheme, ip, port, endpoint.lstrip('/'))
-        logging.info('GET {}'.format(url))
-        request_headers = {}
-        if cluster.auth_enabled:
-            request_headers = cluster.web_auth_default_user.auth_header
-        response = requests.get(url, headers=request_headers)
+        response = cluster.get(path=endpoint, node=ip)
         assert response.ok
         try:
             json_response = response.json()
             logging.info('Response: {}'.format(json_response))
         except ValueError:
-            logging.exception('Coult not deserialized json response from {}\nResponse contents:{}'.format(
-                url, response.content.decode()))
+            logging.exception('Could not deserialize response contents:{}'.format(response.content.decode()))
             raise
-        assert len(json_response) > 0, 'json response is invalid from {}: {}'.format(url, json_response)
+        assert len(json_response) > 0, 'Empty JSON returned from 3DT request'
         return json_response
     return make_request
 
@@ -60,7 +47,7 @@ def test_3dt_health(cluster, make_3dt_request):
 
     # Check all masters 3DT instances on base port since this is extra-cluster request (outside localhost)
     for host in cluster.masters:
-        response = make_3dt_request(host, '/', is_agent=False)
+        response = make_3dt_request(host, '/')
         assert len(response) == len(required_fields), 'response must have the following fields: {}'.format(
             ', '.join(required_fields)
         )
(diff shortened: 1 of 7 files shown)


mellenburg:

cmaloney:
This seems structurally reasonable to me. There's the stuff in #772 outsanding I think. That we need to manually pass thorugh (user=None) as much as we do suggests there's a missing wrapper / binding that out that should happen, although admittedly that was pre-existing. Will do a full code review when #772 has gotten a ship

mellenburg:
Preemptively rebased this on the updated https://github.com/dcos/dcos/pull/772
By simply setting DcosUser.auth_header at init  removed a few user=None. Otherwise, all the user=None calls are places where you just want to see that there is no chance that auth headers are being populated

test_util/cluster_api.py
```diff
@@ -201,55 +202,50 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+        self.dns_host = parse_result.netloc.split(':')[0]
 
         # Make URI never end with /
         self.dcos_uri = dcos_uri.rstrip('/')
 
-    @staticmethod
-    def _marathon_req_headers():
-        return {'Accept': 'application/json, text/plain, */*'}
+        # Generate request functions
+        for method in self.request_methods:
```
cmaloney: @orsenthil is this python canonical / standard way to do this?

packages/dcos-integration-test/extra/test_3dt.py
```diff
@@ -22,29 +20,18 @@ def make_3dt_request(cluster):
...
-        port = {
-            'agent': {'http': 61001, 'https': 61002},
-            'master': {'http': 80, 'https': 443}}['agent' if is_agent else 'master'][scheme]
-        url = '{}://{}:{}/{}'.format(scheme, ip, port, endpoint.lstrip('/'))
-        logging.info('GET {}'.format(url))
-        request_headers = {}
-        if cluster.auth_enabled:
-            request_headers = cluster.web_auth_default_user.auth_header
-        response = requests.get(url, headers=request_headers)
+        response = cluster.get(path=endpoint, node=ip)
```
cmaloney: nit: would have been nice in the commit message to have a comment that this is moving 3dt to go through agent adminrouter. Took me a while to figure that out.

test_util/cluster_api.py
cmaloney: nit: this doesn't quite feel right, but works quite well without a lot of code, reasonably understandable.

cmaloney: https://docs.python.org/3/library/functools.html#functools.partialmethod might be useful.

Could then do something like:
```
get = partialmethod(cluster_request_method_name, 'get')
post = partialmethod(cluster_request_method_name, 'post')
```
on the class for the ones we actually use / care about.

currently this adds a bunch of member functions which are very hard to trace back to their origins (introspection tools in editors and any sort of static analysis will likely fail hard)

test_util/cluster_api.py
```diff
@@ -201,55 +202,50 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
-        return requests.head(self.dcos_uri + path, headers=hdrs)
-
-    def node_get(self, node, path="", params=None, disable_suauth=False, **kwargs):
-        """Execute a GET request against the adminrouter on node."""
-        hdrs = self._suheader(disable_suauth)
-        hdrs.update(kwargs.pop('headers', {}))
-
-        if node in self.masters:
-            port = self.adminrouter_master_port[self.scheme]
+    def cluster_request(self, request_fn):
```
cmaloney: Can we make wrapping this just be binding parameters with partialmethod / partial rather than doing the decorator pattern? 

I find it a lot easier to not have the function in function and just read a couple of parameters being bound away.

```python
def cluster_request(self, request_method, path="", node=None, user=self.web_auth_default_user, port=None, **kwargs):
    ...
    requests.request(method, url, ...)
```

then we can do `self.get = functools.partialmethod(cluster_request, "get")`

test_util/cluster_api.py
```diff
@@ -201,55 +202,50 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+        """Decorator function for making authorized requests within the cluster
+        """
+        def wrapped_request(path="", node=None, user=self.web_auth_default_user, port=None, **kwargs):
```
cmaloney: I'd think `path`, `node` should both be required parameters. `port` as the third parameter, then `user` and `**kwargs`.

Having a cluster pick one specific node is a bit odd. Could _maybe_ default it to the leading master if that's heavily used, but I don't think it particularly is.

test_util/cluster_api.py
```diff
@@ -201,55 +202,50 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+            from default_user (if set)
+            """
+            if node is None:
+                node = self.dns_host
+            if self.auth_enabled and user:
+                hdrs = user.auth_header
+            else:
+                hdrs = {}
+            hdrs.update(kwargs.pop('headers', {}))
+            url = self.get_url(node=node, path=path, port=port)
```
cmaloney: This "get url" using urlunparse definitely doesn't feeling like the right thing to use here. Aren't these all just kwargs we're building for the one call?

How about:
```
url = '{scheme}://{node}/{path}'.format(scheme=self.scheme, '{}:{}'.format(node, port) if port else node, path)
logging.info('{}: {}'.format(method, url)
if self.auth_enabled and user:
    kwargs.setdefault('headers', dict()).setdefault('headers', dict()).update(user.auth_header)
requests.request(method, url, **kwargs)
```
(the format string there will not even need the .format() once we get some of the new python fancy which is coming, https://www.python.org/dev/peps/pep-0498/)

test_util/cluster_api.py
mellenburg: I'm OK with making path a required parameter, but one of the main desired of this features of this harness is "just do a request to whatever the DCOS_DNS_ADDRESS is". It only in very specific cases that you wish to hit a specific master or agent IP, and having to set node=cluster.DCOS_DNS_ADDRESS feels more awkward

cmaloney: I can get the node=None means use the public cluster address / load balancer, definitely needs a bit in the docstring.

test_util/cluster_api.py
mellenburg: Note: one reason I chose this somewhat awkward pattern is that in downstream tests, we need to bind verify=my_ca_cert_path for all these same methods and its nice to do it in a programmatic manner to prevent divergence. Case in point: in the current method with the manually wrapped request methods, PUT takes in **kwargs and never passes it and also breaks the parity with requests lib by mapping the json keyword to payload instead.

mellenburg:
@cmaloney This is ready for another pass.

orsenthil:
I reviewed this too. And this looks good to me. (it has already gone through a rigorous review).

test_util/cluster_api.py
```diff
@@ -396,7 +406,7 @@ def deploy_marathon_app(self, app_definition, timeout=120, check_health=True, ig
             applications. I.E:
                 [Endpoint(host='172.17.10.202', port=10464), Endpoint(host='172.17.10.201', port=1630)]
         """
-        r = self.post('/marathon/v2/apps', app_definition, headers=self._marathon_req_headers())
+        r = self.post('/marathon/v2/apps', json=app_definition, headers=self._marathon_req_headers())
```
cmaloney: future thought: Ideally eventually it might be nice to move this to a `Marathon` API wrapper that we shared with the DC/OS CLI to have native Marathon API calling. Have to balance that a little bit with "want to make sure the Marathon API doesn't change unexpectedly"

PR #775, opened 2016-09-26
Fixed compilation error of mesos-overlay-module.

asridharan:
Removing an unused namespace from the overlay-module, that was causing compilation failure with Mesos HEAD (since that namespace has been removed in the newer version of the API).


asridharan:
reviewed by @jieyu

https://github.com/dcos/mesos-overlay-modules/pull/8

asridharan:
Ready for review.

PR #777, opened 2016-09-26
Add hook for setting mesos user in test

mellenburg:


test_util/cluster_api.py
```diff
@@ -200,6 +200,7 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
                 configured if its value is "true".
             provider: onprem, azure, or aws
             auth_enabled: True or False
+            mesos_user: account that mesos is running under
```
adam-mesos: Really, it's the "default linux user that Marathon will launch its tasks as". Other frameworks can be configured to launch their tasks as different linux users, and even Marathon apps can override the default to launch tasks as other users.

mellenburg: Hmmm, my naming decision was influenced by https://github.com/dcos/dcos/pull/772 which adds a default_user option. So, how about we call this os_default_framework_user?

test_util/cluster_api.py
```diff
@@ -360,7 +362,7 @@ def deploy_test_app_and_check(self, app, test_uuid):
                 msg += "Detailed explanation of the problem: {2}"
                 raise Exception(msg.format(r.status_code, r.reason, r.text))
 
-            assert r.json() == {'username': 'root'}
+            assert r.json() == {'username': self.mesos_user}
```
adam-mesos: Also note that this "operating_environment" check returns the "container user", which will always be 'root' when run in a docker image whose Dockerfile has USER=root (the default, I believe), even if Marathon's default user is 'nobody', although the sandbox would be owned by 'nobody', and the "host machine user" is still 'nobody'.
We should also test for the "host machine user", which is what `--mesos_user` actually affects.

mellenburg: Oh, good catch; I didnt know that. Fortunately this helper only works on the same test_app which will not change docker user, so easy to cover

test_util/cluster_api.py
```diff
@@ -338,6 +340,11 @@ def get_test_app_in_docker(self, ip_per_container):
         return app, test_uuid
 
     def deploy_test_app_and_check(self, app, test_uuid):
+        marathon_user = self.mesos_user
+        # if using docker containerizer, reported user will be docker user
+        if app['container'] and app['container']['type'] == 'DOCKER':
+            marathon_user = 'root'
+        marathon_user = self.default_os_user
```
adam-mesos: Seems like you should have an `else` here, rather than always overriding the `marathon_user` with the `default_os_user`.

test_util/cluster_api.py
```diff
@@ -338,6 +340,11 @@ def get_test_app_in_docker(self, ip_per_container):
         return app, test_uuid
 
     def deploy_test_app_and_check(self, app, test_uuid):
+        marathon_user = self.mesos_user
```
adam-mesos: `mesos_user` doesn't exist anymore, does it? Also, you set `marathon_user` to `default_os_user` below.

test_util/cluster_api.py
```diff
@@ -187,7 +187,7 @@ def _authenticate(self):
         self.superuser_auth_cookie = r.cookies['dcos-acs-auth-cookie']
 
     def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
-                 dns_search_set, provider, auth_enabled, username, password):
+                 dns_search_set, provider, auth_enabled, username, password, default_os_user='root'):
```
adam-mesos: Why do you need to provide the value 'root' for `default_os_user` here and in `conftest.py`? Seems like only one should be necessary. Defining the same constant in 2 places is rarely necessary.

test_util/cluster_api.py
```diff
@@ -338,6 +340,11 @@ def get_test_app_in_docker(self, ip_per_container):
         return app, test_uuid
 
     def deploy_test_app_and_check(self, app, test_uuid):
+        marathon_user = self.mesos_user
+        # if using docker containerizer, reported user will be docker user
```
greggomann: Could you expand on this comment a little? I think it could be helpful to explain that the testing service provided by the test app checks the username within the task's container; when using the docker containerizer, this user is dependent on the image being used.

test_util/cluster_api.py
mellenburg: I did this so I could avoid editing the vagrant test. Editing so its a required arg

mellenburg:
@adam-mesos updated, so PTAL

test_util/cluster_api.py
```diff
@@ -200,6 +200,7 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
                 configured if its value is "true".
             provider: onprem, azure, or aws
             auth_enabled: True or False
+            default_os_user: default user that marathon will launch tasks under
```
adam-mesos: marathon/metronome

test_util/cluster_api.py
```diff
@@ -338,6 +340,18 @@ def get_test_app_in_docker(self, ip_per_container):
...
 
     def deploy_test_app_and_check(self, app, test_uuid):
+        """This method deploys the test server app and then
+        pings its /operating_environment endpoint to retrieve the container
+        user.
+
+        In a docker container this will be whatever is set as USER (defaults to root)
+        In a mesos container, this will be the marathon user
+        """
+        # if using docker containerizer, reported user will be docker user
```
adam-mesos: Not sure you need this comment now that you have the big one above

mellenburg:
@adam-mesos Comments address, ready to :ship: when tests finish?

test_util/cluster_api.py
```diff
@@ -338,6 +340,17 @@ def get_test_app_in_docker(self, ip_per_container):
         return app, test_uuid
 
     def deploy_test_app_and_check(self, app, test_uuid):
+        """This method deploys the test server app and then
+        pings its /operating_environment endpoint to retrieve the container
+        user.
+
+        In a docker container this will be whatever is set as USER (defaults to root)
```
cmaloney: nit: This comment suggests the code should be looking up USER from the docker but it doesn't. Maybe something like "Docker containers run things as USER from the Dockerfile, That is currently in these tests always the default, root"".

test_util/cluster_api.py
```diff
@@ -338,6 +340,17 @@ def get_test_app_in_docker(self, ip_per_container):
...
+        In a mesos container, this will be the marathon user
+        """
+        if 'container' in app and app['container']['type'] == 'DOCKER':
+            marathon_user = 'root'
+        else:
+            marathon_user = self.default_os_user
```
cmaloney: Shouldn't this do a app.get('user', self.default_os_user) so the user can be overridden via the marathon.json?

cmaloney:
:ship:

PR #783, opened 2016-09-27
Improve gen resolvconf

sargun:


packages/spartan/extra/gen_resolvconf.py
```diff
@@ -4,32 +4,64 @@
...
+"""
+
 # Constants
 MAX_SERVER_COUNT = 3
 SPARTANS = ['198.51.100.1', '198.51.100.2', '198.51.100.3']
 
+# We assume that RESOLVCONF_DIR_PATH is precreated
+RESOLVCONF_DIR_PATH = '/var/lib/dcos/resolvconf'
+RESOLVCONF_PATH = os.path.join(RESOLVCONF_DIR_PATH, 'resolv.conf')
+DCOS_BIND_MOUNT_POINT = '/tmp/dcos'
```
alberts: I think this is dangerous -- some tmpwatch thing is going to roll through and clean this up.

/run/dcos is probably a much better place for this file.

sargun: But it's only used inside of the private namespace. I thought about using /run, but I'm not sure as to the rules of mkdiring under /run, and also what in /run needs to be accessible for the system to operate correctly.

PR #789, opened 2016-09-28
Wait for ZK quorum before starting tests

mellenburg:


mellenburg:

PR #790, opened 2016-09-28
Handle AWS Client Throttling Error

mellenburg:
-Catches client error and if the error is due to throttling, the poll interval will be increased
-Frees up the API quota by adding a sleep before beginning to poll as we can be reasonable certain a given cloud formation will not come up in less than N minutes


mellenburg:

PR #794, opened 2016-09-28
Accept cookie="" in auth logout test

mellenburg:


cmaloney:
:ship:

jgehrcke:
Please hold this until we have understood where the changes comes from.

packages/dcos-integration-test/extra/test_auth.py
```diff
@@ -42,5 +42,5 @@ def test_logout(auth_cluster):
     """
     r = auth_cluster.get('/acs/api/v1/auth/logout')
     cookieheader = r.headers['set-cookie']
-    assert 'dcos-acs-auth-cookie=;' in cookieheader
+    assert 'dcos-acs-auth-cookie=;' in cookieheader or 'dcos-acs-auth-cookie=""' in cookieheader
```
jgehrcke: Can you add a `;` for the second case?
```
or 'dcos-acs-auth-cookie="";'
```

jgehrcke:
Okay, this is now properly triaged and indeed expected as of the update to CPython 3.5.2 and the corresponding http.cookie stdlib changes this brings along.

@mellenburg I have added one commend and asked for addition of one semicolon. After that this LGTM.

PR #795, opened 2016-09-29
Improvements to support AWS govcloud

lingmann:
* Fix missing NATAmi mapping in the Simple Template for the govcloud region
* Set the S3 Exhibitor ARN to the correct values required for us-gov-west-1
* Add CentOS7 base AMI images in the govcloud region for Advanced Templates

### DIFF
--- a/dcos_installer/backend.py
+++ b/dcos_installer/backend.py
@@ -56,7 +56,8 @@ region_to_endpoint = {
     'ap-northeast-1': 's3-ap-northeast-1.amazonaws.com',
     'eu-central-1': 's3.eu-central-1.amazonaws.com',
     'eu-west-1': 's3-eu-west-1.amazonaws.com',
-    'sa-east-1': 's3-sa-east-1.amazonaws.com'
+    'sa-east-1': 's3-sa-east-1.amazonaws.com',
+    'us-gov-west-1': 's3-us-gov-west-1.amazonaws.com'
 }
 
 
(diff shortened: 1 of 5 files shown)


cmaloney:
More things that will need updating: https://github.com/dcos/dcos/blob/master/gen/installer/aws.py#L19
https://github.com/dcos/dcos/blob/master/dcos_installer/backend.py#L48

lingmann:
@cmaloney thanks for catching that. So the buttons page will not be that useful without jumping through some more hoops. Specifically, the AWS console URL for govcloud must be changed from:

https://console.aws.amazon.com/cloudformation/

to:

https://console.amazonaws-us-gov.com/cloudformation/

And we would also have to start publishing the templates to a govcloud s3 endpoint (you can't reference https://s3-us-west-2.amazonaws.com/downloads.dcos.io/... when launching a stack from a govcloud account.

I'm inclined to just not provide buttons for govcloud. All of the templates will still work if you manually submit them to cloudformation (with the exception of the zen template, since it refers to non-govcloud s3 locations).

lingmann:
Removed the aws.html button support for govcloud due to the region being enough of a snowflake that it is not worth adding at this point.

lingmann:
@cmaloney this is ready for review. In addition to the AWS integration test, I've tested deploying both simple templates and advanced templates from this build to the govcloud region.

PR #796, opened 2016-09-29
 Allow deploying centos advanced templates

mellenburg:


cmaloney:
@branden for review

mellenburg:
CentOs test passes: https://teamcity.mesosphere.io/viewLog.html?buildId=424391&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsAd_2&tab=buildResultsDiv

test_util/aws.py
```diff
@@ -230,7 +230,11 @@ def create(cls, stack_name, boto_wrapper, template_url,
             'PrivateAgentInstanceType': private_agent_type,
             'PrivateSubnet': private_subnet}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_name, boto_wrapper), SSH_INFO['coreos']
+        if template_url.split('/')[-1].startswith('coreos'):
+            os_name = 'coreos'
+        else:
+            os_name = 'centos'
```
branden: If we add a template that uses a different OS that this code doesn't expect, it'd be easier to debug an exception here rather than an SSH error later. It looks like the advanced templates have their own set of aliases for AMIs, so let's add a `CF_OS_SSH_INFO` dict alongside the others with the keys `'coreos'` and `'el7'`. Then this can return `CF_OS_SSH_INFO[template_url.split('/')[-1].split('-')[0]]`. Those `split()`s and indexes assume a lot about the URL structure that isn't asserted elsewhere in this method, so it might be useful to handle `(KeyError, IndexError)` there by raising an exception that mentions the template URL.

mellenburg:
Note: I just realized parsing out the OS in this way wont work with variants as written

mellenburg:
@branden PTAL

test_util/aws.py
```diff
@@ -258,7 +258,16 @@ def create(cls, stack_name, boto_wrapper, template_url,
             'PrivateAgentInstanceType': private_agent_type,
             'PrivateSubnet': private_subnet}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_name, boto_wrapper), SSH_INFO['coreos']
+        try:
+            os_string = template_url.split('/')[-1].split('.')[-2].split('-')[0]
+            ssh_info = CF_OS_SSH_INFO[os_string]
+        except KeyError:
+            logging.exception('No SSH info found corresponding to OS string: {}'.format(os_string))
```
branden: It'd be handy to have the template URL here as well since this error can also be caused by an unexpected URL format. You can use the same handler for both exceptions since they're describing the same error (unexpected template URL).

test_util/aws.py
```diff
@@ -258,7 +258,15 @@ def create(cls, stack_name, boto_wrapper, template_url,
             'PrivateAgentInstanceType': private_agent_type,
             'PrivateSubnet': private_subnet}
         stack = boto_wrapper.create_stack(stack_name, template_url, parameters)
-        return cls(stack.stack.stack_name, boto_wrapper), SSH_INFO['coreos']
+        try:
+            os_string = template_url.split('/')[-1].split('.')[-2].split('-')[0]
```
cmaloney: nit: longer term I'd much rather build this into a "variant" + template name which we break apart, then just do a lookup per general template url shape. For now this works though.

PR #799, opened 2016-09-29
WIP: dcos-history service not returning any data

amitaekbote:
dcos-history is not returning data, current suspicion is that the buffer update thread is not running or running but not updating. Modifying the tests to see if we catch it in integration test run

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,3 +159,4 @@ Lots of docs are still being written. If you have immediate questions please ask
  - Add general theory of stuff that goes in here.
  - PR (guidelines, testing)
  - How to make different sorts of common changes
+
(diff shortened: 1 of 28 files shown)


amitaekbote:
@cmaloney mark this as work in progress

cmaloney:
So I think the bug that was filed is different than what this is doing.

Zero slaves showed up because the first request to mesos there were no agents (which is just fine, they haven't registered yet).

The problem is after the cluster did finish standing up, that latest buffer never seems to get updated / the "most recent" datapoint never changes / gets updated (Although the history service backend data gatherer does keep running)

PR #806, opened 2016-09-30
Fixing dcos history service to fail on exceptions and adding more tests

amitaekbote:
on_starting : Called just before the master process is initialized.

pre_fork : Called just before a worker is forked.


amitaekbote:
@mellenburg  @cmaloney  thoughts on having the tests in test_endpoints vs cluster_api? Currently its doing the same test in both

once the tests pass this should be ready for review

amitaekbote:
@cmaloney ready for review

test_util/cluster_api.py
```diff
@@ -87,6 +87,24 @@ def _wait_for_dcos_history_up(self):
     @retrying.retry(wait_fixed=1000,
                     retry_on_result=lambda ret: ret is False,
                     retry_on_exception=lambda x: False)
+    def _wait_for_dcos_history_data(self):
+        for _ in range(10):
+            # call 10 times to make sure you get something, might need to sleep instead
```
mellenburg: What is the purpose of iterating this ten times? If the request response is bad just return False to trigger the retry

test_util/cluster_api.py
```diff
@@ -87,6 +87,24 @@ def _wait_for_dcos_history_up(self):
...
+            ro = self.get('/dcos-history-service/history/last')
+        # resp_code >= 500 -> backend is still down probably
+        if ro.status_code <= 500:
+            logging.info("DC/OS History is probably getting data")
+            json = ro.json()
+            assert len(json["slaves"]) != 0
```
mellenburg: I would prefer that we just assert that we get all the expected slaves rather than only asserting nonzero, but I could also see the argument that this waiting-style check should not make specific assertions.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -116,15 +117,24 @@ def test_if_pkgpanda_metadata_is_available(cluster):
...
+        assert r.status_code == 200
+        # Make sure some basic fields are present from state-summary which the DC/OS
+        # UI relies upon. Their exact content could vary so don't test the value.
+        json = r.json()
+        assert 'cluster' in json
+        assert 'frameworks' in json
+        assert 'slaves' in json
+        assert 'hostname' in json
+
+        r2 = cluster.get('/dcos-history-service/history/last')
```
mellenburg: Why do we need to get the same JSON again?

packages/dcos-history/extra/history/server.py
```diff
@@ -1,93 +1,8 @@
-import logging
 import os
-import sys
+import history.server_util
```
mellenburg: What is the motivation for splitting out all the business logic into 'server_util' if the resulting server module is almost completely trivial?

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -116,15 +117,24 @@ def test_if_pkgpanda_metadata_is_available(cluster):
...
+        assert r.status_code == 200
+        json2 = r2.json()
+        assert len(json2["slaves"]) != 0
```
mellenburg: Can we check that the number of slaves matches the expected number of slaves?

test_util/cluster_api.py
amitaekbote: I think the retry should be enough, I will get rid of the for loop

packages/dcos-integration-test/extra/test_endpoints.py
amitaekbote: I will get rid of this, the retry should take care of failures

packages/dcos-history/extra/history/server.py
amitaekbote: The motivation for splitting this was we did not want to invoke the thread for unit testing, if name main doesnt work with gunicorn, not having if name main doesnt work for unit tests. Separating is the cleanest way of doing this. We unit test for the state buffer directly without invoking it using app

mellenburg: > "Not having if __name__ == '__main__' doesn't work for unit tests"
I don't follow. Unit tests weren't invoked by calling the server file as main

What I find awkward here is the whole purpose of this module is to create a thread bound to a global variable *on import*, just so we can invoke gunicorn and have this still work. Possible alternatives:
(1) Use a gunicorn setting like post_worker_init to start the thread within the GIL of the worker and enforce that we only have 1 worker. This is exactly what the history service is doing now (flask only supports a single worker)
(2) Don't use gunicorn at all. It seems to be adding a lot more magic. If the goal of this PR is to catch exceptions and fail hard, then its not clear how this PR accomplishes that other than trusting the gunicorn will fulfill this purpose

cmaloney:
in holding for 1.9

packages/dcos-history/extra/history/server.py
cmaloney: The goal long term would be to move to aiohttp and coroutines for the helper rather than mixing threading and python. The old threading is a big hack as well (and that it worked is some very specific trickery). 

All the refactoring and code shape changes this gets get us closer to what that will need to be (see https://github.com/dcos/dcos/blob/master/dcos_installer/async_server.py)

Gunicorn is how we run the other servers we have. That flask doesn't bail on port bind failing is a bug in flask. Can't really fix that.

Doing a "is this port available" check and then doing the "start flask" doesn't actually fix the issue. Flask's loop may still start, someone else grabbed the port between the "check port" and "start flask", flask continues in a loop while not actually listening anywhere.

packages/dcos-history/extra/history/server_util.py
```diff
@@ -0,0 +1,116 @@
+import logging
```
cmaloney: nit: Moving this code to a different file would have been nice to have separate from the code structure changes.

packages/dcos-history/extra/history/server_util.py
```diff
@@ -0,0 +1,116 @@
...
+compress = Compress()
+state_buffer = None
+log = logging.getLogger(__name__)
+add_headers_cb = None
+
+
+try:
+    import dcos_auth_python
+    log.info('dcos_auth_python module detected; applying settings')
+    global add_headers_cb
```
cmaloney: /me is surprised a `global` is needed here. This isn't in a function.

packages/dcos-history/extra/history/server_util.py
```diff
@@ -0,0 +1,116 @@
...
+
+    compress.init_app(app)
+
+    if 'HISTORY_BUFFER_DIR' not in os.environ:
+        sys.exit('HISTORY_BUFFER_DIR must be set!')
+
+    global state_buffer
+    state_buffer = BufferCollection(os.environ['HISTORY_BUFFER_DIR'])
+
+    def update():
```
cmaloney: definitely looks like this update() function could be defined at the top level rather then nested.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -116,15 +117,24 @@ def test_if_pkgpanda_metadata_is_available(cluster):
...
-    assert 'frameworks' in json
-    assert 'slaves' in json
-    assert 'hostname' in json
+    @retry(stop_max_delay=20000, wait_fixed=500)
+    def check_up():
+        r = cluster.get('/dcos-history-service/history/last')
+        assert r.status_code == 200
+        # Make sure some basic fields are present from state-summary which the DC/OS
+        # UI relies upon. Their exact content could vary so don't test the value.
+        json = r.json()
```
cmaloney: `assert {'cluster', 'frameworks', 'slaves', 'hostname'} <= json.keys()`

That way we get a nice structured error message with a diff of the dictionaries if there's an issue.

PR #807, opened 2016-09-30
Switch dcos overlay configuration to be json.dump() rather than hand-coded json

cmaloney:
cc: @sargun @asridharan


asridharan:
makes sense.

🚢

PR #808, opened 2016-09-30
Add cluster_docker_credentials configuration

cmaloney:
The cluster_docker_credentials can either be owned by DC/OS or a sysadmin's
configuration management system (ansible, chef, puppet, salt, etc.).

Core config options:
```
cluster_docker_credentials:
    Dictionary of credentials to pass.

    If unset / left the default, this will cause an empty credentials file to be placed in
    /etc/mesosphere/docker_credentials on DC/OS install. Sysadmin scripts can then edit that file
    and do a `systemctl restart dcos-mesos-slave` / `systemctl restart dcos-mesos-slave-public`.

    It can also be set as: http://mesos.apache.org/documentation/latest/configuration/
    '--docker_config' json format. Note that in config.yaml just write yaml and it will be mapped to
    the JSON format for you. This will cause the docker credentials configuration to be part of the
    DC/OS internal configuration (live in /opt/mesosphere) and the only way to update / change it is
    to roll out a new DC/OS internal configuration.

    More precise behavior can be controlled with the following options:

    cluster_docker_credentials_dcos_owned: <true|false>
        If true, then credentials file will always be in /opt/mesosphere. If false, it will always
        be in /etc/mesosphere/docker_credentials. Defaults to false if and only if the credentials are default / unset.

    cluster_docker_credentials_write_to_etc: <true|false>
        Only used if cluster_docker_credentials_dcos_owned is true. If set to true, then rather than
        writing a docker credentials file to /etc/mesosphere/docker_credentials on DC/OS install, no
        file is written. This is useful if the file is baked into an machine image / AMI and
        installing DC/OS overwriting it will cause problems.

    cluster_docker_credentials_enabled: <true|false>
        Defaults to true. If true, then the mesos --docker_config option will be passed to Mesos. If
        false, it will not.

cluster_docker_registry_url: <url>
    Optional. If set, will configure Mesos' `--docker_registry` flag to the given url. This changes the default URL Mesos uses for pulling docker images. By default https://registry-1.docker.io is used.
```

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -198,8 +198,14 @@ def calculate_rexray_config_contents(rexray_config):
     )
 
 
-def validate_rexray_config(rexray_config):
-    assert isinstance(json.loads(rexray_config), dict), 'Must be a mapping.'
+def validate_json_dictionary(data):
+    # TODO(cmaloney): Pull validate_json() out.
+    try:
+        loaded = json.loads(data)
+        assert isinstance(loaded, dict), "Must be a JSON dictionary. Got a {}".format(type_str(loaded))
+        return loaded
+    except ValueError as ex:
+        raise AssertionError("Must be valid JSON. Got: {}".format(data)) from ex
 
 
 def calculate_gen_resolvconf_search(dns_search):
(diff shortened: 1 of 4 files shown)


pyronicide:
👍  from me.

PR #813, opened 2016-10-02
Add Cluster Marathon Client

mellenburg:
-Adds generic get_client to cluster_api by which one may set an API base, user auth,  and default headers to make requests
-Adds marathon api client with marathon helpers for extended functionality

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -1,5 +1,7 @@
 import uuid
 
+from test_util.marathon import get_test_app, get_test_app_in_docker
+
 
 def test_if_marathon_app_can_be_deployed(cluster):
     """Marathon app deployment integration test
@@ -15,8 +17,7 @@ def test_if_marathon_app_can_be_deployed(cluster):
     "GET /test_uuid" request is issued to the app. If the returned UUID matches
     the one assigned to test - test succeds.
     """
-    app, test_uuid = cluster.get_test_app()
-    cluster.deploy_test_app_and_check(app, test_uuid)
+    cluster.marathon.deploy_test_app_and_check(*get_test_app())
 
 
 def test_if_docker_app_can_be_deployed(cluster):
@@ -25,8 +26,7 @@ def test_if_docker_app_can_be_deployed(cluster):
     Verifies that a marathon app inside of a docker daemon container can be
     deployed and accessed as expected.
     """
-    app, test_uuid = cluster.get_test_app_in_docker(ip_per_container=False)
-    cluster.deploy_test_app_and_check(app, test_uuid)
+    cluster.marathon.deploy_test_app_and_check(*get_test_app_in_docker(ip_per_container=False))
 
 
 def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
@@ -43,7 +43,7 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
     When port mapping is available (MESOS-4777), this test should be updated to
     reflect that.
     """
-    app, test_uuid = cluster.get_test_app()
+    app, test_uuid = get_test_app()
     app['container'] = {
         'type': 'MESOS',
         'docker': {
@@ -56,7 +56,7 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
             'mode': 'RO'
         }]
     }
-    cluster.deploy_test_app_and_check(app, test_uuid)
+    cluster.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
 def test_octarine_http(cluster, timeout=30):
@@ -90,7 +90,7 @@ def test_octarine_http(cluster, timeout=30):
         }]
     }
 
-    cluster.marathon_deploy_and_cleanup(app_definition)
+    cluster.marathon.deploy_and_cleanup(app_definition)
 
 
 def test_octarine_srv(cluster, timeout=30):
@@ -137,7 +137,7 @@ def test_octarine_srv(cluster, timeout=30):
         }]
     }
 
-    cluster.marathon_deploy_and_cleanup(app_definition)
+    cluster.marathon.deploy_and_cleanup(app_definition)
 
 
 def test_pkgpanda_api(cluster):
--- a/packages/dcos-integration-test/extra/test_auth.py
+++ b/packages/dcos-integration-test/extra/test_auth.py
@@ -8,8 +8,13 @@ def auth_cluster(cluster):
     return cluster
 
 
-def test_adminrouter_access_control_enforcement(auth_cluster):
-    r = auth_cluster.get('/acs/api/v1', user=None)
+@pytest.fixture(scope='module')
+def noauth_cluster(auth_cluster):
+    return auth_cluster.get_user_session(None)
+
+
+def test_adminrouter_access_control_enforcement(auth_cluster, noauth_cluster):
+    r = noauth_cluster.get('/acs/api/v1')
     assert r.status_code == 401
     assert r.headers['WWW-Authenticate'] in ('acsjwt', 'oauthjwt')
     # Make sure that this is UI's error page body,
@@ -20,7 +25,7 @@ def test_adminrouter_access_control_enforcement(auth_cluster):
     # Verify that certain locations are forbidden to access
     # when not authed, but are reachable as superuser.
     for path in ('/mesos_dns/v1/config', '/service/marathon/', '/mesos/'):
-        r = auth_cluster.get(path, user=None)
+        r = noauth_cluster.get(path)
         assert r.status_code == 401
         r = auth_cluster.get(path)
         assert r.status_code == 200
@@ -28,9 +33,8 @@ def test_adminrouter_access_control_enforcement(auth_cluster):
     # Test authentication with auth cookie instead of Authorization header.
     authcookie = {
         'dcos-acs-auth-cookie': auth_cluster.web_auth_default_user.auth_cookie}
-    r = auth_cluster.get(
+    r = noauth_cluster.get(
         '/service/marathon/',
-        user=None,
         cookies=authcookie)
     assert r.status_code == 200
 
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -3,6 +3,8 @@ import logging
 import requests
 import retrying
 
+from test_util.marathon import get_test_app, get_test_app_in_docker
+
 
 def ensure_routable(cmd, service_points, timeout=120):
     @retrying.retry(wait_fixed=2000,
@@ -21,11 +23,11 @@ def ensure_routable(cmd, service_points, timeout=120):
 def test_if_minuteman_routes_to_vip(cluster):
     """Test if we are able to connect to a task with a vip using minuteman.
     """
-    origin_app, origin_uuid = cluster.get_test_app()
+    origin_app, origin_uuid = get_test_app()
     origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '1.2.3.4:5000'}
-    with cluster.marathon_deploy_and_cleanup(origin_app):
-        proxy_app, proxy_uuid = cluster.get_test_app()
-        with cluster.marathon_deploy_and_cleanup(proxy_app) as service_points:
+    with cluster.marathon.deploy_and_cleanup(origin_app):
+        proxy_app, proxy_uuid = get_test_app()
+        with cluster.marathon.deploy_and_cleanup(proxy_app) as service_points:
             cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://1.2.3.4:5000/ping'
             ensure_routable(cmd, service_points)()
 
@@ -34,11 +36,11 @@ def test_if_minuteman_routes_to_named_vip(cluster):
     """Test if we are able to connect to a task with a named vip using minuteman.
     """
 
-    origin_app, origin_uuid = cluster.get_test_app()
+    origin_app, origin_uuid = get_test_app()
     origin_app['portDefinitions'][0]['labels'] = {'VIP_0': 'foo:5000'}
-    with cluster.marathon_deploy_and_cleanup(origin_app):
-        proxy_app, proxy_uuid = cluster.get_test_app()
-        with cluster.marathon_deploy_and_cleanup(proxy_app) as service_points:
+    with cluster.marathon.deploy_and_cleanup(origin_app):
+        proxy_app, proxy_uuid = get_test_app()
+        with cluster.marathon.deploy_and_cleanup(proxy_app) as service_points:
             cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://foo.marathon.l4lb.thisdcos.directory:5000/ping'
             ensure_routable(cmd, service_points)()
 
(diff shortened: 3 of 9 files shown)


mellenburg:
@cmaloney This is ready for review. If desired, I can open a new PR just for adding the client

mellenburg:
@alberts  This is critical for unifying tests PTAL

packages/dcos-integration-test/extra/test_service_discovery.py
```diff
@@ -72,7 +74,7 @@ def _service_discovery_test(cluster, docker_network_bridge):
 
     app_definition["constraints"] = [["hostname", "UNIQUE"], ]
 
-    with cluster.marathon_deploy_and_cleanup(app_definition) as service_points:
+    with cluster.marathon().deploy_and_cleanup(app_definition) as service_points:
```
cmaloney: I don't see a good reason not to just make this do a `marathon` / take the marathon fixture. the `cluster.slaves` count we can move into the base cluster fixture if it's not there already.

test_util/cluster_api.py
```diff
@@ -271,244 +273,29 @@ def get_url(self, node, path, port=None):
...
-
-        try:
-            _destroy_complete(r.json()['deploymentId'])
-        except retrying.RetryError:
-            raise Exception("Application destroy failed - operation was not "
-                            "completed in {} seconds.".format(timeout))
+    def get_client(self, path, user='default', default_headers={}):
+        return test_util.helpers.ApiClient(
+            cluster=self,
+            user=self._get_user(user),
```
cmaloney: why do we resolve user here but not just pass in headers? Seems like this should either just take the user name or take the auth headers.

test_util/cluster_api.py
```diff
@@ -271,244 +273,29 @@ def get_url(self, node, path, port=None):
...
 
-    @contextmanager
-    def marathon_deploy_and_cleanup(self, app_definition, timeout=120, check_health=True, ignore_failed_tasks=False):
-        yield self.deploy_marathon_app(
-            app_definition, timeout, check_health, ignore_failed_tasks)
-        self.destroy_marathon_app(app_definition['id'], timeout)
+    def marathon(self, user='default'):
+        return test_util.marathon_api.MarathonApi.get_client(
+            cluster=self,
+            user=self._get_user(user))
```
cmaloney: Why is this pattern different than the get_client one above? seems like all get_client calls should expect the same thing for user.

test_util/cluster_api.py
```diff
@@ -271,244 +273,29 @@ def get_url(self, node, path, port=None):
...
+            user=self._get_user(user),
+            api_base=path,
+            default_headers=default_headers)
 
-    @contextmanager
-    def marathon_deploy_and_cleanup(self, app_definition, timeout=120, check_health=True, ignore_failed_tasks=False):
-        yield self.deploy_marathon_app(
-            app_definition, timeout, check_health, ignore_failed_tasks)
-        self.destroy_marathon_app(app_definition['id'], timeout)
+    def marathon(self, user='default'):
```
cmaloney: `@property`?

test_util/cluster_api.py
```diff
@@ -271,244 +273,29 @@ def get_url(self, node, path, port=None):
...
+        return test_util.marathon_api.MarathonApi.get_client(
+            cluster=self,
+            user=self._get_user(user))
 
     def metronome_one_off(self, job_definition, timeout=300, ignore_failures=False):
         """Run a job on metronome and block until it returns success
         """
         job_id = job_definition['id']
+        metronome_api = self.get_client('/service/metronome/v1')
```
cmaloney: I'd include the trailing slash here.

Also might be handy to make a generic helper that is `cluster.metronome` where the metronome client can be accessed rather than a one-off in this function.

test_util/helpers.py
```diff
@@ -31,6 +30,35 @@ def authenticate(self, cluster):
         logging.info('Authentication successful')
 
 
+class ApiClient:
+    def __init__(self, cluster, user, api_base, default_headers={}):
```
cmaloney: python doesn't like passing dictionaries as default arguments, should make the default `None` and do a `if default_headers is None: default_headers = dict()`

test_util/helpers.py
```diff
@@ -31,6 +30,35 @@ def authenticate(self, cluster):
...
+        self.cluster = cluster
+        self.user = user
+        self.api_base = api_base
+        self.headers = default_headers
+
+    def api_request(self, method, path, **kwargs):
+        """Wrapper users the set user to this API instance's user,
+        provides required headers, appends api base, and passes all else
+        """
+        cluster_request = getattr(self.cluster, method)
```
cmaloney: this doesn't seem like getattr is really needed. Why can't this just do a little bit of wrapping and call `self.cluster.cluster_request(method, ...)`?

test_util/helpers.py
```diff
@@ -31,6 +30,35 @@ def authenticate(self, cluster):
...
+        headers = self.headers
+        if self.user:
+            headers.update(self.user.auth_header)
+        headers.update(kwargs.pop('headers', {}))
+        path = self.api_base + path
+        return cluster_request(path, user=None, headers=headers, **kwargs)
```
cmaloney: that we go from user -> headers -> call a function where we could pass user manually here is odd. Do we need to do the user resolution in this function or could we just let cluster_request() do it?

test_util/helpers.py
```diff
@@ -31,6 +30,35 @@ def authenticate(self, cluster):
...
+
+    get = functools.partialmethod(api_request, 'get')
```
cmaloney: nit: Would be nice if we didn't have to duplicate these between here and the `Cluster` class

test_util/marathon_api.py
```diff
@@ -0,0 +1,242 @@
...
+def get_test_app(custom_port=False):
+    test_uuid = uuid.uuid4().hex
+    app = copy.deepcopy({
+        'id': TEST_APP_NAME_FMT.format(test_uuid),
+        'cpus': 0.1,
+        'mem': 32,
+        'instances': 1,
+        # NOTE: uses '.' rather than `source`, since `source` only exists in bash and this is
+        # run by sh
+        'cmd': '. /opt/mesosphere/environment.export && /opt/mesosphere/bin/python '
```
cmaloney: note: will conflict with dcos-shell patchset

test_util/marathon_api.py
```diff
@@ -0,0 +1,242 @@
...
+        app['container']['docker']['network'] = 'USER'
+        app['ipAddress'] = {'networkName': 'dcos'}
+    else:
+        app['container']['docker']['network'] = 'BRIDGE'
+    return app, test_uuid
+
+
+class MarathonApi(test_util.helpers.ApiClient):
+    @classmethod
+    def get_client(cls, cluster, user):
```
cmaloney: why can't this be in the constructor? Wrap the user provided default_headers with one new header?

test_util/marathon_api.py
```diff
@@ -0,0 +1,242 @@
...
+            if r.status_code != 200:
+                msg = "Test server replied with non-200 reply: '{0} {1}. "
+                msg += "Detailed explanation of the problem: {2}"
+                raise Exception(msg.format(r.status_code, r.reason, r.text))
+
+            r_data = r.json()
+
+            assert r_data['test_uuid'] == test_uuid
+
+            # Test the app is running as root
```
cmaloney: nit: Not caused by this PR but this comment is out of date.

mellenburg:
@cmaloney I did another pass on this PTAL. Downstream looks MUCH better with this revision

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -15,8 +17,8 @@ def test_if_marathon_app_can_be_deployed(cluster):
     "GET /test_uuid" request is issued to the app. If the returned UUID matches
     the one assigned to test - test succeds.
     """
-    app, test_uuid = cluster.get_test_app()
-    cluster.deploy_test_app_and_check(app, test_uuid)
+    app, test_uuid = get_test_app()
+    cluster.marathon.deploy_test_app_and_check(app, test_uuid)
```
cmaloney: nit: `cluster.marathon.deploy_test_app_and_check(*get_test_app())`

packages/dcos-integration-test/extra/test_auth.py
```diff
@@ -20,17 +25,16 @@ def test_adminrouter_access_control_enforcement(auth_cluster):
     # Verify that certain locations are forbidden to access
     # when not authed, but are reachable as superuser.
     for path in ('/mesos_dns/v1/config', '/service/marathon/', '/mesos/'):
-        r = auth_cluster.get(path, user=None)
+        r = auth_cluster.get_user_session(None).get(path)
```
cmaloney: isn't this just `noauth_cluster.get(path)`?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -21,11 +23,11 @@ def _ensure_routable():
...
     """Test if we are able to connect to a task with a vip using minuteman.
     """
-    origin_app, origin_uuid = cluster.get_test_app()
+    origin_app, origin_uuid = get_test_app()
     origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '1.2.3.4:5000'}
-    with cluster.marathon_deploy_and_cleanup(origin_app):
-        proxy_app, proxy_uuid = cluster.get_test_app()
-        with cluster.marathon_deploy_and_cleanup(proxy_app) as service_points:
+    with cluster.marathon.deploy_and_cleanup(origin_app):
+        proxy_app, proxy_uuid = get_test_app()
```
cmaloney: nit / not yours but typically in python when there's a "throwaway" / variable not going to be used (`proxy_uuid`), generally just assign it to `_`, so this would be `proxy_app, _ = cluster.get_test_app()`. (This and the couple others around it are also suggesting returning the uuid isn't actually the best option, but that would definitely be an out of scope change).

test_util/cluster_api.py
```diff
@@ -241,280 +241,43 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
-
-    def get_url(self, node, path, port=None):
-        """Looks up node type and appends the appropriate port and scheme
-        """
-        if node in self.masters or node == self.dns_host:
-            default_port = self.adminrouter_master_port[self.scheme]
-        elif node in self.all_slaves:
-            default_port = self.adminrouter_agent_port[self.scheme]
+    def get_user_session(self, user):
+        new_session = copy.deepcopy(self)
```
cmaloney: why are we copying the self object here? That seems a lot more expensive and unnecessary than necessary... What magic internal state is this needing to change?

test_util/cluster_api.py
```diff
@@ -241,280 +241,43 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
-
-    @contextmanager
-    def marathon_deploy_and_cleanup(self, app_definition, timeout=120, check_health=True, ignore_failed_tasks=False):
-        yield self.deploy_marathon_app(
-            app_definition, timeout, check_health, ignore_failed_tasks)
-        self.destroy_marathon_app(app_definition['id'], timeout)
+            raise Exception('Node {} not found in DC/OS cluster'.format(node))
+
+    def get_client(self, path, default_headers=None):
+        if default_headers is None:
```
cmaloney: looking at this again, ApiClient already does the None -> dict() logic, so doing nothing here and just letting the None fall through into ApiClient's __init__ is plenty

test_util/cluster_api.py
```diff
@@ -525,7 +288,7 @@ def metronome_one_off(self, job_definition, timeout=300, ignore_failures=False):
                         retry_on_result=lambda ret: not ret,
                         retry_on_exception=lambda x: False)
         def wait_for_completion():
-            r = self.get('/service/metronome/v1/jobs/' + job_id, params={'embed': 'history'})
+            r = self.metronome.get('/jobs/' + job_id, params={'embed': 'history'})
```
cmaloney: no leading slash needed here (and the other metronome ones) I think because get_client() includes a `/` at the end.

test_util/helpers.py
```diff
@@ -25,9 +25,9 @@ def __init__(self, auth_json):
         self.auth_cookie = None
 
     def authenticate(self, cluster):
-        assert isinstance(cluster, test_util.cluster_api.ClusterApi), 'Unrecognized cluster object!'
         logging.info('Attempting authentication')
-        r = cluster.post(path='/acs/api/v1/auth/login', user=None, json=self.auth_json)
+        # Get a non-user session so that auth headers are absent
+        r = cluster.get_user_session(None).post(path='/acs/api/v1/auth/login', json=self.auth_json)
```
cmaloney: why doesn't just cluster.post() work here anymore?

test_util/helpers.py
```diff
@@ -37,6 +37,81 @@ def authenticate(self, cluster):
...
 
+class ApiClient:
+
+    def __init__(self, cluster, user, api_base, default_headers=None, ca_cert_path=None):
+        self.cluster = cluster
+        self.user = user
+        self.api_base = api_base
+        if default_headers is None:
+            default_headers = dict()
+        self.headers = default_headers
```
cmaloney: nit: I'd rather be consistent on this naming everywhere, either always default_headers or always headers.

test_util/helpers.py
```diff
@@ -37,6 +37,81 @@ def authenticate(self, cluster):
...
+        self.ca_cert_path = ca_cert_path
+        self.port_mapping = {
```
cmaloney: why is this a member variable and not just a static thing that is accessed?

test_util/helpers.py
```diff
@@ -37,6 +37,81 @@ def authenticate(self, cluster):
...
+        else:
+            role = self.cluster.get_node_role(node)
+
+        if self.api_base:
+            path = '{}/{}'.format(self.api_base.rstrip('/'), path.lstrip('/'))
+
+        if port is None:
+            port = self.port_mapping[role][self.cluster.scheme]
+
+        return '{}://{}/{}'.format(self.cluster.scheme, '{}:{}'.format(node, port), path.lstrip('/'))
```
cmaloney: very tiny nit: can we make this leave of the port if it's port 80 and http or port 443 and https? (Practically leaving it doesn't really harm and is actually more explicit, just from web browsers and the like I'm used to it being folded away).

test_util/marathon.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def get_test_app(custom_port=False):
+    test_uuid = uuid.uuid4().hex
+    app = copy.deepcopy({
+        'id': TEST_APP_NAME_FMT.format(test_uuid),
+        'cpus': 0.1,
+        'mem': 32,
+        'instances': 1,
+        # NOTE: uses '.' rather than `source`, since `source` only exists in bash and this is
```
cmaloney: nit: This one's 100% my fault (need more code reviews of my code...)... this doesn't use `.` or `source` anymore....

test_util/marathon.py
```diff
@@ -0,0 +1,243 @@
...
+    if ip_per_container:
+        app['container']['docker']['network'] = 'USER'
+        app['ipAddress'] = {'networkName': 'dcos'}
+    else:
+        app['container']['docker']['network'] = 'BRIDGE'
+    return app, test_uuid
+
+
+class Marathon(test_util.helpers.ApiClient):
+    def __init__(self, cluster, user, api_base='/marathon/v2', default_headers=None, ca_cert_path=None):
```
cmaloney: nit: Can api_base be `/marathon/v2/` and then all the other calls don't have to put the `/` at the start?

test_util/cluster_api.py
mellenburg: The only thing that changes is the web_auth_default_user. The magic state lies in whether the user has been authenticated yet, which is resolved here by doing that first. I could instead do:
```
if user:
    user.authenticate(self)
return self.__class__(<pass all the necessary args w/ user as the new web_auth_default_user>)
```
but the reason I didnt opt for that is I liked how clean it looked to do copy.deepcopy instead of having to set 11 args

mellenburg:
@cmaloney I've made the obvious changes to this and left comments on where I differed.Specifically:
(1) Use deepcopy instead of returning another instance for brevity
(2) Use a member variable for port mapping to allow APIs to dictate which ports should be used when talking to certain types of nodes
(3) Enforcing that we have no auth headers in our requests when we try to authenticate a user

mellenburg:
@lingmann PTAL

test_util/helpers.py
cmaloney: get_user_session(None) creates a copy of the whole object every time, could we say `if user is none then .post() otherwise do the copy`?

test_util/helpers.py
mellenburg: Point taken on vetoing static member variables. We can return to my settable port mapping when I actually implement an API that would use it.

mellenburg:
@cmaloney Comments have been addressed

cmaloney:
🚢

PR #814, opened 2016-10-03
Custom genconf directory

jcortejoso:
Hi all!

This PR allows overriding the default location of genconf folder during the installation process (default location is $(pwd)/genconf; wherever it has been run). Exporting the environmental variable $LOCAL_GENCONF_FOLDER lets you to set the genconf location somewhere else. This can be handy for using the same bootstrap for deploying different cluster without worrying about moving the installer script dcos_generate_config.sh.

Any comment is welcome. Best regards.


cmaloney:
So this is definitely not a knob I'd want to advertise / document generally, people already get really confused by "/genconf" showing up sometimes, and making it so the user folder doesn't match what gets used I think will cause more confusion. I can get why this would be handy in some tooling (although from my perspective is there any reason this is easier + less confusing than just `mv genconf genconf-a && mv genconf-b genconf` type thing?

cmaloney:
As a note also, the mid-term plan is that dcos_generate_config.sh stops being a self-extracting docker and ideally becomes something like a PyInstaller binary at which point the config.yaml path + output folder could be fully configurable.

jcortejoso:
Hi cmaloney! Thanks a lot for your feedback.

The needed for that in my workflow I run dcos-generate-config.sh from inside a docker container. To avoid running docker-inside-docker I run dcos bootstrap image directly in the host (exposing docker socket inside the first container). Right now I map the same locations out and inside the containers, with it would be very pleasant if I could configure this folder and I would avoid those kinds of hacks.

jcortejoso:
Any thought about last comment?

Thanks!!!

cmaloney:
@jcortejoso sorry it's been a bit getting back to you. The approach I'd like to take is making dcos_generate_config.sh not be a docker at all. I have a proof of concept PR to start that work: https://github.com/dcos/dcos/pull/881

With that it would be possible to just have the path to config.yaml + the output directory be passed in as arguments (rather than the current where we have them not configurable because every time we made them configurable users got stuck / lost / confused when files didn't show up).

A shorter term "temporary" approach would be to make the output location + config.yaml locations configurable so long as the path starts with `/genconf`.

PR #816, opened 2016-10-03
Revert "Handle connection error on dcos-history"

cmaloney:
This reverts commit 9cf07a7ce10e3a3689725ec86c7e37f4a987c2be.

Short term to make the UI look good until a proper fix is completed / can be landed.

cc: @amitaekbote


amitaekbote:
looks good to me , thanks!

PR #830, opened 2016-10-05
Handle version number with decimals

luigiberrettini:
Our prereq installation was failing because we were on RHEL 7.2

It would be nice if instead of a 700MB shell script you created a compressed file containing scripts separated from binaries, so that people can fix errors on the fly.

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -351,7 +351,7 @@ if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
   exit 0
 fi
 
-version=$(cat /etc/*-release | sed -n 's@^VERSION_ID="\(.*\)"$@\\1@p')
+version=$(cat /etc/*-release | sed -n 's@^VERSION_ID="\([0-9]*\)\([0-9\.]*\)"$@\1@p')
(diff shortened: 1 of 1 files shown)


cmaloney:
:ship: 

dcos_generate_config.sh is actually a self-extracting tarball... (self extracting docker container, which is packed as a tarball)

luigiberrettini:
As an example how can I fix what I fixed in this PR once I downloaded dcos_generate_config.sh?
I extract it manually or should I docker exec the self-extracted container?

Thanks!

cmaloney:
As far as that goes it's non-trivial. To just play around with, what you could do is run dcos_generate_config.sh, then do a `less` of the file to get the `docker run` command underlying it. Change that to a docker exec with the entrypoint `/bin/sh`. Edit files at will then use `/installer_internal_wrapper` to run it.

Practically though, development on the installer is much easier before it's the "all in one docker" (it's possible to just do a `./prep_local` into a python virtual env, mock up a couple of variables and be able to run through full flows).

luigiberrettini:
That's why I was suggesting creating a compressed file.

cmaloney:
It was tried long ago. There are a number ofease of use as well as technical issues which get run into

PR #832, opened 2016-10-06
add pkgpanda service `dcos-log`

darkonie:
- Enable system components: DC/OS Log service, DC/OS Log socket on all DC/OS nodes.
- Add [DC/OS Logging service](https://github.com/dcos/dcos-log) as pkgpanda service.
- Bump adminrouter to include the required routes.
- 

[DCOS-10116](https://mesosphere.atlassian.net/browse/DCOS-10116)

- [x] System logs integration tests

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -315,6 +315,10 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log-config.json
+    content: |
+      {
+      }
 {% switch use_mesos_hooks %}
 {% case "true" %}
       MESOS_HOOKS={{ mesos_hooks }}
--- a/packages/bootstrap/extra/dcos_internal_utils/cli.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/cli.py
@@ -59,6 +59,8 @@ bootstrappers = {
     'dcos-minuteman': noop,
     'dcos-navstar': noop,
     'dcos-spartan': noop,
+    'dcos-log-master': noop,
+    'dcos-log-agent': noop,
 }
 
 
--- a/packages/dcos-integration-test/extra/test_dcos_log.py
+++ b/packages/dcos-integration-test/extra/test_dcos_log.py
@@ -1,0 +1,75 @@
+import json
+
+import requests
+
+
+def validate_json_entry(entry: dict):
+    required_fields = set(['fields', 'cursor', 'monotonic_timestamp', 'realtime_timestamp'])
+
+    assert set(entry.keys()) <= required_fields, (
+        "Entry didn't have all required fields. Entry fields: {}, required fields:{}".format(entry, required_fields))
+
+    assert entry['fields'], '`fields` cannot be empty dict. Got {}'.format(entry)
+
+
+def validate_sse_entry(entry):
+    assert entry, 'Expect at least one line. Got {}'.format(entry)
+    entry_json = json.loads(entry.lstrip('data: '))
+    validate_json_entry(entry_json)
+
+
+def check_response_ok(response: requests.models.Response, headers: dict):
+    assert response.ok, 'Request {} returned response code {}'.format(response.url, response.status_code)
+    for name, value in headers.items():
+        assert response.headers.get(name) == value, (
+            'Request {} header {} must be {}. All headers {}'.format(response.url, name, value, response.headers))
+
+
+def test_log_text(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.logs.get('v1/range/?limit=10', node=node)
+        check_response_ok(response, {'Content-Type': 'text/plain'})
+
+        # expect 10 lines
+        lines = filter(lambda x: x != '', response.content.decode().split('\n'))
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
+
+
+def test_log_json(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.logs.get('v1/range/?limit=1', node=node, headers={'Accept': 'application/json'})
+        check_response_ok(response, {'Content-Type': 'application/json'})
+        validate_json_entry(response.json())
+
+
+def test_log_server_sent_events(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.logs.get('v1/range/?limit=1', node=node, headers={'Accept': 'text/event-stream'})
+        check_response_ok(response, {'Content-Type': 'text/event-stream', 'Cache-Control': 'no-cache'})
+        validate_sse_entry(response.content.decode())
+
+
+def test_stream(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.logs.get('v1/stream/?skip_prev=1', node=node, stream=True,
+                                    headers={'Accept': 'text/event-stream'})
+        check_response_ok(response, {'Content-Type': 'text/event-stream', 'Cache-Control': 'no-cache'})
+        lines = response.iter_lines()
+        sse_id = next(lines)
+        assert sse_id, 'First line must be id. Got {}'.format(sse_id)
+        data = next(lines).decode()
+        validate_sse_entry(data)
+
+
+def test_log_proxy(cluster):
+    r = cluster.get('/mesos/master/slaves')
+    check_response_ok(r, {})
+
+    data = r.json()
+    slaves_ids = sorted(x['id'] for x in data['slaves'] if x['hostname'] in cluster.all_slaves)
+
+    for slave_id in slaves_ids:
+        response = cluster.get('/system/v1/agent/{}/logs/v1/range/?skip_prev=10'.format(slave_id))
+        check_response_ok(response, {'Content-Type': 'text/plain'})
+        lines = filter(lambda x: x != '', response.content.decode().split('\n'))
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
--- a/packages/dcos-log/build
+++ b/packages/dcos-log/build
@@ -1,0 +1,54 @@
+#!/bin/bash
+mkdir -p /pkg/src/github.com/dcos
+# Create the GOPATH for the go tool to work properly
+mv /pkg/src/dcos-log /pkg/src/github.com/dcos/
+cd /pkg/src/github.com/dcos/dcos-log
+
+make install
+# Copy the build from the bin to the correct place
+cp -r /pkg/bin/ "$PKG_PATH"
+
+master_service=${PKG_PATH}/dcos.target.wants_master/dcos-log-master.service
+slave_service=${PKG_PATH}/dcos.target.wants_slave/dcos-log-agent.service
+slave_public_service=${PKG_PATH}/dcos.target.wants_slave_public/dcos-log-agent.service
+
+master_socket_service=${PKG_PATH}/dcos.target.wants_master/dcos-log-master.socket
+slave_socket_service=${PKG_PATH}/dcos.target.wants_slave/dcos-log-agent.socket
+slave_public_socket_service=${PKG_PATH}/dcos.target.wants_slave_public/dcos-log-agent.socket
+
+mkdir -p $(dirname $master_service)
+mkdir -p $(dirname $slave_service)
+mkdir -p $(dirname $slave_public_service)
+
+cat <<EOF | tee $master_service
+[Unit]
+Description=DC/OS Log service master: DC/OS logging service
+[Service]
+EnvironmentFile=/opt/mesosphere/environment
+Restart=always
+StartLimitInterval=0
+RestartSec=5
+User=dcos_log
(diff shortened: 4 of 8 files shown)


packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
+import json
+import requests
+
+RANGE_ENDPOINT = '/system/v1/logs/v1/range/'
+STREAM_ENDPOINT = '/system/v1/logs/v1/stream/'
```
mellenburg: Rather than having to add these everywhere, it would nice to add a new client:
in test_util/cluster_api.py
```
@property
def logs(self):
    return self.get_client('/system/v1/logs')
```
and then this test looks like
```
def test_log_text(cluster):
    for node in cluster.masters + cluster.all_slaves:
        response = cluster.logs.get('v1/range/?limit=10', node=node)
        assert response.ok
```

darkonie:
@cmaloney would you take a look at this PR? i am going to add Michael's suggestions right after your review.

gen/dcos-config.yaml
```diff
@@ -315,6 +315,11 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log-config.json
+    content: |
+      {
+        "verbose": true
```
cmaloney: if this is always true why does it need to be config? Why can't it just be the program's behavior/

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def validate_json_entry(entry):
+    required_top_level_fields = ['fields', 'cursor', 'monotonic_timestamp', 'realtime_timestamp']
+
+    assert isinstance(entry, dict), 'Entry must be dict'
```
cmaloney: we're using python 3.5, which means basic type stuff can be done using annotations. In this case `def validate_json_entry(entry: dict)` has the same effect as this assert

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+    assert entry, 'Entry cannot be empty'
+
+    for required_field in required_top_level_fields:
```
cmaloney: why not do a set comparison here? Then we get all the non-matching keys + py.test's assert introspection

`assert set(entry.keys()) <= required_fields, "Entry didn't have all required fields. Entry fields: {}, required fields:{}".format(entry, required_fields)`

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if stream:
+        cursor = lines[0].lstrip('id: ')
+        assert cursor == entry_json['cursor'], 'In streaming mode an entry must have id: Got {}'.format(entry)
+
+
+def test_log_text(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.get(path=RANGE_ENDPOINT + '?limit=10', node=node)
+        assert response.ok
```
cmaloney: should do something like `assert response.ok, "Response code {}, body: {}".format(response.status_code, response.body)`. Currently if that fails, we'll see that the response status code was not equal  to 200, but won't see the response which means it'll be really hard to debug (esp. if it is a flaky failure).

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+        assert response.headers.get('Content-Type') == 'text/plain'
+
+        # expect 10 lines
+        lines = filter(lambda x: x != '', response.content.decode().split('\n'))
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}'.format(len(lines))
```
cmaloney: might make sense to include the lines for debugging

packages/dcos-integration-test/extra/test_dcos_log.py
cmaloney: Looking at the rest of the code, we do this response.ok + content-type header check a whole lot. As such, it probably makes sense to write a helper function `check_response_ok` or something that does those two asserts, and makes sure they have detailed assert messages that make debugging easy when things go wrong.

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+        lines = []
+        try:
+            buffer = ''
+            for line in response.iter_lines():
+                if not line:
+                    lines.append(buffer)
+                    buffer = ''
+                buffer += line.decode() + '\n'
+        except requests.exceptions.ConnectionError:
```
cmaloney: why on connection error is it okay to silently move on? Needs a comment explaining

packages/dcos-log/build
```diff
@@ -0,0 +1,55 @@
...
+mkdir -p /pkg/src/github.com/dcos
+# Create the GOPATH for the go tool to work properly
+mv /pkg/src/dcos-log /pkg/src/github.com/dcos/
+cd /pkg/src/github.com/dcos/dcos-log
+
+make install
+# Copy the build from the bin to the correct place
+cp -r /pkg/bin/ "$PKG_PATH"
+
+master_service=${PKG_PATH}/dcos.target.wants_master/dcos-log.service
```
cmaloney: why do master, slave, and slave_public all have different services with the same name? If the unit file isn't identical, it's a lot easier to debug things with different names.

packages/dcos-log/build
```diff
@@ -0,0 +1,55 @@
...
+cat <<EOF | tee $master_service
+[Unit]
+Description=DC/OS Log service: DC/OS logging service
+[Service]
+EnvironmentFile=/opt/mesosphere/environment
+Restart=always
+StartLimitInterval=0
+RestartSec=5
+User=dcos_log
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log
```
cmaloney: why do agent and master dcos-log need different bootstrapping yet get identical everything else / configuration?

gen/dcos-config.yaml
darkonie: the reason for config here is to be consistent with DC/OS EE. https://github.com/mesosphere/dcos-enterprise/pull/407
in this case i don't need to diverge the systemd unit and will only have to bump the upstream for this service to start.

packages/dcos-log/build
darkonie: the same reason as i mentioned in first comment. To be consistent with DC/OS EE. Unfortunately we need different service accounts for the same service running on master and agent nodes :( I could update the build script but it would mean we'd have to diverge the EE build script.

packages/dcos-log/build
darkonie: unfortunately we need 2 different system accounts for running the same service on master and agent nodes. The bootstrap does not know about roles at all, and we need to communicate the node role somehow. The best way to do this is to run bootstrap script with different parameters, this is how we bootstrap service account now.
cc @alberts

alberts: dcos-log is exactly like 3dt. What makes sense for 3dt more than likely makes sense for dcos-log.

E.g. https://github.com/dcos/dcos/blob/master/packages/3dt/build#L15

darkonie: the difference with 3dt, is that we have JWT code in open DC/OS, which makes it work in both open and ee DC/OS. Less diverge code.

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

darkonie:
@cmaloney @spahl this is green with @mellenburg commit
can i get a code review?

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+            pass
+
+        assert lines
```
alberts: TIL assert []  fails.

orsenthil: Empty collection.

```
>>> bool([])
False
```

cmaloney:
This should be using https://github.com/dcos/dcos/blob/master/PULL_REQUEST_TEMPLATE.md in the PR message

gen/dcos-config.yaml
darkonie: that makes sense, fixed

darkonie:
@cmaloney addressed comments

gen/dcos-config.yaml
```diff
@@ -315,6 +315,10 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log-config.json
```
malnick: Is this empty on purpose?

malnick: It looks like there may have been a mistake in insertion here (contains env variables that don't look associated with the service)

packages/dcos-log/build
malnick: Take a look at the bootstrapping for metrics, I used a single UID for the service and run on two different roles.

cmaloney:
test failures are real, back to wip until they're fixed

gen/dcos-config.yaml
darkonie: see Cody's comment, it's empty on purpose

darkonie:
@cmaloney you are right, i forgot to update signal service integration test. Done

darkonie:
@cmaloney tests passed

packages/dcos-log/buildinfo.json
```diff
@@ -0,0 +1,11 @@
+{
+  "docker": "dcos/dcos-builder:dcos-log_dockerdir-latest",
```
adam-mesos: Why the fancy dcos-builder?

darkonie: magic of custom containers. This builder will use my container to build https://github.com/darkonie/dcos/blob/beb2b1242840327ccd89112b1a03cdf49e8e9265/packages/dcos-log/docker/Dockerfile

darkonie: we are using this approach to build [`dcos-ca`](https://github.com/mesosphere/dcos-enterprise/blob/master/packages/dcos-ca/buildinfo.json)

packages/dcos-log/docker/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.7
```
darkonie: updated PR to use base image and added a commit from https://github.com/dcos/dcos/pull/998

packages/bootstrap/extra/dcos_internal_utils/cli.py
```diff
@@ -59,6 +59,8 @@ def noop(b, opts):
     'dcos-minuteman': noop,
     'dcos-navstar': noop,
     'dcos-spartan': noop,
+    'dcos-log': noop,
```
cmaloney: nit: can we call this `dcos-log-master` for consistency between this, the unit names?

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_log_text(cluster):
+    for node in cluster.masters + cluster.all_slaves:
+        response = cluster.logs.get('v1/range/?limit=10', node=node)
+        check_response_ok(response, {'Content-Type': 'text/plain'})
+
+        # expect 10 lines
+        lines = filter(lambda x: x != '', response.content.decode().split('\n'))
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
```
cmaloney: nit: Would be nice if we could check the text is something sensical / not jumbled bytes, but this works well enough for now. Should be doable once this gets to collecting container logs (Launch a container that prints a couple of very specific lines into the log, make sure that in the overall list of log lines those lines appear in the right order.

darkonie: yes, there will be more tests added to https://github.com/dcos/dcos/pull/971 that will run container and pod and validate the log strings. For now it is difficult to say what the logs would look like

cmaloney:
:ship: assuming tests pass

darkonie:
@cmaloney tests passed

cmaloney:
will get on the next train which will either be this evening or tomorrow morning depending on when I make the next review pass

PR #839, opened 2016-10-07
Replace 'ee' with 'downstream' in tests

branden:


branden:
@cmaloney Ready for review.

PR #842, opened 2016-10-07
Abstracting Region and NAT ami mapping

amitaekbote:
Unit tests pass , more testing required before final review

### DIFF
--- a/gen/installer/aws.py
+++ b/gen/installer/aws.py
@@ -54,6 +54,71 @@ aws_region_names = [
         'id': 'ap-southeast-2'
     }]
 
+
+region_to_ami_map = {
+    'ap-northeast-1': {
+        'coreos': 'ami-965899f7',
+        'stable': 'ami-965899f7',
+        'el7': 'ami-264f8747',
+        'natami': 'ami-55c29e54'
+    },
+    'ap-southeast-1': {
+        'coreos': 'ami-3120fe52',
+        'stable': 'ami-3120fe52',
+        'el7': 'ami-0765bd64',
+        'natami': 'ami-b082dae2'
+    },
+    'ap-southeast-2': {
+        'coreos': 'ami-b1291dd2',
+        'stable': 'ami-b1291dd2',
+        'el7': 'ami-3f1a2c5c',
+        'natami': 'ami-996402a3'
+    },
+    'eu-central-1': {
+        'coreos': 'ami-3ae31555',
+        'stable': 'ami-3ae31555',
+        'el7': 'ami-846e9eeb',
+        'natami': 'ami-204c7a3d'
+    },
+    'eu-west-1': {
+        'coreos': 'ami-b7cba3c4',
+        'stable': 'ami-b7cba3c4',
+        'el7': 'ami-250c7f56',
+        'natami': 'ami-3760b040'
+    },
+    'sa-east-1': {
+        'coreos': 'ami-61e3750d',
+        'stable': 'ami-61e3750d',
+        'el7': 'ami-0e019062',
+        'natami': 'ami-b972dba4'
+    },
+    'us-east-1': {
+        'coreos': 'ami-6d138f7a',
+        'stable': 'ami-6d138f7a',
+        'el7': 'ami-47096750',
+        'natami': 'ami-4c9e4b24'
+    },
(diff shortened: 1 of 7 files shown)


gen/installer/aws.py
```diff
@@ -54,6 +54,83 @@
         'id': 'ap-southeast-2'
     }]
 
+
+region_to_ami_map = [
+    {
+        'name': 'ap-northeast-1',
```
cmaloney: `region`?

gen/installer/aws.py
```diff
@@ -54,6 +54,83 @@
         'id': 'ap-southeast-2'
     }]
 
+
+region_to_ami_map = [
```
cmaloney: as long as we're here want to fix https://dcosjira.atlassian.net/browse/DCOS-12?

gen/installer/aws.py
```diff
@@ -178,6 +269,7 @@ def render_cloudformation_transform(cf_template, transform_func=lambda x: x, **k
         {k: transform_func(v) for k, v in kwds.items()}
     )
 
+    print(template_str)
```
cmaloney: remove debug statement

gen/installer/aws.py
amitaekbote: should be straightforward, I will add it as a separate commit

amitaekbote:
@cmaloney ready for review

gen/installer/aws.py
```diff
@@ -138,6 +215,24 @@ def get_test_session(config=None):
     return release.call_matching_arguments(release.storage.aws.get_session, config, True)
 
 
+def gen_ami_mapping(mappings):
```
cmaloney: nit: if we made region_to_ami_map an actual map ratherthan a list which looks like:
```
{
  'ap-northeast-1': {
    'coreos': '',
    'stable': '',
    'el7': '',
    'natami': ''
  }
}
```

then we could really easily test the ami mapping wihtout copying the AMI ids around:
```python
assert result["ap-northeast-1"] == {'stable': gen.installer.aws.region_to_ami_map['ap-northeast-1']['stable']}
```

which makes the test less fragile (every time someone updates the AMIs currently they have to touch this test which they'll never think of until CI pings them). It also is a lot simpler than what you have to do with the list thing for the same effect:
```python
ap_northeast_1_list = list(filter(lambda x: x['region'] == 'ap-northeast-1', gen.installer.aws.region_to_ami_map))
assert len(ap_northeast_1_list) == 1
assert result["ap-northeast-1"] == {'stable': ap_northeast_1_list[0]['stable']}
```

It also makes gen_ami_mapping a little simpler:
```python
def gen_ami_mapping(mappings):
    # create new dict with required mappings
    # all will have region by default
    for region, amis in region_to_ami_map.items():
        # Build the final dictionary
        final[region] = dict()
        for key in mappings:
          # TODO(cmaloney): natami needs to be mapped to the `default` key for 
          # now. Should update the templates to remove the need.
          final_key = 'default' if key == 'natami' else key
          final[region][final_key] = amis[key]

    # Returning -1:1 to cut the curly braces off the outside of the json dictionary so it formats
    # right into the templates.
    return json.dumps(final, indent=4, sort_keys=True)[1:-1]
```

cmaloney: To build the final dictionary could go even more one line removing the manual iteration over mappings with a dictionary comprehsion (although not necessarily a good thing):
```python
final[region] = {'default' if key == 'natami' else key: amis[key] for key in mappings}
```

gen/test_installer_aws.py
```diff
@@ -0,0 +1,10 @@
+import json
+import gen.installer.aws
+
+
+def test_gen_aws_mapping():
+    result = json.loads("{" + gen.installer.aws.gen_ami_mapping({"stable"}) + "}")
+    # check number of regions
+    assert len(result) == 10
+    # check format of response
+    assert result["ap-northeast-1"] == {'stable': 'ami-965899f7'}
```
cmaloney: the copying of the AMI here is going to make it so anyone updating the AMI mappings (Which we usually do once a release). Be really nice if we programatically grabbed it

cmaloney:
:ship:

PR #843, opened 2016-10-07
Add boto rate limit decorator

mellenburg:
-decrease boto verbosity
-catch another rate exception

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -1,23 +1,20 @@
 #!/usr/bin/env python3
 import logging
-import os
 import time
 
 import boto3
 import retrying
-from botocore.exceptions import ClientError
 
-from test_util.helpers import Host, SshInfo
+from test_util.helpers import Host, retry_boto_rate_limits, SshInfo
 
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
 logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
-logger = logging.getLogger(__name__)
+# AWS verbosity in debug mode overwhelms meaningful logging
+logging.getLogger('botocore').setLevel(logging.INFO)
+log = logging.getLogger(__name__)
 
 VPC_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-cluster-template.json'
 VPC_EBS_ONLY_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-ebs-only-cluster-template.json'
-
-# At time of implementation, VPC averaged 4 minutes and CF 9 minutes
-AWS_WAIT_BEFORE_POLL_MIN = os.getenv('AWS_WAIT_BEFORE_POLL_MIN', 3)
 
 
 def template_by_instance_type(instance_type):
@@ -55,7 +52,7 @@ class BotoWrapper():
     def create_stack(self, name, template_url, user_parameters, deploy_timeout=60):
         """Returns boto stack object
         """
-        logging.info('Requesting AWS CloudFormation...')
+        log.info('Requesting AWS CloudFormation...')
         cf_parameters = []
         for k, v in user_parameters.items():
             cf_parameters.append({'ParameterKey': k, 'ParameterValue': v})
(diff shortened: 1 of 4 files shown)


mellenburg:

test_util/aws.py
```diff
@@ -349,10 +333,11 @@ def delete(self):
         self.stack = self.boto_wrapper.resource('cloudformation').Stack(self.stack.stack_id)
         self.stack.delete()
 
+    @retry_boto_rate_limits
     def get_vpc_host_ips(self):
         reservations = self.boto_wrapper.client('ec2').describe_instances(Filters=[{
```
cmaloney: is it feasible to build this into the boto_wrapper for ec2 perhaps rather than having to remember to hand-wrap every usage of it?

test_util/helpers.py
```diff
@@ -31,6 +33,36 @@ def authenticate(self, cluster):
         logging.info('Authentication successful')
 
 
+def retry_boto_rate_limits(boto_fn, wait=2, timeout=60 * 60):
```
cmaloney: https://docs.python.org/3/library/functools.html#functools.wraps?

test_util/helpers.py
```diff
@@ -31,6 +33,36 @@ def authenticate(self, cluster):
...
+    """Decorator to make boto functions resilient to AWS rate limiting and throttling.
+    If one of these errors is encounterd, the function will sleep for a geometrically
+    increasing amount of time
+    """
+    def ignore_rate_errors(*args, **kwargs):
+        local_wait = wait
```
cmaloney: nit: this assumes (probably safe) that wait and timeout are value types / will always be copied rather than be reference things (strings, dictionaries, lists, etc). I like dropping a copy in this case just to make the target behavior very explicit and always work.

test_util/helpers.py
```diff
@@ -31,6 +33,36 @@ def authenticate(self, cluster):
...
+                    error_code = e.response['Error']['Code']
+                elif isinstance(WaiterError, e):
+                    error_code = e.last_response['Error']['Code']
+                else:
+                    raise
+                if error_code in ['Throttling', 'RequestLimitExceeded']:
+                    logging.warn('AWS API Limiting error: {}'.format(error_code))
+                    logging.warn('Sleeping for {} seconds before retrying'.format(local_wait))
+                    time.sleep(local_wait)
+                    local_timeout -= wait
```
cmaloney: shouldn't timeout take into account time that has passed while boto_fn was running?

test_util/aws.py
mellenburg: I cannot currently think of a more elegant method. At the boto_wrapper level, we just return clients. The issue is a lot (but not all) functions of the clients need to be wrapped for this case.

What I could imagine is changing a client's getattr so that whenever it returns a function it returns a function wrapped with the retry and backoff method, but that feels like a hack

test_util/aws.py
```diff
@@ -74,49 +75,23 @@ def __init__(self, stack_name, boto_wrapper):
...
-                        retry_on_result=lambda res: res is False,
-                        retry_on_exception=lambda ex: False)
-        def wait_and_backoff():
-            nonlocal default_poll_interval
-            try:
-                self._wait_loop(state_1, state_2, default_poll_interval)
-            except ClientError as e:
-                if e.response['Error']['Code'] == 'Throttling':
-                    logging.warn('AWS Client raised a throttling error; increasing poll interval...')
-                    default_poll_interval += 10
```
mellenburg: The reason I decided on this was that I wanted to make decorator that represented "try this function and if it gets this error, wait, and then try again, with waits getting successively longer. That way I could wrap functions that didnt have the concept of repeatedly calling polling AWS.

The effect that this has for functions that repeatedly poll is that we now just poll at a constant interval. back off. poll at the same interval. back off for even longer. repeat. My hand-wavy reasoning is that geometrically increasing backoffs (wherein no polling occurs) are roughly equivalent (in terms of overall API use) to a polling interval that increases linearly after failure

mellenburg: All of that being said, we can definitely have a wrapper that tells the next successive retries to wait even longer, but it felt better to get one general function that handles all cases

test_util/helpers.py
mellenburg: Yes, but I figured it was relatively insignificant in light of a geometric backoff. However, we can track time here if desired

mellenburg:
@cmaloney I've addressed the minor issues in this PR, but I still need feedback on the point of "is there a better way than wrapping each function that could raise a client error?" Still the only other way I see is to modify boto3.session.client to always provide functions that are wrapped. but I'm sure there is plenty of nuance there once I try it that will make that method unfavorable

cmaloney:
re: wrapping each function, this works for now. I'm fairly confident we'll need a better solution long term but this helps for now and isn't too much overhead. Just really easy for people to get wrong. May be that we need to patch boto3 upstream directly or something similar.

test_util/aws.py
```diff
@@ -125,26 +101,30 @@ def wait_loop():
...
 
+    @retry_boto_rate_limits
     def get_stack_events(self):
+        log.debug('Requesting stack events')
         return self.boto_wrapper.client('cloudformation').describe_stack_events(
             StackName=self.stack.stack_id)['StackEvents']
 
-    def wait_for_stack_creation(self):
-        self.wait_for_status_change('CREATE_IN_PROGRESS', 'CREATE_COMPLETE')
+    def wait_for_stack_creation(self, wait_before_poll=3):
```
cmaloney: nit: `wait_before_poll_min`?

mellenburg:
Last comment address, :ship:

cmaloney:
:ship:

PR #844, opened 2016-10-07
Update Azure VM options

lingmann:
- Update the available set of Azure agent VM sizes available to only
  those which meet the minimum DC/OS requirements of >= 2 cores.
- Add support for recently added Azure VM types.
- Update the validation VM size from Standard_A0 to Standard_A2.

### DIFF
--- a/gen/azure/templates/acs.json
+++ b/gen/azure/templates/acs.json
@@ -77,8 +77,6 @@
     },
     "agentVMSize": {
       "allowedValues": [
-        "Standard_A0",
-        "Standard_A1",
         "Standard_A2",
         "Standard_A3",
         "Standard_A4",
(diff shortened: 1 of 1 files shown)


lingmann:
@anhowe could you take a look at this template update which is now using an agent vm size mapping based on your new generator? You'll note I've updated both the vm mappings and the LRS mappings so that they are identical to you latest generator code.

The only thing I have updated regarding the masters is the 'validation' VM size to Standard_A2 since Standard_A0 no longer meets the minimum requirements. We don't yet provide options for customizing the master size and I'd prefer that land in a separate patch.

Please take a look and let me know if you have any feedback. Thanks!

lingmann:
I've identified a problem with adding all of these new instance types... if you select an instance type which is not available in the deploy region, ARM validation does not catch it, and the end-user will not see the cluster fail to launch until very late in the launch process when the agent VMSS fails.

We'll need to investigate this further before this is ready to merge. Some options:

1. Generate templates for each region with mappings for exactly what is supported
2. Reduce the set of combinations to only those which will work (ie. we further limit the number of regions and/or instance types)
3. More advanced template mapping logic to represent these limitations (if possible).

cc: @anhowe

cmaloney:
@lingmann do we still need this or can we close it?

lingmann:
@cmaloney I've removed the new VM sizes from this PR since it introduced additional complications for end-users (VM sizes for agent pools are not validated at ARM template launch time, so failures can easily occur when users select a VM size which is not available in all regions.)

This PR now just includes:
* Remove Standard_A0 and Standard_A1 which don't meet the minimum DC/OS requirements of >= 2 cores.
* Update the validation VM size to Standard_A2 now that Standard_A1 is no longer supported.
* Remove unused variable `storageAccountType`.

When tests are green this is ready for review.

PR #846, opened 2016-10-08
Adding testing options passthrough like the CF templates

mellenburg:
Carrying on work from https://github.com/dcos/dcos/pull/716 (which is on a branch I cannot edit)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,7 @@ setup(
             'ccm-deploy-test=test_util.test_aws_vpc:main',
             'test-aws-cf-deploy=test_util.test_aws_cf:main',
             'test-upgrade-vpc=test_util.test_upgrade_vpc:main',
+            'test-azure-rm-deploy=test_util.azure_test_driver:main',
             'pkgpanda=pkgpanda.cli:main',
             'mkpanda=pkgpanda.build.cli:main',
             'dcos_installer=dcos_installer.cli:main'
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -23,15 +23,18 @@ def validate_env():
     The following environment variables must be set to ensure the template can be deployed successfully and that
     parameters we rely on from a testing perspective are explicitly set and not relying on defaults which may change.
     '''
-    required = ['AZURE_PARAM_linuxAdminUsername', 'AZURE_PARAM_oauthEnabled',
-                'AZURE_PARAM_masterEndpointDNSNamePrefix', 'AZURE_PARAM_agentEndpointDNSNamePrefix',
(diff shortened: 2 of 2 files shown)


mellenburg:

setup.py
```diff
@@ -63,6 +63,7 @@ def get_advanced_templates():
             'ccm-deploy-test=test_util.test_aws_vpc:main',
             'test-aws-cf-deploy=test_util.test_aws_cf:main',
             'test-upgrade-vpc=test_util.test_upgrade_vpc:main',
+            'test-azure-rm-deploy=test_util.azure_test_driver:main',
```
cmaloney: /me will be so happy when we have `dcos-launch destroy` as a binary instead of all these...

test_util/azure_test_driver.py
```diff
@@ -23,15 +23,18 @@ def validate_env():
     The following environment variables must be set to ensure the template can be deployed successfully and that
     parameters we rely on from a testing perspective are explicitly set and not relying on defaults which may change.
     '''
-    required = ['AZURE_PARAM_linuxAdminUsername', 'AZURE_PARAM_oauthEnabled',
-                'AZURE_PARAM_masterEndpointDNSNamePrefix', 'AZURE_PARAM_agentEndpointDNSNamePrefix',
-                'AZURE_TEMPLATE_URL', 'AZURE_PARAM_sshRSAPublicKey']
+    required = [
+        'AZURE_PARAM_linuxAdminUsername',
```
cmaloney: as long as we're reformatting should also alphabetize.

test_util/azure_test_driver.py
```diff
@@ -58,7 +61,16 @@ def get_value(template_parameter):
     return(get_env_params()[template_parameter]['value'])
 
 
-def run():
+def get_test_config():
+    add_env = {}
```
cmaloney: this doesn't seem to handle the `DCOS_AUTH_ENABLED` old case. How does that work for going through CI before and after that change?

Might be helpful to add a "Handle DCOS_AUTH_ENABLED" / oauthEnabled explicitly for the transition.

test_util/azure_test_driver.py
```diff
@@ -209,4 +221,4 @@ def wait_for_delete():
         sys.exit(2)
 
 if __name__ == '__main__':
-    run()
+    main()
```
cmaloney: :+1:

test_util/azure_test_driver.py
mellenburg: Its now passed via TEST_ADD_ENV_DCOS_AUTH_ENABLED but we can specifically handle this case. Every option felt awkward so its all the same to me

mellenburg:
@cmaloney This has been updated. Are there any other point to address?

cmaloney:
:ship:

PR #847, opened 2016-10-08
Fix conftest bugs

mellenburg:
-resiliency mark makes it impossible to run just open_source_tests in downstream; simple env is better
-if logging is declared in cluster fixutre, -s option will not work when selecting specific tests


mellenburg:

PR #851, opened 2016-10-10
pod smoke test

kensipe:


greggomann:
It looks like the test cluster code has changed significantly with this commit: https://github.com/dcos/dcos/commit/0e433ab3630cfdef44a852204a8c0c635f2df019
This new test probably needs some tweaking to accomodate these changes.

kensipe:
@greggomann thanks!   refactored!

test_util/marathon.py
```diff
@@ -202,6 +202,82 @@ def _pool_for_marathon_app(app_id):
...
+        r = self.post('/pods', json=pod_definition)
+        logging.info('Response from marathon: {}'.format(repr(r.json())))
+        assert r.ok
+
+        @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda ret: ret is None,
+                        retry_on_exception=lambda x: False)
+        def _wait_for_deployment(pod_id):
+            req_uri = '/marathon/v2/deployments'
+            r = self.get(req_uri)
```
mellenburg: This should just be self.get('deployments')

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,39 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
...
+                'name': 'ct2',
+                'resources': {'cpus': 0.1, 'mem': 32},
+                'exec': {'command': {'shell': 'while true; do echo -n $PING ' ';sleep 1; done'}}
+            }
+        ],
+        'networks': [{'mode': 'host'}]
+    }
+
+    instances = cluster.marathon.deploy_marathon_pod(pod_definition)
+    assert pod_definition['scaling']['instances'] == instances
```
mellenburg: As written, this will leave the pod if the test fails, which could have unknown consequences for other tests. As such, I would prefer we add a context manager method like the deploy_and_cleanup one for apps.

test_util/marathon.py
```diff
@@ -202,6 +202,82 @@ def _pool_for_marathon_app(app_id):
...
+
+        Returns:
+            Scaling instance count
+        """
+        r = self.post('/pods', json=pod_definition)
+        logging.info('Response from marathon: {}'.format(repr(r.json())))
+        assert r.ok
+
+        @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda ret: ret is None,
```
mellenburg: retrying on return None feels a little awkward to me. Can we do retry on False instead? Otherwise, it seems like you may expect return data["scaling"]["instances"] to be None in some cases.

greggomann:
@kensipe, I wonder if this should be rebased onto the `greg/pods` branch in the Mesosphere repo and merged into that instead? Then we could see the test results when run against the pod-enabled codebase.

kensipe:
@greggomann I don't see a `greg/pods` branch

test_util/marathon.py
```diff
@@ -202,6 +202,85 @@ def _pool_for_marathon_app(app_id):
             raise Exception("Application deployment failed - operation was not "
                             "completed in {} seconds.".format(timeout))
 
+    def deploy_test_pod_and_check(self, pod, test_uuid):
+        with deploy_pod_and_cleanup(pod) as instances:
+            assert pod['scaling']['instances'] == instances
```
mellenburg: I would not include this deploy_test_pod_and_check method as this is only asserting on instance count. Rather, we should block on instance count. The intention behind deploy_test_add_and_check is that there is a well known app that tests a variety of features and we want that variety checked whereas the pod for this case is truly a smoke test and trivial to check

test_util/marathon.py
```diff
@@ -202,6 +202,85 @@ def _pool_for_marathon_app(app_id):
...
+        def _wait_for_pod_deployment(pod_id):
+            r = self.get('/deployments')
+            data = r.json()
+            if len(data) > 0:
+                logging.info('Waiting for pod to be deployed %s', repr(data))
+                return False
+            # deployment complete
+            r = self.get('/pods' + pod_id)
+            data = r.json()
+            return data["scaling"]["instances"]
```
mellenburg: Also, I would prefer that we have a return a value similar to what deploy_app does. E.G. can we return all the service points for all the containers? That way I can deploy the dedicated test app inside of a pod and have use most of the same pipes.

test_util/marathon.py
```diff
@@ -202,6 +202,85 @@ def _pool_for_marathon_app(app_id):
             raise Exception("Application deployment failed - operation was not "
                             "completed in {} seconds.".format(timeout))
 
+    def deploy_test_pod_and_check(self, pod, test_uuid):
+        with deploy_pod_and_cleanup(pod) as instances:
+            assert pod['scaling']['instances'] == instances
+
+    def deploy_pod(self, pod_definition, timeout=300, check_health=False):
```
mellenburg: Do you know if pods take significantly longer to deploy? we set timeouts to 120s for apps and would be nice to minimize this from the start

test_util/marathon.py
```diff
@@ -202,6 +202,85 @@ def _pool_for_marathon_app(app_id):
...
+            data = r.json()
+            return data["scaling"]["instances"]
+
+        try:
+            return _wait_for_pod_deployment(pod_definition['id'])
+        except retrying.RetryError:
+            raise Exception("Pod deployment failed - operation was not "
+                            "completed in {} seconds.".format(timeout))
+
+    def destroy_pod(self, pod_id, timeout=300):
```
mellenburg: Same as above. Should it really be greater than 120?

test_util/marathon.py
```diff
@@ -214,7 +293,7 @@ def destroy_app(self, app_name, timeout=120):
         @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
                         retry_on_result=lambda ret: not ret,
                         retry_on_exception=lambda x: False)
-        def _destroy_complete(deployment_id):
+        def _destroy_pod_complete(deployment_id):
```
mellenburg: why is this changing? This isnt destroying a pod

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,38 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
...
+
+    test_uuid = uuid.uuid4().hex
+
+    pod_definition = {
+        'id': '/integration-test-pods-{}'.format(test_uuid),
+        'scaling': {'kind': 'fixed', 'instances': 1},
+        'environment': {'PING': 'PONG'},
+        'containers': [
+            {
+                'name': 'ct1',
```
mellenburg: Nit: would be nice to use test_server here and pair it with something, so we can check service routing, but that can be a future PR so we get this smoke test in sooner

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,38 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
     cluster.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
+def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(cluster):
+    """Marathon pods deployment integration test using the Mesos Containerizer
```
mellenburg: nit: I would prefer that this test uses containers with health checks. If the execs totally fail, we won't know as the deployment will still consume and instance and appear OK. A simple exec of touch foo and health check of test -f foo to verify

test_util/marathon.py
kensipe: should take < 3s  I'll adjust to 60s

test_util/marathon.py
kensipe: wow... I'm not sure what happened there.  this was resolved and the correct method name refactored.

adam-mesos:
We can close this PR in favor of #856 now that these tests are included there

PR #852, opened 2016-10-10
Minuteman support for named vips in metric tags.

aeyakovenko:


aeyakovenko:
this is ready for review, @cmaloney @sargun

cmaloney:
DC/OS 1.9 targeted?

cmaloney:
Is there a way we can test / check this new feature is actually working?

aeyakovenko:
this is tested in this integration test, https://github.com/mesosphere/dcos-enterprise/pull/319

cmaloney:
@aeyakovenko with it only in a place people here can't see, I can't tell when making changes to this repository if it's badly broken or not. Really need something so if I revert your change to the package the test fails in this repository.

aeyakovenko:
@sargun @cmaloney Do we have any general integration tests for minuteman already in dcos?  Those should verify that minuteman is working properly.  Otherwise this change enables an enterprise only feature, it would be hard to test for it outside of the enterprise component that uses it, since minuteman only exposes erlang endpoints.

aeyakovenko:
@cmaloney @sargun, enterprise tests pass as well,
https://github.com/mesosphere/dcos-enterprise/pull/319

is there anything else blocking this?

sargun:
@spahl @orsenthil

cmaloney:
:ship:

PR #853, opened 2016-10-11
fix getpass in python_test_server.py

alberts:


cmaloney:
Would be much cleaner to add the user in the test which needs this or re-configure the test to use a specific user which does exist by default in the debian:jessie image.

cmaloney:
volume mounting the host's /etc/passwd and the like is definitely not an acceptable solution for htis. If you have something like a Ubuntu 16.04 container running on a CoreOS host, or CentOS host, all three of those OSes have differnt uids for core system users and groups. By mounting in /etc/passwd and /etc/group those get shuffled. The users may not even match properly with passwords because `/etc/shadow` isn't mounted.

I'm against this approach, it's not a good way to make things work.

alberts:
@cmaloney what is a good way? We're really open to suggestions here.

alberts:
For context, we really do mean to check whether the container is running as nobody on the hojst.

PR #855, opened 2016-10-11
[WIP] Added flag in `config.yaml` for turning on http command executor.

asridharan:
This will allow the operator to turn on the HTTP command executor on the agents, instead of old driver based command executors that are currently being used. This is especially useful in cases where mesos (or UCR) containers would running on a virtual network which if Firewalled from the host network. In such scenarios using the HTTP command executor would allow the operator to open a single port on the host network for the executor to register with the agent. Without the HTTP command executor (using the old driver based executor) the operator will pretty much have to open up all ports on the host to be accessible by the virtual network.


cmaloney:
This configuration option doesn't really make sense to expose to end users (it's a very internal knob). We need to be okay with it always being set (or setting it in some security modes but not others. I don't believe the http api is security compatible currently).

asridharan:
@cmaloney thinking aloud, why do you think selection of the HTTP executor is an internal knob? The problem is that the HTTP command executor is experimental. Without all the users a knob, we won't have a mechanism for them to turn it on and off selectively in DC/OS. I don't think having it always on or off might be a viable strategy.

The HTTP command executor starts becoming really important in cases where operators want to run containers on virtual networks which are completely Firewalled from the host network.

vinodkone:
The plan is for http based command executor to be the default instead of driver based command executor which will be deprecated and removed. I would like to make http executor the default once it is baked in our soak cluster for a while and it is available as optional for people using 1.9.

cmaloney:
@asridharan why would a customer care what protocol the executor speaks? Why should they have to think about it? 

From the point of view of a customer, they want CNI or they don't. They just want the default to work whatever they choose. The choice of what protocol the executor speaks is entirely orthagonal to their interests / their software. The customer shouldn't need to generally care that there even is a libprocess API and a HTTP API.

By adding a new flag we also combinatorially increase the complexity of our software configuration. More things that might break when used in certain combinations and are completely untested.

asridharan:
@cmaloney they care cause the existing executor uses the scheduler driver, which would open multiple TCP session (one in each direction) to the agent. This makes it hard to run these executors on virtual network (CNI) when virtual networks need to be firewalled from the host network. 

I agree the customer should not have to worry about this fact, and that is why as @vinodkone mentioned we are moving towards a world where the default executor is the HTTP executor. We are not there yet, and hence the need for this knob. Hopefully, by 1.10, we can move to the HTTP executor as the default.

I also agree with you that adding the flag adds more combinations to test, but not sure how else will we provide this flexibility to the user to try out new pieces of software without giving them a safety net.

cmaloney:
Build reliable dependable software we can trust. if a new feature doesn't have confidence we can ship it in production that's a bug. I really don't like adding permutation because we can't get a high enough confidence level without having a "maybe people will sometimes test this by hand".

Esp. since that comes with a "if you do http api security=strict doesn't work" and "if you do non-http you can't use overlay / CNI stuff". That's  a complicated caveat to explain when things go wrong.

asridharan:
@cmaloney the whole point of having this knob is that the HTTP API is not turned on by default. There will always be features in the software that are experimental and that users need to play around to get some feedback. 

As far as security is concerned I don't think any of that should apply to open DC/OS so not sure why are discussing it in this context?

cmaloney:
Everything in DC/OS is also in Mesosphere Enterprise DC/OS (Enterprise extends open). That I can't use http executors at all in secure mode is going to lead to problems if using them is the only way to make things work.

asridharan:
@cmaloney, again, that's why we have the knob and have to have it off by default. Which implies that the operator really need to know what he is doing to turn it on in order to try it out. He obviously should not be turning it on with security setting in which we know the HTTP executor does not work.

That said, open DC/OS should not change to accommodate DC/OS enterprise, it should be the other way round.

asridharan:
By the way, I am open to suggestion if there are any better ideas of how operators can enable the HTTP command executor on agents, without adding this knob to the installer.

vinodkone:
> "Build reliable dependable software we can trust. if a new feature doesn't have confidence we can ship it in production that's a bug."

I disagree. Software doesn't become reliable out of gate. Software needs to go through testing and baking (both internal and external) before it can be called production ready. Calling a feature production ready without actually having it tested in a real production environment is misleading.

> Esp. since that comes with a "if you do http api security=strict doesn't work" and "if you do non-http you can't use overlay / CNI stuff". That's a complicated caveat to explain when things go wrong.

Not completely true. HTTP command executor works in strict mode.

asridharan:
@vinodkone that would be my bad, didn't realize HTTP command executor works in 'security:strict' mode. I thought since all endpoints need AuthN in 'strict' mode,  the HTTP command executor will also need to do AuthN which I believe isn't implemented yet.

spahl:
Why make a switch and not turn it on by default?

PR #856, opened 2016-10-11
Updated Mesos/Marathon/UI for pods.

greggomann:


packages/mesos/buildinfo.json
```diff
@@ -3,8 +3,8 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/mesosphere/mesos",
-    "ref": "d5746045ac740d5f28f238dc55ec95c89d2b7cd9",
-    "ref_origin" : "dcos-mesos-1.0.x-837bb4e"
+    "ref": "c8e8c70a71b2849ee37d3c6b9f396ceaac73369e",
+    "ref_origin" : "dcos-mesos-master-35ce5042"
```
adam-mesos: If you're going to bump this in dcos master, could you also update https://github.com/dcos/dcos/blob/master/packages/mesos/README.md with the new cherry-pick(s) for 1.1.x / master?

test_util/marathon.py
```diff
@@ -202,6 +202,88 @@ def _pool_for_marathon_app(app_id):
             raise Exception("Application deployment failed - operation was not "
                             "completed in {} seconds.".format(timeout))
 
+
```
adam-mesos: Too many blank lines in this file. @kensipe please clean up so we can pass the CodeQuality check
https://teamcity.mesosphere.io/viewLog.html?buildId=474873&buildTypeId=DcosIo_Dcos_CodeQuality

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,37 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
     cluster.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
+def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(cluster):
+    """Marathon pods deployment integration test using the Mesos Containerizer
```
adam-mesos: Copying @mellenburg 's comment from #851 
> nit: I would prefer that this test uses containers with health checks. If the execs totally fail, we won't know as the deployment will still consume and instance and appear OK. A simple exec of touch foo and health check of test -f foo to verify

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,37 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
...
+
+    test_uuid = uuid.uuid4().hex
+
+    pod_definition = {
+        'id': '/integration-test-pods-{}'.format(test_uuid),
+        'scaling': {'kind': 'fixed', 'instances': 1},
+        'environment': {'PING': 'PONG'},
+        'containers': [
+            {
+                'name': 'ct1',
```
adam-mesos: Copying @mellenburg 's comment from #851 
> Nit: would be nice to use test_server here and pair it with something, so we can check service routing, but that can be a future PR so we get this smoke test in sooner

test_util/marathon.py
```diff
@@ -202,6 +202,88 @@ def _pool_for_marathon_app(app_id):
...
+        def _wait_for_pod_deployment(pod_id):
+            r = self.get('/deployments')
+            data = r.json()
+            if len(data) > 0:
+                logging.info('Waiting for pod to be deployed %s', repr(data))
+                return False
+            # deployment complete
+            r = self.get('/pods' + pod_id)
+            data = r.json()
+            return data["scaling"]["instances"]
```
adam-mesos: Copying @mellenburg 's comments from #851 
> Why do we exit the retry on any number of instances being bought up? I would expect a method like this to block until all the instances declared in the definition have been deployed

> Also, I would prefer that we have a return  value similar to what deploy_app does. E.G. can we return all the service points for all the containers? That way I can deploy the dedicated test app inside of a pod and have use most of the same pipes.

greggomann:
This PR is also still missing the latest Marathon. @aquamatthias it sounds like we have a bugfix in flight for a config issue in snap26, can we bump Marathon here also when snap27 is available?

aquamatthias:
@greggomann Thanks for the hint. The config problem is fixed and  1.4.0-snap27 is released. 
Not sure how I can update this branch - but if you don't mind, please update the marathon part with:
```json
   "marathon": {
      "kind": "url_extract",
      "url": "http://downloads.mesosphere.io/marathon/v1.4.0-snap27/marathon-1.4.0-snap27.tgz",
      "sha1": "524c03b9dee1f9c0efce26fc068d08728d3f5d21"
    },
```
Thanks.

greggomann:
@aquamatthias sure no problem, I just bumped to snap27. Thanks for the build! Just for future reference, this branch is in the `mesosphere` fork so you should be able to push to it if necessary.

packages/mesos/buildinfo.json
greggomann: Readme updated.

gen/calc.py
```diff
@@ -375,7 +375,8 @@ def calculate_config_yaml(user_arguments):
 
 
 def calculate_mesos_isolation(enable_gpu_isolation):
-    isolators = 'cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,docker/runtime,docker/volume'
+    isolators = ('cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,' +
+                 'docker/runtime,docker/volume,volume/sandbox_path')
```
adam-mesos: Maybe worth converting this to an array of strings and then `join()` them with a comma?

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,37 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
...
+    This test verifies that a Marathon pods can be deployed.
+    """
+
+    test_uuid = uuid.uuid4().hex
+
+    pod_definition = {
+        'id': '/integration-test-pods-{}'.format(test_uuid),
+        'scaling': {'kind': 'fixed', 'instances': 1},
+        'environment': {'PING': 'PONG'},
+        'containers': [
```
mellenburg: I still have the same complaint here and I am upgrading it from a nit. Doing an echo test is a complete false sense of security. Just do
```
exec: {command {shell: touch foo}}
healthcheck: {command: {shell: test -f foo}}
```
Unless we have a health check, we have no idea if the deployment is remotely behaving as expected

test_util/marathon.py
```diff
@@ -205,6 +205,84 @@ def _pool_for_marathon_app(app_id):
...
+        def _wait_for_pod_deployment(pod_id):
+            r = self.get('v2/deployments')
+            data = r.json()
+            if len(data) > 0:
+                logging.info('Waiting for pod to be deployed %s', repr(data))
+                return False
+            # deployment complete
+            r = self.get('v2/pods' + pod_id)
+            data = r.json()
+            return data["scaling"]["instances"]
```
mellenburg: same criticism I had before, this should return something at least similar to what deploy_app does. I.E. the endpoints per container. If I deploy the a few python test servers in a pod, I would like this method to return how to reach those deployments

mellenburg: Additionally, this method must not return until the correct number of instances has been deployed. One would not consider something "deployed" if it had not yet met the number of instances required by the deployment definition.

packages/dcos-integration-test/extra/test_applications.py
greggomann: I added simple health checks along these lines for the pod tasks.

gen/calc.py
cmaloney: nit: There's no reason to use the parens and + here, just doing a line continuation would have been enough and probably be more readable.

packages/marathon/buildinfo.json
```diff
@@ -2,8 +2,8 @@
   "requires": ["java"],
   "single_source": {
     "kind": "url_extract",
-    "url": "https://downloads.mesosphere.com/marathon/v1.3.6/marathon-1.3.6.tgz",
-    "sha1": "0d6a3f0abd4188cbb7d9d64e729f16489cb9bc6f"
+    "url": "http://downloads.mesosphere.io/marathon/v1.4.0-snap29/marathon-1.4.0-snap29.tgz",
```
cmaloney: required: Https + downloads.mesosphere.com, part of the standard for all of these.

packages/mesos-modules/buildinfo.json
```diff
@@ -3,7 +3,7 @@
     "single_source" : {
       "kind": "git",
       "git": "https://github.com/dcos/dcos-mesos-modules.git",
-      "ref": "11d8b8fbab3f98112b403059446338a8a3a81190",
+      "ref": "22a6fc3f79d4c9e394c84f66d714d63715bd6a23",
```
cmaloney: This takes this back a commit: https://github.com/dcos/dcos-mesos-modules/commits/master. Why are we going backwards and not forwards in history?

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
             raise Exception("Application deployment failed - operation was not "
                             "completed in {} seconds.".format(timeout))
 
+    def deploy_pod(self, pod_definition, timeout=120):
```
cmaloney: nit: we never change this timeout, why have it as a parameter?

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+            timeout: a time to wait for the pod to reach 'Healthy' status
+                     after which the test should be failed.
+            check_health: wait until Marathon reports tasks as healthy before
+                          returning
+        Returns:
+            Scaling instance count
+        """
+        r = self.post('v2/pods', json=pod_definition)
+        logging.info('Response from marathon: {}'.format(repr(r.json())))
+        assert r.ok
```
cmaloney: must: This produces unusable / impossible to reason about what went wrong in the test result errors (all you get is r.ok was false, not what the status code was or the body which usually has useful info). Need to have an assertion comment which includes at least `r.status_code` and `r.body`. `r.raise_for_status()` is also an acceptable fix, although I prefer the more detailed assert

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+
+            for deployment in r.json():
+                if deployment_id == deployment.get('id'):
+                    logging.info('Waiting for pod to be destroyed')
+                    return False
+            logging.info('Pod destroyed')
+            return True
+
+        r = self.delete('v2/pods' + pod_id)
+        assert r.ok
```
cmaloney: r.ok assertion needs to be updated

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+        Args:
+            pod_id: id of the pod to remove
+            timeout: seconds to wait for destruction before failing test
+        """
+        @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda ret: not ret,
+                        retry_on_exception=lambda x: False)
+        def _destroy_pod_complete(deployment_id):
+            r = self.get('v2/deployments')
+            assert r.ok
```
cmaloney: r.ok assertion needs to be updated

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+            logging.info('Pod destroyed')
+            return True
+
+        r = self.delete('v2/pods' + pod_id)
+        assert r.ok
+
+        try:
+            _destroy_pod_complete(r.headers['Marathon-Deployment-Id'])
+        except retrying.RetryError:
+            raise Exception("Pod destroy failed - operation was not "
```
cmaloney: exceptions being raised based on other exceptions need to be `raise <x> from <original>`

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+            r.raise_for_status()
+            data = r.json()
+            if int(data['scaling']['instances']) != pod_definition['scaling']['instances']:
+                logging.info('Pod is still scaling. Continuing to wait...')
+                return False
+            return data
+        try:
+            return _wait_for_pod_deployment(pod_definition['id'])
+        except retrying.RetryError:
+            raise Exception("Pod deployment failed - operation was not "
```
cmaloney: This needs `from <orignial exception>` added so we can see the whole exception stack when debugging.

greggomann:
@cmaloney thanks for the helpful comments! I've addressed them, let me know what you think.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +59,41 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
...
+            {
+                'name': 'ct1',
+                'resources': {'cpus': 0.1, 'mem': 32},
+                'image': {'kind': 'DOCKER', 'id': 'busybox'},
```
cmaloney: pulling dockers, esp. with mesos containerizer we have found to be very flaky in CI for a mixture of issues. Do we have to use a docker here? Why can't we just use the host binaries / filesystem?

If we have to use a docker, then should use the same image we are using in the one docker pull test we have, `debian:jessie`: https://github.com/dcos/dcos/blob/master/test_util/marathon.py#L61

packages/dcos-ui/buildinfo.json
```diff
@@ -3,7 +3,7 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/dcos/dcos-ui.git",
-    "ref": "b51934df009bd2a86dec605cdc5b34f006feffef",
-    "ref_origin": "master"
+    "ref": "df3fc757e36684c307b55f6baf2085c5fa0120fb",
+    "ref_origin": "release/1.9"
```
cmaloney: This seems to be going back a considerable number of commits (https://github.com/dcos/dcos-ui/compare/df3fc757e36684c307b55f6baf2085c5fa0120fb...b51934df009bd2a86dec605cdc5b34f006feffef). Is that intentional?

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+                            Marathon API
+            check_health: wait until Marathon reports tasks as healthy before
+                          returning
+        Returns:
+            Scaling instance count
+        """
+        timeout = 120
+
+        r = self.post('v2/pods', json=pod_definition)
+        logging.info('Response from marathon: {}'.format(repr(r.json())))
```
cmaloney: nit: the response body might not be json which would cause this to error badly (say there was a permission denied because the JWT token expired so the API response comes from bouncer and isn't json). Putting this after the r.ok would probably make this easier to debug.

test_util/marathon.py
```diff
@@ -205,6 +205,83 @@ def _pool_for_marathon_app(app_id):
...
+            timeout: seconds to wait for destruction before failing test
+        """
+        @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda ret: not ret,
+                        retry_on_exception=lambda x: False)
+        def _destroy_pod_complete(deployment_id):
+            r = self.get('v2/deployments')
+            assert r.ok, 'status_code: {} body: {}'.format(r.status_code, r.body)
+
+            for deployment in r.json():
```
cmaloney: nit: It's a bit weird we can't just ask about the one deployment we want, and instead have to fetch all of them and iterate. That said, might just be how marathon's api is for now.

packages/dcos-ui/buildinfo.json
greggomann: @rcorral @orlandohohmeier we're trying to merge pods into DCOS master. Should we bump dcos-ui to the `release/1.9` branch, or is the dcos-ui master branch working well with pods again?

packages/dcos-integration-test/extra/test_applications.py
cmaloney: We used to deploy a registry and push images to it. Still had a whole lot of issues

greggomann: :(

test_util/marathon.py
mellenburg: Thanks for filing that @greggomann . If you touch this code again, I would recommend adding the JIRA reference into a code comment just so that its harder to forget about

greggomann: Good call, done.

packages/dcos-ui/buildinfo.json
adam-mesos: cc: @mlunoe while Rafa is OOO

mlunoe: Master branch should work well with pods!

greggomann:
I removed the dcos-ui update, hopefully the current dcos-ui SHA is sufficient. @mlunoe @orlandohohmeier: could you let us know if there are any new changes in dcos-ui master that we should pull in for pods?

test_util/marathon.py
```diff
@@ -205,6 +205,85 @@ def _pool_for_marathon_app(app_id):
...
+        @retrying.retry(wait_fixed=2000, stop_max_delay=timeout * 1000,
+                        retry_on_result=lambda ret: ret is False,
+                        retry_on_exception=lambda x: False)
+        def _wait_for_pod_deployment(pod_id):
+            # TODO(greggomann): Revisit this logic. Can we make a request for
+            # info on only our specific deployment? See DCOS-11707.
+            r = self.get('v2/deployments')
+            data = r.json()
+            if len(data) > 0:
+                logging.info('Waiting for pod to be deployed %s', repr(data))
```
jgehrcke: Nit: Using `%s` together with `repr()` is weird. That's what `%r` is for :-).

-> `logging.info('Waiting for pod to be deployed: %r', data)`

cmaloney: @greggomann please send a follow-up PR which fixes this, landing as it is currently.

cmaloney:
@greggomann https://teamcity.mesosphere.io/viewLog.html?buildId=487443&buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr&guest=1 one test failure to investigate so far

test_util/marathon.py
greggomann: Will do; thanks Cody.

PR #857, opened 2016-10-12
Change adminrouter ExecReload to work

cmaloney:
Cuttent one doesn't work when a `systemctl reload dcos-adminrouter` is run (same fix applies to dcos-adminrouter-agent)

Internal JIRA: https://mesosphere.atlassian.net/browse/DCOS-10580


cmaloney:
cc: @jgehrcke @vespian

jgehrcke:
:rocket:

PR #862, opened 2016-10-12
integrating changes made in acs codebase back into dcos

JackQuincy:
Moving large .deb files onto the ephemeral disk for faster read write
times on setup. Making about a 2 min speed up in standing up a cluster.(Not exhaustively tested, but speed up was noticeable in ad-hoc testing.
- Now stored in the  new folder /var/lib/mesos/dcos/
  Making ssh work on port 22 on the master load balancer. Achieved by:
- opening port 2222 for ssh on the master vms by modifiy sshd_config
- NATing port 22 on LB to port 2222 on master 0
- updating rest of networking resources to make that work.
  this is none ideal in a few ways:
- the arm template isn't perfectly optimized, all masters NICs depend on the new NAT rule but only one uses it, minor since the rule comes up very quickly. Didn't just drop the old NAT rule as that would be a breaking change, and wanted to do this in a none breaking manner.
- messes with the sshd_config. Required since NAT rules have to be 1:1 so I couldn't reuse old port. Also uses regex and sed to modify the file. Since we start from a known state it works just none ideal. Couldn't think of a better way to do it in cloud init.

### DIFF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -10,8 +10,8 @@ root:
       Environment=DEBIAN_FRONTEND=noninteractive
       StandardOutput=journal+console
       StandardError=journal+console
-      ExecStartPre=/usr/bin/curl -fLsSv --retry 20 -Y 100000 -y 60 -o /tmp/d.deb https://az837203.vo.msecnd.net/dcos-deps/docker-engine_1.11.2-0~xenial_amd64.deb
-      ExecStart=/usr/bin/bash -c "try=1;until dpkg -D3 -i /tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
+      ExecStartPre=/usr/bin/curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/tmp/d.deb https://az837203.vo.msecnd.net/dcos-deps/docker-engine_1.11.2-0~xenial_amd64.deb
+      ExecStart=/usr/bin/bash -c "try=1;until dpkg -D3 -i /var/tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
   - path: /etc/systemd/system/docker.service.d/execstart.conf
     permissions: "0644"
     content: |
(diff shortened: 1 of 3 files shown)


JackQuincy:
I can't see why the azure integration failed.  The vagrant appears to be a build system issue.   Can someone take a look at the details of the azure failure send them to me?

gen/azure/templates/acs.json
```diff
@@ -528,11 +528,45 @@
...
+        {
+          "id": "[concat(variables('masterSshInboundNatRuleIdPrefix'),'3')]"
+        }
+      ],
+      [
+        {
+          "id": "[concat(variables('masterSshInboundNatRuleIdPrefix'),'4')]"
+        }
+      ]
+    ],
```
lingmann: We'll need to be careful to expand this if the number of masters ever increase beyond 5. However, I think this is fine since I believe the ARM engine will hard fail if an expression evaluates to something indexed outside of the array... so we would catch with a 7 master integration test.

JackQuincy: Agreed, I wanted to do a true false, but arm templates don't allow it. :(

JackQuincy:
A comment about the limit you guys were hitting before. there is a limit but it is much higher than you experienced before. The compute resource provider had a bug which they fixed.  Last I knew you guys had 6000 more characters you could use in your custom data before running into an issue, currently your at about 8000 if I remember correctly.  I'll still try to make the changes you suggested though

gen/azure/cloud-config.yaml
```diff
@@ -59,11 +59,14 @@ runcmd:
...
-    - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /tmp/4.deb https://az837203.vo.msecnd.net/dcos-deps/libltdl7_2.4.6-0.1_amd64.deb
-    - bash -c "try=1;until dpkg -i /tmp/{1,2,3,4}.deb || ((try>9));do echo retry \$((try++));sleep \$((try*try));done"
+    - mkdir /var/lib/mesos/dcos
+    - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/lib/mesos/dcos/1.deb https://az837203.vo.msecnd.net/dcos-deps/libipset3_6.29-1_amd64.deb
+    - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/lib/mesos/dcos/2.deb https://az837203.vo.msecnd.net/dcos-deps/ipset_6.29-1_amd64.deb
+    - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/lib/mesos/dcos/3.deb https://az837203.vo.msecnd.net/dcos-deps/unzip_6.0-20ubuntu1_amd64.deb
+    - curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/lib/mesos/dcos/4.deb https://az837203.vo.msecnd.net/dcos-deps/libltdl7_2.4.6-0.1_amd64.deb
+    - sudo sed -i "s/^Port 22$/Port 22\nPort 2222/1" /etc/ssh/sshd_config
+    - sudo service ssh restart
+    - bash -c "try=1;until dpkg -i /var/lib/mesos/dcos/{1,2,3,4}.deb || ((try>9));do echo retry \$((try++));sleep \$((try*try));done"
```
JackQuincy: changed

gen/azure/cloud-config.yaml
```diff
@@ -74,7 +76,7 @@ bootcmd:
 disk_setup:
   ephemeral0:
     table_type: gpt
-    layout: [ 50, 50 ]
+    layout: [ 25, 50, 25 ]
```
lingmann: I'd like to keep the allocation between docker and mesos partitions to a 50/50 split... so can we change this to: `layout: [ 48, 48, 4 ]`? With the default ephemeral disk size of 50GB, this will give users 2GB of tmp space, which should be plenty of room for the download and extract of DC/OS and the deb packages (~700MB required).

Also, @JackQuincy do you know if it is possible for users to encounter ephemeral disk sizes smaller than 50GB? If so, we may need to increase the percentage for /var/tmp.

lingmann:
@JackQuincy if you could address my comment on the layout and squash this PR into a single commit, I think this will be good to go. Once squashed I'll run a final pass test and we can get this merged. Thanks!!

gen/azure/cloud-config.yaml
lingmann: Ok, this might actually be a problem with some of the new instance types. We have another PR here: https://github.com/dcos/dcos/pull/844 to add support for new VM types (and also limit to only instances which have 2+ cores which is the min required for DC/OS).

According to https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-windows-sizes/ the Standard_F2s only has an 8gb ephemeral disk... in which case we would need to allocate at least 10% for the tmp volume. I'm going to investigate a bit further to see if we can specify both absolute and percentage with the cloud-init disk_setup stanza.

JackQuincy: So internally we are currently planning on restricting vm sizes to only vms that have at least a 100GB ephemeral disk. Does that seem reasonable to you? I could update this template and then set layout to 44,44,2.

lingmann:
@JackQuincy I think I would rather support a wider range of instance types at the expense of some ephemeral disk space. So I'd prefer to do a 45,45,10 layout for now. Also, it appears that one thing missing in this PR is the 550MB DC/OS bootstrap tarball is still downloading to `/tmp`. You can see how it needs to be adjusted in my [test branch here](https://github.com/lingmann/dcos/commit/71595ae14e19777c7cb935dee684fb46c953a91a).

So if you could take care of the layout update, the bootstrap fix, and squashing this all into 1 or 2 commits this will be ready to merge. I'm happy to help with that you don't have time. LMK. Thanks.

JackQuincy:
thanks for the help Jeremy. I just got time to make the last couple changes. Let me know if I did the layout/bootstrap right. I haven't looked at this in a while.

lingmann:
Thanks @JackQuincy this is ready to merge! I've also run tests verifying that ports 22 and 2200+ are forwarding as expected.

PR #868, opened 2016-10-16
Rework gen/ to be made of three primitives: Source, Target, Scope and Resolver

cmaloney:
This reworks the gen a bit to make the library more understandable in terms of what it is doing. It introduces four main primitives:

`Source` is a source of configuration variables. A `Target` is a set of configuration variables which
need to be set to have a complete configuration. The `Resolver` figures out how to use a collection of
sources to fill in a collection of targets and check for conflicts.

`Source`s can be user-provided (a dictionary loaded out of a config file). They can also be provided
by different contexts (AWS templates have a source of AWS which implies things likes resolvers). A
third common provider is deriving from other already given values (master count is derived from a
master list if a master list is given).

`Target`s are a set of variables which need to be set, as well as a set of variables which need to be
set under certain conditions. The "conditional" variables you can think of as a "Switch". Based on
the value of the argument to the switch, one of multiple paths are going to be taken, and all
configuration variables in that particular path are expected to be taken.

The `Resolver` takes a collection of sources and targets, then tries to calculate (and validate) all
the values needed to satisfy the targets. It keeps track of the path to get to a specific value, if
there are any conflicts between different sources (master_count is set to a conflicting / different
value than the length of master_list).

### DIFF
--- a/gen/__init__.py
+++ b/gen/__init__.py
@@ -12,23 +12,23 @@ Operates strictly:
 """
 
 import importlib.machinery
-import inspect
 import json
 import logging as log
 import os
 import os.path
 import textwrap
 from copy import copy, deepcopy
-from functools import partialmethod
 from subprocess import check_call
 from tempfile import TemporaryDirectory
 
 import yaml
 
 import gen.calc
+import gen.internals
 import gen.template
+from gen.exceptions import ValidationError
 from pkgpanda import PackageId
-from pkgpanda.util import load_string, make_tar
+from pkgpanda.util import json_prettyprint, load_string, make_tar, write_json
 
 # List of all roles all templates should have.
 role_names = {"master", "slave", "slave_public"}
@@ -186,44 +186,6 @@ def merge_dictionaries(base, additions):
     return base_copy
 
 
-# Recursively merge to python dictionaries.
-# If both base and addition contain the same key, that key's value will be
-# merged if it is a dictionary.
-# This is unlike the python dict.update() method which just overwrites matching
-# keys.
-def merge_replace_append_dictionaries(base, additions):
-    base_copy = base.copy()
-    for k, v in additions.items():
-        try:
-            if k not in base:
-                base_copy[k] = v
-                continue
-            if isinstance(v, dict) and isinstance(base_copy[k], dict):
-                base_copy[k] = merge_replace_append_dictionaries(base_copy.get(k, dict()), v)
-                continue
-
-            # Append arrays
-            if isinstance(v, list) and isinstance(base_copy[k], list):
-                base_copy[k].extend(v)
-                continue
-
-            # Merge sets
-            if isinstance(v, set) and isinstance(base_copy[k], set):
-                base_copy[k] |= v
-                continue
-
-            # Overwrite if the new one is a string
-            if isinstance(v, str):
-                base_copy[k] = v
-                continue
-
-            # Unknown types
-            raise ValueError("Can't merge type {} into type {}".format(type(v), type(base_copy[k])))
-        except ValueError as ex:
-            raise ValueError("{} inside key {}".format(ex, k)) from ex
-    return base_copy
-
-
 def load_templates(template_dict):
     result = dict()
     for name, template_list in template_dict.items():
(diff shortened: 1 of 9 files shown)


gen/template.py
```diff
@@ -378,32 +380,28 @@ def render_ast(ast):
 
         return render_ast(self.ast)
 
-    def get_scoped_arguments(self):
+    def make_target(self):
         def variables_from_ast(ast, blacklist):
```
branden: This should be renamed to `target_from_ast` to match the new return type.

gen/__init__.py
```diff
@@ -267,209 +231,20 @@ def render_templates(template_dict, arguments):
     return rendered_templates
 
 
-# Load all the un-bound variables in the templates which need to be given values
-# in order to convert the templates to go from jinja -> final template. These
-# are effectively the set of DC/OS parameters.
+# Collect the un-bound / un-set variables from all the given templates to build
+# the schema / configuration target. The templates and their structure serve
+# as the schema for what configuration a user must provide.
 def get_parameters(template_dict):
```
branden: `target_from_templates`?

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+}
+
+
+def get_function_parameters(function):
+    return set(inspect.signature(function).parameters)
+
+
+def validate_arguments_strings(arguments: dict):
+    errors = dict()
+    # Validate that all keys and vlaues of arguments are strings
```
orsenthil: prefer killing this comment.

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+class Source:
+    def __init__(self, entry=None, is_user=False,):
+        self.setters = dict()
+        self.validate = list()
+        self.is_user = is_user
+        if entry:
+            self.add_entry(entry, False)
+
+    def add_setter(self, name, value, is_optional, conditions):
+        self.setters.setdefault(name, list()).append(Setter(name, value, is_optional, conditions, self.is_user))
```
orsenthil: Do we need to see if `name` is a list or list like object?

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+            del_setter(name)
+
+        for name, cond_options in scope.get('conditional', dict()).items():
+            if name in self.setters:
+                raise NotImplementedError("Should conditional setters overwrite all setters?")
+
+    def add_entry(self, entry, replace_existing):
+        if replace_existing:
+            self.remove_setters(entry)
+
```
orsenthil: Kill empty line.

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+class CalculatorError(Exception):
+    def __init__(self, message, chain=[]):
+        assert isinstance(message, str)
+        assert isinstance(chain, list)
+        self.message = message
+        self.chain = chain
+        super().__init__(message)
+
+
+# Depth first search argument calculator. Detects cycles, as well as unmet
```
orsenthil: You could move this into function docstrings.

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+# dependencies.
+# TODO(cmaloney): Separate chain / path building when unwinding from the root
```
orsenthil: It might be a good practice to create a JIRA ticket against this. Usually, TODO's just languish in the code, a strategy is required to prevent it.

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+            self._calculate_target(sub_scope.cases[choice])
+
+        # Perform all multi-argument validations
+        for parameter_set, validate_fn in self._multi_arg_validate.items():
+            # Build up argument map for validate function. If any arguments are
+            # unset then skip this validate function.
+            kwargs = dict()
+            skip = False
+            for parameter in parameter_set:
+                if parameter not in self._arguments or self._arguments[parameter] is None:
```
orsenthil: I would parenthesize this in order to prevent any tricky operator precedence rule that can go un-noticed.

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+                continue
+
+            # Call the validation function, catching AssertionErrors and turning them into errors in
+            # the error dictionary.
+            try:
+                validate_fn(**kwargs)
+            except AssertionError as ex:
+                self._errors[parameter_set] = ex.args[0]
+
+        # TODO(cmaloney): Return per-target results with the path to calculate each argument and the
```
orsenthil: JIRA for TODO?

gen/internals.py
```diff
@@ -0,0 +1,464 @@
...
+        if len(self._errors) or len(self._unset):
+            raise ValidationError(self._errors, self._unset)
+
+        return self._arguments
+
+
+def resolve_configuration(sources: list, targets: list, user_arguments: dict):
+
+    # Make sure all user provided arguments are strings.
+    # TODO(cmaloney): Loosen this restriction  / allow arbitrary types as long
```
orsenthil: JIRA for TODO?

cmaloney: I don't believe generally making JIRAs for TODOs creates useful data / input points. We should at some point make a tool which gathers TODOs and presents them, and maybe gives an easy button to make a JIRA for them, but manually keeping TODOs in sync with JIRA is going to be a very long task without a lot of benefit.

It's also not the standard for the codebase at this point in time.

gen/internals.py
cmaloney: I'm a fan of leaving vertical whitespace between things (esp. after ifs). Practically things like the whitespace comes down to a whole lot of personal preference, so I tend to try and avoid it in code reviews (our codebase isn't consistent, and trying to get it consistent isn't of great value). Tox / code quality checks enforce the major whitespace bits that are cared about / required for landing.

gen/internals.py
cmaloney: this isn't doing anything like that with it, it's just saying `if the dictionary setters has an entry for name already do nothing, otherwise create an entry and set it to be an empty list`. We then append to that list. setdefault() is doing an "initialize if uninitialized".

gen/internals.py
cmaloney: This was explicitly added previously to help with confusion people had

cmaloney:
@branden: Yes, resolver roughly maps to DFSArgumentCalculator. In future patches in this series, there will be some `.resolve()` functions which should show up (`config.resolve([target1, target2], [source1, source2])`. It's also because the core entry to the "guts" of gen is now `gen.internals.resolve_configuration()`.

The gen.generate() + gen.validate() functions are fairly out of date / not matching what they're doing (they've never been great names), and there's a decent chance it'll turn into a `gen.resolve()` which does a resolve_configuration, then the "load templates" + "render cluster packages" is part of a different subsystem then the "gen.internals" / configuration resolution stuff (That it moved to gen.internals was just to get it out of the base gen for now. Probably won't be it's final resting place)

PR #877, opened 2016-10-20
Ship all container logs to journald

kaysoky:
Should be merged only after the appropriate UI modifications and logging service have been added.

* Activates the journald container logger module, which replaces the logrotate container logger module.


malnick:
This is a blocker for https://github.com/dcos/dcos/pull/832

malnick:
Better yet, can we ship this as "inactive" for the timebeing, so we do not tie to many things to this single PR?

cmaloney:
This PR has some issues currently, just looking briefly: Building and shipping our own systemd-journal is a no-go

kaysoky:
@cmaloney How do you propose we get the journald library?

cmaloney:
@kaysoky it is shipped with systemd always or rely on the specific, stable APIs to interact with journald.

kaysoky:
The headers aren't available even if systemd is available.  And the build system is a non-systemd container, so it won't be able to link unless we have the systemd library somewhere.

On CentOS, setup is usually `yum install systemd-devel`.

kaysoky:
Is adding `apt-get install -y libsystemd-journal-dev` to the docker file acceptable?

malnick:
@cmaloney I want to clarify this, we need the journald headers not because we are shipping a new journald to the host systems, we need the headers in our code directly. @darkonie Also needs these headers for the journald API PR, referenced above. Both projects use them, but this does not affect system journald on the hosts to which we install DC/OS.

packages/mesos-modules/build
```diff
@@ -1,15 +1,53 @@
 #!/bin/bash
+set -o nounset -o pipefail
+
+# Update the image and download build tools.
+apt-get update -qq
+apt-get install -y autoconf libtool make pkg-config
```
cmaloney: This sort of thing has to happen in a docker build, it is not reliable / re-producible, which means this package build with it is not reproducible.

The dco-ca package has an example of this currently: https://github.com/mesosphere/dcos-enterprise/tree/master/packages/dcos-ca. (Packages can define a Dockerfile to build the build env for that package)

The other option is to add it to the base DC/OS package build docker.

The docker is built once, and as long as the Dockerfile doesn't change forever re-used, so that individual package builds are hopefully deterministic.

Ideally for these though we _really_ need to use the standard / same docker we use for building:
1. mesos dependencies
2. mesos
3. mesos-modules

If those dockers differ, than odd incompatibilities and crashes start showing up.

packages/mesos-modules/build
```diff
@@ -1,15 +1,53 @@
...
+
+# Use an older version of GCC (older than the one in `build-essential`)
+# so that we're ABI-compatible with CentOS 7.
+apt-get install -y gcc-4.8 g++-4.8
```
cmaloney: the GCC version is not what matters here. We need to use the same build environment top to bottom. There's a lot that can cause binary incompatiblities outside of gcc version.

packages/mesos-modules/build
```diff
@@ -1,15 +1,53 @@
...
+  libsvn-dev
+
 pushd /pkg/src/mesos-modules/
 ./bootstrap
 popd
 
-source /opt/mesosphere/environment.export
+# Install the systemd headers.
+# This is needed by one of the modules.
+apt-get install -y libsystemd-dev libsystemd-journal-dev
```
cmaloney: can't do apt-get install inside build scripts, it is not reproducible at all.

packages/mesos-modules/build
kaysoky: The dependency chain for `mesos-modules` is:
```
 mesos-modules
      |
    mesos
/      |     \
|  libevent   curl
\      |      / \
 --- openssl   python-requests
                     |
                  python
```
`openssl` and `python` in particular are dependencies for tons of other packages, so instead of selectively adding a new build image, it seems better to change the default image (the second option you referred to).

cmaloney: the openssl, curl, python we bundle in DC/OS, shouldn't use the ones from the base image docker.

kaysoky:
Closed for now, as I split apart the build system changes from the module changes (into separate PRs)

malnick:
NOTE: This PR needs to go into master at the same time that @mlunoe ships the UI bits. Both of those PR's require #923 to be on master first, but that component can ship independently.

CC @spahl

cmaloney:
That would mean that all these bits have to be reverted at the same time if there's an issue. Is there any way to make it so pieces can land independently? (Log to both places or something similar?). Practically we also need to think about what happens at upgrade:

What happens when I use the new UI to access old hosts / logs because masters are upgraded but not agents?

kaysoky:
While I haven't seen the UI in action, the UI should be able to go in independently of this change.  I assume the masters (and hence the UI) are generally upgraded before the agents.

When the UI hits an old version of the DC/OS agent, it should continue displaying the sandbox contents via `/files/browse` and `/files/read`.  The `stdout` and `stderr` file will still appear in those sandboxes.

cmaloney:
@kaysoky / @malnick is this pr still relevant?

kaysoky:
Yes.  The other logging changes (to the UI, CLI, and services) need to go in first, as they should be backwards compatible with the existing logging system; as well as the journald-based one.

spahl:
closing in favor of https://github.com/dcos/dcos/pull/1045

PR #880, opened 2016-10-22
Update dcos_installer/backend.py to use s3 region when aws_template_upload is true

BenWhitehead:
When config.yaml provided to dcos_installer/backend.py contains `aws_template_upload: true` `aws_template_storage_region_name` will be set to the actual region of the specified s3 bucket.

The actual region of the s3 bucket is determined by calling `get_bucket_location` using boto3 s3 client.

### DIFF
--- a/dcos_installer/backend.py
+++ b/dcos_installer/backend.py
@@ -10,7 +10,6 @@ import botocore.exceptions
 import gen
 import gen.calc
 import gen.installer.aws
-import gen.internals
 import release
 import release.storage.aws
 import release.storage.local
@@ -105,6 +104,14 @@ def validate_aws_bucket_access(aws_template_storage_region_name,
                 aws_template_storage_bucket_path, aws_template_storage_bucket, ex)) from ex
 
 
+def validate_aws_template_storage_access_key_id(aws_template_storage_access_key_id):
+    assert aws_template_storage_access_key_id, "Must be non-empty"
+
+
+def validate_aws_template_storage_secret_access_key(aws_template_storage_secret_access_key):
+    assert aws_template_storage_secret_access_key, "Must be non-empty"
+
+
 def calculate_reproducible_artifact_path(config_id):
     return 'config_id/{}'.format(config_id)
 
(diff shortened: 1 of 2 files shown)


BenWhitehead:
rebased on `master`

BenWhitehead:
@cmaloney All green and ready for review

tests/test_installer_backend.py
```diff
@@ -1,15 +1,41 @@
...
+import gen.installer.aws
+import release
 from dcos_installer import backend
 from dcos_installer.config import Config, make_default_config_if_needed, to_config
 
 os.environ["BOOTSTRAP_ID"] = "12345"
 
 
+@pytest.fixture(scope='module')
+def config():
```
cmaloney: Just as a note, discussed this offline previously, sadly this is the easiest way to get the fixtures here for now. Ideally we should have one copy.

@mellenburg any idea? (tried putting them in a common import but that blows up in fun ways like the upstream mixed with downstream tests  / fixture stuff you ran into)

tests/test_installer_backend.py
```diff
@@ -221,16 +247,57 @@ def test_do_configure(tmpdir):
...
+    create_fake_build_artifacts(aws_base_config, tmpdir)
+
+    with tmpdir.as_cwd():
+        assert backend.do_aws_cf_configure() == 0
+
+
+valid_storage_config = """---
+master_list:
+ - 127.0.0.1
+aws_template_storage_access_key_id: {}
```
cmaloney: Would be good to use named parameters in the format string here to make it more readable what is going where (And easier to extended in the future)

Rather than `{}` can do `{name}` and then `.format(name="foobar")`

dcos_installer/backend.py
```diff
@@ -10,7 +10,6 @@
 import gen
 import gen.calc
 import gen.installer.aws
-import gen.internals
```
BenWhitehead: I think the import was seen as not needed because the fully qualified name was being used at the use site. I can add it back in if you want, but it doesn't seem to be necessary.

dcos_installer/backend.py
```diff
@@ -127,8 +126,28 @@ def calculate_cloudformation_s3_url(bootstrap_url, config_id):
...
+        aws_template_storage_secret_access_key,
+        aws_template_storage_bucket):
+
+    session = boto3.session.Session(
+        aws_access_key_id=aws_template_storage_access_key_id,
+        aws_secret_access_key=aws_template_storage_secret_access_key)
+
+    try:
+        location_info = session.client('s3').get_bucket_location(Bucket=aws_template_storage_bucket)
+        loc = location_info["LocationConstraint"]
```
BenWhitehead: will cleanup

tests/test_installer_backend.py
BenWhitehead: will do

dcos_installer/backend.py
```diff
@@ -127,8 +126,32 @@ def calculate_cloudformation_s3_url(bootstrap_url, config_id):
     return '{}/config_id/{}'.format(bootstrap_url, config_id)
 
 
+def calculate_aws_template_storage_region_name(
+        aws_template_storage_access_key_id,
+        aws_template_storage_secret_access_key,
+        aws_template_storage_bucket):
+
+    if not aws_template_storage_access_key_id:
```
cmaloney: nit: These would be better as single-parameter validation functions (Single parameter ones are guaranteed to always be evaluated before any multi-parameter ones. There are two "phases" more or less)

dcos_installer/backend.py
```diff
@@ -127,8 +126,28 @@ def calculate_cloudformation_s3_url(bootstrap_url, config_id):
     return '{}/config_id/{}'.format(bootstrap_url, config_id)
 
 
+def calculate_aws_template_storage_region_name(
+        aws_template_storage_access_key_id,
+        aws_template_storage_secret_access_key,
+        aws_template_storage_bucket):
+
+    session = boto3.session.Session(
+        aws_access_key_id=aws_template_storage_access_key_id,
```
BenWhitehead: Validation functions added and wired into the `Source`

PR #881, opened 2016-10-22
dcos-spaceport: cluster configuration, assembly, and launch, and maintenance code

cmaloney:
This replaces `dcos_generate_config.sh` with a PyInstaller binary (linux-only for now) which is called `dcos-spaceport`. The `dcos-spaceport` should have the functionality of dcos_generate_config.sh (although restructuring the command line a bit to cleanup some old tech debt), while also bring in a few more members of the family
1. `dcos-spaceport launch`: What was being worked on as dcos-launch, as a subcommand.
2. `dcos-spaceport upgrade node`: Upgrade or create a script to upgrade an individual node in a DC/OS cluster
3. `dcos-spaceport config-update`: Guidance around migrating from current config.yaml to the new / next config.yaml.

The goal would be to cleanup some old installer tech debt, as well as "unify" the code into a single cohesive goal, re-arranging it a bit to make more sense. The code structure I have been thinking of as a goal to ehad towards:

```
spaceport
  internals/
    history/  # (dcos-history-service)
    bootstrap/  # (Common code from the bootstrap package to share with downstreams)
    pkgpanda/  # (pkgpanda/)
  config/  # (what was gen + dcos_installer/config.py)
  launch/  # (what was test_util/launch, as well as possibly the ssh and web installers)
    ssh/     # (the ssh library, might just go in internals)
  release/  # The release/ tools
packages/  # Just what it is today
```

### DIFF
--- a/spaceport/cli.py
+++ b/spaceport/cli.py
@@ -1,0 +1,20 @@
+"""DC/OS Spaceport: Configuration, Assemble, Launch, and Update DC/OS
+
+Usage:
+  dcos-spaceport launch
+  dcos-spaceport configure
+  dcos-spaceport configure aws-advanced
+  dcos-spaceport configure azure-advanced
+  dcos-spaceport configure ssh
+  dcos-spaceport web
+"""
+
+from docopt import docopt
+
+
+def main():
+    docopt(__doc__)
+
+
+if __name__ == '__main__':
+    main()
--- a/build_dcos_spaceport.sh
+++ b/build_dcos_spaceport.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+#
+# Simple helper script to build dcos-spaceport binary for Linux
+# NOTE: this needs to be kept in sync with gen.installer.bash::make_dcos_spaceport()
+
+set -x
+set -o errexit -o pipefail
+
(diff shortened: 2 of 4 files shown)


spaceport/cli.py
```diff
@@ -0,0 +1,22 @@
...
+
+Usage:
+  dcos-spaceport launch
+  dcos-spaceport configure
+  dcos-spaceport configure aws
+  dcos-spaceport configure azure
+  dcos-spaceport configure ssh
+  dcos-spaceport build <web>
+  dcos-spaceport update node
+  dcos-spaceport config-update
```
amitaekbote: update-config?

cmaloney:
Can't just do `dcos` because that's what the DC/OS cli uses.

That said, when there's a binary:
`dcos-foo`

then the CLI will make it available as a subcommand:
`dcos foo`, so `dcos-spaceport` if you have the cli installed will be runnable as `dcos spaceport`

cmaloney:
Needs code so this is built like dcos-launch with every dc/os build

orsenthil:
@cmaloney, please associate a JIRA ticket with this PR.

PR #882, opened 2016-10-24
Add dcos-launch

mellenburg:
Initial implementation of DC/OS Launch Utility
-only covers launching generic AWS CF simple templates. Can technically launch any generic cloudformation template, but it all features will only work for a simple deployment
-Covers create, wait, describe, test, and delete
-test optionally takes command line args for interactively working on tests

# Checklist

### DIFF
--- a/gen/build_deploy/bash.py
+++ b/gen/build_deploy/bash.py
@@ -12,6 +12,7 @@ import gen.template
 import pkgpanda.util
 from gen.calc import calculate_environment_variable
 from gen.internals import Source
+from pkgpanda.build.src_fetchers import GitLocalSrcFetcher
 from pkgpanda.util import logger
 
 
@@ -654,12 +655,18 @@ def make_installer_docker(variant, bootstrap_id, installer_bootstrap_id):
 def make_dcos_launch():
     # NOTE: this needs to be kept in sync with build_dcos_launch.sh
     work_dir = py.path.local.mkdtemp()
+    # the tree will not be in the working dir if run from downstream
+    tree_src = './'
+    if os.path.exists('ext/upstream'):
+        tree_src = './ext/upstream'
+    src_info = {'kind': 'git_local', 'rel_path': tree_src}
+    this_tree = GitLocalSrcFetcher(src_info, None, os.getcwd())
+    this_tree.checkout_to(str(work_dir))
+    # put the launch spec into the root of the tree before running pyinstaller
     work_dir.join('dcos-launch.spec').write(pkg_resources.resource_string(__name__, 'bash/dcos-launch.spec'))
-    work_dir.join('test_util').ensure(dir=True)
-    work_dir.join('test_util').join('launch.py').write(pkg_resources.resource_string('test_util', 'launch.py'))
     with work_dir.as_cwd():
-        subprocess.check_call(['pyinstaller', 'dcos-launch.spec'])
-    subprocess.check_call(['mv', str(work_dir.join('dist').join('dcos-launch')), "dcos-launch"])
+        subprocess.check_call(['pyinstaller', '--log-level=DEBUG', 'dcos-launch.spec'])
+    subprocess.check_call(['mv', str(work_dir.join('dist').join('dcos-launch')), 'dcos-launch'])
     work_dir.remove()
 
     return "dcos-launch"
--- a/gen/build_deploy/bash/dcos-launch.spec
+++ b/gen/build_deploy/bash/dcos-launch.spec
@@ -1,4 +1,12 @@
-a = Analysis(['test_util/launch.py'])
+# Warning:
+#   - hidden import must be used as there is a bug in pyinstaller
+#     https://github.com/pyinstaller/pyinstaller/issues/2185
+#   - data must be decalared explicitly if not a .py file
+#   - Building will suck up the local SSL .so and package it
+#     with the final exe. Ensure build system has OpenSSL 1.0.2g or greater
+a = Analysis(['test_util/launch.py'],
+             hiddenimports=['html.parser'],
+             datas=[('gen/ip-detect/*.sh', 'gen/ip-detect/')])
 pyz = PYZ(a.pure, a.zipped_data, cipher=None)
 exe = EXE(
     pyz,
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -48,8 +48,8 @@ class BotoWrapper():
     def create_key_pair(self, key_name):
         """Returns private key of newly generated pair
         """
-        key = self.resource('ec2').KeyPair(key_name)
-        return key.key_material
+        key = self.client('ec2').create_key_pair(KeyName=key_name)
+        return key['KeyMaterial']
 
     def delete_key_pair(self, key_name):
         self.resource('ec2').KeyPair(key_name).delete()
--- a/test_util/launch.py
+++ b/test_util/launch.py
@@ -1,20 +1,292 @@
-"""
+"""DC/OS Launch
 
 Usage:
-dcos-launch create [--wait] [--dump-info=cluster_info.json] <config=config.yaml>
-dcos-launch wait <path=cluster_info.json>
-dcos-launch describe <path=cluster_info.json>
-dcos-launch delete <path=cluster_info.json>
+  dcos-launch create [--config-path=<path> --info-path=<path>]
+  dcos-launch wait [--info-path=<path>]
+  dcos-launch describe [--info-path=<path>]
+  dcos-launch pytest [--info-path=<path> --env=<envlist>] [--] [<pytest_extras>]...
+  dcos-launch delete [--info-path=<path>]
+
+Commands:
+  create    Reads the file given by --config-path, creates the cluster described therein
+            and finally dumps a JSON file to the path given in --info-path which can then
+            be used with the wait, describe, and delete calls.
+  wait      Block until the cluster is up and running.
+  describe  Return additional information about the composition of the cluster.
+  pytest    Runs integration test suite on cluster. Can optionally supply options and arguments to pytest
+  delete    Destroying the provided cluster deployment.
+
+Options:
+  --config-path=<path>  Path for config to create cluster from [default: config.yaml].
+  --info-path=<path>    JSON file output by create and consumed by wait, describe,
+                        and delete [default: cluster_info.json].
+  --env=<envlist>       Specifies a comma-delimited list of environment variables to be
+                        passed from the local environment into the test environment.
 """
+import abc
+import copy
+import os
+import sys
 
 from docopt import docopt
 
+import ssh.tunnel
+import test_util.runner
+from pkgpanda.util import json_prettyprint, load_json, load_string, load_yaml, write_json, YamlParseError
+from test_util.aws import BotoWrapper, DcosCfSimple
 
-def main():
-    docopt(__doc__)
 
-    raise NotImplementedError()
+class LauncherError(Exception):
+    def __init__(self, error, msg):
+        self.error = error
+        self.msg = msg
+
+    def __repr__(self):
+        return '{}: {}'.format(self.error, self.msg)
+
+
+class AbstractLauncher(metaclass=abc.ABCMeta):
+    def create(self, config):
+        raise NotImplementedError()
+
+    def wait(self, info):
+        raise NotImplementedError()
+
+    def describe(self, info):
+        raise NotImplementedError()
+
+    def delete(self, info):
+        raise NotImplementedError()
+
+    def ssh_from_config(self, info):
+        raise NotImplementedError()
+
+    def test(self, info):
+        raise NotImplementedError()
+
+
+class AwsCloudformationLauncher(AbstractLauncher):
+    def __init__(self, boto_wrapper):
+        self.boto_wrapper = boto_wrapper
+
+    def create(self, config):
+        check_keys(config, ['stack_name', 'template_url'])
+        ssh_info = self.ssh_from_config(config)
+        # NOTE: even if parameters not given, ssh_from_config will add KeyName
+        # parameter to config['parameters']
+        self.boto_wrapper.create_stack(
+            config['stack_name'], config['template_url'], config['parameters'])
+        return {
+            'type': 'cloudformation',
+            'stack_name': config['stack_name'],
+            'provider': {
+                'region': config['provider_info']['region'],
+                'access_key_id': config['provider_info']['access_key_id'],
+                'secret_access_key': config['provider_info']['secret_access_key']},
+            'ssh': ssh_info}
+
+    def wait(self, info):
+        # TODO: should this support the case where the cluster is being updated?
+        cf = self.get_stack(info)
+        status = cf.get_stack_details()['StackStatus']
+        if status == 'CREATE_IN_PROGRESS':
+            cf.wait_for_stack_creation(wait_before_poll_min=0)
+        elif status == 'CREATE_COMPLETE':
+            return
+        else:
+            raise LauncherError('WaitError', 'AWS Stack has entered unexpected state: {}'.format(status))
+
+    def describe(self, info):
+        desc = copy.copy(info)
+        cf = self.get_stack(info)
+        desc.update({
+            'masters': convert_host_list(cf.get_master_ips()),
+            'private_agents': convert_host_list(cf.get_private_agent_ips()),
+            'public_agents': convert_host_list(cf.get_public_agent_ips())})
+        return desc
+
+    def delete(self, info):
+        if info['ssh']['delete_with_stack']:
+            self.boto_wrapper.delete_key_pair(info['ssh']['key_name'])
+        self.get_stack(info).delete()
+
+    def ssh_from_config(self, config):
+        """
+        In AWS simple deploys, SSH is only used for running the integration test suite.
+        In order to use SSH with AWS, one must use an SSH key pair that was previously created
+        in and downloaded from AWS. The private key cannot be obtained after its creation, so there
+        are three supported possibilities:
+        1. User has no preset key pair and would like one to be created for just this instance.
+            The keypair will be generated and the private key and key name will be added to the
+            ssh_info of the cluster info JSON. This key will be marked in the ssh_info for deletion
+            by dcos-launch when the entire cluster is deleted. SSH user name cannot be inferred so
+            it must still be provided in the ssh_info of the config
+        2. User has a preset AWS KeyPair and no corresponding private key. Testing will not be possible
+            without the private key, so ssh_info can be completely omitted with no loss.
+        3. User has a preset AWS KeyPair and has the corresponding private key. The private key must be
+            pointed to with private_key_path in the config ssh_info along with user.
+
+        Case 2 and 3 require specifying key name. The key name can be provided to dcos-launch in two ways:
+        1. Passed directly to the template like other template parameters.
+        ---
+        parameters:
+          - ParameterKey: KeyName
+            ParameterValue: my_key_name
+        ssh_info:
+          user: foo
+          private_key_path: path_to_my_key
+        2. Provided with the other SSH paramters:
+        ---
+        ssh_info:
+          user: foo
+          key_name: my_key_name
+          private_key_path: path_to_my_key
+
+        This method will take the config dict, determine if a key name is provided, if necessary, generate
+        the key and amend the config, and return the ssh_info for the cluster info JSON
+        """
+        generate_key = False
+        key_name = None
+        private_key = None
+        # Native AWS parameters take precedence
+        if 'parameters' in config:
+            for p in config['parameters']:
+                if p['ParameterKey'] == 'KeyName':
+                    key_name = p['ParameterValue']
+        # check ssh_info data
+        if 'ssh_info' in config:
+            if key_name is None:
+                key_name = config['ssh_info'].get('key_name', None)
+            elif 'key_name' in config['ssh_info']:
+                raise LauncherError('ValidationError',
+                                    'Provide key name as either a parameter or in ssh_info; not both')
+            if 'private_key_path' in config['ssh_info']:
+                if key_name is None:
+                    raise LauncherError('ValidationError', 'If a private_key_path is provided, '
+                                        'then the KeyName template parameter must be set')
+                private_key = load_string(config['ssh_info']['private_key_path'])
+        if not key_name:
+            key_name = config['stack_name']
+            generate_key = True
+            private_key = self.boto_wrapper.create_key_pair(key_name)
+            cf_params = config.get('parameters', list())
+            cf_params.append({'ParameterKey': 'KeyName', 'ParameterValue': key_name})
+            config['parameters'] = cf_params
+        return {
+            'delete_with_stack': generate_key,
+            'private_key': private_key,
+            'key_name': key_name,
+            'user': config.get('ssh_info', dict()).get('user', None)}
+
+    def test(self, info, test_cmd):
+        ssh_info = info['ssh']
+        for param in ['user', 'private_key']:
+            if param not in ssh_info:
+                raise LauncherError('MissingInfo', 'Cluster was created without providing ssh_info: {}'.format(param))
+        details = self.describe(info)
+        test_host = details['masters'][0]['public_ip']
+        # TODO: section should be generalized for use by other pr
+        with ssh.tunnel.tunnel(ssh_info['user'], ssh_info['private_key'], test_host) as test_tunnel:
+            return test_util.runner.integration_test(
+                tunnel=test_tunnel,
+                dcos_dns=test_host,
+                master_list=[m['private_ip'] for m in details['masters']],
+                agent_list=[a['private_ip'] for a in details['private_agents']],
+                public_agent_list=[a['private_ip'] for a in details['public_agents']],
+                aws_access_key_id=info['provider']['access_key_id'],
+                aws_secret_access_key=info['provider']['secret_access_key'],
+                region=info['provider']['region'],
+                test_cmd=test_cmd)
+
+    def get_stack(self, info):
+        """Returns the correct class interface depending how the AWS CF is configured
+        NOTE: only supports Simple Cloudformation currently
+        """
+        try:
+            return DcosCfSimple(info['stack_name'], self.boto_wrapper)
+        except Exception as ex:
+            raise LauncherError('StackNotFound', '{} is not accessible'.format(info['stack_name'])) from ex
+
+
+def get_launcher(launcher_type, provider_info):
+    """Returns a launcher given a launcher type (string) and a dictionary of
+    appropriate provier_info
+    """
+    if launcher_type == 'cloudformation':
+        check_keys(provider_info, ['region', 'access_key_id', 'secret_access_key'])
+        boto_wrapper = BotoWrapper(
+            provider_info['region'], provider_info['access_key_id'], provider_info['secret_access_key'])
+        return AwsCloudformationLauncher(boto_wrapper)
+    else:
+        raise LauncherError('UnsupportedAction', 'Launcher type not supported: {}'.format(launcher_type))
+
+
+def convert_host_list(host_list):
+    """ Makes Host tuples more readable when using describe
+    """
+    return [{'private_ip': h.private_ip, 'public_ip': h.public_ip} for h in host_list]
+
+
+def check_keys(user_dict, key_list):
+    missing = [k for k in key_list if k not in user_dict]
+    if len(missing) > 0:
+        raise LauncherError('MissingInput', 'The following keys were required but '
+                            'not provided: {}'.format(repr(missing)))
+
+
+def do_main(args):
+    if args['create']:
+        info_path = args['--info-path']
+        if os.path.exists(info_path):
+            raise LauncherError('InputConflict', 'Target info path already exists!')
+        config = load_yaml(args['--config-path'])
+        check_keys(config, ['type', 'provider_info', 'this_is_a_temporary_config_format_do_not_put_in_production'])
+        write_json(info_path, get_launcher(config['type'], config['provider_info']).create(config))
+        return 0
+
+    info = load_json(args['--info-path'])
+    check_keys(info, ['type', 'provider'])
+    launcher = get_launcher(info['type'], info['provider'])
+
+    if args['wait']:
+        launcher.wait(info)
+        print('Cluster is ready!')
+        return 0
+
+    if args['describe']:
+        print(json_prettyprint(launcher.describe(info)))
+        return 0
+
+    if args['pytest']:
+        test_cmd = 'py.test'
+        if args['--env'] is not None:
+            if '=' in args['--env']:
+                # User is attempting to do an assigment with the option
+                raise LauncherError('OptionError', "The '--env' option can only pass through environment variables "
+                                    "from the current environment. Set variables according to the shell being used.")
+            var_list = args['--env'].split(',')
+            check_keys(os.environ, var_list)
+            test_cmd = ' '.join(['{}={}'.format(e, os.environ[e]) for e in var_list]) + ' ' + test_cmd
+        if len(args['<pytest_extras>']) > 0:
+            test_cmd += ' ' + ' '.join(args['<pytest_extras>'])
+        launcher.test(info, test_cmd)
+        return 0
+
+    if args['delete']:
+        launcher.delete(info)
+        return 0
+
+
+def main(argv=None):
+    args = docopt(__doc__, argv=argv, version='DC/OS Launch v.0.1')
+
+    try:
+        return do_main(args)
+    except (LauncherError, YamlParseError) as ex:
+        print('DC/OS Launch encountered an error!')
+        print(repr(ex))
+        return 1
 
 
 if __name__ == '__main__':
-    main()
+    sys.exit(main())
--- a/test_util/test_launch.py
+++ b/test_util/test_launch.py
@@ -1,0 +1,316 @@
+import json
+from contextlib import contextmanager
+
+import pytest
+import yaml
+from docopt import DocoptExit
+
+import ssh.tunnel
+import test_util
+from pkgpanda.util import load_json
+from test_util.helpers import Host
+from test_util.launch import get_launcher, LauncherError, main
+
+MOCK_SSH_KEY_DATA = 'ssh_key_data'
+MOCK_KEY_NAME = 'my_key_name'
+
+
+def check_cli(cmd):
+    assert main(cmd) == 0, 'Command failed! {}'.format(' '.join(cmd))
+
+
+def check_success(capsys, tmpdir, config_str):
+    """
+    Runs through the required functions of a launcher and then
+    runs through the default usage of the script for a
+    given config path and info path, ensuring each step passes
+    if all steps finished successfully, this parses and returns the generated
+    info JSON and stdout description JSON for more specific checks
+    """
+    # Test launcher directly first
+    config = yaml.safe_load(config_str)
+    launcher = get_launcher(config['type'], config['provider_info'])
+    info = launcher.create(config)
+    launcher.wait(info)
+    launcher.describe(info)
+    launcher.test(info, 'py.test')
+    launcher.delete(info)
+
+    # add config to disk and make info path for CLI testing
+    config_path = tmpdir.join('my_specific_config.yaml')  # test non-default name
+    config_path.write(config_str)
+    config_path = str(config_path)
+    info_path = str(tmpdir.join('my_specific_info.json'))  # test non-default name
+
+    # Now check launcher via CLI
+    check_cli(['create', '--config-path={}'.format(config_path), '--info-path={}'.format(info_path)])
+    # use the info written to disk to ensure JSON parsable
+    info = load_json(info_path)
+    # General assertions about info
+    assert 'type' in info
+    assert 'provider' in info
+    assert 'ssh' in info
+    assert 'user' in info['ssh']
+    assert 'private_key' in info['ssh']
+
+    check_cli(['wait', '--info-path={}'.format(info_path)])
+
+    # clear stdout capture
+    capsys.readouterr()
+    check_cli(['describe', '--info-path={}'.format(info_path)])
+    # capture stdout from describe and ensure JSON parse-able
+    description = json.loads(capsys.readouterr()[0])
+
+    # general assertions about description
+    assert 'masters' in description
+    assert 'private_agents' in description
+    assert 'public_agents' in description
+
+    check_cli(['pytest', '--info-path={}'.format(info_path)])
+
+    check_cli(['delete', '--info-path={}'.format(info_path)])
+
+    return info, description
+
+
+@contextmanager
+def mocked_context(*args, **kwargs):
+    """ To be directly patched into an ssh.tunnel invocation to prevent
+    any real SSH attempt
+    """
+    yield type('Tunnelled', (object,), {})
+
+
+@pytest.fixture
+def mocked_test_runner(monkeypatch):
+    monkeypatch.setattr(ssh.tunnel, 'tunnel', mocked_context)
+    monkeypatch.setattr(test_util.runner, 'integration_test', lambda *args, **kwargs: 0)
+
+
+@pytest.fixture
+def mocked_ssh_key_path(tmpdir):
+    ssh_key_path = tmpdir.join('ssh_key')
+    ssh_key_path.write(MOCK_SSH_KEY_DATA)
+    return str(ssh_key_path)
+
+
+@pytest.fixture
+def mocked_aws_cf_simple_backend(monkeypatch, mocked_test_runner):
+    """Does not include SSH key mocking
+    """
+    # mock create
+    monkeypatch.setattr(test_util.aws.BotoWrapper, 'create_stack', lambda *args: None)
+    # mock wait
+    monkeypatch.setattr(test_util.aws.CfStack, 'get_stack_details', lambda _: {'StackStatus': 'CREATE_COMPLETE'})
+    # mock describe
+    monkeypatch.setattr(test_util.aws.DcosCfSimple, 'get_master_ips',
+                        lambda _: [Host('127.0.0.1', '12.34.56')])
+    monkeypatch.setattr(test_util.aws.DcosCfSimple, 'get_private_agent_ips',
+                        lambda _: [Host('127.0.0.1', None)])
+    monkeypatch.setattr(test_util.aws.DcosCfSimple, 'get_public_agent_ips',
+                        lambda _: [Host('127.0.0.1', '12.34.56')])
+    # mock delete
+    monkeypatch.setattr(test_util.aws.DcosCfSimple, 'delete', lambda _: None)
+    monkeypatch.setattr(test_util.aws.BotoWrapper, 'delete_key_pair', lambda *args: None)
+    # mock config
+    monkeypatch.setattr(test_util.aws.BotoWrapper, 'create_key_pair', lambda *args: MOCK_SSH_KEY_DATA)
+
+
+@pytest.fixture
+def mocked_aws_cf_simple(mocked_ssh_key_path, mocked_aws_cf_simple_backend):
+    return """
+---
+this_is_a_temporary_config_format_do_not_put_in_production: yes_i_agree
+type: cloudformation
+template_url: http://us-west-2.amazonaws.com/downloads
+stack_name: foobar
+provider_info:
+  region: us-west-2
+  access_key_id: asdf09iasdf3m19238jowsfn
+  secret_access_key: asdf0asafawwa3j8ajn
+ssh_info:
+  user: core
+  private_key_path: {}
+parameters:
+  - ParameterKey: KeyName
+    ParameterValue: default
+  - ParameterKey: AdminLocation
+    ParameterValue: 0.0.0.0/0
+  - ParameterKey: PublicSlaveInstanceCount
+    ParameterValue: 1
+  - ParameterKey: SlaveInstanceCount
+    ParameterValue: 5
+""".format(mocked_ssh_key_path)
+
+
+def test_aws_cf_simple(capsys, tmpdir, mocked_aws_cf_simple):
+    """Test that required parameters are consumed and appropriate output is generated
+    """
+    info, desc = check_success(capsys, tmpdir, mocked_aws_cf_simple)
+    # check AWS specific info
+    assert info['type'] == 'cloudformation'
+    assert 'stack_name' in info
+    assert 'region' in info['provider']
+    assert 'access_key_id' in info['provider']
+    assert 'secret_access_key' in info['provider']
+    assert info['ssh']['key_name'] == 'default'
+    assert info['ssh']['delete_with_stack'] is False
+    assert info['ssh']['private_key'] == MOCK_SSH_KEY_DATA
+    assert info['ssh']['user'] == 'core'
+    # check that description is updated with info
+    assert 'stack_name' in desc
+
+
+@pytest.mark.usefixtures('mocked_aws_cf_simple')
+def test_aws_cf_simple_make_key(capsys, tmpdir):
+    """ Same mocked backend as other aws_cf_simple tests, but marginally different config
+    Test that required parameters are consumed and appropriate output is generated
+    """
+    config = """
+---
+this_is_a_temporary_config_format_do_not_put_in_production: yes_i_agree
+type: cloudformation
+template_url: http://us-west-2.amazonaws.com/downloads
+stack_name: foobar
+provider_info:
+  region: us-west-2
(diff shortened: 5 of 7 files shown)


mellenburg:
@cmaloney This is ready for a first pass review.

mellenburg:
Note: the PR will be able to support Azure deployments with the addition of https://github.com/dcos/dcos/pull/914

mellenburg:
@orsenthil PTAL

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
+
+def do_create(config_path, info_path):
+    """Reads a YAML configuration file from config_path, creates a cluster accordingly, and
+    then dumps the necessary attributes in a JSON file at info_path
+    """
+    assert not os.path.exists(info_path), 'There is already a cluster info at {}'.format(info_path)
+    config = yaml.load(load_string(config_path))
+    assert 'this_is_a_temporary_config_format_do_not_put_in_production' in config
+    write_json(info_path, get_launcher(config).create())
+    return 0
```
orsenthil: why `return 0` ? If the for `sys.exit` code, I think, it is better to do sys.exit(0) explicitly than have a return value here.

gen/installer/bash.py
```diff
@@ -638,12 +638,15 @@ def fill_template(base_name, format_args):
 def make_dcos_launch():
     # NOTE: this needs to be kept in sync with build_dcos_launch.sh
     work_dir = py.path.local.mkdtemp()
+    if os.path.exists('ext/upstream'):
+        git_src = 'file://{}/ext/upstream'.format(os.getcwd())
+    else:
+        git_src = 'file://{}'.format(os.getcwd())
```
orsenthil: Pattern/idiom comment. How about
```
git_src = 'file://{}'.format(os.getcwd())
if os.path.exists('ext/upstream'):
     git_src = os.path.join(git_src, 'ext/upstream')
```

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
+import sys
+import yaml
 from docopt import docopt
+from pkgpanda.util import load_json, load_string, write_json
+
+from test_util.aws import BotoWrapper, DcosCfSimple
+
+
+class AbstractLauncher(metaclass=abc.ABCMeta):
+    def __init__(self):
```
cmaloney: I don't think there is value in forcing Launchers to have a custom `__init__` (as this does). If we had something like a `dcos-docker` or `dcos-vagrant` launcher, those would have less or very different things (if anything in the init)

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
+        raise NotImplementedError()
+
+    def describe(self):
+        raise NotImplementedError()
+
+    def delete(self):
+        raise NotImplementedError()
+
+
+class AwsLauncher(AbstractLauncher):
```
cmaloney: `AwsCloudformationLauncher`?

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
+        elif status == 'CREATE_COMPLETE':
+            pass
+        else:
+            raise Exception('')
+        print('Cluster is ready!')
+
+    def describe(self):
+        cf = self.get_stack()
+        return {
+            'masters': convert_host_list(cf.get_master_ips()),
```
cmaloney: How does this work if it's a blind cloudformation template? (I see this assumes AWS CF Simple, but would be better not to for the generic AWS CloudFormation Launcher).

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
+        """Returns the correct class interface depending how the AWS CF is configured
+        NOTE: only supports Simple Cloudformation currently
+        """
+        return DcosCfSimple(self.config['stack_name'], self.boto_wrapper)
+
+
+def get_launcher(config):
+    """Returns a launcher given a python dictionary
+    """
+    if 'cloudformation' in config:
```
cmaloney: Can this be a `configuration['type'] == "cloudformation"` or something we switch on rather than magical keys to search for?

test_util/launch.py
```diff
@@ -1,19 +1,145 @@
...
 
 def main():
-    docopt(__doc__)
+    args = docopt(__doc__, version='DC/OS Launch 1.0')
+
+    if args['create']:
+        sys.exit(do_create(args['--config-path'], args['--info-path']))
+
+    cluster_info = load_json(args['--info-path'])
+
```
cmaloney: I would argue the config should be loaded here rather than in the individual do_* functions.

mellenburg: I don't follow. I load the only config the following functions need with:
```
cluster_info = load_json(args['--info-path'])
```
unless you were just referring to do_create loading its own config, in which case, sure, Ill make that change, but it must only be within the args['create'] pathway so that the other calls can run with just cluster info

test_util/launch.py
mellenburg: This was taken directly from https://mesosphere.atlassian.net/browse/DCOS-10194

I'm fine with type: cloudformation

test_util/launch.py
mellenburg: Given that we said this should only support CFSimple I wanted to leave that decision until we add Advanced CF. This class will never do plain Vpc (rather, we will have a launcher for plain Vpc and a launcher for generic onprem which will work together for the onprem-in-vpc case).

How would you propose differentiating between simple and advanced with nothing more than stack name? Checking the parameters and looking at the provided template URL and then judge based on the name (won't allow changing the template name for custom deployments)? Checking if the stack resources include the advanced stacks?

test_util/launch.py
```diff
@@ -1,19 +1,136 @@
...
+import sys
+import yaml
 from docopt import docopt
+from pkgpanda.util import load_json, load_string, write_json
+
+from test_util.aws import BotoWrapper, DcosCfSimple
+
+
+class AbstractLauncher(metaclass=abc.ABCMeta):
+    def create(self):
```
cmaloney: these should take all relavant parameters.

test_util/launch.py
```diff
@@ -1,19 +1,136 @@
...
+
+    def describe(self):
+        raise NotImplementedError()
+
+    def delete(self):
+        raise NotImplementedError()
+
+
+class AwsCloudformationLauncher(AbstractLauncher):
+    def __init__(self, config):
```
cmaloney: I think we need to workout whether the config is required for all options / operations or not. This code as it is currently is suggesting that yes, the config.yaml (because it contains the AWS Credentials) is necessary for any / all usage of AwsCloudformation stack, so the CLI needs to always provide / specify it.

I'm fine with that as a design, but we probably need to adapt the CLI docopt string to it.

Another approach we could go with is "create" is the only thing that needs the config, then _everything_ (including credentials) which is critical to continued operation is stored in the cluster_info, and only the cluster_info is needed at that point to create a Launcher.

test_util/launch.py
```diff
@@ -1,19 +1,136 @@
...
+        self.get_stack().delete()
+
+    def get_stack(self):
+        """Returns the correct class interface depending how the AWS CF is configured
+        NOTE: only supports Simple Cloudformation currently
+        """
+        return DcosCfSimple(self.config['stack_name'], self.boto_wrapper)
+
+
+def get_launcher(config):
```
cmaloney: this says it takes a config, but sometimes we pass it a cluster_info object which isn't the config.

gen/installer/bash.py
```diff
@@ -638,12 +638,15 @@ def fill_template(base_name, format_args):
 def make_dcos_launch():
     # NOTE: this needs to be kept in sync with build_dcos_launch.sh
     work_dir = py.path.local.mkdtemp()
+    git_src = 'file://{}'.format(os.getcwd())
+    if os.path.exists('ext/upstream'):
+        git_src = os.path.join(git_src, 'ext/upstream')
+    # clone so the in-place build doesn't affect this tree
+    subprocess.check_call(['git', 'clone', '--progress', git_src, str(work_dir)])
```
mellenburg: I do not understand the value of using fetch_git. That function either checks out a mirror clone if the destination doesnt exist or it updates an already existing git repo to have an updated origin.

What I need here is a copy of the working tree so that the spec file can grab test_util/launch.py and any other data (uses local path, not package_data association). The reason for this is when we want to do more advanced things (like running an installer against a VPC) we will need the utilities that pull in these files.

So, we can leave this out entirely for now and go back to just copying the test_util/launch.py file to our working dir, but if/when we do need to have a local copy of the tree, why would I use a function that requires me to make extra function calls? 
1, need to make a new path string that is not a dir yet
2. fetch_git
3. git checkout whatever current HEAD is inherited from cloning the local tree
4. go to objects dir to build
VS
1. subprocess git clone (not a mirror) or recursive copy local tree
2. build in resultant dir

cmaloney: No local changes can be cloned in git, so we should actually be validating there aren't any. That's what the GitLocal source fetcher does: https://github.com/dcos/dcos/blob/master/pkgpanda/build/src_fetchers.py#L129

Otherwise _magically_ in the user's build any local / uncommitted changes they have won't show up, which is painful.

In terms of code, writing two functions which do the same thing is duplication. Each one can have bugs. Writing a couple lines that does the "simple clone" is duplicating the clone code. Git's api, and the `--mirror` actually is critical for correctness in some situations, there is a reason it was added to `fetch_git` (And can actually dig through the commit history if you want to find out why).

test_util/aws.py
```diff
@@ -55,7 +55,7 @@ def create_stack(self, name, template_url, user_parameters, deploy_timeout=60):
         log.info('Requesting AWS CloudFormation...')
         cf_parameters = []
         for k, v in user_parameters.items():
-            cf_parameters.append({'ParameterKey': k, 'ParameterValue': v})
+            cf_parameters.append({'ParameterKey': k, 'ParameterValue': str(v)})
```
mellenburg: I tried rewriting this with a function that pulls the template in advance, looks in the key section, checks that they keys the user provided are there, and then looks at the parameter type in the template and casts the parameter as that type if necessary. However, AWS template parameter types do NOT apply to python. Every parameters MUST be a string. This is especially confusing given that Azure does the opposite, enforcing parameter types in python to be consistent with the parameter type in the template

For example. here we see a boto validation error because I passed two values as floats being as the parameter type in the template is "Number" and the dcos say "Numbers can be integers or floats". 
https://teamcity.mesosphere.io/viewLog.html?buildId=480558&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsDe_2&tab=buildLog#_focus=164

So, no matter what parameters you are passing, they need to be string here, thus, this is the best place to double check that.

mellenburg:
@cmaloney I fixed the build steps to use GitLocalSrcFetcher and added the same "no uncommitted changes" check to the build_dcos_launch.sh script.

The only other remaining issue was my casting of every param as string, however on more investigation it turns out this is the ONLY valid parameter type while using boto to create a stack, so that is actually what we want to do.

Is this about ready to ship?

gen/installer/bash.py
```diff
@@ -638,12 +639,18 @@ def fill_template(base_name, format_args):
 def make_dcos_launch():
     # NOTE: this needs to be kept in sync with build_dcos_launch.sh
     work_dir = py.path.local.mkdtemp()
+    # the tree will not be in the working dir if run from downstream
+    tree_src = './'
+    if os.path.exists('ext/upstream'):
+        tree_src = os.path.join(tree_src, 'ext/upstream')
```
cmaloney: why not just `tree_src = './ext/upstream'`. Seems simpler to read / understand

test_util/launch.py
```diff
@@ -1,19 +1,139 @@
...
+    return [{'private_ip': h.private_ip, 'public_ip': h.public_ip} for h in host_list]
 
 
 def main():
-    docopt(__doc__)
+    args = docopt(__doc__, version='DC/OS Launch 1.0')
+
+    if args['create']:
+        info_path = args['--info-path']
+        assert not os.path.exists(info_path), 'There is already a cluster info at {}'.format(info_path)
```
cmaloney: shouldn't be an assertion / this is part of main program flow. Assertions are for programmer integrity checking (not meant for user error messages). This should probably be a `if not os.path.exists()` / `print(error)` / `exit(-`)`. Could maybe do an exception + global try/catch/print for the "LauncherException".

test_util/launch.py
```diff
@@ -1,19 +1,139 @@
...
+        config = yaml.load(load_string(args['--config-path']))
+        assert 'this_is_a_temporary_config_format_do_not_put_in_production' in config,\
+            'Configs used by dcos-launch must have the non-production flag'
+        write_json(info_path, get_launcher(config['type'], config['provider_info']).create())
+        sys.exit(0)
+
+    info = load_json(args['--info-path'])
+
+    if args['wait']:
+        get_launcher(info['type'], info['provider_info']).wait()
```
cmaloney: why not get_launcher generally right after the info line? All these functions call it with the same function.

test_util/launch.py
```diff
@@ -1,19 +1,139 @@
...
+        print('Cluster is ready!')
+        sys.exit(0)
+
+    if args['describe']:
+        print(json.dumps(get_launcher(info['type'], info['provider_info']).describe(), indent=4))
```
cmaloney: https://github.com/dcos/dcos/blob/237b9afc52e0dd258dce4303292d205f60d07ca0/pkgpanda/util.py#L147

test_util/aws.py
cmaloney: Can we take:

```yaml
parameters:
 - ParameterKey: foo
   ParameterValue: bar
 - ParameterKey: baz
   ParameterValue: bang
````

instead. I'm fine with doing the string conversion, but really want this to be direct pass-through to AWS / none of us manipulating. The str() effectively just ends up being "yaml types aren't what we want".

File: test_util/launch.py (hunk 15 above)
Comment (cmaloney) [inline test_util/launch.py]: Technically should also wrap pulling the fields out of the json so that we get good error messages if they aren't set / the info_json doesn't have them. (Error: `no field 'foo' in 'cluster_info.json` is a much nicer message than a python index exception)

File: test_util/test_launch.py [hunk 17]
```diff
@@ -0,0 +1,37 @@
...
+  PublicSlaveInstanceCount: 1
+  SlaveInstanceCount: 5
+"""
+    config_path = tmpdir.join('config.yaml')
+    config_path.write(aws_simple_config)
+    info_path = tmpdir.join('cluster_info.json')
+    config = yaml.load(load_string(str(config_path)))
+    write_json(str(info_path), get_launcher(config['type'], config['provider_info']).create())
+    # Read dumped cluster info and ensure required keys are present
+    cluster_info = load_json(str(info_path))
```
cmaloney: Can we also do a test of describe?

cmaloney:
I want to expand out the baseline testing + testing harness quite a bit here, should be able to hash it out in person tomorrow pretty quickly.

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
-"""
+"""DC/OS Launch
 
 Usage:
-dcos-launch create [--wait] [--dump-info=cluster_info.json] <config=config.yaml>
-dcos-launch wait <path=cluster_info.json>
-dcos-launch describe <path=cluster_info.json>
-dcos-launch delete <path=cluster_info.json>
+  dcos-launch create [--config-path=<path>] [--info-path=<path>]
```
orsenthil: Instead of `--config-path` would the term `--cluster-config-yaml=<path> / --aws-template-yaml=<path>` be more suitable?

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+  dcos-launch wait [--info-path=<path>]
+  dcos-launch describe [--info-path=<path>]
+  dcos-launch delete [--info-path=<path>]
+
+Commands:
+  create    reads the file given by --config-path, creates the cluster described therein
+            and finally dumps a JSON file to the path given in --info-path which can then
+            be used with the wait, describe, and delete calls
+  wait      Block until the cluster is up and running
```
orsenthil: In all these descriptions, you could add a `.` at the end of lines.

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+        self.error = error
+        self.msg = msg
+
+    def __repr__(self):
+        return '{}: {}'.format(self.error, self.msg)
+
+
+class AbstractLauncher(metaclass=abc.ABCMeta):
+    def create(self, config):
+        raise NotImplementedError()
```
orsenthil: No action required.
We need not do a `()` for raising exceptions without arguments. Just ` raise NotImplementedError` is fine. The rest of the code is using the same style with empty `()`, so it is alright.

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+        self.get_stack(info).delete()
+
+    def get_stack(self, info):
+        """Returns the correct class interface depending how the AWS CF is configured
+        NOTE: only supports Simple Cloudformation currently
+        """
+        return DcosCfSimple(info['stack_name'], self.boto_wrapper)
+
+
+def get_launcher(launcher_type, provider_info):
```
orsenthil: `laucher_type: str, provider_info: dict`   and you could remove those descriptions in the docstrings.

gen/build_deploy/bash.py
```diff
@@ -638,12 +640,18 @@ def fill_template(base_name, format_args):
 def make_dcos_launch():
     # NOTE: this needs to be kept in sync with build_dcos_launch.sh
     work_dir = py.path.local.mkdtemp()
+    # the tree will not be in the working dir if run from downstream
```
orsenthil: what does `downstream` mean here? I assume, `dcos/dcos` repo. It could mentioned explicitly.

gen/build_deploy/bash/dcos-launch.spec
```diff
@@ -1,4 +1,12 @@
-a = Analysis(['test_util/launch.py'])
+# Warning:
+#   - hidden import must be used as there is a bug in pyinstaller
+#     https://github.com/pyinstaller/pyinstaller/issues/2185
+#   - data must be decalared explicitly if not a .py file
+#   - Building will suck up the local SSL .so and package it
+#     with the final exe. Ensure build system has OpenSSL 1.0.2g or greater
```
orsenthil: TODO: Add a pre-req checker that verifies OpenSSL 1.0.2g+
It is highly unlikely that someone will read this, the first time they are excitedly trying dcos-launch.

test_util/aws.py
```diff
@@ -24,6 +24,13 @@ def template_by_instance_type(instance_type):
         return VPC_TEMPLATE_URL
 
 
+def param_dict_to_aws_format(user_parameters):
+    final_parameters = []
+    for k, v in user_parameters.items():
+        final_parameters.append({'ParameterKey': k, 'ParameterValue': str(v)})
+    return final_parameters
+
```
orsenthil: in single line using comprehensions? 

`return [{"ParameterKey": k, "ParameterValue: str(v)} for k, v in user_parameters.items()]`

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+        status = cf.get_stack_details()['StackStatus']
+        if status == 'CREATE_IN_PROGRESS':
+            cf.wait_for_stack_creation(wait_for_poll_min=0)
+        elif status == 'CREATE_COMPLETE':
+            return
+        else:
+            raise LauncherError('WaitError', 'AWS Stack has entered unexpected state: {}'.format(status))
+
+    def describe(self, info):
+        # include the basic info
```
orsenthil: you could kill this comment.

test_util/test_launch.py
```diff
@@ -0,0 +1,156 @@
...
+from test_util.launch import get_launcher, main
+
+
+def check_cli(cmd):
+    assert main(cmd.split(' ')) == 0, 'CLI args failed! {}'.format(cmd)
+
+
+def check_success(capsys, tmpdir, config_str):
+    """
+    Runs through the required functions of a laucnher and then
```
orsenthil: spell fix: `launcher`

gen/build_deploy/bash.py
cmaloney: upstream is dcos/docs, downstream has a treeinfo.json that points to an upstream.

gen/build_deploy/bash/dcos-launch.spec
cmaloney: This is only run by our CI / not run by people trying to build this themselves generally. In the case someone is doing local development and iterating on dcos-launch, then this pulling in whatever the local openssl is should be fine.

test_util/launch.py
cmaloney: The dcos/dcos codebase currently has 14 instances of throwing NotImplementedError, and all of those use () for actually constructing an instance of the class, rather than throwing the exception class itself.

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+            'type': 'cloudformation',
+            'stack_name': config['stack_name'],
+            'provider_info': {
+                'region': config['provider_info']['region'],
+                'access_key_id': config['provider_info']['access_key_id'],
+                'secret_access_key': config['provider_info']['secret_access_key']}}
+
+    def wait(self, info):
+        # TODO: should this support the case where the cluster is being updated?
+        cf = self.get_stack(info)
```
cmaloney: can we make sure this has a clean error if the stack doesn't exist?

test_util/launch.py
orsenthil: Yup. It's more a preference thing (ref: http://stackoverflow.com/a/16709222/18852) and consistency within the code-base determines the preference.

test_util/test_launch.py
```diff
@@ -0,0 +1,156 @@
...
+from docopt import DocoptExit
+
+import test_util
+from pkgpanda.util import load_json
+from test_util.helpers import Host
+from test_util.launch import get_launcher, main
+
+
+def check_cli(cmd):
+    assert main(cmd.split(' ')) == 0, 'CLI args failed! {}'.format(cmd)
```
cmaloney: any particular reason we can't pass through the argument array?

test_util/test_launch.py
```diff
@@ -0,0 +1,156 @@
...
+def check_success(capsys, tmpdir, config_str):
+    """
+    Runs through the required functions of a laucnher and then
+    runs through the default usage of the script for a
+    given config path and info path, ensuring each step passes
+    if all steps finished successfully, this parses and returns the generated
+    info JSON and stdout description JSON for more specific checks
+    """
+    # Test launcher directly first
+    config = yaml.load(config_str)
```
cmaloney: there's a `load_yaml` should use now, produces better error messages

test_util/launch.py
```diff
@@ -1,20 +1,169 @@
...
+
+def main(argv):
+    args = docopt(__doc__, argv=argv, version='DC/OS Launch 1.0')
+
+    try:
+        if args['create']:
+            info_path = args['--info-path']
+            if os.path.exists(info_path):
+                raise LauncherError('InputConflict', 'Target info path already exists!')
+            config = yaml.load(load_string(args['--config-path']))
```
cmaloney: there's a `load_yaml` should use now, produces better error messages

test_util/test_launch.py
```diff
@@ -0,0 +1,156 @@
...
+    assert 'provider_info' in info
+
+    check_cli('wait --info-path={}'.format(info_path))
+
+    # clear stdout capture
+    capsys.readouterr()
+    check_cli('describe --info-path={}'.format(info_path))
+    # capture stdout from describe and ensure JSON parse-able
+    description = json.loads(capsys.readouterr()[0])
+
```
cmaloney: Should add a test for
 - What happens with an already-deleted cluster?
 - What happens with a cluster that doesn't actually exist?

tox.ini
```diff
@@ -55,6 +55,7 @@ passenv =
 deps =
   dnspython
   pytest
+  PyYAML
```
cmaloney: shouldn't this be picked up from the setup.py in this repository?

mellenburg: setup.py has a very serious flaw where we cannot declare dependencies whose pypi name is significantly different from its module import name.

See: https://mesosphere.atlassian.net/browse/DCOS-11783

If you just add PyYaml to setup, then when using any of the entrypoints, you will be greeted with a an exception saying "could not import module 'pyyaml'"

test_util/launch.py
mellenburg: - cluster-config-yaml is not suitable as this may not necessarily launch "clusters" strictly speaking
- aws-template-yaml is definitely not suitable as this will not be limited to aws and the config.yaml is not a special aws-template config.

I could go for --config-yaml-path instead but with json being a subset of yaml, you could imagine this taking a JSON file.

Perhaps --launch-config-path could work, but it just seems redundant as its there is not another config that this accepts.

I feel this should remain config-path. I loathe unnecessarily long option names and the options description below can be used to add more explanation if thats truly necessary

mellenburg: Adding 'output' does not help. The info file that is produced by create is consumed by wait, delete, and describe. So, the user would use one special arg to point to where they want to the info json to appear and then a totally different arg to point to where they want that same thing to be found.

Furthermore, I believe this is well covered by the documentation that is printed out when you do --help:
```
  --info-path=<path>    JSON file output by create and consumed by wait, describe,
                        and delete [default: cluster_info.json].
```

mellenburg:
I am folding the launch work in here directly as I see no good argument for keeping that interface separate. As such, this is now blocked by https://github.com/dcos/dcos/pull/989

mellenburg:
This PR has now been rebased to rely upon https://github.com/dcos/dcos/pull/1088

mellenburg:
@spahl @lingmann This is ready for review. Can you please update from the 'holding' tag?

lingmann:
@mellenburg done, pinging some people to review #1088

test_util/launch.py
```diff
@@ -1,20 +1,257 @@
...
+
+def check_keys(user_dict, key_list):
+    missing = [k for k in key_list if k not in user_dict]
+    if len(missing) > 0:
+        raise LauncherError('MissingInput', 'The following keys were required but'
+                            'not provided: {}'.format(repr(missing)))
+
+
+def main(argv):
+    args = docopt(__doc__, argv=argv, version='DC/OS Launch 1.0')
```
orsenthil: TODO: It might be a good idea to consolidate on the command line parses we use in dcos. We use a variety.

```
$ ag -l 'docopt|argparse|click'
bootstrap/extra/dcos_internal_utils/cli.py
dcos-image-deps/buildinfo.json
dcos-oauth/extra/dcos_add_user.py
flask/build
flask/buildinfo.json
python-docopt/buildinfo.json
```

Might be a good idea to consolidate on [onething](https://realpython.com/blog/python/comparing-python-command-line-parsing-libraries-argparse-docopt-click/#conclusion).

Also, curious to know about your decision to choose/introduce docopt here.

test_util/test_aws_cf_launch_test.py
```diff
@@ -0,0 +1,50 @@
+"""WARNING: DO NOT MERGE THIS FILE
```
orsenthil: This file LGTM.

You could remove this in your next update to this PR.

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+from pkgpanda.util import load_json
+from test_util.helpers import Host
+from test_util.launch import get_launcher, LauncherError, main
+
+MOCK_SSH_KEY_DATA = 'ssh_key_data'
+MOCK_KEY_NAME = 'my_key_name'
+
+
+def check_cli(cmd):
+    assert main(cmd.split(' ')) == 0, 'CLI args failed! {}'.format(cmd)
```
orsenthil: It's not really `CLI args failed here`. The entire process is run and compared with the `exit 0` output.

Perhaps you can say, `"Command Failed: {}".format(cmd)`

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+                        lambda _: [Host('127.0.0.1', '12.34.56')])
+    # mock delete
+    monkeypatch.setattr(test_util.aws.DcosCfSimple, 'delete', lambda _: None)
+    monkeypatch.setattr(test_util.aws.BotoWrapper, 'delete_key_pair', lambda *args: None)
+    # mock config
+    monkeypatch.setattr(test_util.aws.BotoWrapper, 'create_key_pair', lambda *args: MOCK_SSH_KEY_DATA)
+
+
+@pytest.fixture
+def mock_aws_cf_simple(mock_aws_cf_simple_backend, mock_ssh_key_path):
```
orsenthil: `mock_aws_cf_simple_backend` is not used in this method, do you think using `@pytest.mark.usefixtures` will be preferable? http://doc.pytest.org/en/latest/fixture.html#using-fixtures-from-classes-modules-or-projects

(Basically, that only states states the dependency as a decorator, when we really dont need the object that is sent as argument here)

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+
+@pytest.fixture
+def mock_ssh_key_path(tmpdir):
+    ssh_key_path = tmpdir.join('ssh_key')
+    ssh_key_path.write(MOCK_SSH_KEY_DATA)
+    return str(ssh_key_path)
+
+
+@pytest.fixture
+def mock_aws_cf_simple_backend(monkeypatch, mock_test_runner):
```
orsenthil: `@pytest.mark.usefixtures("mock_test_runner")`  instead of `mock_test_runner` as a arg?

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+    assert 'secret_access_key' in info['provider']
+    assert info['ssh']['key_name'] == 'default'
+    assert info['ssh']['delete_with_stack'] is False
+    assert info['ssh']['private_key'] == MOCK_SSH_KEY_DATA
+    assert info['ssh']['user'] == 'core'
+    # check that description is updated with info
+    assert 'stack_name' in desc
+
+
+def test_aws_cf_simple_make_key(capsys, tmpdir, mock_aws_cf_simple):
```
orsenthil: `mock_aws_cf_simple` unused, `@pytest.mark.usefixtures` instead?

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+        assert exinfo.value.error == 'StackNotFound'
+
+    info = aws_launcher.create(config)
+    check_stack_error('wait', (info,))
+    check_stack_error('describe', (info,))
+    check_stack_error('delete', (info,))
+    check_stack_error('test', (info, 'py.test'))
+
+
+def test_aws_ssh_key_handling(mock_aws_cf_simple, mock_ssh_key_path):
```
orsenthil: `mock_aws_cf_simple` unused, `@pytest.mark.usefixtures` instead?

test_util/test_launch.py
```diff
@@ -0,0 +1,302 @@
...
+def test_noop():
+    """Ensure docopt exit (displays usage)
+    """
+    with pytest.raises(DocoptExit):
+        main([])
+    with pytest.raises(DocoptExit):
+        main(['foobar'])
+
+
+def test_conflicting_path(tmpdir, mock_aws_cf_simple):
```
orsenthil: Observation: This is very costly test for testing a 'conflicting path` constraint.  Ideally, no cluster should be created for testing this constraint, but in this style, we are creating one cluster for the first `(main['create'])`

(Is it a good idea to refactor code to test these 'constrains of dcos-launch` without actually creating a cluster. If you think is good idea, perhaps this can be revisited with a TODO.

test_util/test_launch.py
orsenthil: [minor]: I think, `mock_` is a overloaded term here. Although correct, we generally associate mocks with an object that can be mutated and used for assertions. Here, this function is more of a `aws_cf_simple_test_config`

test_util/test_aws_cf_launch_test.py
mellenburg: No, this file is not to be included as intended with all my warnings. The only reason this is in python and not pure bash is that I wanted to hack into the current CI commands to get a status check on this PR. We do not want these sloppy runner scripts. That's the whole point of dcos launch. We do not want to be creating a new interface everytime we want to make a marginally different new test. One of the biggest motivations of making dcos-launch is that we can just plug this into our CI:
```
#!/bin/bash
wget $DCOS_LAUNCH_URL
set -o errexit
cat <<EOF > config.yaml
---
this_is_a_temporary_config_format_do_not_put_in_production: yes_i_agree
type: cloudformation
template_url: $template_url
stack_name: $stack_name
provider_info:
  region: us-west-2
  access_key_id: $access_key_id
  secret_access_key: $secret_access_key
ssh_info:
  user: core
parameters:
  - ParameterKey: AdminLocation
    ParameterValue: 0.0.0.0/0
  - ParameterKey: PublicSlaveInstanceCount
    ParameterValue: 1
  - ParameterKey: SlaveInstanceCount
    ParameterValue: 2
EOF
./dcos-launch create --config-path=config.yaml
./dcos-launch wait
./dcos-launch describe
./dcos-launch test -- py.test -s -rs -vv --teamcity
./dcos-launch delete
```
In this code block, you are _only_ dependent upon being able to download one executable and every necessary setting to control this test is listed in one place. You can replace any part of this with parameters at anytime without committing code changes.

If we keep this file that I say must go, the testing looks like this:
```
# get artifacts from previous test, which is TC specific
# create and activate a new virtual environment
# install everything in that virtual environment
# read in a bunch of envvars that are set elsewhere in CI
test-aws-cf-launch
```
Now we need completely rebuild our software just to test a different configuration (e.g. 4 slaves)

test_util/launch.py
cmaloney: Please not another standard. argparse and docopt are what is used currently throughout the codebase. Which is used depends on what the shape of the API being made is.

mellenburg: Hm. I don't think its a bad idea to try to use one method of python arg handling, but I also feel its a arbitrary goal with little return-on-investment once we have a bunch of functional command line parsers already implemented. I.E. I think this belongs in the category of fixing something that is not broken.

As for the reason why I chose docopt: is that its documentation is perfectly in line with its function and boilerplate is absolutely minimal. The command line handling is not what I am concerned with in this PR; its the functionality that concerns me. If we decide we want this to have a more robust CLI, then we could look into that after we have created the launchers for all the providers

Furthermore, I don't really think there is such thing as a single python CLI that should be used. 

Docopt is the easiest to parse visually in the least amount of code. Ignore that example in the link you posted; its terrible. If you want to do fancy validation within the CLI, then just don't use docopt. Docopt is pointless if your final code looks that ugly IMO.

Argeparse is great if you just need to parse something extremely simple and want a script that can run on any machine with vanilla python and no modules.

Click is probably the 'best' option if you want fine-grained control. However, this will require a bunch of boiler plate and adding a new dependency to DC/OS (which is just one more unnecessary thing for now)

test_util/test_launch.py
mellenburg: Sure, I guess it helps to know that the fixture is a non-interactive dependency

test_util/test_launch.py
mellenburg: I am confused with how this test is costly. None of these tests actually create a cluster; its all mocked out in the backend so that the aws code always does nothing

test_util/launch.py
```diff
@@ -1,20 +1,257 @@
...
+            'key_name': key_name,
+            'user': config.get('ssh_info', dict()).get('user', None)}
+
+    def test(self, info, test_cmd):
+        ssh_info = info['ssh']
+        for param in ['user', 'private_key']:
+            if param not in ssh_info:
+                raise LauncherError('MissingInfo', 'Cluster was created without providing ssh_info: {}'.format(param))
+        details = self.describe(info)
+        # FIX ME clean this up when done
```
mellenburg: Possibly. What I intended here was "TODO: generalize this" so I should at least make that change.

test_util/launch.py
orsenthil: Acknowledge. I hear that it is not 'worth the effort'.

test_util/test_launch.py
mellenburg: I cannot use pytest.mark.usefixtures in this case. Apparently this is only good for test cases because marking fixtures does not make a huge amount of sense in pytest. So don't try to tell anyone else to do this as it simply wont work :P

test_util/test_launch.py
mellenburg: Same as above, pytest.mark.usefixtures cannot be used on a fixture

test_util/test_launch.py
orsenthil: Haha. Interesting, I didn't know that.  I assumed that 'fixture' injection via decorators will be valid mechanism of specifying dependency irrespective of the module being a test case or not.

mellenburg:
@orsenthil I fixed your issues where possible. Is this ready to :ship: ?

orsenthil:
Yes, this looks good to me. 🚢

test_util/test_launch.py
mellenburg: Both of these situations are covered with this test I added: https://github.com/mellenburg/dcos/blob/dcos-launch/test_util/test_launch.py#L249

test_util/test_launch.py
mellenburg: load_yaml takes a filename, not a string:
https://github.com/dcos/dcos/blob/master/pkgpanda/util.py#L135

gen/build_deploy/bash.py
```diff
@@ -654,12 +655,18 @@ def fill_template(base_name, format_args):
...
+    tree_src = './'
+    if os.path.exists('ext/upstream'):
+        tree_src = './ext/upstream'
```
cmaloney: nit: This works for now, kinda ugly but functional. Relies on magic knowledge between how the two repos code works...

gen/build_deploy/bash/dcos-launch.spec
```diff
@@ -1,4 +1,12 @@
...
+a = Analysis(['test_util/launch.py'],
+             hiddenimports=['html.parser'],
+             datas=[('gen/ip-detect/*.sh', 'gen/ip-detect/')])
```
cmaloney: does this need the yaml templates that make up the target config gen uses? What is it internally using the ip-detect for?

test_util/aws.py
```diff
@@ -48,8 +48,8 @@ def resource(self, name):
     def create_key_pair(self, key_name):
         """Returns private key of newly generated pair
         """
-        key = self.resource('ec2').KeyPair(key_name)
-        return key.key_material
+        key = self.client('ec2').create_key_pair(KeyName=key_name)
```
cmaloney: why is the client better than using the resource way of doing this? Doesn't seem like it would use less API calls or anything

test_util/launch.py
```diff
@@ -1,20 +1,282 @@
...
 
 Usage:
-dcos-launch create [--wait] [--dump-info=cluster_info.json] <config=config.yaml>
-dcos-launch wait <path=cluster_info.json>
-dcos-launch describe <path=cluster_info.json>
-dcos-launch delete <path=cluster_info.json>
+  dcos-launch create [--config-path=<path>] [--info-path=<path>]
+  dcos-launch wait [--info-path=<path>]
+  dcos-launch describe [--info-path=<path>]
+  dcos-launch test [--info-path=<path>] [--] [<test_command>]...
```
cmaloney: This test isn't intended for people not developing DC/OS, while dcos-launch in general _is_ intended for people doing that. I feel somewhat that `test` shouldn't be a command at all on dcos-launch (it's ambiguous to people not actively working on DC/OS internals). I feel very strongly that this not be called `test` at the very least. Something like `run-internal-integration-tests` would make a lot more sense and make it more evident this isn't running arbitrary tests, and isn't shakedown either.

test_util/launch.py
```diff
@@ -1,20 +1,282 @@
...
+import sys
 
 from docopt import docopt
 
+import ssh.tunnel
+import test_util.runner
+from pkgpanda.util import json_prettyprint, load_json, load_string, load_yaml, write_json
+from test_util.aws import BotoWrapper, DcosCfSimple
+
+DCOS_LAUNCH_VERSION = '0.1'
```
cmaloney: Why does this have independent versioning / need a versioning streem of its own? Isn't doing any sort of independent release process from the core DC/OS release, and as far as I'm aware there's no plan to do that. I'd drop the versioning string altoghether.

test_util/launch.py
```diff
@@ -1,20 +1,282 @@
...
+    """ Makes Host tuples more readable when using describe
+    """
+    return [{'private_ip': h.private_ip, 'public_ip': h.public_ip} for h in host_list]
+
+
+def check_keys(user_dict, key_list):
+    missing = [k for k in key_list if k not in user_dict]
+    if len(missing) > 0:
+        raise LauncherError('MissingInput', 'The following keys were required but'
+                            'not provided: {}'.format(repr(missing)))
```
cmaloney: Might make sense to also check the key values are strings (or at least simple types, or cast them all to strings for the user?) (https://github.com/dcos/dcos/blob/master/gen/__init__.py#L43)?

An aside / likely out of scope nit is that could switch to using the config class (https://github.com/dcos/dcos/blob/master/dcos_installer/config.py) from the installer which can alredy do that as a helper function.

test_util/launch.py
```diff
@@ -1,20 +1,282 @@
...
+
+
+def main(argv):
+    args = docopt(__doc__, argv=argv, version=DCOS_LAUNCH_VERSION)
+
+    try:
+        if args['create']:
```
cmaloney: nit: Rather than have this all be "indented"  by the try (getting to high levels of indent here) might make sense to move the dispatch to a second helper function this calls.

test_util/launch.py
```diff
@@ -1,20 +1,282 @@
...
+            return 0
 
-def main():
-    docopt(__doc__)
+        if args['delete']:
+            launcher.delete(info)
+            return 0
 
-    raise NotImplementedError()
+    except LauncherError as ex:
```
cmaloney: Should also except YamlParseError (https://github.com/dcos/dcos/blob/master/pkgpanda/util.py#L131) which is raied when load_yaml fails to parse a file. 

May also make sense to teach load_yaml to catch https://docs.python.org/3/library/exceptions.html#OSError and re-throw it as a YamlParseError with details like the filename.

test_util/test_launch.py
cmaloney: nit: may want to use `safe_load` to prevent doing things users can't / make sure we're parsing yaml in the same way user yaml will be parsed then.

test_util/test_launch.py
```diff
@@ -0,0 +1,310 @@
...
+    launcher.delete(info)
+
+    # add config to disk and make info path for CLI testing
+    config_path = tmpdir.join('my_specific_config.yaml')  # test non-default name
+    config_path.write(config_str)
+    config_path = str(config_path)
+    info_path = str(tmpdir.join('my_specific_info.json'))  # test non-default name
+
+    # Now check launcher via CLI
+    check_cli('create --config-path={} --info-path={}'.format(config_path, info_path))
```
cmaloney: please use arg arrays rather than relying on parsing (if for some reason someone had a space in their TMPDIR (happens sometimes on some platforms) then this would break. `check_cli(['create', '--config-path={}'.format(config_path), ['--info-path={}'.format(info_path)]` where check_cli then just passes the arg array to main is about the same difficulty to write and introduces no ambiguity.

The whole codebase to date uses direct argument passing. There are many articles on why you should do it in python around subprocess.

test_util/test_launch.py
```diff
@@ -0,0 +1,310 @@
...
+@pytest.mark.usefixtures('mocked_aws_cf_simple')
+def test_aws_ssh_key_handling(mocked_ssh_key_path):
+    provider_info = {
+        'region': 'us-west-2',
+        'access_key_id': 'foo',
+        'secret_access_key': 'bar'}
+    aws_launcher = get_launcher('cloudformation', provider_info)
+    # Test most minimal config (nothing given and key is generated)
+    ssh_info = aws_launcher.ssh_from_config({'stack_name': 'foo'})
+    assert ssh_info['delete_with_stack'] is True
```
cmaloney: nit: I find it a lot better to compare the whole dictionary exactly. That way you check for
1. any keys that got added you didn't expect (some testing code that added a key then people downstream start depending on it...)
2. you find out all the keys that are off in a nice diff that py.test gives, rather than the first key which is off only. There's also much nicer errors if ssh_info doesn't contain an expected error (Was on the right side but isn't on the left side and heres the value which was expected) rather than what you'll currently get which is a KeyError exception and none of the keys of the dictoinary printed.

cmaloney:
As an aside, an interesting next step validation would be to change (one by one) the  custom test entrypoints (https://github.com/dcos/dcos/blob/master/setup.py#L65 and friends) to have different internal code that just calls dcos-launch then dcos-test. Ideally change each one as their backend gets supported by dcos-launch.

test_util/aws.py
mellenburg: The former actually doesnt work. This was untested code. The key material can only be retrieved from the client method

gen/build_deploy/bash/dcos-launch.spec
mellenburg: ip-detect scripts can be left off for now, but they will be required when we do VPC installs (which this PR originally included and hence this line). So, I would just rather leave this here as onprem will be one of the first expansions for this

gen is not yet invoked directly by dcos-launch so this missing yaml is not a problem currently

mellenburg:
@cmaloney I actually had included with this PR a commit that was a sample of the basic shell script that pulls in launch, cats out the config with environment injection for AWS creds, then runs create, wait, describe, test, and delete, and patched that into the test-aws-cf entrypoint (it worked great). However, adding that script would break the integration tests for advanced installs, so it can't be added until all AWS is covered like the current script there.

Here is sample of dcos-launch running from a previous build of this PR that had said commit:
https://teamcity.mesosphere.io/viewLog.html?buildId=532605&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsDe_2&tab=buildLog#_focus=37&state=37

cmaloney:
My thought wouldn't be adding a new script, just keeping the old wrapper script and replacing all the code with two simple calls to main() in dcos-launch directly (potentially, would keep it lower effort than downloading).

Definitely if we need things like advanced support before it's feasible, then that should be next steps.

test_util/launch.py
mellenburg: This does have the capability to run any arbitrary test (when the user provides the specific arbitrary command) so the name run-internal-integration-tests is also a little bit misleading

test_util/launch.py
mellenburg: This function sole purpose is to see if a list of strings is present as keys inside a dictionary and raise a single error for all the missing keys at once. I do not think it would make a lot of sense to mutate that dictionary here as it can apply to a broad scope of targets, irrespective of key-values

test_util/launch.py
lingmann: How about using a keyword of `state` or `status` here, since that is what this file represents?

mellenburg: Do you mean switching to use 'state-path' or 'status-path'? Or do you just mean including 'state' or 'status' in the description?

test_util/launch.py
```diff
@@ -1,20 +1,288 @@
...
+  wait      Block until the cluster is up and running.
+  describe  Return additional information about the composition of the cluster.
+  pytest    Runs integration test suite on cluster. Can optionally supply options and arguments to pytest
+  delete    Destroying the provided cluster deployment.
+
+Options:
+  --config-path=<path>  Path for config to create cluster from [default: config.yaml].
+  --info-path=<path>    JSON file output by create and consumed by wait, describe,
+                        and delete [default: cluster_info.json].
+  --env=<envlist>       Specifies a list of environment variables to be passed from
```
branden: This needs to mention that it's a comma-delimited list.

test_util/launch.py
```diff
@@ -1,20 +1,288 @@
...
+        print('Cluster is ready!')
+        return 0
+
+    if args['describe']:
+        print(json_prettyprint(launcher.describe(info)))
+        return 0
+
+    if args['pytest']:
+        test_cmd = 'py.test '
+        if args['--env'] is not None:
```
branden: Let's `assert '=' not in args['--env']` here so that the command fails ASAP if someone mistakenly tries to define an env var within the `dcos-launch pytest` invocation.

test_util/launch.py
```diff
@@ -1,20 +1,288 @@
...
+        launcher.wait(info)
+        print('Cluster is ready!')
+        return 0
+
+    if args['describe']:
+        print(json_prettyprint(launcher.describe(info)))
+        return 0
+
+    if args['pytest']:
+        test_cmd = 'py.test '
```
branden: Nit: Do we need this trailing space if we aren't passing `<pytest_extras>`? If not then we should be adding it when we build that string below. `test_cmd += ' ' + ' '.join(...)`

PR #886, opened 2016-10-24
Teach configuration to pass through internal load balancer DNS for certificate generation

cmaloney:
This makes it possible to do things like issue certs properly inside the cluster
for all the addresses of a master. Previously that was only possible for AWS
clusters because the integration was AWS only.

Builds on #883, Train 91 (fixes a bug that existed before and after that train, but in a piece of code #868 reworks some)

https://mesosphere.atlassian.net/browse/DCOS-10934


cmaloney:
note: this will require downstream changes

cmaloney:
cc: @alberts

PR #889, opened 2016-10-25
[TESTING] Added rlimit support.

jieyu:
This patch added rlimit support to Mesos. It is based on 1.2 Mesos head 16f479d.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -375,10 +375,19 @@ def calculate_config_yaml(user_arguments):
 
 
 def calculate_mesos_isolation(enable_gpu_isolation):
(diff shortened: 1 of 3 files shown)


cmaloney:
This pr hasn't been updated recently, assuming it was a testing branch and isn't needed anymore.

adam-mesos:
Correct. This testing PR has been superseded by https://github.com/dcos/dcos/pull/977 which has an even more recent Mesos master.
@jieyu Do we need to add `posix/rlimits` to the default DCOS isolators, or is this something customers would configure themselves? Should we add a config.yaml flag for it?

PR #894, opened 2016-10-26
Account for AWS under-provisioning

mellenburg:
CI tests are failing because Cloudformations report as "CREATE_COMPLETE" yet they have less than the number of instances required by the deployment. Eventually those instances seem to appear and the logs don't indicate that our client could have done any better with the information returned. See:
https://teamcity.mesosphere.io/viewLog.html?buildId=456276&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp

Thus, keep retrying until we get the expected number of nodes.

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -126,6 +126,12 @@ class CfStack():
     def wait_for_stack_deletion(self, wait_before_poll_min=3):
         self.wait_for_status_change('DELETE_IN_PROGRESS', 'DELETE_COMPLETE', wait_before_poll_min)
 
+    def get_parameter(self, param):
+        for p in self.stack.parameters:
+            if p['ParameterKey'] == param:
+                return p['ParameterValue']
+        return None
+
 
 class DcosCfSimple(CfStack):
     @classmethod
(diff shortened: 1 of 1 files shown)


amitaekbote:
Lets see if this happens again? two times is not enough to throw in more retries.

mellenburg:
@amitaekbote This has already happened a third time:
https://teamcity.mesosphere.io/viewLog.html?buildId=453465&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp
https://teamcity.mesosphere.io/viewLog.html?buildId=456154&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp

Those are just recent examples, I've been seeing stuff like this (returns with fewer nodes than requested) for months. It peaks sporadically, but its a one more fault to be tolerant to

mellenburg:
@cmaloney This is still alarmingly common: https://teamcity.mesosphere.io/viewLog.html?buildId=456696&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp

Might want to merge sooner rather than later to save some CI cycles

mellenburg:
As @amitaekbote pointed out to me, tagging is a separate event from instance creation and so AWS is likely to report that a bunch of instances are ready before it tags them. Its odd that the gap between tagging and creation is so often instantaneous enough that we usually dont have this issue...

In any case, this is a simple fix to unblock, but going forward I will add utilities for a stack to find its own subnet/vpc and thereby make associations that way

PR #899, opened 2016-10-26
Upgrade marathon to 1.3.5

jasongilanfarr:


cmaloney:
failures are related to not being based on latest `dcos/dcos` master, hand-fixed when pulling into Train 93

PR #902, opened 2016-10-27
Update dcos_installer/backend.py::calculate_aws_template_storage_region_name to handle empty value for `us-east-1`

BenWhitehead:
The aws s3 api will return an empty value when the buckets region is `us-east-1` [1]. calculate_aws_template_storage_region_name has been updated to return the value `us-east-1` when this happens.

[1] http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html#RESTBucketGETlocation-responses-response-elements

### DIFF
--- a/dcos_installer/backend.py
+++ b/dcos_installer/backend.py
@@ -145,7 +145,14 @@ def calculate_aws_template_storage_region_name(
 
     try:
         location_info = session.client('s3').get_bucket_location(Bucket=aws_template_storage_bucket)
-        return location_info["LocationConstraint"]
+        loc = location_info["LocationConstraint"]
+        if loc is None or loc.strip() == "":
+            # If a buckets region is in fact 'us-east-1' the response from the api will actually be an empty value?!
+            # Rather than returning the empty value on to we set it to 'us-east-1'.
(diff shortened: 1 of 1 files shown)


mellenburg:
:ship: Gotta love the special region

dcos_installer/backend.py
```diff
@@ -145,7 +145,14 @@ def calculate_aws_template_storage_region_name(
 
     try:
         location_info = session.client('s3').get_bucket_location(Bucket=aws_template_storage_bucket)
-        return location_info["LocationConstraint"]
+        loc = location_info["LocationConstraint"]
+        if loc:
```
orsenthil: ```if "" == loc.strip():``` is better.

```
>>> s = "      "
>>> bool(s)
True
>>> s = ""
>>> bool(s)
False
```

orsenthil: Oh, in that case, shouldn't it be `if loc is None` ?

This review makes the case that, even after reading the description, I could not  get what is being checked.

mellenburg: Fair point, ```# If a buckets region is in fact 'us-east-1' the response from the api will actually be an empty value?!``` should be changed to say ```will actually return None```

orsenthil:
ship-it. I am fine with this change. 

My code-review comments were from the perspective of the reader trying to guess the value of the `loc` if it was None or an empty string, and trying to be explicit.

PR #903, opened 2016-10-27
Revert "Make dcos-spartan-watchdog start later rather than have a random 60 second sleep"

cmaloney:
This reverts commit ecfe9a57ac9e61e84ad84161c8854f51aa12f417.

There still isn't a good explanation _why_ sleep is better than a delay. The behavior hasn't
really been explained.

That said, this fixes the issue for one customer at present, although we still need to come up with
a reasonable reason for that and _proper_ fixes. `sleep` is rarely the correct solution to software
problems, just a band-aid...

I'm really not a fan of this, only doing it because we need a fix today.

# Issues

[DCOS-10809](https://mesosphere.atlassian.net/browse/DCOS-10809)

# Checklist

 - [ ] No: Included a test which will fail if code is reverted but test is not
 - [ x ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)


sargun:
👍

PR #905, opened 2016-10-27
Change the default build image to Ubuntu 16

kaysoky:
This PR does:
- Alphabetizes the list of dependencies in the Dockerfile
- Bumps the base image from `ubuntu:14.04.4` to `ubuntu:16.04`
  - Changes `gcc` to `gcc-4.8` and `g++-4.8` (the same version that is currently being used) and set `CC` and `CXX` environment variables appropriately.
  - Bumps the package for `linux-headers` to match the updated kernel
  - Add `libpcre++-dev` dependency for adminrouter/nginx (the installed library is placed in the same location as before)
  - Add `xutils-dev` dependency for the `makedepend` utility
- `s/CXXFLAGS/CPPFLAGS/` in the main Python build, needed to find and build the `_ssl` module
- Removed extraneous `sudo`s from the Cosmos and Marathon build scripts

### DIFF
--- a/packages/cosmos/build
+++ b/packages/cosmos/build
@@ -4,7 +4,7 @@ mkdir -p "$PKG_PATH/lib/"
 mkdir -p "$PKG_PATH/usr/"
 
 cp -rp /pkg/src/cosmos/cosmos-*-one-jar.jar "${PKG_PATH}/usr/cosmos.jar"
-sudo chmod -R o+r "${PKG_PATH}/usr"
+chmod -R o+r "${PKG_PATH}/usr"
 
 cosmos_service="${PKG_PATH}/dcos.target.wants_master/dcos-cosmos.service"
 mkdir -p $(dirname "$cosmos_service")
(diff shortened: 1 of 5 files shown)


kaysoky:
@cmaloney Ready for review!

packages/cosmos/build
```diff
@@ -4,7 +4,7 @@ mkdir -p "$PKG_PATH/lib/"
 mkdir -p "$PKG_PATH/usr/"
 
 cp -rp /pkg/src/cosmos/cosmos-*-one-jar.jar "${PKG_PATH}/usr/cosmos.jar"
-sudo chmod -R o+r "${PKG_PATH}/usr"
```
cmaloney: :+1:

pkgpanda/docker/dcos-builder/Dockerfile
```diff
@@ -1,39 +1,46 @@
...
-  libpopt-dev \
-  libnl-3-dev \
-  libnl-genl-3-dev \
-  linux-headers-3.19.0-21-generic \
-  pkg-config \
-  gettext-base \
-  unzip
+  zlib1g-dev
+
+ENV CC=/usr/bin/gcc-4.8
```
cmaloney: why can't / don't we want to move to the new compiler version?

cmaloney: Sounds good for now. Definitely we're going to need to bundle libstdc++ in the future (And maybe libc as well).

PR #910, opened 2016-10-28
Minuteman ipvs

aeyakovenko:
work in progress


cmaloney:
@aeyakovenko is this still being worked on or can it be closed?

PR #913, opened 2016-10-29
Cleanup logging

mellenburg:
- Many of our builds logs are extremely hard to parse
- Test logs should pump out status logging info continuously so sudden failure causes can be diagnosed

### DIFF
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -187,7 +187,7 @@ def main():
                 provider='azure',
                 test_dns_search=False,
                 add_env=get_test_config(),
-                pytest_cmd=os.getenv('DCOS_PYTEST_CMD', "py.test -rs -vv -m 'not ccm' ") + os.getenv('CI_FLAGS', ''))
+                pytest_cmd=os.getenv('DCOS_PYTEST_CMD', "py.test -vv -s -rs -m 'not ccm' ") + os.getenv('CI_FLAGS', ''))
         test_successful = True
     except Exception as ex:
         traceback.print_exc()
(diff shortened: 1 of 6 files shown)


mellenburg:
@cmaloney This is ready for review. I watched the logs as the tests were running to confirm that info is streaming out now. Thus, if the cluster fails to come up, one will finally be able to see the cause in the logs

ssh/ssh_tunnel.py
```diff
@@ -88,13 +88,13 @@ def write_to_remote(self, src, dst):
...
         cmd = self.ssh_cmd + ['-C', '-p', str(self.port), self.target, 'cat>' + dst]
-        logger.debug('Running socket write: ' + ' '.join(cmd))
+        log.info('Running socket write: ' + ' '.join(cmd))
         with open(src, 'r') as fh:
             check_call(cmd, stdin=fh)
 
     def close(self):
         close_tunnel = self.ssh_cmd + ['-p', str(self.port), '-O', 'exit', self.target]
-        logger.debug('Closing SSH Tunnel: ' + ' '.join(close_tunnel))
+        log.debug('Closing SSH Tunnel: ' + ' '.join(close_tunnel))
```
cmaloney: I think this should become log.info() as well. We log.info() the creation, and if this happens unexpectedly early bad things happen.

test_util/azure_test_driver.py
```diff
@@ -17,6 +18,10 @@
 from ssh.ssh_tunnel import SSHTunnel
 from test_util.test_runner import integration_test
 
+logging.getLogger("msrest").setLevel(logging.INFO)
+logging.getLogger("msrest").setLevel(logging.INFO)
```
cmaloney: duplicate line

cmaloney:
Currently this has both `-vv -s -rs` and `-rs -s -vv`. Can we make it only do one of them?

mellenburg:
@cmaloney This PR has been updated so that every test now calls 'pytest -vv -s -rs'

cmaloney:
:ship:

PR #914, opened 2016-10-29
Refactor azure

mellenburg:
Our azure test utility is currently only good for running an integration test. This commit will split out its functionality so we can eventually move towards a uniform dcos-cluster testing interface


mellenburg:
@lingmann PTAL

test_util/azure.py
```diff
@@ -0,0 +1,170 @@
...
+            raise Exception("Template verification failed!")
+
+        log.info('Starting template deployment')
+        deploy_poller = azure_wrapper.rmc.deployments.create_or_update(
+            group_name, deployment_name, deployment_properties)
+
+        @retry(stop_max_delay=45 * 60 * 1000,
+               retry_on_result=lambda res: res is False,
+               retry_on_exception=lambda ex: ex is not KeyboardInterrupt)
+        def poll_deploy():
```
cmaloney: shouldn't polling be separate from "start it"? That way we can connect to a deployment already in process?

test_util/azure.py
```diff
@@ -0,0 +1,170 @@
...
+
+        try:
+            deployment_result = poll_deploy()
+        except:
+            if rollback:
+                azure_wrapper.rmc.resource_groups.delete(group_name)
+            log.exception('Polling raise an exception!')
+            raise Exception('TemplateDeployment failed!')
+
+        outputs = {}
```
cmaloney: `outputs = {k: v['value'] for k, v in deployment_result.properties.outputs.items}`?

test_util/azure.py
```diff
@@ -0,0 +1,170 @@
...
+        # Remove worthless value tags from outputs
+        for k, v in deployment_result.properties.outputs.items():
+            outputs[k] = v['value']
+        new_deploy = cls(group_name, azure_wrapper)
+        new_deploy.outputs = outputs
```
cmaloney: if outputs should be assigned like this, shouldn't it be something that the class itself gets / builds? It's odd this is setting a member right after constructing it.

test_util/azure.py
```diff
@@ -0,0 +1,170 @@
...
+        return new_deploy
+
+    def get_ip_bucket(self, node_type, force_scan=False):
```
cmaloney: can we make this a lazy property instead of the magic non-set -> set thing? http://stackoverflow.com/a/3013910

Seems like it would help both here and the AWS case if we did that, and make the code easier to read. Basically on first call calculate and store the value, keep it around for later.

test_util/test_azure.py
```diff
@@ -0,0 +1,139 @@
...
+                provider='azure',
+                test_dns_search=False,
+                add_env=options.add_env,
+                pytest_cmd=options.pytest_cmd)
+    finally:
+        if result == 0:
+            log.info('Test successsful! Deleting CloudFormation...')
+        else:
+            logging.warning('Test exited with an error')
+        if options.azure_cleanup or result == 0:
```
cmaloney: Can we maybe use https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack rather than having to write the big try / finally with complicated ending?

test_util/test_azure.py
```diff
@@ -0,0 +1,139 @@
...
+        admin_name=options.linux_user,
+        oauth_enabled=options.oauth_enabled,
+        vm_size=options.vm_size,
+        agent_count=options.num_agents,
+        name_suffix=options.name_suffix,
+        vm_diagnostics_enabled=options.vm_diagnostics_enabled,
+        rollback=options.azure_cleanup)
+    result = 1
+    try:
+        with SSHTunnel(options.linux_user, options.ssh_key_path, arm_deploy.outputs['masterFQDN'], port=2200) as t:
```
cmaloney: Can we add a `from_azure` like https://github.com/dcos/dcos/blob/master/test_util/cluster.py#L169 then call https://github.com/dcos/dcos/blob/master/test_util/cluster.py#L472

mellenburg: I won't say we shouldn't do that, but there needs to be a log of changes to cluster to get that to work. The biggest issue is that azure launches hosts behind this load balancer that maps nodes to a certain port. Thus, the address for SSHing is totally different from the DC/OS node address. The cluster object has a lot of little checks for SSH and assumes is port 22. Thus, we would need to:
- create a special utility for managing nodes in azure
- redefine the Host named tuple to be port aware and make sure the SSH utility plays nice
- create a class for daisy-chaining from master.

the last option is my favorite. One only has to define how to access a master, and then the user can script everything else in terms of the same connection strings as what DC/OS will see. This even helps for the onprem case where private nodes just so happen to have public IPs. However, it feels a little ouf of-scope to add that

mellenburg:
@cmaloney This is ready for review again
@lingmann PTAL

cmaloney:
Azure failure seems to be actual failure: https://teamcity.mesosphere.io/viewLog.html?buildId=481307&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssAzureIntegrati_2&tab=buildLog#_focus=216

mellenburg:
@cmaloney oops, I guess I have a few more flakes to bump into. I figured that when you poll on deployment_operations, that all operations would be known/listed once the deployment began, but apparently, there is a case in which all the deployment_operations returned are succeeded but the deployment still has more operation to do. Anyway, I made it a bit simpler with this fix
Note: I already have to retry on just retrieving the template because the deployment being query-able is not atomic to starting it

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+        template_params = {
+            'sshRSAPublicKey': public_key,
+            # must be lower or validation will fail
+            'masterEndpointDNSNamePrefix': master_prefix.lower(),
+            'agentEndpointDNSNamePrefix': agent_prefix.lower(),
+            'linuxAdminUsername': admin_name,
+            'agentVMSize': vm_size,
+            'agentCount': agent_count,
+            'nameSuffix': name_suffix,
+            'oauthEnabled': repr(oauth_enabled).lower(),
```
cmaloney: we should put a type check on oauth_enabled. I'm guessing this makes it be `true`/`false` but just reading this code that isn't obvious

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class TemplateDeployment:
+    def __init__(self, group_name, azure_wrapper):
+        # each template deployment is a resource group
+        self.group_name = group_name
+        self.azure_wrapper = azure_wrapper
+
+    @classmethod
+    def create(
```
cmaloney: Can we add type annotations here? Really hard to tell what many of these should be. is vm_size a str or an int? Same goes for things like agent_count, oauth_enabled.

test_util/test_azure.py
```diff
@@ -0,0 +1,141 @@
...
+        object: generic object used for cleanly passing options through the test
+
+    Raises:
+        AssertionError: if any environment variables or resources are missing
+            or do not conform
+    """
+    options = type('Options', (object,), {})()
+
+    # Required
+    options.public_ssh_key = calculate_environment_variable('AZURE_PUBLIC_SSH_KEY')
```
cmaloney: This seems to have lost the generic `AZURE_PARAM` passing through that the old  version had to gather all the azure parameters. It's also changing the env variable API as part of quite a bit of code motion, which I'd rather not do.

cmaloney: Also note the old code allowed editing / setting arbitrary azure template parameters, either "integer" type ones or string type ones. I think that's a valuable feature to keep around.

test_util/test_azure.py
```diff
@@ -0,0 +1,141 @@
...
+    options.vm_diagnostics_enabled = os.getenv('AZURE_VM_DIAGNOSTICS_ENABLED', 'true') == 'true'
+    options.azure_cleanup = os.getenv('AZURE_CLEANUP', 'true') == 'true'
+    options.ci_flags = os.getenv('CI_FLAGS', '')
+
+    add_env = {}
+    prefix = 'TEST_ADD_ENV_'
+    for k, v in os.environ.items():
+        if k.startswith(prefix):
+            add_env[k.replace(prefix, '')] = v
+    options.add_env = add_env
```
cmaloney: The DCOS_AUTH_ENABLED is no longer passed here

test_util/test_azure.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    options = type('Options', (object,), {})()
+
+    # Required
+    options.public_ssh_key = calculate_environment_variable('AZURE_PUBLIC_SSH_KEY')
+    options.subscription_id = calculate_environment_variable('AZURE_SUBSCRIPTION_ID')
+    options.tenant_id = calculate_environment_variable('AZURE_TENANT_ID')
+    options.client_id = calculate_environment_variable('AZURE_CLIENT_ID')
+    options.client_secret = calculate_environment_variable('AZURE_CLIENT_SECRET')
+    options.template_url = calculate_environment_variable('DCOS_TEMPLATE_URL')
```
cmaloney: Why did this change from AZURE_TEMPLATE_URL to DCOS_TEMPLATE_URL? It has to be an azure template...

test_util/azure_test_driver.py
```diff
@@ -1,226 +0,0 @@
...
-    # client_secret = os.environ.get('AZURE_CLIENT_SECRET')
-    group_name = 'testing' + ''.join(random.choice('01234567890abcdef') for n in range(10))
-    deployment_name = 'deployment{}'.format(uuid.uuid4().hex)
-
-    rmc = ResourceManagementClient(credentials, subscription_id)
-
-    template_parameters = get_env_params()
-
-    # Output resource group
-    print("Resource group name: {}".format(group_name))
```
cmaloney: These are really helpful / necessary for debugging, need to keep logging them.

test_util/azure_test_driver.py
```diff
@@ -1,226 +0,0 @@
...
-    print("Deployment name: {}".format(deployment_name))
-
-    azure_cluster = {
-        'resource_group_name': group_name,
-        'deployment_name': deployment_name}
-    pkgpanda.util.write_json('azure-cluster.json', azure_cluster)
```
cmaloney: This azure-cluster.json is necessary so that we can properly delete on error / have a two-stage job where if the first times out in TeamCity, the second will always attempt a delete

test_util/azure_test_driver.py
```diff
@@ -1,226 +0,0 @@
...
-
-    # Create a new resource group
-    print("Creating new resource group in location: {}".format(location))
```
cmaloney: this printing is also needed for debugging

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+            'agentVMSize': vm_size,
+            'agentCount': agent_count,
+            'nameSuffix': name_suffix,
+            'oauthEnabled': repr(oauth_enabled).lower(),
+            # oauth uses string, vm diagnostic uses bool
+            'enableVMDiagnostics': vm_diagnostics_enabled}
+        log.info('Provided template parameters: {}'.format(template_params))
+        # add worthless required 'value' fields
+        template_parameters = {k: {'value': v} for k, v in template_params.items()}
+        deployment_properties = DeploymentProperties(
```
cmaloney: This needs to be in something that will try / catch / cleanup the resource_Group if an error happens. Maybe an ExitStack and disarming it on success, like https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack.pop_all?

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+            template_link=TemplateLink(uri=template_uri),
+            mode=DeploymentMode.incremental,
+            parameters=template_parameters)
+        # Use RPC against azure to validate the ARM template is well-formed
+        log.info('Validating template deployment')
```
cmaloney: this comment and the template verification failed need to have the same message

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+        try:
+            result = azure_wrapper.rmc.deployments.validate(
+                group_name, deployment_name, properties=deployment_properties)
+            if result.error:
+                for details in result.error.details:
+                    log.error('{}: {}'.format(details.code, details.message))
+                raise Exception("Template verification failed!")
```
cmaloney: please include the details inside the exception. Having the exception without the details isn't very useful, esp. since they're right there.

test_util/azure.py
```diff
@@ -0,0 +1,198 @@
...
+            raise
+
+        log.info('Starting template deployment')
+        azure_wrapper.rmc.deployments.create_or_update(
+            group_name, deployment_name, deployment_properties)
+
+        new_deploy = cls(group_name, azure_wrapper)
+        return new_deploy
+
+    def wait_for_deployment(self, timeout=60 * 60):
```
cmaloney: these timeouts are radiccally changed from the roiginal code (45 attempts, each one minute apart). Please do the old behavior. Can talk about changing it, but this PR already has a lot moving in it, and I'd rather keep things as small as possible.

test_util/test_azure.py
mellenburg: I really dislike passing this explicitly at all. I would prefer if this is what we use instead: https://github.com/dcos/dcos/pull/989

test_util/azure.py
mellenburg: These details can be pretty verbose and so I used the logger to naturally split them out into multiple lines. I could instead do:
```
error_msgs = '\n'.join(['{}: {}'.format(d.code, d.message) for d in result.error.details])
```

test_util/test_azure.py
mellenburg: I wasn't a huge fan of the need to declare the parameter type as a part of the environment name (or at least the way we were doing it). Specifically, oauth enabled is set with a string and vmdiagnostics is set with a bool. Also, I just realized that what I have here will always fail when running against a dcos template instead of a acs tempalte, so what I would like to do is what I already tried for AWS before I realized all params are strings there: pull the given template down, look at the parameters section, verify that all user params are in the template json, cast according the json, and then go ahead.

mellenburg:
@lingmann @spahl can we remove the 'holding' tag and put this back to 'ready for review'? This is a big step in getting azure ready for dcos-launch

lingmann:
@cmaloney can you refresh your review? @mellenburg can you suggest/ping one other person for review? I'm happy to do it as long as you don't mind waiting until Tuesday.

test_util/azure_test_driver.py
lingmann: Resource group logging appears to be handled here: https://github.com/dcos/dcos/pull/914/files#diff-62f8a6125c076a8784110f740fc8a8d3R68

test_util/test_azure.py
```diff
@@ -0,0 +1,139 @@
...
+            master_list=dcos_resource_group.get_master_ips(),
+            agent_list=dcos_resource_group.get_private_ips(),
+            public_agent_list=dcos_resource_group.get_public_ips(),
+            test_cmd=options.test_cmd)
+    if result == 0:
+        log.info('Test successsful!')
+    else:
+        logging.warning('Test exited with an error')
+    if options.ci_flags:
+        result = 0  # Wipe the return code so that tests can be muted in CI
```
lingmann: This is not working correctly. The last couple of builds on this PR have failed the tests and yet TC is still showing the overall test as successful... https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssAzureIntegrati_2&buildId=536571#_state=290&focus=543

test_util/helpers.py
```diff
@@ -211,3 +213,19 @@ def remove_file():
     # Attempt to remove the file upon normal interpreter exit.
     atexit.register(remove_file)
     return temp_path
+
+
+def lazy_property(property_fn):
```
lingmann: Would like to see some documentation on this function and also some tests to validate it's behavior since it's doing "fancy" things.

test_util/azure.py
```diff
@@ -0,0 +1,196 @@
...
+                'agentEndpointDNSNamePrefix': agent_prefix.lower(),
+                'linuxAdminUsername': admin_name,
+                'agentVMSize': vm_size,
+                'agentCount': agent_count,
+                'nameSuffix': name_suffix,
+                'oauthEnabled': repr(oauth_enabled).lower(),
+                # oauth uses string, vm diagnostic uses bool
+                'enableVMDiagnostics': vm_diagnostics_enabled}
+            log.info('Provided template parameters: {}'.format(template_params))
+            # add worthless required 'value' fields
```
lingmann: Why are we doing it if the values are "worthless"? Comment should be improved.

mellenburg: Will change to "arbitrary required 'value' dictionary assignments

test_util/test_azure.py
mellenburg: For tests failing and showing green that is tracked in: https://mesosphere.atlassian.net/browse/DCOS-12772

In this case I just fat fingered it so it should be working on this next push

mellenburg:
@lingmann This is all fixed up again!

Sample output showing that Azure is actually running the tests:
https://teamcity.mesosphere.io/viewLog.html?buildId=538121&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssAzureIntegrati_2&tab=buildLog#_focus=252&state=252

test_util/test_azure.py
amitaekbote: did we fix the azure test with the same fix as mentioned in https://mesosphere.atlassian.net/browse/DCOS-12772?

test_util/runner.py
```diff
@@ -44,9 +44,9 @@ def integration_test(
         'AWS_SECRET_ACCESS_KEY=' + aws_secret_access_key,
         'AWS_REGION=' + region]
 
-    pytest_cmd = """'source /opt/mesosphere/environment.export &&
+    pytest_cmd = """ "source /opt/mesosphere/environment.export &&
```
mellenburg: This change has already been merged here:
https://github.com/dcos/dcos/commit/1a35a34b395f98527795f50a93acdc8af5702008

test_util/test_azure.py
mellenburg: It was the same thing that this https://github.com/dcos/dcos/commit/1a35a34b395f98527795f50a93acdc8af5702008

lingmann:
@mellenburg I tried to run this by hand following the same steps used with in CI job, and it complained about a missing env var:
> AssertionError: AZURE_PUBLIC_SSH_KEY must be a set environment variable

I'm assuming this is one of the new requirements that is not yet captured in our CI job? Also, can you rebase this on master? Thanks

mellenburg:
@lingmann You need to look at the params set outside of the job:
https://teamcity.mesosphere.io/admin/editBuildParams.html?id=template:ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAzureAr

we do this so we can trigger tests with customized values without having to change the test configuration everywhere. So, the CI currently captures this

test_util/azure.py
```diff
@@ -0,0 +1,196 @@
...
+                group_name,
+                ResourceGroup(location=location))
+            stack.callback(azure_wrapper.rmc.resource_groups.delete, group_name)
+            log.info('Resource group created: {}'.format(group_name))
+
+            template_params = {
+                'sshRSAPublicKey': public_key,
+                # must be lower or validation will fail
+                'masterEndpointDNSNamePrefix': master_prefix.lower(),
+                'agentEndpointDNSNamePrefix': agent_prefix.lower(),
```
lingmann: Remove the .lower() since the template already does this for us on these input params.

test_util/test_azure.py
```diff
@@ -0,0 +1,139 @@
...
+    path for the SSH key to be used with a preexiting cluster.
+    Defaults to 'default_ssh_key'
+
+CI_FLAGS: string (default=None)
+    If provided, this string will be passed directly to py.test as in:
+    py.test -vv CI_FLAGS integration_test.py
+
+TEST_ADD_ENV_*: string (default=None)
+    Any number of environment variables can be passed to integration_test.py if
+    prefixed with 'TEST_ADD_ENV_'. The prefix will be removed before passing
```
lingmann: Expand docstring to include all the configurable options.

mellenburg:
@lingmann I addressed all comments. PTAL

PR #919, opened 2016-11-01
Enforce python 3.4 in builds

mellenburg:
We are supposed to be using 3.4 everywhere, but some things still end up being run on 3.5 and then are confused by the artifacts. This commit will enforce explicit use of python 3.4 for the build environment
(disregard branch name)

### DIFF
--- a/build_local.sh
+++ b/build_local.sh
@@ -30,7 +30,7 @@ EOF
 fi
 
 # Create a python virtual environment to install the DC/OS tools to
-python3 -m venv /tmp/dcos_build_venv
+python3.4 -m venv /tmp/dcos_build_venv
 . /tmp/dcos_build_venv/bin/activate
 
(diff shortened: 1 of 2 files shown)


orsenthil:
The tox is going to complain because we still reference py34 in tox

https://github.com/mellenburg/dcos/blob/4635ed9de5435b19068b23872efdbaa1bea20d07/tox.ini#L25

* You need higher version of tox
* And reference py35 in tox for the tests to run with python 3.5

mellenburg:
@orsenthil I'm actually going to try to enforce 3.4 everywhere as its really the inconsistency between the two that breaks everything and then I'll amend this once that is all set

orsenthil:
That sounds better to me.
1. Enforce consistency.
2. Upgrade next while following 1.

PR #920, opened 2016-11-01
Combine the overlay module into the `mesos-modules` package.

kaysoky:
* Moves all code from [the HEAD of mesos-overlay-modules](https://github.com/dcos/mesos-overlay-modules/commit/c0ed33b607444ad831b1c3ce55abb9124fb176e8) into [dcos-mesos-modules](https://github.com/dcos/dcos-mesos-modules)
* `dcos-mesos-modules` includes code for a new module (journald container logger), but it is not hooked up.
  * This module needs `libsystemd-dev` to build.
* Bumps Mesos SHA to include some recent module interface changes.


kaysoky:
Ready for review 📦 !

packages/mesos-modules/build
```diff
@@ -1,15 +1,29 @@
 #!/bin/bash
+set -o nounset -o pipefail
```
cmaloney: these don't need to be set / are automatically set for all pkgpanda builds (https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L1173)

packages/mesos-modules/build
```diff
@@ -1,15 +1,29 @@
...
+
 pushd /pkg/src/mesos-modules/
 ./bootstrap
 popd
 
-source /opt/mesosphere/environment.export
+export LD_LIBRARY_PATH=/opt/mesosphere/lib
```
cmaloney: it's better to source environment.export here to get other important things dependencies might set. Sourcing environment.export will also always set LD_LIBRARY_PATH

packages/mesos-modules/build
```diff
@@ -1,15 +1,29 @@
...
+
+# Add `protoc` to the path.
+export PATH=/opt/mesosphere/lib/mesos/3rdparty/bin:$PATH
```
cmaloney: why can't we move / copy / symlink into `$PKG_PATH/bin` in the mesos package if this is critical (or just everything in the 3rdparty/bin folder)?

packages/mesos/buildinfo.json
```diff
@@ -3,8 +3,8 @@
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/mesosphere/mesos",
-    "ref": "d5746045ac740d5f28f238dc55ec95c89d2b7cd9",
-    "ref_origin" : "dcos-mesos-1.0.x-837bb4e"
+    "ref": "fd011fcce61ecf4c4d4cc3cde09a11c1470ff1b6",
```
cmaloney: need an explicit ack from the pods and the like that this is the direction we want / need to go for 1.9. To date we've been intentionally keeping pods a bit separate.

kaysoky: I assume we will eventually want to bump Mesos to 1.1.0 or beyond, regardless of when the pod feature is shipped.

I can take this commit out if I tweak some of the module code (i.e. use an older version of the interfaces).

packages/mesos-modules/build
kaysoky: :thinking: I suppose it is odd to have binaries in the `lib` directory.  Seems reasonable to move into the binary folder.

kaysoky:
Revisions posted.  

The ` test_composition.test_signal_service` seems potentially flaky, but I need to dig in and check.

packages/mesos/build
```diff
@@ -20,7 +20,14 @@ LIBS="-lssl -lcrypto" LDFLAGS="-L/opt/mesosphere/active/openssl/lib" "/pkg/src/m
 make -j$NUM_CORES
 
 make install
+
 cp src/java/target/mesos-*.jar "$libdir"
+
+# This is copied during `make install` as per the configure flag
+# `--enable-install-module-dependencies`.  Modules that have their
+# own protobufs may wish to use this during their builds.
+mv $PKG_PATH/lib/mesos/3rdparty/bin/protoc $PKG_PATH/bin/protoc
```
cmaloney: I think adding an extra symlink here would be better than doing a move. That way things which know / understand how mesos does it's 3rdparty structure natively work right, as well as things which don't.

cmaloney:
:ship:

PR #922, opened 2016-11-02
integration test: print 3dt error

darkonie:


cmaloney:
@darkonie could we changes this to do a `r.raise_for_status()` (should give good enough structured results) and land it? Trying to remove the `response.ok` asserts from everywhere, they make debugging hard when tests break / are flaky.

PR #926, opened 2016-11-02
Clean up admin ports doc

branden:
This PR removes an irrelevant port and applies consistent formatting and naming.

# Issues

N/A

# Checklist

### DIFF
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -38,7 +38,6 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 8181: dcos-exhibitor
  - 9990: dcos-cosmos
  - 15055: dcos-history-service
- - 62500: Networking API
 
 ### UDP
 
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -12,14 +12,14 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 61421: dcos-minuteman
  - 62053: dcos-spartan
  - 62080: dcos-navstar
- - 62501 - Spartan
- - 62502 - Navstar
- - 62503 - Minuteman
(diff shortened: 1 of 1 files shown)


branden:
Pinging @sargun for review.

branden:
@sargun which systemd unit provides "DCOS Overlay" (port 64000)?

docs/admin-ports.md
```diff
@@ -12,9 +12,9 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 61421: dcos-minuteman
  - 62053: dcos-spartan
  - 62080: dcos-navstar
- - 62501 - Spartan
- - 62502 - Navstar
- - 62503 - Minuteman
+ - 62501: dcos-spartan
+ - 62502: dcos-navstar
+ - 62503: dcos-minuteman
```
sargun: This is a dup?

docs/admin-ports.md
```diff
@@ -38,7 +38,6 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 8181: dcos-exhibitor
  - 9990: dcos-cosmos
  - 15055: dcos-history-service
- - 62500: Networking API
```
branden: I don't have a good answer for that. I could copy the doc and keep this port, but without a process to sync that version with this one, it'll just go out of date.

docs/admin-ports.md
sargun: Ah.

sargun: 👍

sargun:
The kernel uses it, no systemd unit.

docs/admin-ports.md
alberts: @branden is going to add a doc for just for enterprise-only ports

PR #930, opened 2016-11-03
Various CentOS 7 AMI fixes and improvements.

lloesche:
This modifies default packer settings and supporting scripts for the CentOS 7 AMI.

The current CentOS 7 AMI uses 8GB on / of which 4GB is in use by the OS and DC/OS. On our prod cluster's leader the remaining 4GB were filled with /var/log data within a day.

This PR:
- fixes a race in `dcos_vol_setup.sh` where the Kernel could sometimes take longer to pick up the newly created partition than it takes the script to proceed with `mkfs.xfs` (`dcos_vol_setup.sh[689]: /dev/xvdf1: No such file or directory`).
- makes the `partprobe` call optional. The script sometimes fails right now because the new partition is being picked up before the partprobe call and make the device busy (`Error: Error informing the kernel about modifications to partition /dev/xvde1 -- Device or resource busy.`). Since the partition previously doesn't exist, if xfs is able to format it we're all good. However when spinning up large clusters (100 nodes+) in my tests I would sometimes but rarely run into a situation where the partition blockdevice (e.g. /dev/xvde1) exists but mkfs.xfs fails and actually required a node reboot before the partition was usable. To fix this we will have to refactor the entire volume setup mechanism to one where we prepare the volumes, detect if they require a reboot, reboot the node if needed and then resume the formating and mounting. That's however beyond the scope of this PR. The way it's behaving now already increases reliability of volume setup drastically. Whereas previously out of 50 nodes ~5-10 wouldn't have all four EBS volumes mounted it's now only one or none.
- ensures we actually mount the newly created filesystem instead of `mount -a`
- adds a 20G partition at /var/log
- turns off rsyslog (/var/log/messages)
- turns on persistent systemd journal (mkdir /var/log/journal)
- allows for 10G of the 20G in /var/log to be used by the journal
- increases the journald per service rate limit to 15k/30s (500/s) up from 1000/30s (33/s)
- raises the /var/lib/docker partition 20G -> 100G to make it better suited for production use
- raises /var/lib/mesos and /dcos/volume0 20G -> 50G
- removes the tty requirement for sudo
- adds `nfs-utils` package so nodes are AWS EFS ready
- cleans up some `yum` syntax (officially the order is `yum [options] [command] ...` even though reverting command and options works)
- cleans up an `rm` typo

The storage changes increase the cost per node by $0,53/day on current GP2 EBS volumes but make the image more prod usable imho. I can remove those though if not wanted.

### DIFF
--- a/cloud_images/centos7/dcos_vol_setup.sh
+++ b/cloud_images/centos7/dcos_vol_setup.sh
@@ -8,9 +8,9 @@ function usage {
 cat <<USAGE
  USAGE: $(basename "$0") <device> <mount_location>
 
-  This script will partition, format, and persistently mount the device to the specified location.
+  This script will format, and persistently mount the device to the specified location.
   It is intended to run as an early systemd unit (local-fs-pre.target) on AWS to set up EBS volumes.
-  It will only execute if <device> is not yet partitioned.
+  It will only execute if <device> doesn't already contain a filesystem.
 
  EXAMPLES:
 
@@ -35,29 +35,22 @@ function main {
     exit 1
   fi
 
-  echo "Waiting for $device to come online"
-  until test -b "$device"; do sleep 1; done
-  partition=${device}1
-  if [[ ! -b "$partition" ]]
+  echo -n "Waiting for $device to come online"
+  until test -b "$device"; do sleep 1; echo -n .; done
+  echo
+  local formated
+  mkfs.xfs $device > /dev/null 2>&1 && formated=true || formated=false
+  if [ "$formated" = true ]
   then
-    echo "Partition $partition not detected: creating partitions"
-    parted -s -a optimal "$device" -- \
-      mklabel gpt \
-      mkpart primary xfs 1 -1
-    partprobe "$device" > /dev/null 2>&1 || :
-    echo "Waiting for $partition to be detected by the Kernel"
-    until test -b "$partition"; do sleep 1; done
-    echo "Formatting: $partition"
-    mkfs.xfs "$partition" > /dev/null
-    echo "Setting up partition mount"
+    echo "Setting up device mount"
     mkdir -p "$mount_location"
-    fstab="$partition $mount_location xfs defaults 0 2"
+    fstab="$device $mount_location xfs defaults 0 2"
     echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
     if [ "$mount_location" = "/var/log" ]; then
-      echo "Preparing $partition by migrating logs from $mount_location"
+      echo "Preparing $device by migrating logs from $mount_location"
       mkdir -p /var/log-prep
-      mount "$partition" /var/log-prep
+      mount $device /var/log-prep
       mkdir -p /var/log-prep/journal
       systemctl is-active chronyd > /dev/null && systemctl stop chronyd || :
       systemctl is-active tuned > /dev/null && systemctl stop tuned || :
@@ -68,17 +61,19 @@ function main {
       rmdir /var/log-prep
       rm -rf /var/log
       mkdir -p /var/log
-      echo "Mounting: $partition to $mount_location"
-      until grep ^$partition /etc/mtab > /dev/null; do sleep 1; mount "$mount_location"; done
+      echo -n "Mounting: $device to $mount_location"
+      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
+      echo
       systemctl restart systemd-journald || :
       systemctl is-enabled tuned > /dev/null && systemctl start tuned || :
       systemctl is-enabled chronyd > /dev/null && systemctl start chronyd || :
     else
-      echo "Mounting: $partition to $mount_location"
-      until grep ^$partition /etc/mtab > /dev/null; do sleep 1; mount "$mount_location"; done
+      echo -n "Mounting: $device to $mount_location"
+      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
+      echo
     fi
   else
-    echo "Partition $partition detected: no action taken"
+    echo "Device $device contains a filesystem: no action taken"
     exit
   fi
 }
--- a/cloud_images/centos7/dcos_vol_setup.sh
+++ b/cloud_images/centos7/dcos_vol_setup.sh
@@ -5,18 +5,23 @@ export device=${1:-}
 export mount_location=${2:-}
 
 function usage {
-cat <<USAGE
- USAGE: $(basename "$0") <device> <mount_location>
+cat <<EOUSAGE
+USAGE: $(basename "$0") <device> <mount_location>
 
-  This script will format, and persistently mount the device to the specified location.
-  It is intended to run as an early systemd unit (local-fs-pre.target) on AWS to set up EBS volumes.
-  It will only execute if <device> doesn't already contain a filesystem.
+ This script will format, and persistently mount the device to the specified
+ location. It is intended to run as an early systemd unit (local-fs-pre.target)
+ on AWS to set up EBS volumes.
 
- EXAMPLES:
+ If <mount_location> is /var/log the script will migrate existing data to the
+ new filesystem.
 
-  $(basename "$0") /dev/xvde /dcos/volume1
+ It will only execute if <device> doesn't already contain a filesystem.
 
-USAGE
+EXAMPLES:
+
+ $(basename "$0") /dev/xvde /dcos/volume1
+
+EOUSAGE
 }
 
 for i in "$@"
--- a/cloud_images/centos7/dcos_vol_setup.sh
+++ b/cloud_images/centos7/dcos_vol_setup.sh
@@ -35,6 +35,8 @@ function main {
     exit 1
   fi
 
+  echo "Waiting for $device to come online"
+  until test -b "$device"; do sleep 1; done
   partition=${device}1
   if [[ ! -b "$partition" ]]
   then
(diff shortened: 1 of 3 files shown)


cmaloney:
@lloesche can you rebase on the current master? that should cleanup a lot of the failures

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -35,23 +35,48 @@ function main {
     exit 1
   fi
 
+  echo "Waiting for $device to come online"
+  until test -b "$device"; do sleep 1; done
```
cmaloney: can we sleep and echo here (that way get sort of a log how long it takes)?

Same goes for the other 'Waiting for'.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -35,23 +35,48 @@ function main {
...
+    partprobe "$device" > /dev/null 2>&1 || :
+    echo "Waiting for $partition to be detected by the Kernel"
+    until test -b "$partition"; do sleep 1; done
     echo "Formatting: $partition"
-    mkfs.xfs "$partition" >/dev/null
+    mkfs.xfs "$partition" > /dev/null
     echo "Setting up partition mount"
     mkdir -p "$mount_location"
-    fstab="$partition $mount_location xfs defaults,nofail 0 2"
+    fstab="$partition $mount_location xfs defaults 0 2"
```
cmaloney: why drop the nofail here?

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -35,23 +35,48 @@ function main {
...
     echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
-    echo "Mounting: $partition to $mount_location"
-    mount -a
+    if [ "$mount_location" = "/var/log" ]; then
+      echo "Preparing $partition by migrating logs from $mount_location"
```
cmaloney: what is the value in the "migrating" logs rather than just making a new empty directory / cleaning out the existing directory?

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -35,23 +35,48 @@ function main {
...
+      mkdir -p /var/log-prep
+      mount "$partition" /var/log-prep
+      mkdir -p /var/log-prep/journal
+      systemctl is-active chronyd > /dev/null && systemctl stop chronyd || :
```
cmaloney: shouldn't these be reliably running? They're on by default presumably in the CentOS 7 base AMI

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -35,23 +35,48 @@ function main {
...
+    mkfs.xfs "$partition" > /dev/null
     echo "Setting up partition mount"
     mkdir -p "$mount_location"
-    fstab="$partition $mount_location xfs defaults,nofail 0 2"
+    fstab="$partition $mount_location xfs defaults 0 2"
     echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
-    echo "Mounting: $partition to $mount_location"
-    mount -a
+    if [ "$mount_location" = "/var/log" ]; then
```
cmaloney: what is the value of the conditional here? I don't see how we run this with "Sometimes it's a mount and sometimes it isn't"

The condition just seems like dead code which will bit-rot.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -21,7 +21,7 @@ sysctl -w net.ipv6.conf.all.disable_ipv6=1
 sysctl -w net.ipv6.conf.default.disable_ipv6=1
 
 echo ">>> Installing DC/OS dependencies and essential packages"
-yum install --assumeyes --tolerant perl tar xz unzip curl bind-utils net-tools ipset libtool-ltdl rsync
+yum -y --tolerant install perl tar xz unzip curl bind-utils net-tools ipset libtool-ltdl rsync nfs-utils
```
cmaloney: why do we need nfs-utils? It's not part of the base DC/OS dependencies

cloud_images/centos7/install_prereqs.sh
```diff
@@ -33,12 +33,23 @@ Type=oneshot
...
 WantedBy=local-fs.target
 EOF
 systemctl enable dcos_vol_setup
 
+echo ">>> Disable rsyslog"
+systemctl disable rsyslog
+
+echo ">>> Set journald limits"
+mkdir -p /etc/systemd/journald.conf.d/
+echo -e "[Journal]\nRateLimitBurst=15000\nRateLimitInterval=30s\nSystemMaxUse=10G" > /etc/systemd/journald.conf.d/dcos-el7.conf
```
cmaloney: why don't we just use / get this from the snippet we already always ship with DC/OS? Why do we need a special snippet here different than what we're shipping on all the other platforms?

https://github.com/dcos/dcos/blob/master/gen/cloud-config.yaml#L12

I really don't want different forms of our templates to have slightly different behaviors.

cloud_images/centos7/packer.json
```diff
@@ -22,18 +22,50 @@
...
+        "delete_on_termination": true
+      },
+      {
+        "device_name": "/dev/sdh",
+        "volume_type": "gp2",
         "volume_size": 20,
         "delete_on_termination": true
+      }
+    ],
+    "launch_block_device_mappings": [
```
cmaloney: why do we need both launch launch_block_device_mappings as well as ami_block_device_mappings now?

cloud_images/centos7/dcos_vol_setup.sh
lloesche: Because it's good to see those errors when the device isn't there for some reason.

cloud_images/centos7/dcos_vol_setup.sh
lloesche: It checks if the filesystem that's about to be created will be mounted to /var/log. If so it executes a branch of code that migrates the existing logs to the new filesystem.

cloud_images/centos7/dcos_vol_setup.sh
lloesche: I'm stopping (and later restarting) all processes that have open filehandles to /var/log before migrating and then deleting the data.

cloud_images/centos7/dcos_vol_setup.sh
lloesche: It preserves previous logs and their permissions. /var/log/chrony for instance belongs to user chrony. This way we don't have to track what was there before and owned by whom. Even when we update the base AMI it'll just be preserved.

cloud_images/centos7/dcos_vol_setup.sh
lloesche: I had that in at first but dropped it. It was just adding noise to the log and the journal is timestamped anyways so can tell how long it took by the delta between this and the next echo.

cloud_images/centos7/packer.json
lloesche: That was something @lingmann added when we were debugging the missing EBS devices on some of the nodes that got spun up. I believe he got it from our Teamcity AMI. It had something to do with when the device would be available to the system. @lingmann?

lloesche:
Re: Why does /var/log need to be a separate partition
There's a couple of low traffic system services writing directly to /var/log. Though they shouldn't be an issue. I also thought that DC/OS itself is writing there but now that I'm checking I see that in 1.8 we switched everything to /var/lib/dcos/... instead of /var/log/mesos/.
Not sure how I feel about that. The issue is that when the filesystem Zookeeper is running on is full ZK breaks (still runs and accepts connections but doesn't answer any requests) and with it the entire Master node (mesos-master and bouncer/adminrouter).
We had that happen on our prod cluster which meant that ever 5th request to the WebUI or API would result in an error.
Instead of writing to /var/lib/docs/ where we also store the app data itself we should probably write logs to /var/log/dcos/.

lloesche:
I changed the code to use the devices directly instead of first creating a single partition and then using that.

This way we can't run into the
```
Error: Error informing the kernel about modifications to partition /dev/xvde1 -- Device or resource busy.  This means Linux won't know about any changes you made to /dev/xvde1 until you reboot -- so you shouldn't mount it or use it in any way before rebooting.
```
issue and don't have to refactor the entire process into a multi stage one that involves rebooting the node before storage becomes usable.

cloud_images/centos7/dcos_vol_setup.sh
lingmann: Let's add some details to the usage section about this special case. We should call out that when the script is called with a mount location of `/var/log` the contents will be migrated and certain services will be restarted to accomplish the migration.

cloud_images/centos7/install_prereqs.sh
lingmann: @lloesche and I had a chat about this and we agreed that making the `nfs-utils` package available in our default centos image will make it easier for anyone trying to migrate legacy applications or run Jenkins (most of our Jenkins users are backing their storage by NFS). I would like to continue to encourage and educate users about the limitations with NFS as central storage for a DC/OS cluster, but at the end of the day I think it's ok to let them make this judgement call. Also they would still have to set up the NFS mounts on their cluster so this won't be something they start using without some explicit action (and hopefully review of EFS limitations) on their part.

cloud_images/centos7/packer.json
lingmann: Yup, both sections are needed to ensure that the instance always start with the extra volumes available (required at launch time vs best effort basis).

lloesche:
@lingmann I added the /var/log information to the USAGE output. I also reformatted the output to not exceed 80 characters.

lloesche:
@lingmann I added the journald config again that allows it to use up to 15GB of the 20GB /var/log filesystem. Rate Limits get set by default DC/OS installation.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -33,12 +33,23 @@ Type=oneshot
 ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvde /var/lib/mesos
 ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdf /var/lib/docker
 ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdg /dcos/volume0
+ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdh /var/log
```
cmaloney: This seems to be saying that /var/log which was conditional above must always be mounted like this...

cloud_images/centos7/install_prereqs.sh
lloesche: Why can't it be a framework's custom executor that's doing the mounting? Maybe RexRay will one day have a featureset like k8s' PersistentVolumes and support NFS. In any case whatever the user uses to mount a NFS volume he's going to require the nfs-utils package.

cloud_images/centos7/install_prereqs.sh
lloesche: It's not conditional as in it might or might not be mounted. It will always be mounted like this.
The conditional `if [ "$mount_location" = "/var/log" ]; then` just executes some extra steps that migrate the existing log data to the new filesystem.

cmaloney:
approving for now, still not a fan of the `/var/log` stuff, but want this to be off the plate before next week.

PR #933, opened 2016-11-03
Use resources to find instances in AWS util

mellenburg:
Using tags for finding instances in groups resulted in failures as tagging is not guaranteed at deployment time, but resources are guaranteed to be created. See: https://mesosphere.atlassian.net/browse/DCOS-11009

### DIFF
--- a/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
+++ b/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
@@ -1,4 +1,5 @@
 import os
+from functools import partial
 
 import pytest
 
@@ -32,20 +33,20 @@ def dcos_launchpad(cluster):
 @pytest.mark.last
 def test_agent_failure(dcos_launchpad, cluster, vip_apps):
     # make sure the app works before starting
+    @retry_boto_rate_limits
+    def get_running_agents(group_name):
+        return [i for i in dcos_launchpad.get_auto_scaling_instances(group_name)
+                if i.state['Name'] == 'running']
+
     test_util.helpers.wait_for_pong(vip_apps[0][1], 120)
     test_util.helpers.wait_for_pong(vip_apps[1][1], 10)
-
-    agents = [i['InstanceId'] for i in dcos_launchpad.
-              get_group_instances(['PublicSlaveServerGroup', 'SlaveServerGroup'], state_list=['running'])]
+    agents = [i.instance_id for i in get_running_agents('PublicSlaveServerGroup') +
+              get_running_agents('SlaveServerGroup')]
 
     # Agents are in autoscaling groups, so they will automatically be replaced
     dcos_launchpad.boto_wrapper.client('ec2').terminate_instances(InstanceIds=agents)
     waiter = dcos_launchpad.boto_wrapper.client('ec2').get_waiter('instance_terminated')
     retry_boto_rate_limits(waiter.wait)(InstanceIds=agents)
-
-    def get_running_agents():
-        return dcos_launchpad.get_group_instances(
-            ['PublicSlaveServerGroup', 'SlaveServerGroup'], state_list=['running'])
 
     # Tell mesos the machines are "down" and not coming up so things get rescheduled.
     down_hosts = [{'hostname': slave, 'ip': slave} for slave in cluster.all_slaves]
@@ -58,13 +59,15 @@ def test_agent_failure(dcos_launchpad, cluster, vip_apps):
     cluster.post('/mesos/machine/down', json=down_hosts).raise_for_status()
 
     # Wait for replacements
-    test_util.helpers.wait_for_len(get_running_agents, len(cluster.all_slaves), 600)
+    test_util.helpers.wait_for_len(partial(get_running_agents, 'SlaveServerGroup'), len(cluster.slaves), 600)
+    test_util.helpers.wait_for_len(
+        partial(get_running_agents, 'PublicSlaveServerGroup'), len(cluster.public_slaves), 600)
 
     # Reset the cluster to have the replacement agents
-    cluster.slaves = sorted([agent.private_ip for agent in
-                             dcos_launchpad.get_private_agent_ips(state_list=['running'])])
-    cluster.public_slaves = sorted([agent.private_ip for agent in
-                                    dcos_launchpad.get_public_agent_ips(state_list=['running'])])
+    cluster.slaves = sorted([agent.private_ip_address for agent in
+                             get_running_agents('SlaveServerGroup')])
+    cluster.public_slaves = sorted([agent.private_ip_address for agent in
+                                    get_running_agents('PublicSlaveServerGroup')])
     cluster.all_slaves = sorted(cluster.slaves + cluster.public_slaves)
 
     # verify that everything else is still working
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -25,7 +25,7 @@ def template_by_instance_type(instance_type):
 
 
 def instances_to_hosts(instances):
-    return [Host(i['PrivateIpAddress'], i['PublicIpAddress'] if 'PublicIpAddress' in i else None) for i in instances]
+    return [Host(i.private_ip_address, i.public_ip_address) for i in instances]
 
 
 class BotoWrapper():
@@ -70,6 +70,7 @@ class CfStack():
     def __init__(self, stack_name, boto_wrapper):
         self.boto_wrapper = boto_wrapper
         self.stack = self.boto_wrapper.resource('cloudformation').Stack(stack_name)
+        self._host_cache = {}
 
     def wait_for_status_change(self, state_1, state_2, wait_before_poll_min, timeout=60 * 60):
         """
@@ -127,10 +128,34 @@ class CfStack():
         self.wait_for_status_change('DELETE_IN_PROGRESS', 'DELETE_COMPLETE', wait_before_poll_min)
 
     def get_parameter(self, param):
+        """Returns param if in stack parameters, else returns None
+        """
         for p in self.stack.parameters:
             if p['ParameterKey'] == param:
                 return p['ParameterValue']
-        return None
+        raise KeyError('Key not found in template parameters: {}. Parameters: {}'.
+                       format(param, self.stack.parameters))
+
+    @retry_boto_rate_limits
+    def get_auto_scaling_instances(self, logical_id):
+        """ Get instances in ASG with logical_id. If logical_id is None, all ASGs will be used
+        Will return instance objects as describd here:
+        http://boto3.readthedocs.io/en/latest/reference/services/ec2.html#instance
+
+        Note: there is no ASG resource hence the need for this method
+        """
+        ec2 = self.boto_wrapper.resource('ec2')
+        return [ec2.Instance(i['InstanceId']) for asg in self.boto_wrapper.client('autoscaling').
+                describe_auto_scaling_groups(
+                    AutoScalingGroupNames=[self.stack.Resource(logical_id).physical_resource_id])
+                ['AutoScalingGroups'] for i in asg['Instances']]
+
+    def get_hosts_cached(self, group_name, refresh=False):
+        if refresh or group_name not in self._host_cache:
+            host_list = instances_to_hosts(self.get_auto_scaling_instances(group_name))
+            self._host_cache[group_name] = host_list
+            return host_list
+        return self._host_cache[group_name]
 
 
 class DcosCfSimple(CfStack):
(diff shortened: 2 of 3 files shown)


mellenburg:

test_util/aws.py
```diff
@@ -170,30 +208,28 @@ def empty_and_delete_s3_bucket_from_stack(self):
...
-        filters = [{'Name': 'tag:aws:cloudformation:stack-name', 'Values': [self.stack.stack_name]}]
-        if group_list:
-            filters.append({'Name': 'tag:aws:cloudformation:logical-id', 'Values': group_list})
-        if state_list:
-            filters.append({'Name': 'instance-state-name', 'Values': state_list})
-        reservations = self.boto_wrapper.client('ec2').describe_instances(
-            Filters=filters)['Reservations']
-        return [instance for res in reservations for instance in res['Instances']]
+        instance_ids = []
+        for asg in self.get_auto_scaling_groups():
```
cmaloney: any reason not to use: `asg_resource = self.stack.Resource(group_name)` to get the CF Resource, then `self.get_autoscaling_instances(self, asg_resource.physical_resource_id)`?

which should in theory return a list of https://boto3.readthedocs.io/en/latest/reference/services/ec2.html#instance

mellenburg:
@cmaloney This is ready for another pass
@orsenthil @amitaekbote PTAL

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -58,13 +58,15 @@ def get_running_agents():
...
+    test_util.helpers.wait_for_len(partial(get_running_agents, 'SlaveServerGroup'), len(cluster.slaves), 600)
+    test_util.helpers.wait_for_len(
+        partial(get_running_agents, 'PublicSlaveServerGroup'), len(cluster.public_slaves), 600)
 
     # Reset the cluster to have the replacement agents
-    cluster.slaves = sorted([agent.private_ip for agent in
-                             dcos_launchpad.get_private_agent_ips(state_list=['running'])])
-    cluster.public_slaves = sorted([agent.private_ip for agent in
-                                    dcos_launchpad.get_public_agent_ips(state_list=['running'])])
+    cluster.slaves = sorted([agent.private_ip_address for agent in
```
cmaloney: nit / for future: Would be nice if we could somehow just tell the ClusterApi object that "here's the new set of slaves" and it would deal with the sorted() bit. It's odd that this has to magically know they're sorted, etc. Would be nice to keep all that normalization and list validation (no repeats, no masters, etc) in one place

test_util/aws.py
```diff
@@ -127,11 +127,36 @@ def wait_for_stack_deletion(self, wait_before_poll_min=3):
         self.wait_for_status_change('DELETE_IN_PROGRESS', 'DELETE_COMPLETE', wait_before_poll_min)
 
     def get_parameter(self, param):
+        """Returns param if in stack parameters, else returns None
```
cmaloney: if we want this behavior, than this function needs to work exactly like `.get()` (https://docs.python.org/3/library/stdtypes.html#dict.get). Don't want sort of that API. (Namely: Need to be able to pass the default).

That said, in all of our cases hard failing / raising an exception if not found would probably be better since we're using this not to introspect and find out what shapes things are, but rather to get out mandatory chunks of templates we launch, where we _should_ know the template.

test_util/aws.py
```diff
@@ -274,46 +283,23 @@ def delete(self, delete_vpc=False):
...
+        return CfStack(self.stack.Resource(resource_name).physical_resource_id, self.boto_wrapper)
 
-    @retry_boto_rate_limits
-    def get_substack_hosts(self, substack):
-        instances = self.boto_wrapper.client('ec2').describe_instances(
-            InstanceIds=self.get_substack_instance_ids(substack))['Reservations'][0]['Instances']
-        return instances_to_hosts(instances)
+    def get_substack_ips(self, substack_name):
+        return instances_to_hosts(
+            self.get_resource_stack(substack_name).get_auto_scaling_instances(None))
```
cmaloney: why can't we know the stack name here?

I really don't like the "sometimes we'll aggregate" for you behavior, in the instances where we're launching our templates, should always know exactly the AWS ASG resource name provided by the templates.

test_util/aws.py
```diff
@@ -127,11 +127,36 @@ def wait_for_stack_deletion(self, wait_before_poll_min=3):
...
+                if r.resource_type == resource_type]
+
+    @retry_boto_rate_limits
+    def get_auto_scaling_instances(self, logical_id):
+        """ Get instances in ASG with logical_id. If logical_id is None, all ASGs will be used
+        Will return instance objects as describd here:
+        http://boto3.readthedocs.io/en/latest/reference/services/ec2.html#instance
+        """
+        if logical_id is None:
+            asg_list = [asg.physical_resource_id for asg in
```
cmaloney: I'm not a fan of having a list of ASGs, or auto-finding all of them if one isn't provided, it's a really weird API shape, and very unexpected behavior if somehow a None accidentally got into this (Now you give my master, agent, and public agent instances if I accidentally made a simple programming error with the simple templates?)

At the very least the logical id -> list of instances code should be one function
doing it for all asgs inside the stack would be another function which calls out to that one.

test_util/aws.py
```diff
@@ -170,30 +195,14 @@ def empty_and_delete_s3_bucket_from_stack(self):
...
 
-    def get_master_ips(self, state_list=None):
-        instances = self.get_group_instances(['MasterServerGroup'], state_list)
-        return instances_to_hosts(instances)
-
-    def get_public_agent_ips(self, state_list=None):
-        instances = self.get_group_instances(['PublicSlaveServerGroup'], state_list)
-        return instances_to_hosts(instances)
+    def get_master_ips(self):
+        return instances_to_hosts(self.get_auto_scaling_instances('MasterServerGroup'))
```
cmaloney: this should probably be caching

test_util/aws.py
```diff
@@ -170,30 +195,14 @@ def empty_and_delete_s3_bucket_from_stack(self):
...
 
-    def get_private_agent_ips(self, state_list=None):
-        instances = self.get_group_instances(['SlaveServerGroup'], state_list)
-        return instances_to_hosts(instances)
+    def get_public_agent_ips(self):
+        return instances_to_hosts(self.get_auto_scaling_instances('PublicSlaveServerGroup'))
```
cmaloney: this should probably be caching

test_util/aws.py
```diff
@@ -170,30 +195,14 @@ def empty_and_delete_s3_bucket_from_stack(self):
...
-        filters = [{'Name': 'tag:aws:cloudformation:stack-name', 'Values': [self.stack.stack_name]}]
-        if group_list:
-            filters.append({'Name': 'tag:aws:cloudformation:logical-id', 'Values': group_list})
-        if state_list:
-            filters.append({'Name': 'instance-state-name', 'Values': state_list})
-        reservations = self.boto_wrapper.client('ec2').describe_instances(
-            Filters=filters)['Reservations']
-        return [instance for res in reservations for instance in res['Instances']]
+    def get_private_agent_ips(self):
+        return instances_to_hosts(self.get_auto_scaling_instances('SlaveServerGroup'))
```
cmaloney: this should probably be caching

test_util/aws.py
cmaloney: should these be filtering for running like https://github.com/dcos/dcos/pull/933/files#diff-0ba5c9c4483f599539d64bc5a299545dR37 does?

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -58,13 +58,15 @@ def get_running_agents():
...
+                             get_running_agents('SlaveServerGroup')])
+    cluster.public_slaves = sorted([agent.private_ip_address for agent in
```
cmaloney: Why can't we call
DcosCfSimple.get_public_agent_ips()
and
DcosCfSimple.get_private_agent_ips()
https://github.com/dcos/dcos/pull/933/files#diff-3bcac1662148998e548482825464c035R204

test_util/aws.py
```diff
@@ -336,18 +322,8 @@ def delete(self):
...
     def get_vpc_host_ips(self):
-        expected_instances = int(self.get_parameter('ClusterSize'))
-        reservations = self.boto_wrapper.client('ec2').describe_instances(Filters=[{
-            'Name': 'tag-value', 'Values': [self.stack.stack_name]}])['Reservations']
-        log.debug('Reservations for {}: {}'.format(self.stack.stack_id, reservations))
-        instances = [instance for res in reservations for instance in res['Instances']]
-        if len(instances) != expected_instances:
-            return False
-        return instances_to_hosts(instances)
+        return instances_to_hosts(self.get_auto_scaling_instances(None))
```
cmaloney: we should know this stack resource name. We built the template.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: Can't call those methods because they currently return Host NamedTuples and this needs plain private IP strings. I don't mind making everything use Host tuples instead, but that should be in a separate PR as it will be many changes to touch everywhere cluster.slaves and what not is called (but well worth it)

test_util/aws.py
mellenburg: re: caching how would we ever deal with nodes that are terminated or added? Perhaps check to see if any stack new stack events have occurred. If not, we don't need to worry about refreshing

The running filter is tricky because last time I tried that I was getting only a subset of the expected nodes after the cloudformation reported "CREATE_COMPLETE" because apparently the labeling of the instances as 'running' is sometimes a little bit delayed. That was an old implementation so I can try it again to prove whether or not this is really a problem.

test_util/aws.py
mellenburg: see above; how do we decide when to refresh and pickup terminated or added nodes?

test_util/aws.py
mellenburg: see above

test_util/aws.py
mellenburg: We do. Its always CentOS server group regardless of what OS is used. I can do that here and add a comment that says "hey, this template is extremely misleading and will label any OS CentOS". It felt better to just take all the instances so this all doesn't blow up if someone puts a more accurate name than "CentosServerGroup" at that template URL.

test_util/aws.py
mellenburg: We can. I just thought it looked more elegant to say "I know this stack has a single ASG, so take whatever it is" rather than:
```
def get_public_agent_ips(self):
    return instances_to_hosts( self.get_resource_stack('PublicAgentStack').get_auto_scaling_instances('PublicAgentStackServerGroup'))
```
I don't object to changing though

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -32,21 +33,20 @@ def dcos_launchpad(cluster):
 @pytest.mark.last
 def test_agent_failure(dcos_launchpad, cluster, vip_apps):
     # make sure the app works before starting
+    def get_running_agents(group_name):
```
mellenburg: I think you're right. The issue with using resource objects instead of client objects for everything is that way boto is implemented, when you ask for these properties, it just calls out to client.describe_instances() with a specific filter. So each i.state would be a new client call. 

This is actually pretty crucial... the nice thing about keeping this the way it was is that you never need to wrap functions after you provide the instance info because it because the client calls are all made in safe method.  This unleashes the possibility of any dependent usage causing rate limit failures.

So, maybe we should wrap boto clients directly. I.E. patching at runtime. It will be a big ugly patch, but its the only way to allow the usage of these resource objects without taking a step backwards in CI stability

test_util/aws.py
mellenburg: See below. This was motivated from the confusing naming of the VPC ASG and then it just allowed me to have a slightly cleaner looking implementation in the AdvancedCf, but I am fine with changing to have more hard failures.

test_util/aws.py
mellenburg: Will switch to fail hard if not found

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
cmaloney: I think for now / this pr just adding the decorator is good enough.

Practically I _think_ when you construct the instance object with the instance ID the api call is made, and then .instance_id is accessing the local cache variable (although we'd have to verify).

Long term definitely we want to wrap boto clients directly. I _think_ http://boto3.readthedocs.io/en/latest/guide/events.html provides an event we could use (Creating client class, creating resource class) to wrap the objects, or maybe even hook into overriding the retry behavior. But I think we can safely enough postpone that until later.

test_util/aws.py
cmaloney: For caching, if hosts randomly appear / disappear currently a lot of the code (and all of our tests) I think are currently quite broken, they expect a static list of hosts / no host failures while running.

My biggest concern with caching is if we don't, everytime someone calls get_master_ips or get_public_agent_ips they're implicitly making a whole bunch of AWS API calls, which could get us rate limited.

cmaloney: Having people explicitly be able to pass through a "fetch new" / "update cache" parameter to force a re-calculation would make a lot of sense to me.

test_util/aws.py
cmaloney: I'd rather be precise, might make a helper function to make that a little shorter to write, but I like having things be a little more coupled, and that way if for some reason there are 5 ASGs rather than the just one you expect, can actually do something sane / error.

test_util/aws.py
cmaloney: Lets fix the template then to just be "ServerGroup" or something similar then. For now leaving the bad name with a TODO would be enough for getting this done

mellenburg:

test_util/aws.py
```diff
@@ -127,10 +128,34 @@ def wait_for_stack_deletion(self, wait_before_poll_min=3):
...
+        """
+        ec2 = self.boto_wrapper.resource('ec2')
+        return [ec2.Instance(i['InstanceId']) for asg in self.boto_wrapper.client('autoscaling').
+                describe_auto_scaling_groups(
+                    AutoScalingGroupNames=[self.stack.Resource(logical_id).physical_resource_id])
+                ['AutoScalingGroups'] for i in asg['Instances']]
+
+    def get_hosts_cached(self, group_name, refresh=False):
+        if refresh or group_name not in self._host_cache:
+            host_list = instances_to_hosts(self.get_auto_scaling_instances('SlaveServerGroup'))
```
cmaloney: typo: `'SlaveServerGroup'` -> `group_name`

cmaloney:
:ship:

PR #934, opened 2016-11-04
make GCC 4.8 the default in dcos-builder to avoid downstream issues

alberts:
Make GCC 4.8 the default in dcos-builder to avoid downstream issues.

Could consider using update-alternatives here, but Ubuntu doesn't do that by default.

Can do http://askubuntu.com/questions/26498/choose-gcc-and-g-version if we feel strongly.

### DIFF
--- a/pkgpanda/docker/dcos-builder/Dockerfile
+++ b/pkgpanda/docker/dcos-builder/Dockerfile
@@ -40,7 +40,12 @@ RUN apt-get -qq update && apt-get -y install \
   xz-utils \
   zlib1g-dev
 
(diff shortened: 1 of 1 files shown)


alberts:
https://github.com/mesosphere/dcos-enterprise/pull/390 looks pretty good -- just rerunning one flaky test now.

alberts:
downstream build and tests passed.

cmaloney:
@alberts please only land things in Trains. there were several other PRs that were ready to land, which should have all landed at the same time. Individual PRs should not be merged alone.

cc: @spahl / @lingmann

PR #936, opened 2016-11-04
Teamcity message integration

BenWhitehead:
Add pkgpanda.util.MessageLogger

* MessageLogger provides a set of methods that improve logging when ran
  in the context of TeamCity.
  * The first type of method is `scope`, which allows the user to define
    a "block" where all other log messages created within the scope of the
    context managed method will be "in" that block and will be collapsed
    when viewing the log in TeamCity.
  * The second type of methods set specific log levels on the messages
    being logged, and in the case of error allows providing some error
    details.
  * When any of these methods are used in a context determined not to be
    TeamCity (determined by `is_running_uder_teamcity()`) the scope
    messages will not be printed, and the log level messages will be
    simplified removing the teamcity specific elements

Refactorings
============
1. create new package `teamcity-messages`. Rather than teamcity-messages
   only being added by tox, it is now a primary dependency of pkgpanda and
   dcos-image.
2. Add usage of new "scope" and "level" messages into pkgpanda and
   dcos-image for builds


pkgpanda/build/__init__.py
```diff
@@ -19,8 +19,8 @@
 from pkgpanda.constants import RESERVED_UNIT_NAMES
 from pkgpanda.exceptions import FetchError, PackageError, ValidationError
 from pkgpanda.util import (check_forbidden_services, download_atomic, load_json,
-                           load_string, make_file, make_tar, rewrite_symlinks,
-                           write_json, write_string)
+                           load_string, logger, make_file, make_tar,
```
orsenthil: minor request: Hanging indent with sorted names will be better here instead of 3 column labels in the import.

BenWhitehead: I'm not sure I understand what you mean here.

BenWhitehead:
@cmaloney @orsenthil This is ready for review

[here](https://teamcity.mesosphere.io/viewLog.html?buildId=469363&buildTypeId=DcosIo_Dcos_BuildPr&tab=buildLog#_state=92,522&focus=523) is what the new log is like after the update.

build_teamcity
```diff
@@ -2,58 +2,80 @@
 set -x
 set -o errexit -o pipefail
 
-# Fail quickly if docker daemon is not up
-systemctl status docker
+function now { date +"%Y-%m-%dT%H:%M:%S.000" | tr -d '\n' ;}
+function _scope_opened() {
+    set +x
```
cmaloney: add a comment around why +x wrapping here?

build_teamcity
```diff
@@ -2,58 +2,80 @@
...
-rm -rf dcos-release.config.yaml
-rm -rf artifacts/
-rm -f packages/*.active.json
-rm -f packages/bootstrap.latest
-rm -f packages/*.bootstrap.tar.xz
-rm -f CHANNEL_NAME
-rm -rf build/env
-rm -f dcos_generate_config*.sh
-rm -rf wheelhouse/
+function main() {(
```
cmaloney: I don't see what the value of turning this into a function main() is? Seems like it mainly just indents the code more and adds a call to it. No code reuse or the like gained.

gen/installer/aws.py
```diff
@@ -532,50 +533,51 @@ def get_button(template_name, s3_url):
...
-    # Button page linking to the basic templates.
-    button_page = gen_buttons(build_name, reproducible_artifact_path, tag, commit, variant_arguments)
-    yield {
-        'channel_path': 'aws.html',
-        'local_content': button_page,
-        'content_type': 'text/html; charset=utf-8'}
-
-    # This renders the infra template only, which has no difference between CE and EE
-    yield from gen_supporting_template()
+    with logger.scope("Creating AWS installer"):
```
cmaloney: Why does this need to be inside each do_create function? Why can't we do it at the site that calls all of these that has the module name as well as the do_create function? Can pretty easily add the name to use as an extra piece of data it pulls out of the module:

https://github.com/dcos/dcos/blob/master/release/__init__.py#L401

Saves a lot of indenting everywhere, making code easier / simpler to read and harder to get wrong (or accidentally forget). Forgetting becomes a hard error.

pkgpanda/build/__init__.py
```diff
@@ -111,7 +111,7 @@ def get_src_fetcher(src_info, cache_dir, working_directory):
 
         return pkgpanda.build.src_fetchers.all_fetchers[kind](**args)
     except ValidationError as ex:
-        raise BuildError("Validation error when fetching sources for package: {}".format(ex))
+        raise BuildError("Validation error when fetching sources for package: {}".format(src_name)) from ex
```
cmaloney: We want the exception here (it tells us _why_ the fetch failed). Please don't make these sorts of changes when there's so much code motion already. Should be separate PRs.

pkgpanda/build/__init__.py
orsenthil: Ignore my previous comment. I meant an identation like this.

```
from pkgpanda.util import (
        check_forbidden_services,
        download_atomic,
        load_json,
        ...)
```

This was popularized by `go` and `gofmt` and has practical utility value. Some good libraries like `requests` use this style.  Then I realized our style checker supports this in an awkward way. So, let's just ignore my earlier comment.

BenWhitehead:
@cmaloney Feedback incorporated. Helper methods used for the large methods that had large indentation blocks.

I also rebased on master to pick-up the fixed for the flake8 failures.

gen/installer/bash.py
```diff
@@ -659,26 +659,29 @@ def do_create(tag, build_name, reproducible_artifact_path, commit, variant_argum
...
-        bootstrap_installer_name = '{}installer'.format(pkgpanda.util.variant_prefix(variant))
-        bootstrap_installer_id = all_bootstraps[bootstrap_installer_name]
-
-        installer_filename = make_installer_docker(variant, bootstrap_info['bootstrap_id'], bootstrap_installer_id)
-
-        yield {
-            'channel_path': 'dcos_generate_config.{}sh'.format(pkgpanda.util.variant_prefix(variant)),
-            'local_path': installer_filename
-        }
+    with logger.scope("building dcos_generate_config.sh"):
```
cmaloney: Can we move this inside the loop over variants? Practically we aren't always building dcos_generate_config.sh, sometimes it's `dcos_generate_config.ee.sh`. Would also be nice to not have the "Building installer for variant" line / just have the one block printing.

cmaloney: https://github.com/dcos/dcos/blob/master/gen/installer/bash.py#L673 prints out the actual filename which will be produced in the particular variant build.

packages/pytest/buildinfo.json
```diff
@@ -1,15 +1,10 @@
 {
-  "requires": ["python"],
+  "requires": ["python", "python-teamcity-messages"],
   "sources": {
```
cmaloney: should switch to single_source since this now only has a single source.

packages/pytest/build
```diff
@@ -3,6 +3,4 @@ source /opt/mesosphere/environment.export
 export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.5/site-packages"
 mkdir -p "$LIB_INSTALL_DIR"
 
-for package in pytest teamcity-messages; do
-  pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/$package
-done
+pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/pytest
```
cmaloney: If moving to a single python package build, please copy / paste the standard template (the pytest in this line should be $PKG_NAME): 

https://github.com/dcos/dcos/blob/master/packages/python-docopt/build#L6

packages/python-teamcity-messages/build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+source /opt/mesosphere/environment.export
+export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.5/site-packages"
+mkdir -p "$LIB_INSTALL_DIR"
+
+pip3 install --no-deps --install-option="--prefix=$PKG_PATH" --root=/ /pkg/src/teamcity-messages
```
cmaloney: same as above, please copy/paste the standard template: https://github.com/dcos/dcos/blob/master/packages/python-docopt/build#L6

packages/python-teamcity-messages/buildinfo.json
```diff
@@ -0,0 +1,10 @@
+{
+    "requires": ["python"],
+    "sources": {
```
cmaloney: Package with a single source should use `single_source`.

pkgpanda/build/__init__.py
```diff
@@ -778,8 +778,18 @@ def get_build_ids(self):
 
 
 def build(package_store, name, variant, clean_after_build, recursive=False):
+    variant_str = pkgpanda.util.variant_str(variant)
```
cmaloney: Should use https://github.com/dcos/dcos/blob/master/pkgpanda/util.py#L33 / variant_name in this printing. That will always make a non-empty string, and should always be printed.

pkgpanda/build/__init__.py
```diff
@@ -778,8 +778,18 @@ def get_build_ids(self):
...
+    if variant_str:
+        msg = "Building package {} variant {}".format(name, variant_str)
+    else:
+        msg = "Building package {}".format(name)
+    with logger.scope(msg):
+        print(msg)
```
cmaloney: Why do we need both the logger and a print here? Sometimes we do both, sometimes we only do one. From my perspective shouldn't also need to print if I'm using the logger.

pkgpanda/build/__init__.py
```diff
@@ -1182,23 +1192,24 @@ def clean():
...
-    # Bundle the artifacts into the pkgpanda package
-    tmp_name = pkg_path + "-tmp.tar.xz"
-    make_tar(tmp_name, cache_abs("result"))
-    os.rename(tmp_name, pkg_path)
-    print("Package built.")
+        # Bundle the artifacts into the pkgpanda package
+        tmp_name = pkg_path + "-tmp.tar.xz"
+        make_tar(tmp_name, cache_abs("result"))
+        os.rename(tmp_name, pkg_path)
+        print("Package built.")
```
cmaloney: this shouldn't be inside the "building tarball" block. It's stating that the package build overall reached the end (do need to keep it as a print, provides feedback vs. the "building package" scope just ending).

pkgpanda/util.py
```diff
@@ -279,3 +282,72 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+class PrintLogger:
+    def customMessage(self, text, status, errorDetails='', flowId=None):  # noqa: N802, N803
+        print("{}: {} {}".format(status, text, errorDetails))
+
+    def progressMessage(self, message):  # noqa: N802, N803
+        pass
+
+    def blockOpened(self, name, flowId=None):  # noqa: N802, N803
+        pass
```
cmaloney: Shouldn't this print something like "starting: \<name>"?

pkgpanda/util.py
```diff
@@ -279,3 +282,72 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    def blockClosed(self, name, flowId=None):  # noqa: N802, N803
+        pass
```
cmaloney: Shouldn't this print something like "ending: \<name>"

release/__init__.py
```diff
@@ -298,11 +299,8 @@ def make_stable_artifacts(cache_repository_url):
 
     # TODO(cmaloney): Rather than guessing / reverse-engineering all these paths
     # have do_build_packages get them directly from pkgpanda
-    try:
+    with logger.scope("Building packages"):
         all_completes = do_build_packages(cache_repository_url)
-    except pkgpanda.build.BuildError as ex:
-        print("ERROR Building packages:", ex, file=sys.stderr)
```
cmaloney: it's critical we still catch / pretty-print the error message here. Changes usability quite a bit.

release/__init__.py
```diff
@@ -803,7 +808,7 @@ def main():
     options = parser.parse_args()
     if not hasattr(options, 'action'):
         parser.print_help()
-        print("ERROR: Must use a subcommand")
+        logger.error("Must use a subcommand")
```
cmaloney: this needs to print out `ERROR:` / something similar for basic human CLI usage still. Same for the "Release configuration file" one below.

BenWhitehead:
@cmaloney Addressed all comments, rebased and build is green. Should be ready to merge.

pkgpanda/util.py
```diff
@@ -279,3 +282,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    @contextmanager
+    def _block(self, log, name, flow_id):
+        log.blockOpened(name, flowId=flow_id)
+        log.progressMessage(name)
+        yield
+        log.blockClosed(name, flowId=flow_id)
+
+    @contextmanager
+    def scope(self, name, flow_id=None):
```
orsenthil: I like everything about this.

Have you thought about naming the verb `scope` something other than `scope`? While reading, "scope" tends to give an impression that term has an actionable context, while this is denoting `folded` or `collapsed` format.

Other terms that I can think of include `purview`, `view` , `fold`, `block` etc.

cmaloney: Scope is the classic term for programming languages for this concept. Things happen inside a "scope" in C/C++. You can open and close arbitrary scopes in C and C++ using `{}` which is what this is doing / giving effectively.

Everything must happen _inside_ of a scope (the context manager / `with`). Inside that scope, you can create sub-scopes. The scopes stack to give layered namespaces.

pkgpanda/util.py
```diff
@@ -279,3 +282,74 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+    def error(self, text, flow_id=None, error_details=''):
+        self._custom_message(text=text, status='ERROR', flow_id=flow_id, error_details=error_details)
+
+    def failure(self, text, flow_id=None):
+        self._custom_message(text=text, status='FAILURE', flow_id=flow_id)
+
+
+class PrintLogger:
+    def customMessage(self, text, status, errorDetails='', flowId=None):  # noqa: N802, N803
```
cmaloney: :+1:  for specific noqa error/warning ignoring

release/__init__.py
```diff
@@ -398,26 +399,27 @@ def make_channel_artifacts(metadata):
...
-            if 'packages' in built_resource:
-                for package in built_resource['packages']:
-                    artifacts.append(get_gen_package_artifact(package))
-            else:
-                assert 'packages' not in built_resource
-                artifacts.append(built_resource)
-
-        # TODO(cmaloney): Check the provider artifacts adhere to the artifact template.
-        artifacts += provider_data.get('artifacts', list())
+        with logger.scope("Creating {} installer".format(module.__name__)):
```
cmaloney: Can we change this to "Creating {} deploy tools" (see: https://github.com/dcos/dcos/pull/978)

release/__init__.py
```diff
@@ -429,6 +431,11 @@ def make_abs(path):
 
 
 def do_build_docker(name, path):
+    with logger.scope("dcos/dcos-builder"):
```
cmaloney: can we include name here? For some versions of DC/OS we commonly build multiple dockers with this code.

pkgpanda/util.py
orsenthil: It was for the external behavior, not the language concept.
Editors which can collapse the bunch of text for better viewing  under `+` sign calls the behavior as folding. teamcity-messages seem to call it as block (while using `with` statement).

No strong objection here.

BenWhitehead: The idea I had here is that in the code it would read as "With this logging scope run..."  The idea being that all stuff that happened within the block would be happening in that log scope.

PR #946, opened 2016-11-10
Upgrade Navstar & Minuteman

sargun:
 - This change fixes quite a few issues with Minuteman and Navstar
 - It fixes the performance and scalability issues with Minuteman and Navstar
 - It fixes the DNS polling issues
 - It switches to the IPVS-based datapath for Minuteman

# Issues
https://mesosphere.atlassian.net/browse/DCOS-10256
https://mesosphere.atlassian.net/browse/DCOS-11667
https://mesosphere.atlassian.net/browse/DCOS-9754
https://mesosphere.atlassian.net/browse/DCOS-11024
https://mesosphere.atlassian.net/browse/DCOS-11300

# Checklist

  - This doesn't change any functionality, but merely changes the implementation of it. The existing tests are valid

 - [x] Change Log from last: 
  * https://github.com/dcos/navstar/pull/31
  * https://github.com/dcos/minuteman/pull/61
  * https://github.com/dcos/minuteman/pull/65
 - [x] Test Results: In PRs above
 - [x] Code Coverage: In PRs above

### DIFF
--- a/packages/dcos-integration-test/extra/util/python_test_server.py
+++ b/packages/dcos-integration-test/extra/util/python_test_server.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import urllib.parse
 from http.server import BaseHTTPRequestHandler, HTTPServer
+from socketserver import ThreadingMixIn
 
 import requests
 
@@ -283,6 +284,13 @@ class TestHTTPRequestHandler(BaseHTTPRequestHandler):
             self.send_error(404, 'Not found', 'Endpoint is not supported')
 
 
+# We use a HTTPServer with the ThreadingMixIn in order to handle stalled HTTP requess. Without this, the behaviour
+# is to handle requests serially. If a connection is severed, this can result in the server hanging for the TCP
+# timeout before answering another requests...leading to failures.
+class ThreadingSimpleServer(ThreadingMixIn, HTTPServer):
+    pass
+
+
 def _verify_environment():
     """Verify that the enviroment is sane and can be used by the test_server"""
     if TEST_UUID_VARNAME not in os.environ:
@@ -300,7 +308,7 @@ def start_http_server(listen_port):
 
     logging.info("HTTP server is starting, port: "
                  "{}, test-UUID: '{}'".format(listen_port, os.environ[TEST_UUID_VARNAME]))
-    test_server = HTTPServer(('', listen_port), TestHTTPRequestHandler)
+    test_server = ThreadingSimpleServer(('', listen_port), TestHTTPRequestHandler)
 
     def sigterm_handler(_signo, _stack_frame):
         test_server.server_close()
--- a/packages/minuteman/buildinfo.json
+++ b/packages/minuteman/buildinfo.json
@@ -4,15 +4,14 @@
     "minuteman": {
       "kind": "git",
       "git": "https://github.com/dcos/minuteman.git",
-      "ref": "54b665a30deb8ee7c3230140273e580d98826ad7",
+      "ref": "e6226c74a860d7cc1b7661862cb14f714426ba58",
       "ref_origin": "master"
     }
   },
   "sysctl": {
       "dcos-minuteman": {
-          "net.netfilter.nf_conntrack_tcp_be_liberal": "1",
-          "net.netfilter.ip_conntrack_tcp_be_liberal": "1",
-          "net.ipv4.netfilter.ip_conntrack_tcp_be_liberal": "1"
+          "net.ipv4.vs.snat_reroute": "1",
+          "net.ipv4.vs.conntrack": "1"
       }
   }
 }
--- a/packages/minuteman/extra/dcos-minuteman.service
+++ b/packages/minuteman/extra/dcos-minuteman.service
@@ -1,0 +1,23 @@
+[Unit]
+Description=Layer 4 Load Balancer: DC/OS Layer 4 Load Balancing Service
+
+[Service]
+Restart=always
+StartLimitInterval=0
+RestartSec=5
+WorkingDirectory=/opt/mesosphere/active/minuteman/minuteman
+Environment=HOME=/opt/mesosphere
+EnvironmentFile=/opt/mesosphere/etc/minuteman-erl.env
+EnvironmentFile=/opt/mesosphere/etc/check_time.env
+EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=-/opt/mesosphere/etc/minuteman.env
+EnvironmentFile=-/run/dcos/etc/minuteman_auth.env
+ExecStartPre=/opt/mesosphere/bin/check-time
+ExecStartPre=/usr/bin/env modprobe ip_vs_wlc
+ExecStartPre=/opt/mesosphere/bin/setup_iptables.sh
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-minuteman
+ExecStartPre=/bin/ping -c1 ready.spartan
+ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/minuteman/mnesia
+ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/minuteman/lashup
+ExecStart=/opt/mesosphere/active/minuteman/minuteman/bin/minuteman-env foreground
--- a/packages/minuteman/extra/setup_iptables.sh
+++ b/packages/minuteman/extra/setup_iptables.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -euo pipefail
+set -x
+
+# http://stackoverflow.com/questions/2870992/automatic-exit-from-bash-shell-script-on-error
+# Any subsequent(*) commands which fail will cause the shell script to exit immediately
+
+## Cleanup old minuteman (NFQUEUE) rules
+iptables --wait -D FORWARD -p tcp -m set --match-set minuteman dst,dst -m tcp --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable || true
+iptables --wait -D OUTPUT -p tcp -m set --match-set minuteman dst,dst -m tcp --tcp-flags FIN,SYN,RST,ACK SYN -j REJECT --reject-with icmp-port-unreachable || true
+iptables --wait -t raw -D PREROUTING -p tcp -m set --match-set minuteman dst,dst -m tcp --tcp-flags FIN,SYN,RST,ACK SYN -j NFQUEUE --queue-balance 50:58 || true
+iptables --wait -t raw -D OUTPUT -p tcp -m set --match-set minuteman dst,dst -m tcp --tcp-flags FIN,SYN,RST,ACK SYN -j NFQUEUE --queue-balance 50:58 || true
+
+RULE="POSTROUTING -m ipvs --ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment Minuteman-IPVS-IPTables-masquerade-rule -j MASQUERADE"
+
+if ! iptables --wait -t nat -C ${RULE}; then
+  echo "Minuteman IPTables Rule not found: ${RULE}"
+  echo "Inserting Rule"
+  iptables --wait -t nat -I ${RULE} && echo "Rule successfully inserted"
+else
+  echo "Minuteman IPTables rule found"
+fi
+
--- a/packages/navstar/build
+++ b/packages/navstar/build
@@ -5,7 +5,7 @@ source /opt/mesosphere/environment
 
 export CFLAGS="-I/opt/mesosphere/include -I/opt/mesosphere/active/libsodium/include"
 export LDFLAGS="-L/opt/mesosphere/lib -L/opt/mesosphere/active/libsodium/lib -Wl,-rpath=/opt/mesosphere/active/libsodium/lib"
-export CXXFLAGS=CFLAGS
+export CPPFLAGS=${CFLAGS}
 
 
 pushd /pkg/src/navstar
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -59,7 +59,7 @@ def test_ip_per_container(cluster):
 
     with cluster.marathon.deploy_and_cleanup(app_definition, check_health=True) as service_points:
         app_port = app_definition['container']['docker']['portMappings'][0]['containerPort']
-        cmd = '/opt/mesosphere/bin/curl -s -f http://{}:{}/ping'.format(service_points[1].ip, app_port)
+        cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}:{}/ping'.format(service_points[1].ip, app_port)
         ensure_routable(cmd, service_points)()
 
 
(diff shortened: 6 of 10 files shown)


sargun:
@spahl

alberts:
Needs review according to @cmaloney

packages/minuteman/extra/dcos-minuteman.service
```diff
@@ -0,0 +1,23 @@
...
+RestartSec=5
+WorkingDirectory=/opt/mesosphere/active/minuteman/minuteman
+EnvironmentFile=/opt/mesosphere/etc/minuteman-erl.env
+EnvironmentFile=/opt/mesosphere/etc/check_time.env
+EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=-/opt/mesosphere/etc/minuteman.env
+EnvironmentFile=-/run/dcos/etc/minuteman_auth.env
+ExecStartPre=/opt/mesosphere/bin/check-time
+ExecStartPre=/usr/bin/env modprobe ip_vs_wlc
+ExecStartPre=/opt/mesosphere/bin/setup_iptables.sh
```
branden: Would it be worth cleaning up this IPtables rule in an `ExecStopPost`? If this service cleans up after itself, that might make upgrading easier, like if we change `setup_iptables.sh` in a subsequent release.

packages/dcos-integration-test/extra/util/python_test_server.py
```diff
@@ -300,7 +305,7 @@ def start_http_server(listen_port):
 
     logging.info("HTTP server is starting, port: "
                  "{}, test-UUID: '{}'".format(listen_port, os.environ[TEST_UUID_VARNAME]))
-    test_server = HTTPServer(('', listen_port), TestHTTPRequestHandler)
+    test_server = ThreadingSimpleServer(('', listen_port), TestHTTPRequestHandler)
```
branden: Why this change? Not seeing any changes to tests that impact concurrency.

sargun: If the first request, of any of the healthcheck connections fail, the server gets "stuck"

sargun: It's resulted in a bunch of false failures.

packages/minuteman/extra/dcos-minuteman.service
sargun: Sure. I think it'll kill connections in flight during the upgrade, but tell me what you think, and I'll implement it.

spahl: Yeah that is probably not a good idea

spahl: Does this have to be a shell script outside? Why can't the minuteman code take care of that one?

packages/minuteman/extra/dcos-minuteman.service
```diff
@@ -0,0 +1,23 @@
...
+StartLimitInterval=0
+RestartSec=5
+WorkingDirectory=/opt/mesosphere/active/minuteman/minuteman
+EnvironmentFile=/opt/mesosphere/etc/minuteman-erl.env
+EnvironmentFile=/opt/mesosphere/etc/check_time.env
+EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=-/opt/mesosphere/etc/minuteman.env
+EnvironmentFile=-/run/dcos/etc/minuteman_auth.env
+ExecStartPre=/opt/mesosphere/bin/check-time
+ExecStartPre=/usr/bin/env modprobe ip_vs_wlc
```
spahl: Previously the minuteman code called iptables. That in turn caused it to automatically load the module inside of the iptables/ipset code (since it loads missing modules by itself).

This version of minuteman does not shell out to iptables/ipset anymore but still requires the same modules so we need to load it at some point.

I think it would be cleaner to have minuteman load the module itself and hard fail if it does not exist versus putting this  logic in the init code.

packages/minuteman/buildinfo.json
```diff
@@ -4,15 +4,14 @@
...
       "ref_origin": "master"
     }
   },
   "sysctl": {
       "dcos-minuteman": {
-          "net.netfilter.nf_conntrack_tcp_be_liberal": "1",
-          "net.netfilter.ip_conntrack_tcp_be_liberal": "1",
-          "net.ipv4.netfilter.ip_conntrack_tcp_be_liberal": "1"
+          "net.ipv4.vs.snat_reroute": "1",
+          "net.ipv4.vs.conntrack": "1"
```
spahl: Can we get a bit more details about what changed here? How does it affect existing users?

packages/minuteman/extra/dcos-minuteman.service
sargun: I disagree. I think putting this logic in the Systemd unit is a bit cleaner. Part of this work is to move us to be able to run Minuteman without root privileges. From my understanding, that's been an ask for some time now. This removes any CAP_SYS_ADMIN need from Minuteman.

packages/minuteman/buildinfo.json
sargun: As mentioned in the PR, we converted the code to us the IPVS datapath. Existing users will experience a seamless transition.

sargun:
RE: Does this have to be a shell script outside? Why can't the minuteman code take care of that one?
See: CAP_SYS_ADMIN

packages/minuteman/extra/dcos-minuteman.service
cmaloney: Several "secured" distributions/operating enviornments I'm familiar with will hard-fail software which tries to do a modprobe of a kernel module. If we need to do things like ensure a module is loaded, probably need to use `/etc/modprobe.d` or a similar mechanism to get them loaded by the pieces which are supposed to, as well as adding it to the documentation as a system requirement.

spahl: This would be an improvement but my point was that moving to modprobe here (or in the process) is no different than the existing behavior where it is required and loaded by iptables on the fly (I looked at the code, there is nothing magical, it just loads missing netfilter modules). Nothing new will be breaking in that regard vs older versions of DC/OS.

sargun:
This is in preparation to help support a setuid binary. We have a binary called `bin/procket`. Once we have this in, the next step is to either setcap, or setuid that binary, and then minuteman can run as a normal user.

spahl:
That makes sense yes.

packages/minuteman/extra/dcos-minuteman.service
cmaloney: Running modprobe explicitly is caught by things like selinux, and results in hard failure. Implicitly loading the module because you asked for an API which needs it is not

packages/minuteman/extra/dcos-minuteman.service
```diff
@@ -0,0 +1,23 @@
...
+ExecStartPre=/opt/mesosphere/bin/setup_iptables.sh
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-minuteman
+ExecStartPre=/bin/ping -c1 ready.spartan
+ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/usr/bin/env mkdir -p /var/lib/dcos/minuteman/mnesia
```
cmaloney: This would all end up getting owned by root because the ExecStartPre is run as root, which isn't what we want. Minuteman process probably needs to do the mkdir as part of it's startup.

packages/minuteman/extra/dcos-minuteman.service
```diff
@@ -0,0 +1,23 @@
...
+ExecStartPre=/usr/bin/env mkdir -p /var/lib/dcos/minuteman/lashup
+ExecStart=/opt/mesosphere/active/minuteman/minuteman/bin/minuteman-env foreground
+Environment=HOME=/opt/mesosphere
```
cmaloney: Why isn't this environment variable with all the `EnvironmentFile` entries?

packages/minuteman/extra/setup_iptables.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
```
cmaloney: For all the bash scripts in DC/OS have a standard set of options we use:
`set -euo pipefail`
`set -x`

packages/minuteman/extra/setup_iptables.sh
```diff
@@ -0,0 +1,16 @@
...
+# Any subsequent(*) commands which fail will cause the shell script to exit immediately
+
+RULE="POSTROUTING -m ipvs --ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment Minuteman-IPVS-IPTables-masquerade-rule -j MASQUERADE"
+
+if ! iptables --wait -t nat -C ${RULE}; then
+  echo "Minuteman IPTables Rule not found: ${RULE}"
+  echo "Inserting Rule"
+  iptables --wait -t nat -I ${RULE} && echo "Rule successfully inserted"
+else
+  echo "Minuteman IPTables rule found"
```
cmaloney: how do we know the existing rule is correct? Can we write a program which looks at the iptables rules, see if the right one is there and if it isn't reset / replace with the right one?

packages/navstar/build
```diff
@@ -5,7 +5,7 @@ source /opt/mesosphere/environment
 
 export CFLAGS="-I/opt/mesosphere/include -I/opt/mesosphere/active/libsodium/include"
 export LDFLAGS="-L/opt/mesosphere/lib -L/opt/mesosphere/active/libsodium/lib -Wl,-rpath=/opt/mesosphere/active/libsodium/lib"
-export CXXFLAGS=CFLAGS
+export CXXFLAGS=${CFLAGS}
```
cmaloney: https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html -- CPPFLAGS should be set if you want both CFLAGS and CXXFLAGS to be set

packages/minuteman/extra/dcos-minuteman.service
spahl: iptables calls the actual "modprobe" binary. Exactly the same way. So this does not introduce any new failure scenarios.

https://git.netfilter.org/iptables/tree/libxtables/xtables.c#n356

sargun: @cmaloney Can you point me to how iptables uses these native APIs?

packages/minuteman/extra/dcos-minuteman.service
sargun: RE: Cleanliness, running as root. Also, shelling out to iptables is something I'd rather avoid. Running it outside seems "safer"

packages/minuteman/extra/dcos-minuteman.service
sargun: That's fine. These got created on 1.7 as root. When we have the new unit that converts them to a non-superuser, we'll chown.

packages/navstar/build
sargun: Ah, I got this from 74760244. Should I change it across the board in another PR?

packages/minuteman/extra/setup_iptables.sh
cmaloney: 1. Customers don't have a rule with the same name already existing which causes our software to behave very oddly and be really hard to debug why it failed
2. The rule didn't get messed up via interactions with things like firewalld which reset all the iptables rules periodically
3. the rule didn't get messed up via interactions with docker bridge networking
4. the rule didn't get messed up via interactions with mesos overlay code

Ideally want to verify the rule / chain exists and is doing the right thing. And if it isn't say if it's a form we recognize (Either none or one of the previous versions) and update / fix it.

spahl: The commit now matches enough of those conditions IMHO.

- it deletes the previous minuteman rules (the older version that uses NFQUEUE <=1.8.x)
- is uses -C to check if a very specific rule exists: RULE="POSTROUTING -m ipvs --ipvs --vdir ORIGINAL --vmethod MASQ -m comment --comment Minuteman-IPVS-IPTables-masquerade-rule -j MASQUERADE". That rule even has a comment that would make it quite identifiable.
- if the specific rule is not there it adds the one it needs to ensure it keeps working.

packages/dcos-integration-test/extra/util/python_test_server.py
cmaloney: nit: would be nice to have this as a comment in the code so everyone who ever looks at this code in the future knows / can understand the reasoning. Right now they'd have to find the commit via blame then look up the PR comments to find it.

packages/minuteman/extra/setup_iptables.sh
cmaloney: How are we verifying / testing that? This adds all this bash code which does that in theory, but we don't actually test that it works like that, esp. since some of the conditions around upgrade are harder to get into.

packages/minuteman/extra/dcos-minuteman.service
```diff
@@ -0,0 +1,23 @@
...
+EnvironmentFile=-/opt/mesosphere/etc/minuteman.env
+EnvironmentFile=-/run/dcos/etc/minuteman_auth.env
+ExecStartPre=/opt/mesosphere/bin/check-time
+ExecStartPre=/usr/bin/env modprobe ip_vs_wlc
+ExecStartPre=/opt/mesosphere/bin/setup_iptables.sh
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-minuteman
+ExecStartPre=/bin/ping -c1 ready.spartan
+ExecStartPre=/bin/ping -c1 leader.mesos
+ExecStartPre=/usr/bin/env mkdir -p /var/lib/dcos/minuteman/mnesia
+ExecStartPre=/usr/bin/env mkdir -p /var/lib/dcos/minuteman/lashup
```
cmaloney: These shouldn't be using /us/bin/env, just use `/usr/bin/mkdir`. This code also very much should live inside the service itself.

sargun: Why? On Ubuntu (which I think we're planning on supporting at some point), mkdir is in `/bin`, not `/usr/bin`. This hurts portability. What's the benefit of using `/usr/bin/mkdir` over env?

cmaloney: The standard in this codebase is `/usr/bin/mkdir`, that works on the Ubuntu we do for Azure. Things being added / modified in this codebase needs to follow the standard.

sargun: How is `/usr/bin/mkdir` better than `/usr/bin/env mkdir`?

lingmann: @sargun there are multiple systemd instances that are already using `/usr/bin/mkdir` and we'd like to stay consistent / reduce variability. Furthermore, env is not as explicit and introduces potential differences in how the systemd exec might fail.

If you'd like to provide another PR which updates *all* of the hard-coded systemd paths in dcos to use `/usr/bin/env` then we can discuss that in a separate PR but I think it makes sense to follow our recommendation and not introduce a new convention here.

sargun: I'll make the change as a matter of consistency.

sargun:
🚢

sargun:
🚢🚢🚢🚢  🚢    🚢 🚢 🚢🚢🚢🚢🚢     🚢 🚢🚢🚢🚢🚢 
  🚢      🚢    🚢 🚢 🚢    🚢    🚢   🚢   
   🚢🚢🚢🚢  🚢🚢🚢🚢🚢🚢 🚢 🚢    🚢    🚢   🚢   
       🚢 🚢    🚢 🚢 🚢🚢🚢🚢🚢     🚢   🚢   
  🚢    🚢 🚢    🚢 🚢 🚢         🚢   🚢   
   🚢🚢🚢🚢  🚢    🚢 🚢 🚢         🚢   🚢

sargun:
```
  🚢🚢🚢🚢    🚢        🚢  🚢  🚢🚢🚢🚢🚢          🚢  🚢🚢🚢🚢🚢  
🚢            🚢        🚢  🚢  🚢        🚢        🚢      🚢      
  🚢🚢🚢🚢    🚢🚢🚢🚢🚢🚢  🚢  🚢        🚢        🚢      🚢      
          🚢  🚢        🚢  🚢  🚢🚢🚢🚢🚢          🚢      🚢      
🚢        🚢  🚢        🚢  🚢  🚢                  🚢      🚢      
  🚢🚢🚢🚢    🚢        🚢  🚢  🚢                  🚢      🚢 
```

PR #950, opened 2016-11-10
Remove cluster from ApiClient

mellenburg:
-Circular dependencies are bad. This only worked because python doesnt evaluate until runtime
-Allows ApiClient to be generic
-Adds a convenience constructor to make the Api from a cluster object

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -142,13 +142,13 @@ def test_octarine_srv(cluster, timeout=30):
 
 def test_pkgpanda_api(cluster):
 
-    def get_and_validate_package_ids(node, uri):
-        r = cluster.get(node=node, path=uri)
+    def get_and_validate_package_ids(node, path):
+        r = cluster.get(node=node, path=path)
         assert r.status_code == 200
         package_ids = r.json()
         assert isinstance(package_ids, list)
         for package_id in package_ids:
-            r = cluster.get(node=node, path=uri + package_id)
+            r = cluster.get(node=node, path=path + package_id)
             assert r.status_code == 200
             name, version = package_id.split('--')
             assert r.json() == {'id': package_id, 'name': name, 'version': version}
--- a/test_util/cluster_api.py
+++ b/test_util/cluster_api.py
@@ -1,4 +1,5 @@
 import copy
+import functools
 import logging
 import os
 from urllib.parse import urlparse
@@ -10,6 +11,10 @@ import retrying
 
 import test_util.helpers
 import test_util.marathon
+
+ADMINROUTER_PORT_MAPPING = {
+    'master': {'http': 80, 'https': 443},
+    'agent': {'http': 61001, 'https': 61002}}
 
 
 def get_args_from_env():
@@ -30,7 +35,7 @@ def get_args_from_env():
     assert os.environ['DCOS_PROVIDER'] in ['onprem', 'aws', 'azure']
 
     return {
-        'dcos_uri': os.environ['DCOS_DNS_ADDRESS'],
+        'dcos_url': os.environ['DCOS_DNS_ADDRESS'],
         'masters': os.environ['MASTER_HOSTS'].split(','),
         'public_masters': os.environ['PUBLIC_MASTER_HOSTS'].split(','),
         'slaves': os.environ['SLAVE_HOSTS'].split(','),
@@ -223,14 +228,15 @@ class ClusterApi(test_util.helpers.ApiClient):
         will raise exception if authorization fails
         """
         self.web_auth_default_user.authenticate(self)
+        self.default_headers.update(self.web_auth_default_user.auth_header)
 
-    def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
+    def __init__(self, dcos_url, masters, public_masters, slaves, public_slaves,
                  dns_search_set, provider, auth_enabled, default_os_user,
                  web_auth_default_user=None, ca_cert_path=None):
         """Proxy class for DC/OS clusters.
 
         Args:
-            dcos_uri: address for the DC/OS web UI.
+            dcos_url: address for the DC/OS web UI.
             masters: list of Mesos master advertised IP addresses.
             public_masters: list of Mesos master IP addresses routable from
                 the local host.
@@ -244,9 +250,17 @@ class ClusterApi(test_util.helpers.ApiClient):
                 Note: user must be authenticated explicitly or call self.wait_for_dcos()
             ca_cert_path: (str) optional path point to the CA cert to make requests against
         """
+        # URL must include scheme
+        assert dcos_url.startswith('http')
+        parse_result = urlparse(dcos_url)
+        self.scheme = parse_result.scheme
+        self.dns_host = parse_result.netloc.split(':')[0]
+
+        # Make URL never end with /
+        self.dcos_url = dcos_url.rstrip('/')
+
         super().__init__(
-            cluster=self,
-            user=web_auth_default_user,
+            default_host_url=self.dcos_url,
             api_base=None,
             ca_cert_path=ca_cert_path)
         self.masters = sorted(masters)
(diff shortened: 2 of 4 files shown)


mellenburg:

test_util/cluster_api.py
```diff
@@ -244,9 +249,17 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
                 Note: user must be authenticated explicitly or call self.wait_for_dcos()
             ca_cert_path: (str) optional path point to the CA cert to make requests against
         """
+        # URI must include scheme
```
jgehrcke: In our case, that URI will _always_ be a URL. That's why I liked calling it a URL :-).

The fact that `URI must include scheme` actually is an artifact of this.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -142,13 +142,14 @@ def test_octarine_srv(cluster, timeout=30):
 
 def test_pkgpanda_api(cluster):
 
-    def get_and_validate_package_ids(node, uri):
-        r = cluster.get(node=node, path=uri)
+    def get_and_validate_package_ids(node, path):
```
cmaloney: :+1: to the function signature change

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -142,13 +142,14 @@ def test_octarine_srv(cluster, timeout=30):
...
+        node_url = cluster.get_node_url(node)
```
cmaloney: the two calls here are a considerable API degredation from previously. Maybe cluster() needs to override / inject into the ClusterApi() a wrapper which makes it so that it will take either "host_url", _or_ if given the keyword parameter `node=` it will calculate host_url from that (both given or neither given being an error).

test_util/cluster_api.py
```diff
@@ -263,39 +276,56 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+
     def get_client(self, path, default_headers=None):
-        return test_util.helpers.ApiClient(
+        return test_util.helpers.ApiClient.from_cluster(
             cluster=self,
-            user=self.web_auth_default_user,
             api_base=path,
-            ca_cert_path=self.ca_cert_path,
-            default_headers=default_headers)
+            default_headers=None)
```
cmaloney: why is this not passing through `self.default_headers if default_headers is None else default_headers`?

Lets people set them to empty if they want / need, provides the same responses they would have gotten by hand-prefixing their requests.

test_util/helpers.py
```diff
@@ -50,74 +49,50 @@ def authenticate(self, cluster):
...
         self.default_headers = default_headers
         self.ca_cert_path = ca_cert_path
 
-    def api_request(self, method, path, node=None, port=None, **kwargs):
+    @classmethod
+    def from_cluster(cls, cluster, api_base, default_headers=None):
+        if default_headers is None:
+            default_headers = {}
+        if cluster.auth_enabled and cluster.web_auth_default_user:
+            default_headers.update(cluster.web_auth_default_user.auth_header)
```
cmaloney: why not just copy over cluster.default_headers always?

test_util/helpers.py
```diff
@@ -50,74 +49,50 @@ def authenticate(self, cluster):
...
-            port: see get_url()
+            host_url: override the client's default host URL
             **kwargs: any optional arguments to be passed to requests.request
         """
         headers = copy.copy(self.default_headers)
-        if self.cluster.auth_enabled and self.cluster.web_auth_default_user:
-            headers.update(self.cluster.web_auth_default_user.auth_header)
 
-        if self.ca_cert_path:
+        if self.ca_cert_path and 'verify' not in kwargs:
```
cmaloney: why does this get the verify part added now?

test_util/marathon.py
```diff
@@ -9,7 +9,17 @@
 
 import test_util.helpers
 
+DEFAULT_API_BASE = '/marathon/v2'
```
cmaloney: a whole bunch of pods stuff is going to want slightly different stuff from /marathon/v2 I think (Although should confirm with that team).

test_util/marathon.py
```diff
@@ -9,7 +9,17 @@
...
 TEST_APP_NAME_FMT = '/integration-test-{}'
+REQUIRED_HEADERS = {'Accept': 'application/json, text/plain, */*'}
+
+
+def update_headers(headers):
+    if headers:
```
cmaloney: `if headers is None`? This code seems complicated for what it does. Why isn't just a `headers.update(REQUIRED_HEADERS)` good enough?

packages/dcos-integration-test/extra/test_applications.py
mellenburg: I thought wrapping a wrapper was bad but having to call cluster twice in one go is a little worse I now feel, so I will give it a shot

test_util/helpers.py
mellenburg: Confused myself. Just copying cluster.default_headers is the right thing to do

test_util/helpers.py
mellenburg: this allows people to make requests to ApiClient with verify=False or verify=my_other_custom_ca. Otherwise, the verify kwarg is always overridden, which makes for a bad requests wrapper

test_util/marathon.py
mellenburg: the point of this function is just to handle the repeated case where default_headers=None so you don't have to rewrite:
```
if default_headers is None:
    default_headers = {}
```
but its not much of a savings

test_util/marathon.py
mellenburg: Should there not be a default api version? I just figured the pods tests can then specifically ask for a marathon api with a different version

test_util/cluster_api.py
mellenburg: Definitely a bug. Thanks for catching that.

test_util/helpers.py
cmaloney: My point was it's an odd code change for _this_ PR, seems unrelated to the core goal of this PR and doesn't have a comment about why it changed in the code.

cmaloney: `# Set verify only if it isn't already set so callers can override the CA Cert`

test_util/marathon.py
cmaloney: I'd rather have the api base be `/marathon`, people are familiar with `v2/apps` and friends as the API endpoints they call, rather than just "apps".

test_util/helpers.py
```diff
@@ -48,85 +47,56 @@ def authenticate(self, cluster):
...
-        """
-        Args:
-            path: (str) URL path to request if self.api_base is set it will be prepended
-            node: (str) the hostname of the node to be requested from, if node=None,
-                then public cluster address (see environment DCOS_DNS_ADDRESS)
-            port: (int) port to be requested at. If port=None, the default port
-                for that given node type will be used
-        Returns:
-            fully-qualified URL string for this API
+    def wrapped_request(self, request_fn, *args, **kwargs):
```
mellenburg: @cmaloney What do you think about doing it this way?

test_util/cluster_api.py
```diff
@@ -263,39 +276,76 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
-            cluster=self,
-            ca_cert_path=self.ca_cert_path,
-            user=self.web_auth_default_user)
+        marathon_client = test_util.marathon.Marathon(
+            default_host_url=self.dcos_url,
+            default_os_user=self.default_os_user,
+            default_headers=self.default_headers,
+            ca_cert_path=self.ca_cert_path)
+        # give spawned clients the node= request option
+        marathon_client.request_wrapper = self.request_wrapper
```
mellenburg: @cmaloney having to manually attach a new wrapper is a little clunky, but it still seems better than getting another class

mellenburg:
@orsenthil PTAL

test_util/helpers.py
```diff
@@ -48,85 +47,56 @@ def authenticate(self, cluster):
...
-
-    get = functools.partialmethod(api_request, 'get')
-    post = functools.partialmethod(api_request, 'post')
-    put = functools.partialmethod(api_request, 'put')
-    delete = functools.partialmethod(api_request, 'delete')
-    options = functools.partialmethod(api_request, 'options')
-    head = functools.partialmethod(api_request, 'head')
-    patch = functools.partialmethod(api_request, 'patch')
-    delete = functools.partialmethod(api_request, 'delete')
+        return self.request_wrapper(request_fn, *args, **kwargs)
```
orsenthil: I reviewed this piece too. The change is easy to understand. So, this is fine with me.

test_util/cluster_api.py
```diff
@@ -263,39 +276,76 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+                for that given node type will be used
+        Returns:
+            fully-qualified URL string for this API
+        """
+        if node in self.masters:
+            role = 'master'
+        elif node in self.all_slaves:
+            role = 'agent'
+        else:
+            raise Exception('Node {} is not recognized within the DC/OS cluster'.format(node))
```
orsenthil: Specific exception instead of a generic one?

test_util/cluster_api.py
```diff
@@ -263,39 +276,76 @@ def __init__(self, dcos_uri, masters, public_masters, slaves, public_slaves,
...
+
+    def get_client(self, path):
+        new_client = test_util.helpers.ApiClient(
+            default_host_url=self.dcos_url,
             api_base=path,
             ca_cert_path=self.ca_cert_path,
-            default_headers=default_headers)
+            default_headers=self.default_headers)
+        # give spawned clients the node= reqest option
+        new_client.request_wrapper = self.request_wrapper
```
orsenthil: Is there better way for this? ( why we need to extend the client and a property attribute (line 341)with this additional method?

(this part is slightly confusing and bit difficult to determine the flow explicitly. The comments (which give details  about business logic ( something about node= options) is not indicative either).

test_util/cluster_api.py
mellenburg: Which specific one would you prefer? I prefer generic Exceptions with specific messages until we actually are trying to catch the exception.

test_util/cluster_api.py
mellenburg: The new_client has its own self.request wrapper from ApiClient, however, when we spawn this from a ClusterApi, we want the request signature to match ClusterApi. The difference in the two is that in ClusterApi, our request wrapper checks for the keyword 'node' and can translate that to a host_url before just calling out the ApiClient's request_wrapper. By setting the new_client request wrapper to be ClusterApi's request wrapper, they too can specify a node.

Does that explanation suffice in comment form or is there something you can suggest to restructure? I tried a few approaches and this is the cleanest yet

PR #955, opened 2016-11-14
Fix old pep8 errors

mellenburg:
PyCodeStyle got updated:
https://github.com/PyCQA/pycodestyle/blob/master/CHANGES.txt#L24
causing:
https://teamcity.mesosphere.io/viewLog.html?buildId=473704&buildTypeId=DcosIo_Dcos_CodeQuality&tab=buildResultsDiv


tox.ini
```diff
@@ -30,6 +30,7 @@ deps =
   flake8-import-order==0.9.2
   pep8-naming
   teamcity-messages
+  pycodestyle==2.1.0
```
orsenthil: 👍  for pinning the versions.

In a separate PR, the other dependencies can be pinned to specific versions.

spahl:
:ship:

PR #956, opened 2016-11-15
Remove the providers switch in gen/calc.py

cmaloney:
This removes the `providers` switch in gen/calc.py, pushing all the data which
was somewhat hackishly combined there down into the correct installer components
using the "Source" concept of gen to provide the values. This means moving onprem,
AWS, and Azure to use "Source" directly and add the source to the set of sources
when calling gen.generate rather than setting a magic key by hand and relying
on that to "imply" the right configuration.

This should provide better error messages for all cases: There are less bad
paths which we can tell the user about for possibly setting configuration. In
addition, it makes lite binding configuration easier to implement, as all the
pieces start being more "bound" to the one particular installer they effect.

From my point of view this also makes the code easier to reason about than the
previous nested conditionals which selected which options to use.

The patchset also removes manual argument combining from aws and azure template generation, instead teaching them to pass around sources, and just add new sources with more arguments.

This makes those template generators easier to work with, and harder to mess up, as it's more directly obvious what data came from where, and it's impossible to accidentally overwrite a variable.

That prepares do_create to be able to switch from  being given variant_arguments as a parameter to instead taking variant_sources, the advantage then being that from top to bottom things don't modify the arguments in place, it's obvious where various values came from in the composition, and code is generally simplified.

Unfortunately do_create() can't be migrated currently because cloudformation_s3_url inspects the arguments in a weird way. That is entirely fixable, but a different sort of work than this patchset (Spreading Sources / removing manual argument dictionaries).

Also includes some drive-by fixes, see the individual commits for details.

# Issues

# Checklist


orsenthil:
Review comments:

* How was this tested and validated?
* The test changes as see cover `onprem_source`, but the code makes multiple other changes too. (even internals.py)
* Does this change warrant a manual testing and has that be done to verify? 
* I have no "code" related comments. Those look good to me.

cmaloney:
- Validation: The fact that the integration tests run and still generate / install a DC/OS that functions. We test all these installer codepaths on every single PR
 - There are lots of individual commits in the PR. Each one is atomic, some are really, really tiny. Their individual commit messages talk about what is happening in each commit. The gen.internals changes for instance: https://github.com/dcos/dcos/pull/956/commits/b65552be447019fd9c62a00d915a8e9986cc67f5, https://github.com/dcos/dcos/pull/956/commits/6b09153ef97f8728ba42c0b4c0247cc971a01bb6
 - As far as hand validation, I'm pretty sure the automated validation is good enough, as it tests every code path this touches, combined with verifying through code review that sources all have _exactly_ the same config inside them as before I think there is sufficient evidence to tell if this is working or not.

cmaloney:
@orsenthil updated, please take another pass

gen/build_deploy/aws.py
```diff
@@ -534,31 +596,23 @@ def do_create(tag, build_name, reproducible_artifact_path, commit, variant_argum
 
     for bootstrap_variant, variant_base_args in variant_arguments.items():
         # Setup base arguments
```
branden: This comment can be removed.

gen/__init__.py
```diff
@@ -415,18 +423,24 @@ def add_builtin(name, value):
     temporary_str = 'DO NOT USE THIS AS AN ARGUMENT TO OTHER ARGUMENTS. IT IS TEMPORARY'
     add_builtin('expanded_config', temporary_str)
 
-    return [base_source], targets, templates
+    # Note: must come last so the hash of the "base_source" this is beign added to contains all the
+    # variables but this.
+    add_builtin('sources_id', hash_checkout([hash_checkout(source.make_id()) for source in sources]))
```
branden: This could be hidden inside a `Source.finalize()` method. That method could set a flag that enforces no added variables after it's been called.

gen/build_deploy/aws.py
```diff
@@ -485,10 +485,7 @@ def gen_templates(arguments):
     with logger.scope("Validating CloudFormation"):
         validate_cf(cloudformation)
 
-    return gen.Bunch({
-        'cloudformation': cloudformation,
-        'results': results
-    })
+    return ResultTuple(cloudformation, results)
```
cmaloney: I like using `.` to lookup guaranteed members rather than doing the dictionary indexing. `_as_artifact_and_pkg` also currently requires the API of having the `.` syntax, which we can change but I'd like to not pull into this PR

branden: This PR changes `gen.build_deploy.azure.gen_templates()` to return a dict and changes its member lookup from `.` to `[]` (https://github.com/dcos/dcos/pull/956/files#diff-f4dc0083a55e6bc92a7e71ea4bd8d6e6R252). Make it return a `NamedTuple` for consistency?

cmaloney: https://github.com/dcos/dcos/pull/1013

PR #957, opened 2016-11-15
Switch remaining pieces to Python 3.5

cmaloney:
Switches build and unit tests to run in python 3.5 rather than python 3.4, unifying all dc/os internals and build to always be python 3.5.

This PR should help flush out remaining things that break.

# Issues

# Checklist


orsenthil:
The create-release-pr error is genuine.

```
venv: error: unrecognized arguments: --always-copy
```

After that  is fixed , this change looks good to me.

PR #958, opened 2016-11-15
gen: Add the concept of a Resolvable, Resolve config per-Target

cmaloney:
- Targets now get finalized with results. This also means they need to be constructued per-use.
 - validate_configuration() no longer exists, as resolve_configuration does exactly the same thing,
   users of validate_configuration() are migrated to just use resolve_configuration() once.
 - The per-target results aren't yet plugged into various places, that will come in a later pr
 - Introduce a notion of variables "visited" to calculate a particular variable.
 - Introduce a concept of a "Resolvable". A resolvable tracks the resolution state of a particular configuration value. (Unresolved, being resolved, error). It will eventually also include a "late resolution" (late binding) state. The introduction centralizes the logic for tracking that, and makes it harder to get wrong. Resolvables also keep track of what caused them to come into existence / get resolved.
 - Switches / sub-scopes now generate single-argument "validate" functions rather than having custom logic to validate their values when we reached the switch they were used in (See yield_implicit_validates)
 - Renames DFSArgumentCalculator to Resolver. A resolver is now only usable once, takes a set of setters (taken from a set of sources) and targets, then resolves them all at once. The goal being that the resolution of all targets produces a single, consistent / valid against eachother full final configuration. Resolver now owns / builds the status dictionary itself, and makes it accessible to callers
 - Adds a new unit test for gen internals code

Things this works towards enabling
 - Typed parameters. "Resolvables" can be focused around a type (as discovered by python type annotations on function parameters)
 - "nested" configuration. Currently all config must be one top level dictionary. In theory though we could use nested dictionaries and flatten them ({"a": {"b": {"c": "d"}}} could be accessed as `a_b_c`.
 - Late-binding. Resolvables do a much better job of tracking / keeping contained the resolution state, code which was currently spread about DFSArgumentCalculator previously. The code to track "unset" and spread it through the tree, is very similar to the code needed to track "late bound" and that property spreading / chaining through the tree.
 - "secret" / private data such as aws keys in config.yaml. Things like the AWS keys should be explicitly protected / marked so they don't show up in things like the user config.yaml inside the cluster / on every host. If anything would make them show up, that should result in an error.

# Issues

 - Late-binding configuration
 - Secrets in dcos-release.config.yaml vs. config.yaml for AWS Advanced templates
 - Enabling dcos-launch to use config.yaml

# Checklist

### DIFF
--- a/gen/internals.py
+++ b/gen/internals.py
@@ -1,10 +1,11 @@
 import copy
 import inspect
 import logging
-from functools import partialmethod
+from contextlib import contextmanager
+from functools import partial, partialmethod
 
 from gen.exceptions import ValidationError
-from pkgpanda.util import json_prettyprint
+
 
 log = logging.getLogger(__name__)
 
@@ -26,11 +27,22 @@ def validate_arguments_strings(arguments: dict):
         raise ValidationError(errors, set())
 
 
+# TODO (cmaloney): Python 3.5, add checking valid_values is Iterable[str]
+def validate_one_of(val: str, valid_values) -> None:
+    """Test if object `val` is a member of container `valid_values`.
+    Raise a AssertionError if it is not a member. The exception message contains
+    both, the representation (__repr__) of `val` as well as the representation
+    of all items in `valid_values`.
+    """
+    if val not in valid_values:
+        options_string = ', '.join("'{}'".format(v) for v in valid_values)
+        raise AssertionError("Must be one of {}. Got '{}'.".format(options_string, val))
+
+
 class Setter:
 
     # NOTE: value may either be a function or a string.
-    def __init__(self, name, value, is_optional, conditions, is_user):
-        assert isinstance(conditions, list)
+    def __init__(self, name: str, value, is_optional: bool, conditions: list, is_user: bool):
         self.name = name
         self.is_optional = is_optional
         self.conditions = conditions
@@ -97,6 +109,7 @@ class Target:
     def __init__(self, variables=None, sub_scopes=None):
         self.variables = variables if variables else set()
         self.sub_scopes = sub_scopes if sub_scopes else dict()
+        self._arguments = None
 
     def add_variable(self, variable: str):
         self.variables.add(variable)
@@ -106,6 +119,30 @@ class Target:
             self.sub_scopes[scope.name] += scope
         else:
             self.sub_scopes[scope.name] = scope
+
+    def finalize(self, arguments: dict):
+        assert self._arguments is None, "finalize should only be called once. If it was called " \
+            "more than once likely some code is re-using a target rather than creating a new " \
+            "instance for every resolve_configuration() call"
+
+        # TODO(cmaloney): Walk the tree of things that every argument is
+        # dependent upon to get the "full" set of arguments.
+        self._arguments = arguments
+
+    @property
+    def arguments(self):
+        assert self._arguments is not None, "Must only be called after finalize()"
+        return self._arguments
+
+    def yield_validates(self):
+        # Recursively walk the target / sub scope tree and yield a
+        # validate function for each and every switch.
+        for name, scope in self.sub_scopes.items():
+            # Note: It's important scope.cases.keys() is evaluated now, not later
+            # (is evaluated before the validate is called)
+            yield name, partial(validate_one_of, valid_values=scope.cases.keys())
+            for sub_target in scope.cases.values():
+                yield from sub_target.yield_validates()
 
     def __iadd__(self, other):
         assert isinstance(other, Target), "Internal consistency error, expected Target but got {}".format(type(other))
@@ -129,7 +166,7 @@ class Target:
 
 
 class Source:
-    def __init__(self, entry=None, is_user=False,):
+    def __init__(self, entry=None, is_user=False):
         self.setters = dict()
         self.validate = list()
         self.is_user = is_user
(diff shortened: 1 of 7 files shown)


cmaloney:
this is ready for review

gen/internals.py
```diff
@@ -184,28 +212,131 @@ def add_entry(self, entry, replace_existing):
...
+        self._get_stack = get_stack
+        self._finalized = False
+
+    def __missing__(self, key):
+        assert not self._finalized, "No missing keys should be accessed"
+
+        value = self[key] = Resolvable(key, self._get_stack())
+        return value
+
+    def finalize(self):
```
cmaloney: The goal would be not to act like a `dict` but rather to not be able to add or remove things. Effectively become a `frozenset` of arguments. Unfortunately python doesn't allow set / frozenset to have the key not match the whole object identity...

gen/internals.py
```diff
@@ -237,138 +372,234 @@ def all_conditions_met(setter):
...
+            # only ever has one value / resolvable.
+            finalized_arguments.update(sub_target.arguments)
+
+        target.finalize(finalized_arguments)
+
+    # Force calculation of all arguments by accessing the arguments in this
+    # scope and recursively all sub-scopes.
+    def resolve(self):
+        assert not self._resolved, "Resolvers should only be resolved once"
+        self._resolved = True
```
cmaloney: So there is a function which looks just like that: https://github.com/dcos/dcos/blob/master/gen/internals.py#L374. The "Resolver" class is an internal implementation detail of it. No one other than that one function, resolve_configuration, should ever see / deal with an unresolved Resolver class.

as far as why a class: I think it makes a lot easier to track the common state, as well as environment (things like ArgumentStack, context, and all the validation functions). They all are located together, and easily accessible by the members.

The resolver class once resolved, then provides a very precise interface to retrieve how the resolution ran (the status_dict), as well as all arguments which were calculated (.arguments), and ideally even indexing into the targets it was given eventually (currently it's easier for callers to just look at the resolver objects they passed in because the python objects are passed by reference. Eventually I want to attach names to the targets, which should make for better error messages).

Another advantage of the class, is that when something goes wrong, all the state of the world is centralized, which makes it a lot easier to print out what that state is, rather than having multiple inter-dependent objects, or trying to track / pass it through as multiple function arguments (or even one big function argument passed to a half dozen functions). This also means that when we find configuration errors (things like the unset configuration, or when a validation function fails), that all the state / world is available and around

The class also takes multiple things in, and gives multiple things out. The goal is that the resolver has a simple API to always use it. It keeps track of partially complete state (If an error happens halfway through resolving, it could be inspected / probed because all that state is there). When there are issues in resolving configuration, the bits are always all there. This in particular I think will be helpful to resolve: https://github.com/dcos/dcos/blob/master/dcos_installer/async_server.py#L116. Doing that requires we keep track of what caused what, and can always trace back to the individual fields present in the configuration ui of the web installer.

Arguably, could make the resolve() call part of the constructor, which would give you more of the API you're asking about, but from my perspective it is an important different change to the class from just setting up the environment required for resolving.

gen/internals.py
```diff
@@ -418,49 +649,19 @@ def resolve_configuration(sources: list, targets: list, user_arguments: dict):
...
-            assert isinstance(key, str)
-            messages[key] = {'message': msg}
-
-        for keys, msg in to_do.items():
-            for name in keys:
-                # Skip ones we already have one message for.
-                if name in messages:
-                    continue
+    resolver = Resolver(setters, validate, targets)
+    resolver.resolve()
```
cmaloney: Talked about this in my above comment.

Practically, I think it makes the code easier to understand / reason about rather than having the one uber function. There is a setup that builds up the state, then a call to a recursive function which uses all that common state to do it's work. Unresolved Resolver() is not an external API. Having a constructor / __init__ function do a lot of work is a bit odd unless it's a Resource Acquisition Is Initialization function (Think a class that holds an fd and wants to guarantee if you have an instance of that class, you have an open fd, and when that class destructs, the fd will always be closed)

gen/internals.py
```diff
@@ -184,28 +212,131 @@ def add_entry(self, entry, replace_existing):
...
+    # validate function for each and every switch.
+    for name, scope in target.sub_scopes.items():
+        # Note: It's important scope.cases.keys() is evaluated now, not later
+        # (is evaluated before the validate is called)
+        yield name, partial(validate_one_of, valid_values=scope.cases.keys())
+        for sub_target in scope.cases.values():
+            yield from yield_implicit_validates(sub_target)
+
+
+class ArgumentSet(dict):
```
cmaloney: Practically this is trying _really_ hard to act more like a set than a dictionary (The key and the value have the same identity always, the key's identity is derived from the value being held). If I could I'd use a custom key function and wrap a set, but unfortunately python doesn't allow / have a way to do that.

gen/internals.py
```diff
@@ -184,28 +212,131 @@ def add_entry(self, entry, replace_existing):
...
+        assert self._state == 'resolved', "is_resolved() must return true for this function to " \
+            "be called. State: {}, is_resolved: {}".format(self._state, self.is_resolved)
+
+        return self._value
+
+    def __str__(self):
+        return "<Resolvable name: {}, state: {}>".format(self.name, self._state)
+
+
+def yield_implicit_validates(target):
```
cmaloney: Moving to be a member function but keeping the same name. Not having to hand-aggregate the recursion / trampoline up lists because yield will do it for us is really handy

cmaloney:
@branden pulled validation out into a class in a second commit, adds a bit of lines of code and the API is a bit wonky (The multi-arg validate results have to be yielded out while the single-arg ones don't). Not sure if it's worth doing at this point.

gen/internals.py
```diff
@@ -184,51 +221,203 @@ def add_entry(self, entry, replace_existing):
...
+                # validate things that are never used or given)
+                if parameter not in arguments:
+                    skip = True
+                    break
+
+                # Exit if the parameter resolved to an error
+                resolvable = arguments[parameter]
+                if resolvable.is_error:
+                    skip = True
+                    break
```
branden: This reads more easily when condensed, since both checks have the same outcome and contain a jump.

```
# Skip running these validation functions if a required parameter wasn't provided or resolved to an error.
if parameter not in arguments or arguments[parameter].is_error:
    skip = True
    break
```

branden: We could also get rid of this inner `for` loop to allow a direct `continue` on the outer loop.

```
for parameter_set, validate_fns in self._multi_arg_validate.items():
    # Skip running these validation functions if a required parameter wasn't provided or resolved to an error.
    if any(p not in arguments or arguments[p].is_error for p in parameter_set):
        continue
    kwargs = {p: arguments[p].value for p in parameter_set}
```

gen/internals.py
```diff
@@ -184,51 +221,203 @@ def add_entry(self, entry, replace_existing):
...
+        validate_fns = self._validate_by_arg.get(name)
+        if validate_fns is not None:
+            for validate_fn in validate_fns:
+                validate_fn(value)
+
+    # TODO(cmaloney): The distance between the validate_single and multi_arg_validate interface,
+    # while necessary for efficient functioning currently, is showing that there is tension between
+    # how / when each is run. Moving the multi-argument validation to as soon as possible after an
+    # argument set is finalized rather than one big pass at the end would likely make it much
+    # cleaner.
```
cmaloney: The problem with that is we need to do single-argument validate functions as soon as possible after a value is calculated / before we .resolve() a Resolvable with it. 

That's necessary, so that invalid values never are accessible for other calculation functions or validation functions, where their invalid / incorrect value may cause issues (Code that takes resolvers among other things should be able to assume resolvers is a json list).

Ideally, as soon as we had all the arguments for a multi-arg validate, we would check that as well. Unfortunately, that is more complicated code. The simpler code to do it for multi-argument validates would be to basically say
 - Every time an argument has a new value tell validator about it
 - Have validator iterate through every validation function which takes the new argument as one of it's parameters and if all the parameters are set, then validate it.
 - Call the validate functions, sorting the single-argument validate functions before the multi-argument validate functions with the same signature.

Doing that though I think is out of scope for just what this code is doing, definitely adds more complexity / logic. The current evaluation method is not that complicated: Single arg ones are validated instantly, which is pretty simple to do. If a single-arg validate finds a value to be invalid, the argument won't show up in the multi_argument_validate argument dict as resolved and hence multi-arg validates won't get called

Multi-arg ones are validated at the end of resolution, where we simply iterate once over all the multi-argument validate functions, checking if the validate function arguments are a subset of all the arguments, and if so calling it and yielding out the errors.

gen/internals.py
cmaloney: this is changing this logic, which is extending the scope of the code changed here, was trying to avoid changing this logic as much as possible because that means more things which could break inside this change / more diff to read / evaluate.

branden: 👍 I didn't realize this code wasn't new.

PR #960, opened 2016-11-16
Add test for hash_files_in_folder, fix several usages with absolute paths in codebase

cmaloney:
Add test for hash_files_in_folder, fix several usages with absolute paths in codebase
    
Absolute paths with hash_folder resulted in a non-reproducible hash since the absolute path changes from host to host (where the checkout is changes). This makes hash_folder assertion fail if it's given an absolute path, and gives a helper function which can be used to make absolute paths into relative paths by changing into the directory before running hash_folder

# Issues

[DCOS-275](https://dcosjira.atlassian.net/browse/DCOS-275)

# Checklist

### DIFF
--- a/pkgpanda/build/__init__.py
+++ b/pkgpanda/build/__init__.py
@@ -7,7 +7,8 @@ import random
 import shutil
 import string
 import tempfile
-from os import mkdir
+from contextlib import contextmanager
+from os import chdir, getcwd, mkdir
 from os.path import exists
 from subprocess import CalledProcessError, check_call, check_output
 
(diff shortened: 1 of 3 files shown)


cmaloney:
@vespian can you review / take a look? Test we talked about long ago. Doesn't include things like symlink checking currently, but we currently don't have any symlinks in the repository other than inside the pkgpanda tests, so I think that is fine for now.

pkgpanda/build/__init__.py
```diff
@@ -449,13 +450,52 @@ def hash_checkout(item):
...
+def hash_files_in_folder(directory):
+    assert not directory.startswith('/'), \
+        "For the hash to be reproducible on other machines relative paths must always be used. " \
+        "Got path: {}".format(directory)
+    directory = directory.rstrip('/')
     file_hash_dict = {}
+    # TODO(cmaloney): Disallow symlinks as they're hard to hash, people can symlink / copy in their
+    # build steps if needed.
     for root, dirs, filenames in os.walk(directory):
+        assert not root.startswith('/')
```
vespian: Just a nit: I am wondering if there are os.walk implementations that will give you absolute paths for non-absolute argument.

pkgpanda/build/__init__.py
```diff
@@ -449,13 +450,52 @@ def hash_checkout(item):
...
         for name in filenames:
-            filename = root + '/' + name
-            file_hash_dict[filename] = pkgpanda.util.sha1(filename)
-    return hash_checkout(file_hash_dict)
+            path = root + '/' + name
```
vespian: `os.path.join()` ?

pkgpanda/build/__init__.py
```diff
@@ -449,13 +450,52 @@ def hash_checkout(item):
...
+            base = path[len(directory) + 1:]
+            file_hash_dict[base] = pkgpanda.util.sha1(path)
+
+        # If the directory has files inside of it, then it'll be picked up implicitly. by the files
+        # or folders inside of it. If it contains nothing, it wouldn't be picked up but the existence
+        # is important, so added it with a value for it's hash not-makeable via sha1 (empty string).
```
vespian: :100:

pkgpanda/build/test_util.py
```diff
@@ -0,0 +1,86 @@
+import pkgpanda.build
+
+
+def test_hash_files_in_folder(tmpdir):
+    hash_files_in_folder = pkgpanda.build.hash_files_in_folder
+
+    with tmpdir.as_cwd():
+        # Empty folder
```
vespian: Nit: It's just my opinion, but generally I am a big fan of small independent tests instead of "scenario" tests with intermediate steps depending on each others. I would have probably split this one into at least 5 separate tests.

pkgpanda/build/__init__.py
```diff
@@ -449,13 +450,52 @@ def hash_checkout(item):
         raise NotImplementedError("{} of type {}".format(item, type(item)))
 
 
-def hash_folder(directory):
+def hash_files_in_folder(directory):
```
vespian: Docstring ?

pkgpanda/build/__init__.py
cmaloney: Unfortunately os.path.join has some odd behavior at times. Earlier we rstrip() `/` from directory/root so this should always be safe.

pkgpanda/build/test_util.py
cmaloney: from my perspective, if the first test fails, _all_ the other ones are guaranteed to fail, and seeing their independent failure state will not provide useful help when debugging.

There would be some utility in making it easier to hand-run the different tests one by one for small iteration, but practically this is doing a "building up" / exercising more parts from case to case, hence why I made it one test.

Independent things get independent tests. 100 tests which will all fail because of the same single bug from my perspective produce a lot more noise than 1 test which builds up.

vespian: Nit cntd.

https://pragprog.com/magazines/2012-01/unit-tests-are-first -> check section "I for isolated"
http://ricardogeek.com/docs/clean_code.pdf -> "Unit tests" chapter if you are have time for a longer read.

cmaloney: Ii have disagreements with that FIRST document, haven't read the other one yet. I'd also note it's arguing that you should only test a _single_ behavior inside of a single test.

In this case, we're testing one behavior, and each test adds some new exploration of that feature / corner.

I don't think the code would be better by me copy/pasting or writing more test helper functions to split it into chunks, and it wouldn't produce more useful output to engineers who later work on the code. It would add more volume to the test, and more complexity / pieces which must be simultaneously comprehended to understand the test. It wouldn't make debugging the test particularly simpler / quicker, since the test is already very fast, and it demonstrably builds from one step to the next.

vespian: Feel free to merge - it's not something that I think it's important enough to discuss endlessly and die for if necessary ;)

I was following these guidelines in i.e. Datastore tests - feel free to have a look and see if this is more readable for you: https://github.com/mesosphere/bouncer/blob/master/tests/datastore/test_datastore.py#L278

PR #962, opened 2016-11-16
Move missed tests from tests/ directory

cmaloney:
There were two missed tests in #938. One gets folded into dcos_installer/test_async_server.py (it was small). Other file gets moved to where it should have been originally.

# Issues

# Checklist


amitaekbote:
lgtm

PR #964, opened 2016-11-17
pinned versions for dependencies.

orsenthil:
- Pins down the version dependencies.
 - This prevents future bugs when things break under us.

# Issues

[DCOS-523](https://dcosjira.atlassian.net/browse/DCOS-523)

If there any comments on the motivation, please discuss it in the DCOS-523


cmaloney:
Please validate these all exactly match what all the packages where we bundle them use.

orsenthil:
The way these were generated was.

* Do the local build of DCOS on a new VirtualEnv.
* Pick up the packages from setup.py and write them into requirements.txt exactly as given in setup.py
* Create a new Virtualenv with package using requirements.txt
* `pip freeze` that for the versions and update the setup.py

If anything further needs to be done, let me know.

cmaloney:
Inside of https://github.com/dcos/dcos/tree/master/packages we have many of these packages for what we're actually shipping in DC/OS, and they do download very specific versions. Those should be the versions setup.py pins to if we want to pin, although I don't think it's a good use of the time currently. 

It doesn't provide a lot of value, it does make more work, and more pieces we have to maintain. Someone needs to maintain and operate the pypi-notifier. It needs to be verified regularly the notifier is still functioning. The notifier needs to be updated / maintained over time.

Right now, the breakage we see ocassionally is that the code style stuff gets slight changes, and when that happens we have two options. If it's hard to fix, we pin and file a JIRA. It gets into the workflow as a high-priority task we will resolve, and it gets resolved quickly. If it's easy to fix we do that, there's a PR for it. Both of those have a turnaround time of about an hour or less from first problem seen to landing.

orsenthil:
Yeah, I agree that the value add is not enormous.
I will try to setup https://github.com/cenkalti/pypi-notifier internally and  see if that report is beneficial. Let this PR stay open till the pypi-notifier is setup.

jgehrcke:
Still of relevance, I suppose?

orsenthil:
This is really stale PR. Closing this. The underlying problem of unpinned dependency is still relevant and we will tackle in a separate PR.

PR #967, opened 2016-11-17
DCOS-524:install-prereqs actually doesn't disable SElinux on CentOS

xinxian0458:
 - Disable SElinux on CentOS7 when executing `sudo bash dcos_generate_config.sh --install-prereqs`
 - It can disable SElinux correctly on CentOS7
 - sed -i will break hard or symbolic links when -i is used on such a file, should add --follow-symlinks option. (https://www.gnu.org/software/sed/manual/sed.html)

# Issues

[DCOS-524](https://dcosjira.atlassian.net/browse/DCOS-524)
...

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

sed -i will break the link between /etc/sysconfig/selinux and /etc/selinux/config.

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -363,7 +363,7 @@ if [ -f /opt/dcos-prereqs.installed ]; then
 fi
 
 sudo setenforce 0 && \
-sudo sed -i 's/^SELINUX=.*/SELINUX=disabled/g' /etc/sysconfig/selinux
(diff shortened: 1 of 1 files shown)


cmaloney:
@xinxian0458 can you squash the two commits into one commit? There's only one change here (so should be one commit). With that this should be able to land.

xinxian0458:
@cmaloney I finished it.

cmaloney:
:ship:

this should get pulled into a Train and land later today

PR #968, opened 2016-11-17
WIP: Greg/pods strict test failure

greggomann:
This PR is being used to test a pytest expected failure before merging it into the pods branch.

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -1,3 +1,4 @@
+import pytest
 import uuid
 
 from test_util.marathon import get_test_app, get_test_app_in_docker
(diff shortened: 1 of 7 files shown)


packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -59,6 +65,37 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(cluster):
     cluster.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
+def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(cluster):
```
greggomann: Hey @mellenburg, this test is a bit tricky because when it gets run as part of the Enterprise integration tests, we expect it to fail only in strict security mode. I'm trying to figure out how we can specify that expected failure here, or perhaps there is a way to override this test in Enterprise? Any thoughts?

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -29,6 +30,11 @@ def test_if_docker_app_can_be_deployed(cluster):
     cluster.marathon.deploy_test_app_and_check(*get_test_app_in_docker(ip_per_container=False))
 
 
+@pytest.mark.xfail(
+    dcos.config['security'] == 'strict',
```
greggomann: Unfortunately it looks like we don't have the `dcostests` module in this repo.

mellenburg: Please never use dcostests in anything new. I have duplicated all functionality of that code into the harness in this repo.

PR #969, opened 2016-11-18
Generate node upgrade script

amitaekbote:
 Introducing a new upgrade script using pkgpanda commands. More details at 
https://mesosphere.atlassian.net/browse/DCOS-11435
https://mesosphere.atlassian.net/browse/DCOS-11437
# Issues

[DCOS-11435](https://mesosphere.atlassian.net/browse/DCOS-11435)
...

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

### DIFF
--- a/dcos_installer/backend.py
+++ b/dcos_installer/backend.py
@@ -13,7 +13,7 @@ import gen.calc
 import release
 import release.storage.aws
 import release.storage.local
-from dcos_installer import config_util
+from dcos_installer import config_util, upgrade
 from dcos_installer.config import Config, normalize_config_validation
 from dcos_installer.constants import CONFIG_PATH, GENCONF_DIR
 
@@ -25,6 +25,17 @@ def print_messages(messages):
         log.error('{}: {}'.format(key, error))
 
 
+def validate_gen(config):
+    """ Returns True on success
+    """
+    validate = config.do_validate(include_ssh=False)
+    if len(validate) > 0:
+        for key, error in validate.items():
+            log.error('{}: {}'.format(key, error))
+        return False
+    return True
+
+
 def do_configure(config_path=CONFIG_PATH):
     """Returns error code
 
@@ -33,13 +44,32 @@ def do_configure(config_path=CONFIG_PATH):
     """
     config = Config(config_path)
 
-    validate_gen = config.do_validate(include_ssh=False)
-    if len(validate_gen) > 0:
-        for key, error in validate_gen.items():
-            log.error('{}: {}'.format(key, error))
+    validation = validate_gen(config)
+
+    if not validation:
         return 1
 
     config_util.do_configure(config)
+
+    return 0
+
+
+def generate_node_upgrade_script(installed_cluster_version, config_path=CONFIG_PATH):
+
+    if installed_cluster_version is None:
+        print('Must provide the version of the cluster upgrading from')
+        return 1
+
+    config = Config(config_path)
+    validation = validate_gen(config)
+    if not validation:
+        return 1
+    gen_out = config_util.onprem_generate(config)
+    config_util.make_serve_dir(gen_out)
+
+    # generate the upgrade script
+    upgrade.generate_node_upgrade_script(gen_out, installed_cluster_version)
+
     return 0
 
 
--- a/dcos_installer/cli.py
+++ b/dcos_installer/cli.py
@@ -202,6 +202,10 @@ def dispatch(args):
         sys.stdout.buffer.write(byte_str + b'\n')
         sys.exit(0)
 
+    if args.action == 'generate-node-upgrade-script':
+        status = backend.generate_node_upgrade_script(args.installed_cluster_version)
+        sys.exit(status)
+
     if args.action in dispatch_dict_simple:
         action = dispatch_dict_simple[args.action]
         if action[1] is not None:
@@ -228,7 +232,7 @@ def dispatch(args):
     sys.exit(1)
 
 
-class PasswordAction(argparse.Action):
+class ArgsAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         assert self.option_strings[0][:2] == '--'
         setattr(namespace, 'action', self.option_strings[0][2:])
@@ -245,7 +249,7 @@ def get_argument_parser():
 
     mutual_exc.add_argument(
         '--hash-password',
-        action=PasswordAction,
+        action=ArgsAction,
         dest='password',
         metavar='password',
         nargs='?',
@@ -254,11 +258,20 @@ def get_argument_parser():
 
     mutual_exc.add_argument(
         '--set-superuser-password',
-        action=PasswordAction,
+        action=ArgsAction,
         metavar='password',
         dest='password',
         nargs='?',
         help='Hash the given password and store it as the superuser password in config.yaml'
+    )
+
+    mutual_exc.add_argument(
+        '--generate-node-upgrade-script',
+        action=ArgsAction,
+        metavar='installed_cluster_version',
+        dest='installed_cluster_version',
+        nargs='?',
+        help='Generate a script that upgrades DC/OS nodes running installed_cluster_version'
     )
 
     parser.add_argument(
--- a/dcos_installer/config_util.py
+++ b/dcos_installer/config_util.py
@@ -11,17 +11,23 @@ from dcos_installer.constants import BOOTSTRAP_DIR, CLUSTER_PACKAGES_PATH, SERVE
 log = logging.getLogger(__name__)
 
 
-def do_configure(config):
-    gen_out = gen.generate(config.as_gen_format(), extra_sources=[gen.build_deploy.bash.onprem_source])
+def onprem_generate(config):
+    return gen.generate(config.as_gen_format(), extra_sources=[gen.build_deploy.bash.onprem_source])
 
+
+def make_serve_dir(gen_out):
     subprocess.check_call(['mkdir', '-p', SERVE_DIR])
     gen.build_deploy.bash.generate(gen_out, SERVE_DIR)
 
     # Get bootstrap from artifacts
-    # TODO(cmaloney): Switch to use a local storage provider like do_aws_configure does.
     fetch_bootstrap(gen_out.arguments['bootstrap_id'])
     # Write some package metadata
     pkgpanda.util.write_json(CLUSTER_PACKAGES_PATH, gen_out.cluster_packages)
+
+
+def do_configure(config):
+    gen_out = onprem_generate(config)
+    make_serve_dir(gen_out)
 
 
 def do_move_atomic(src_dir, dest_dir, filenames):
--- a/dcos_installer/test_backend.py
+++ b/dcos_installer/test_backend.py
@@ -67,6 +67,37 @@ def test_set_superuser_password(tmpdir):
         assert passlib.hash.sha512_crypt.verify('foo', config['superuser_password_hash'])
 
 
+def test_generate_node_upgrade_script(tmpdir):
+    upgrade_config = """
+---
+# The name of your DC/OS cluster. Visable in the DC/OS user interface.
+cluster_name: 'DC/OS'
+master_discovery: static
+exhibitor_storage_backend: 'static'
+resolvers:
+- 8.8.8.8
+- 8.8.4.4
+ssh_port: 22
+process_timeout: 10000
+bootstrap_url: file:///opt/dcos_install_tmp
+master_list: ['10.0.0.1', '10.0.0.2', '10.0.0.5']
+"""
+
+    create_fake_build_artifacts(upgrade_config, tmpdir)
+
+    output = subprocess.check_output(['dcos_installer', '--generate-node-upgrade-script', 'fake'], cwd=str(tmpdir))
+    assert output.decode('utf-8').splitlines()[-1].split("Node upgrade script URL: ", 1)[1]\
+                                                  .endswith("dcos_node_upgrade.sh")
+
+    try:
+        subprocess.check_output(['dcos_installer', '--generate-node-upgrade-script'], cwd=str(tmpdir))
+    except subprocess.CalledProcessError as e:
+        print(e.output)
+        assert e.output.decode('ascii') == "Must provide the version of the cluster upgrading from\n"
+    else:
+        raise Exception("Test passed, this should not pass without specifying a version number")
+
+
 def test_version(monkeypatch):
     monkeypatch.setenv('BOOTSTRAP_VARIANT', 'some-variant')
     version_data = subprocess.check_output(['dcos_installer', '--version']).decode()
--- a/dcos_installer/upgrade.py
+++ b/dcos_installer/upgrade.py
@@ -1,0 +1,134 @@
+'''
+Generating node upgrade script
+'''
+
+import subprocess
+import uuid
+
+import gen.build_deploy.util as util
+import gen.calc
+import gen.template
+from dcos_installer.constants import SERVE_DIR
+from pkgpanda.util import write_string
+
+
+node_upgrade_template = """
+#!/bin/bash
+#
+# BASH script to upgrade DC/OS on a node
+#
+# Metadata:
+#   dcos image commit: {{ dcos_image_commit }}
+#   generation date: {{ generation_date }}
+# Copyright 2017 Mesosphere, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit -o nounset -o pipefail
+
+source /opt/mesosphere/environment.export
+
+# Check if this is a terminal, and if colors are supported, set some basic
+# colors for outputs
+if [ -t 1 ]; then
+    colors_supported=$(tput colors)
+    if [[ $colors_supported -ge 8 ]]; then
+    RED='\e[1;31m'
+    BOLD='\e[1m'
+    NORMAL='\e[0m'
+    fi
+fi
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root" 1>&2
+    exit 1
+fi
+
+
+# check for version of dc/os upgrading from
+version=`grep "version" /opt/mesosphere/etc/dcos-version.json | cut -d '"' -f 4`
+if [ $version != {{ installed_cluster_version }} ]; then
+   echo "ERROR: Expecting to upgrade DC/OS from {{ installed_cluster_version }} to {{ installer_version }}.\
+                Version found on node: $version"
+   exit 1
+fi
+
+echo "Upgrading DC/OS agent {{ installed_cluster_version }}  -> {{ installer_version }}"
+
+
+pkgpanda fetch --repository-url={{ bootstrap_url }} {{ cluster_packages }}
+
+
+pkgpanda activate --no-block {{ cluster_packages }}
+
+# check if we are on a master node
+if [ -f /etc/mesosphere/roles/master ]; then
+   # run exhibitor migration script here
+   until dcos-shell dcos-exhibitor-migrate-perform
+   do
+        status=$?
+        case $status in
+             1) echo "Waiting for exhibitor endpoint"
+                sleep 10
+                ;;
+             2) echo "Could not read from exhibitor"
+                echo "Contact Support"
+                exit
+                ;;
+             4) echo "Rolling update in progress"
+                exit
+                ;;
+             8) echo "At least one config value does not have the expected pre migration value,
+                      \ and automatic migration can not take place"
+                echo "Contact Support"
+                exit
+                ;;
+             16) echo "Attempting to start the rolling update failed due to a non 200 response from exhibitor"
+                 echo "Contact Support"
+                 exit
+                 ;;
+        esac
+   done
+fi
+
+"""
+
+
+def generate_node_upgrade_script(gen_out, installed_cluster_version, serve_dir=SERVE_DIR):
+
+    # installed_cluster_version: Current installed version on the cluster
+    # installer_version: Version we are upgrading to
+
+    bootstrap_url = gen_out.arguments['bootstrap_url']
+
+    installer_version = gen.calc.entry['must']['dcos_version']
+
+    package_list = ' '.join(package['id'] for package in gen_out.cluster_packages.values())
+
+    bash_script = gen.template.parse_str(node_upgrade_template).render({
+        'dcos_image_commit': util.dcos_image_commit,
+        'generation_date': util.template_generation_date,
+        'bootstrap_url': bootstrap_url,
+        'cluster_packages': package_list,
+        'installed_cluster_version': installed_cluster_version,
+        'installer_version': installer_version})
+
+    upgrade_script_path = '/upgrade/' + uuid.uuid4().hex
+
+    subprocess.check_call(['mkdir', '-p', serve_dir + upgrade_script_path])
+
+    write_string(serve_dir + upgrade_script_path + '/dcos_node_upgrade.sh', bash_script)
+
+    print("Node upgrade script URL: " + bootstrap_url + upgrade_script_path + '/dcos_node_upgrade.sh')
+
+    return 0
--- a/test_util/cluster.py
+++ b/test_util/cluster.py
@@ -320,18 +320,14 @@ def install_dcos(
         installer.postflight()
 
 
-def upgrade_dcos(cluster, installer_url, add_config_path=None):
+def upgrade_dcos(cluster, installer_url, version, add_config_path=None):
 
-    def upgrade_host(tunnel, role, bootstrap_url):
-        # Download the install script for the new DC/OS.
-        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
+    def upgrade_host(tunnel, role, bootstrap_url, upgrade_script_path):
+        # Download the upgrade script for the new DC/OS.
+        tunnel.remote_cmd(curl_cmd + ['--remote-name', upgrade_script_path])
 
-        # Remove the old DC/OS.
-        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
-        tunnel.remote_cmd(['sudo', 'rm', '-rf', '/opt/mesosphere', '/etc/mesosphere'])
-
-        # Install the new DC/OS.
-        tunnel.remote_cmd(['sudo', 'bash', 'dcos_install.sh', '-d', role])
+        # Upgrade to the new DC/OS.
+        tunnel.remote_cmd(['sudo', 'bash', 'dcos_node_upgrade.sh'])
 
     @retry(
         wait_fixed=(1000 * 5),
@@ -372,6 +368,11 @@ def upgrade_dcos(cluster, installer_url, add_config_path=None):
             add_config_path=add_config_path,
             rexray_config_preset=cluster.rexray_config_preset,
         )
+
+        # Generate node upgrade script
+        output = installer.generate_node_upgrade_script(version)
+        upgrade_script_path = output.decode('utf-8').splitlines()[-1].split("Node upgrade script URL: ", 1)[1]
+
         # Remove docker (and associated journald) restart from the install
         # script. This prevents Docker-containerized tasks from being killed
         # during agent upgrades.
@@ -399,7 +400,7 @@ def upgrade_dcos(cluster, installer_url, add_config_path=None):
         for host in hosts:
             logging.info('Upgrading {}: {}'.format(role_name, repr(host)))
             with cluster.ssher.tunnel(host) as tunnel:
-                upgrade_host(tunnel, role, bootstrap_url)
+                upgrade_host(tunnel, role, bootstrap_url, upgrade_script_path)
 
             wait_metric = {
                 'master': 'registrar/log/recovered',
--- a/test_util/dcos_api_session.py
+++ b/test_util/dcos_api_session.py
@@ -388,3 +388,9 @@ class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
         )
         r.raise_for_status()
         return r.text
+
+    def get_version(self):
+        version_metadata = self.get('/dcos-metadata/dcos-version.json')
+        version_metadata.raise_for_status()
+        data = version_metadata.json()
+        return data["version"]
--- a/test_util/installer_api_test.py
+++ b/test_util/installer_api_test.py
@@ -250,7 +250,9 @@ class DcosCliInstaller(AbstractDcosInstaller):
                 # expected behavior
                 pass
         else:
-            print(self.tunnel.remote_cmd(cmd, timeout=MAX_STAGE_TIME))
+            output = self.tunnel.remote_cmd(cmd, timeout=MAX_STAGE_TIME)
+            print(output)
+            return output
 
     def genconf(
             self, master_list, agent_list, public_agent_list, ssh_user, ssh_key,
@@ -331,3 +333,7 @@ class DcosCliInstaller(AbstractDcosInstaller):
 
     def postflight(self, expect_errors=False):
         self.run_cli_cmd('--postflight', expect_errors=expect_errors)
+
+    def generate_node_upgrade_script(self, version, expect_errors=False):
+        # tunnel run_cmd calls check_output which returns the output hence returning this
+        return self.run_cli_cmd("--generate-node-upgrade-script " + version, expect_errors=expect_errors)
--- a/dcos_installer/test_cli.py
+++ b/dcos_installer/test_cli.py
@@ -48,6 +48,10 @@ def test_set_arg_parser():
     assert parser.password is None
     assert parser.action == 'set-superuser-password'
 
+    parser = parse_args(['--generate-node-upgrade-script', 'fake'])
+    assert parser.installed_cluster_version == 'fake'
+    assert parser.action == 'generate-node-upgrade-script'
+
     # Can't do two at once
     with pytest.raises(SystemExit):
         parse_args(['--validate', '--hash-password', 'foo'])
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -320,6 +320,7 @@ class VpcClusterUpgradeTest:
 
             dcos_api_install.wait_for_dcos()
 
+        installed_version = dcos_api_install.get_version()
         healthcheck_app = create_marathon_healthcheck_app(healthcheck_app_id)
         dns_app = create_marathon_dns_app(dns_app_id, healthcheck_app_id)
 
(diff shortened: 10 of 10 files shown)


amitaekbote:
Work in progress
1. sudo wont work because source wont work for sudo
2. unit test is useless except for manually testing for now
3. integration test needed with a package list

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,112 @@
+"""Generates a bash script for upgrading by hand """
```
branden: This code should be in `gen/installer/bash.py`.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,112 @@
...
+
+import gen.installer.util as util
+import gen.template
+from dcos_installer.config import Config
+from dcos_installer.constants import CONFIG_PATH
+from pkgpanda.util import (if_exists, load_json, write_string)
+
+
+node_upgrade_template = """
```
amitaekbote: I agree, we could take out some of the boilerplate stuff, i am not sure how much value that adds since we dont intend to do upgrades this way in the long run

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,112 @@
...
+function main() {
+    # source the environment required to run the pkgpanda commands
+    source /opt/mesosphere/environment.export
+
+    # check we get pkgpanda once we have the environment setup
+    check_command_exists pkgpanda
+
+    {{ pkgpanda_fetch }}
+
+    {{ pkgpanda_activate }}
```
amitaekbote: sure

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,112 @@
...
+}
```
amitaekbote: what other commands do we intend to run?

amitaekbote:
The py35 stuff is its own pr, was waiting for it to be merged, ill rebase and it should go away

dcos_installer/upgrades.py
branden: We'll at least be moving the ZooKeeper data dir by moving and/or symlinking directories and making requests against the Exhibitor UI.

amitaekbote: Ok cool, the testing will get that stuff in place

branden: FYI that'll only happen on masters, so this script will need to do different things based on the contents of `/etc/mesosphere/roles/`.

dcos_installer/cli.py
```diff
@@ -285,6 +290,15 @@ def get_argument_parser():
         help='Hash the given password and store it as the superuser password in config.yaml'
     )
 
+    mutual_exc.add_argument(
+        '--generate-node-upgrade-script',
+        action=PasswordAction,
```
amitaekbote: Do not want to rewrite passwordaction, rename it maybe?

dcos_installer/backend.py
```diff
@@ -43,6 +43,28 @@ def do_configure(config_path=CONFIG_PATH):
...
+
+    config = Config(config_path)
+
+    validate_gen = config.do_validate(include_ssh=False)
+    if len(validate_gen) > 0:
+        for key, error in validate_gen.items():
+            log.error('{}: {}'.format(key, error))
+        return 1
+
+    gen_out = config_util.do_configure(config)
```
amitaekbote: gen_out helps not regenerate the same things over again to reference in upgrades. this function is basically do_configure.

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,38 @@ def test_set_superuser_password(tmpdir):
         assert passlib.hash.sha512_crypt.verify('foo', config['superuser_password_hash'])
 
 
+def test_generate_node_upgrade_script(tmpdir):
```
amitaekbote: this makes sure the function is not crashing and burning, ill add check for return 0 . 

I would like suggestions as to what makes sense to add as unit tests.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,103 @@
...
+        setup_packages_to_activate.append(package)
+
+    return ' '.join(setup_packages_to_activate)
+
+
+def generate_node_upgrade_script(gen_out, current_version, serve_dir=SERVE_DIR):
+
+    bootstrap_url = gen_out.arguments['bootstrap_url']
+
+    # remove once i have late binding stuff
```
amitaekbote: once cluster_packages has the entire list this wont be needed

test_util/cluster.py
```diff
@@ -381,15 +381,13 @@ def master_upgrade_order(cluster):
         )]
 
     def upgrade_host(tunnel, role, bootstrap_url):
-        # Download the install script for the new DC/OS.
-        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
+        # Download the upgrade script for the new DC/OS.
+        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + 'upgrade.latest'])
+        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/upgrade/' +
```
amitaekbote: testing this right now.

test_util/cluster.py
```diff
@@ -381,15 +381,13 @@ def master_upgrade_order(cluster):
...
+                                      "'`cat upgrade.latest`'" + '/dcos_node_upgrade.sh'])
 
-        # Remove the old DC/OS.
-        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
-        tunnel.remote_cmd(['sudo', 'rm', '-rf', '/opt/mesosphere', '/etc/mesosphere'])
-
-        # Install the new DC/OS.
-        tunnel.remote_cmd(['sudo', 'bash', 'dcos_install.sh', '-d', role])
+        # Upgrade to the new DC/OS.
+        tunnel.remote_cmd(['sudo', 'bash', 'dcos_node_upgrade.sh', '-d', role])
```
branden: The upgrade script as designed takes no args. It should determine the node's role from the contents of `/etc/mesosphere/roles`.

amitaekbote: ah good point, i will need to pass the args, i changed the code so it takes args.

cmaloney: It tells us if a machine should be setup as a master / agent / public agent (master, slave, or slave_public role for pkgpanda), which changes the set of systemd services we start and how they're configured.

dcos_installer/backend.py
```diff
@@ -43,6 +43,28 @@ def do_configure(config_path=CONFIG_PATH):
     return 0
 
 
+def generate_node_upgrade_script(current_version, config_path=CONFIG_PATH):
+
+    if current_version is None:
+        log.error('Must provide the current version of the cluster')
+        return 0
```
BenWhitehead: We probably want this to be non-zero if it's a fatal error.

dcos_installer/config_util.py
```diff
@@ -23,6 +23,8 @@ def do_configure(config):
     # Write some package metadata
     pkgpanda.util.write_json(CLUSTER_PACKAGES_PATH, gen_out.cluster_packages)
 
+    return gen_out
```
BenWhitehead: Nit: rename method from `do_configure` to something more appropriate to signify the behavior change of it now returning something.

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,38 @@ def test_set_superuser_password(tmpdir):
...
+
+    cluster_package_sample = '''
+{
+  "dcos-config": {
+    "filename": "packages/dcos-config/dcos-config--setup_652c02f8bc9611362ffd4657a01f4cb29b2ff68b.tar.xz",
+    "id": "dcos-config--setup_652c02f8bc9611362ffd4657a01f4cb29b2ff68b"
+  },
+  "dcos-metadata": {
+    "filename": "packages/dcos-metadata/dcos-metadata--setup_652c02f8bc9611362ffd4657a01f4cb29b2ff68b.tar.xz",
+    "id": "dcos-metadata--setup_652c02f8bc9611362ffd4657a01f4cb29b2ff68b"
```
BenWhitehead: Are these real package ids going to hurt us when we change the base build of dcos?

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,132 @@
...
+
+node_upgrade_template = """
+#!/bin/bash
+#
+# BASH script to upgrade DC/OS on a node
+#
+# Metadata:
+#   dcos image commit: {{ dcos_image_commit }}
+#   generation date: {{ generation_date }}
+# Copyright 2016 Mesosphere, Inc.
```
BenWhitehead: `s/2016/2017/g`

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,132 @@
...
+fi
+
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root" 1>&2
+    exit 1
+fi
+
+
+# check for version of dc/os upgrading from
+version=`grep "version" /opt/mesosphere/etc/dcos-version.json | cut -d '"' -f 4`
```
BenWhitehead: Could this grep break if the json was minified? Do we care or can we enforce that dcos-version.json will always be pretty-printed?

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,132 @@
...
+if [ $version != {{ dcos_version }} ]; then
+   echo "Not the correct version"
+   exit 0
```
BenWhitehead: If this is an error the exit should probably be non-zero

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,132 @@
...
+
+# check if we are on a master node
+if [ -f /etc/mesosphere/roles/master ]; then
+   # run exhibitor migration script here
+   until dcos-shell dcos-exhibitor-migrate-perform
+   do
+        status=$?
+        case $status in
+             1) echo "Waiting for exhibitor endpoint"
+                sleep 1
```
BenWhitehead: I might do a longer sleep.
Something like
```
echo "Waiting for exhibitor endpoint, will try again in 10 seconds"
sleep 10
```

test_util/cluster.py
```diff
@@ -372,6 +371,11 @@ def wait_for_mesos_metric(cluster, host, key, value):
             add_config_path=add_config_path,
             rexray_config_preset=cluster.rexray_config_preset,
         )
+
+        logging.info("Generating node upgrade script")
+        # Generate node upgrade script
```
branden: nit: This comment doesn't add much value. Pretty clear that an upgrade script is being generated here.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,113 @@
...
+        'cluster_packages': package_list,
+        'dcos_version': current_version})
+
+    hash_string = hash_str(bash_script)
+    upgrade_script_path = '/upgrade/' + hash_string
+
+    subprocess.check_call(['mkdir', '-p', serve_dir + upgrade_script_path])
+
+    write_string(serve_dir + upgrade_script_path + '/dcos_node_upgrade.sh', bash_script)
+    write_string(serve_dir + '/upgrade.latest', hash_string)
```
branden: Is this generated just for the purpose of the upgrade test?

test_util/cluster.py
```diff
@@ -323,15 +323,14 @@ def install_dcos(
 def upgrade_dcos(cluster, installer_url, add_config_path=None):
 
     def upgrade_host(tunnel, role, bootstrap_url):
-        # Download the install script for the new DC/OS.
-        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
+        # Download the upgrade script for the new DC/OS.
+        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/upgrade.latest'])
+        upgrade_id = tunnel.remote_cmd(['cat', 'upgrade.latest'])
```
branden: This test needs to reflect how our upgrade docs will look once published. According to the design doc, we'll be asking the user to get the URL from the installer's stdout, so this test will need to get that URL the same way.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,113 @@
...
+        setup_packages_to_activate.append(cluster_packages[package]['id'])
+
+    return ' '.join(setup_packages_to_activate)
+
+
+def generate_node_upgrade_script(gen_out, current_version, serve_dir=SERVE_DIR):
+
+    bootstrap_url = gen_out.arguments['bootstrap_url']
+
+    package_list = get_package_list(gen_out.cluster_packages)
```
branden: This can be condensed into a pretty readable one-liner: `package_list = ' '.join(package['id'] for package in gen_out.cluster_packages.values()`

dcos_installer/config_util.py
branden: This looks like an opportunity for a refactor. Instead of calling `gen.generate()` at the top of the function and returning its `gen_out` at the bottom, we could decompose this a bit and separate config generation from populating `genconf/serve/`.

Quick sketch:

```
def onprem_generate(config):
    return gen.generate(config.as_gen_format(), extra_sources=[gen.build_deploy.bash.onprem_source])

def make_serve_dir(gen_out):
    subprocess.check_call(['mkdir', '-p', SERVE_DIR])
    gen.build_deploy.bash.generate(gen_out, SERVE_DIR)
    # etc.

# You can do this:
make_serve_dir(onprem_generate(config))

# Or if you need to reuse gen_out, you can do this:
gen_out = onprem_generate(config)
make_serve_dir(gen_out)
do_something_else_with(gen_out)
```

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,113 @@
+'''
```
branden: nit: I think `upgrade.py` would be more consistent with the other module names in this directory.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,113 @@
...
+       exit 0
+    fi
+
+    pkgpanda fetch --repository-url={{ bootstrap_url }} {{ cluster_packages }}
+
+    pkgpanda activate {{ cluster_packages }}
+
+    # check if we are on a master node
+    if [ -f /etc/mesosphere/roles/master ]; then
+       # run exhibitor migration script here
```
branden: nit: This comment can probably go.

dcos_installer/upgrades.py
```diff
@@ -0,0 +1,113 @@
...
+    exit 1
+fi
+
+
+function main() {
+
+    # check for version of dc/os upgrading from
+    version=`grep "version" /opt/mesosphere/etc/dcos-version.json | cut -d '"' -f 4`
+    if [ $version != {{ dcos_version }} ]; then
+       echo "Not the correct version"
```
branden: This would be a pretty confusing error message to run into. This could be more verbose, like: `Expecting to upgrade from DC/OS version {{ dcos_version }}. Found $version.`

dcos_installer/cli.py
```diff
@@ -254,13 +259,22 @@ def get_argument_parser():
...
         metavar='password',
         dest='password',
         nargs='?',
         help='Hash the given password and store it as the superuser password in config.yaml'
     )
 
+    mutual_exc.add_argument(
+        '--generate-node-upgrade-script',
+        action=ArgsAction,
+        metavar='current_version',
```
branden: `current` could be vague to the user -- are we asking about the current version of the dgc or the current version that's running on the cluster? Maybe `installed_version` or `upgrade_from_version` would be clearer?

dcos_installer/backend.py
```diff
@@ -43,6 +43,28 @@ def do_configure(config_path=CONFIG_PATH):
...
+
+    config = Config(config_path)
+
+    validate_gen = config.do_validate(include_ssh=False)
+    if len(validate_gen) > 0:
+        for key, error in validate_gen.items():
+            log.error('{}: {}'.format(key, error))
+        return 1
```
branden: This stanza is copied from `do_configure()`. Factor out into a helper function?

amitaekbote: yup i had a comment on it too, will clean it up

dcos_installer/test_backend.py
amitaekbote: nope, just stuff for testing. it makes sure the script gets generated. check my comment above

dcos_installer/upgrades.py
amitaekbote: the upgrade.latest? This was supposed to help differentiate between different runs of the generate-node-upgrade-script on the same config.

dcos_installer/cli.py
amitaekbote: current_installed_version?

dcos_installer/upgrades.py
amitaekbote: I can take of not pretty printed json but anything else has a chance to break , ill check if we can ensure dcos-version.json will always be pretty printed, might be cleaner than more sed here

dcos_installer/upgrades.py
amitaekbote: ah thanks, this was part of the earlier hack and this is much cleaner

test_util/cluster.py
amitaekbote: Going forward do we test our old way of upgrades or just do this one? I need to either add a new test , currently I am modifying the one that existed.

dcos_installer/upgrades.py
amitaekbote: Nope the doc does not mention it, added it during one of the testing cycles. what do you think?

branden: Seems like it ought to be unnecessary given that the design calls for writing upgrade output to a randomly name directory.

test_util/cluster.py
branden: We should have a single upgrade test that verifies the upgrade procedure supported by the code living alongside it. According to our design, that procedure will require copying the URL from stdout.

dcos_installer/upgrades.py
amitaekbote: why randomly named directory? I would just write it to serve/upgrade/? If its randomly generated directory I would keep what I have. upgrade.latest just points to the last created directory

amitaekbote: Also we are doing the same thing that bootstrap.latest does. If we create a random directory every time we run the script, adding upgrade.latest might help customers keep track of which one to look for.

amitaekbote: Offline discussion conclusion: Separate out upgrade.latest in a pr of its own and figure out if we actually need it.

amitaekbote:
@BenWhitehead @branden this is ready for another pass

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,24 @@ def test_set_superuser_password(tmpdir):
...
+    with tmpdir.as_cwd():
+        genconf_dir = tmpdir.join('genconf')
+        genconf_dir.ensure(dir=True)
+        make_default_config_if_needed('genconf/config.yaml', config_extra)
+        genconf_dir.join('ip-detect').write('#!/bin/bash\necho 127.0.0.1')
+        tmpdir.join('artifacts/bootstrap/12345.bootstrap.tar.xz').write('contents_of_bootstrap', ensure=True)
+        tmpdir.join('artifacts/bootstrap/12345.active.json').write('{"active": "contents"}', ensure=True)
+
+    output = subprocess.check_output(['dcos_installer', '--generate-node-upgrade-script', '1.8.7'], cwd=str(tmpdir))
+    assert output.decode('ascii').strip('\n').split("Node upgrade script URL: ", 1)[1].endswith("dcos_node_upgrade.sh")
```
branden: It'd be safer to decode this as UTF-8. It might be even saferer to not decode the output at all and compare against the literal bytes.

Let's also enforce that the URL is on the very last line of the output, to make sure users don't have to sift through text to get to it. I think this does verify that, but indirectly and in a way that'll fail with an unclear `IndexError` if the URL line prefix isn't there. Looks like we can split by line and then compare against the last line.

dcos_installer/test_backend.py
branden: Testing that a missing version number causes the script to exit with an error status would be useful.

dcos_installer/config.py
```diff
@@ -45,11 +45,11 @@ def normalize_config_validation(messages):
     return validation
 
 
-def make_default_config_if_needed(config_path):
+def make_default_config_if_needed(config_path, config_extra=""):
```
branden: This muddles the function's purpose -- it's no longer default config if you're adding to it, and if you're passing `config_extra`, you'll probably not want this function to return without writing it out. Looking at the call site, looks like you can just build your config and call `write_string()` on it directly.

dcos_installer/config_util.py
```diff
@@ -11,19 +11,26 @@
 log = logging.getLogger(__name__)
 
 
-def do_configure(config):
-    gen_out = gen.generate(config.as_gen_format(), extra_sources=[gen.build_deploy.bash.onprem_source])
+def onprem_generate(config):
+    return gen.generate(config.as_gen_format(), extra_sources=[gen.build_deploy.bash.onprem_source])
+
 
+def configure_gen(gen_out):
```
branden: The purpose of this function isn't very clear from its name. `make_serve_dir()`?

dcos_installer/backend.py
```diff
@@ -25,6 +25,14 @@ def print_messages(messages):
         log.error('{}: {}'.format(key, error))
 
 
+def validate_gen(config):
+    validate_gen = config.do_validate(include_ssh=False)
```
branden: This var name shadows the function's name, which can be confusing.

dcos_installer/cli.py
```diff
@@ -17,6 +17,7 @@
 from dcos_installer.installer_analytics import InstallerAnalytics
 from dcos_installer.prettyprint import PrettyPrint, print_header
 
+
```
branden: Why the newline?

dcos_installer/backend.py
```diff
@@ -33,13 +41,26 @@ def do_configure(config_path=CONFIG_PATH):
...
-            log.error('{}: {}'.format(key, error))
-        return 1
+    validate_gen(config)
 
     config_util.do_configure(config)
+
+    return 0
+
+
+def generate_node_upgrade_script(current_version, config_path=CONFIG_PATH):
```
branden: Let's call this `installed_version` like the CLI arg to make it consistent.

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,24 @@ def test_set_superuser_password(tmpdir):
...
+    """
+    config_extra = "\nmaster_list: ['10.0.0.1', '10.0.0.2', '10.0.0.5']\n"
+
+    with tmpdir.as_cwd():
+        genconf_dir = tmpdir.join('genconf')
+        genconf_dir.ensure(dir=True)
+        make_default_config_if_needed('genconf/config.yaml', config_extra)
+        genconf_dir.join('ip-detect').write('#!/bin/bash\necho 127.0.0.1')
+        tmpdir.join('artifacts/bootstrap/12345.bootstrap.tar.xz').write('contents_of_bootstrap', ensure=True)
+        tmpdir.join('artifacts/bootstrap/12345.active.json').write('{"active": "contents"}', ensure=True)
```
branden: I think you can replace these with the `create_config()` and `create_fake_build_artifacts()` functions from https://github.com/dcos/dcos/pull/1132. I think this unit test might actually fail without the artifacts helper, since after #1132 genconf will fail without the presence of a `complete.latest.json` build artifact.

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+    package_list = ' '.join(package['id'] for package in gen_out.cluster_packages.values())
+
+    bash_script = gen.template.parse_str(node_upgrade_template).render({
+        'dcos_image_commit': util.dcos_image_commit,
+        'generation_date': util.template_generation_date,
+        'bootstrap_url': bootstrap_url,
+        'cluster_packages': package_list,
+        'dcos_version': current_version})
+
+    hash_string = hash_str(bash_script)
```
branden: It's hard to tell, but I think this means that successive re-runs of this command with the same version against the same `genconf/` directory will output the upgrade script to the same location, regardless of any config changes, since script content doesn't depend on config. Probably not what we want -- the purpose of the random directory is to prevent successive runs from overwriting each other. The only requirement for this directory in the design doc is that it be "randomly named", so I don't think we need to name it based on its content like this. A truly random string like a `uuid.uuid4().hex` would fit the bill here.

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root" 1>&2
+    exit 1
+fi
+
+
+# check for version of dc/os upgrading from
+version=`grep "version" /opt/mesosphere/etc/dcos-version.json | cut -d '"' -f 4`
+if [ $version != {{ dcos_version }} ]; then
+   echo "Not the correct version, please rerun upgrade script with the installed dcos version ${version}"
```
branden: I don't think this error message should be suggesting steps to remediate, it doesn't have enough context to be able to do that. And it's more likely that the user would regenerate the upgrade script with the right version, not change the version of the upgrade cluster they're upgrading from. This should just state the version it's expecting and the version it's observing on the node and exit.

test_util/cluster.py
```diff
@@ -322,16 +324,12 @@ def install_dcos(
 
 def upgrade_dcos(cluster, installer_url, add_config_path=None):
 
-    def upgrade_host(tunnel, role, bootstrap_url):
-        # Download the install script for the new DC/OS.
-        tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
+    def upgrade_host(tunnel, role, bootstrap_url, upgrade_script_path):
+        # Download the upgrade script for the new DC/OS.
+        tunnel.remote_cmd(curl_cmd + ['--remote-name', upgrade_script_path + 'dcos_node_upgrade.sh'])
```
branden: This would break if we ever renamed the upgrade script. This function should take the complete URL that was printed to stdout.

dcos_installer/test_backend.py
amitaekbote: I will fix this as part of the merge, it is also used in other places in this file

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,35 @@ def test_set_superuser_password(tmpdir):
...
+- 8.8.8.8
+- 8.8.4.4
+ssh_port: 22
+process_timeout: 10000
+bootstrap_url: file:///opt/dcos_install_tmp
+master_list: ['10.0.0.1', '10.0.0.2', '10.0.0.5']
+"""
+
+    with tmpdir.as_cwd():
+        create_fake_build_artifacts(upgrade_config, tmpdir)
```
branden: What's the purpose of changing dirs into the tmpdir here?

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+if [[ $EUID -ne 0 ]]; then
+    echo "This script must be run as root" 1>&2
+    exit 1
+fi
+
+
+# check for version of dc/os upgrading from
+version=`grep "version" /opt/mesosphere/etc/dcos-version.json | cut -d '"' -f 4`
+if [ $version != {{ dcos_version }} ]; then
+   echo "Version specified is not correct. Version expected ${version}"
```
branden: This should include both versions for easier debugging, both installed and expected. It also shouldn't say what's correct -- this script won't know if it's the wrong script being run on the intended cluster, or the right script being run on an unintended cluster, or maybe they're both wrong. It needs to state what's going on as clearly and succinctly as possible so the operator can adjust accordingly.

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+                ;;
+             16) echo "Attempting to start the rolling update failed due to a non 200 response from exhibitor"
+                 echo "Contact Support"
+                 exit
+                 ;;
+        esac
+   done
+fi
+
+echo "Node upgrade complete"
```
branden: This message shouldn't be here, the upgrade isn't complete until it's validated, which this script doesn't do. The design doc also has us printing out the version before we start the upgrade. Please re-read the design doc and adjust this script's output to match it.

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,128 @@
...
+if [ $version != {{ dcos_version }} ]; then
+   echo "Version specified is not correct. Version expected ${version}"
+   exit 1
+fi
+
+echo "Fetching packages"
+
+pkgpanda fetch --repository-url={{ bootstrap_url }} {{ cluster_packages }}
+
+echo "Activating packages"
```
branden: These messages are too verbose. Probably not necessary, if anything they should be hidden behind a `-v` flag.

dcos_installer/cli.py
```diff
@@ -202,6 +202,12 @@ def dispatch(args):
         sys.stdout.buffer.write(byte_str + b'\n')
         sys.exit(0)
 
+    if args.action == 'generate-node-upgrade-script':
+        messages = backend.generate_node_upgrade_script(args.installed_version)
+        if messages:
+            sys.exit(1)
+        sys.exit(0)
```
branden: Shouldn't this just be returning the exit code returned by `generate_node_upgrade_script()`?

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,35 @@ def test_set_superuser_password(tmpdir):
...
+
+    output = subprocess.check_output(['dcos_installer', '--generate-node-upgrade-script', 'fake'], cwd=str(tmpdir))
+    assert output.decode('utf-8').splitlines()[-1].split("Node upgrade script URL: ", 1)[1]\
+                                                  .endswith("dcos_node_upgrade.sh")
```
branden: Can we split this by line instead? We shouldn't have to hard-code the name of the script here, we should just make sure there is one in the output.

```
url_line = output.decode('utf-8').strip().split('\n')[-1]
assert url_line.startswith('Node upgrade script URL: ')
upgrade_script_path = url_line.split(':', 1)[1]
```

test_util/cluster.py
```diff
@@ -372,6 +368,11 @@ def wait_for_mesos_metric(cluster, host, key, value):
             add_config_path=add_config_path,
             rexray_config_preset=cluster.rexray_config_preset,
         )
+
+        # Generate node upgrade script
+        output = installer.generate_node_upgrade_script()
+        upgrade_script_path = output.decode('utf-8').strip('\n').split("Node upgrade script URL: ", 1)[1]
```
branden: See my note above about splitting by line to get the upgrade script URL --  if the URL were for some reason in the middle of the output here, the `upgrade_script_path` would include a bunch of unrelated text and the test wouldn't fail until we attempt to curl. If the URL isn't in the right place or the right format, we should fail immediately rather than waiting for the curl.

test_util/installer_api_test.py
```diff
@@ -331,3 +332,8 @@ def deploy(self, expect_errors=False):
 
     def postflight(self, expect_errors=False):
         self.run_cli_cmd('--postflight', expect_errors=expect_errors)
+
+    def generate_node_upgrade_script(self, expect_errors=False):
+        output = StringIO()
+        self.run_cli_cmd("--generate-node-upgrade-script 1.8.7", expect_errors=expect_errors, stdout=output)
```
branden: https://github.com/dcos/dcos/pull/969#discussion_r100621120

> Each time a new release gets GA'ed, upgrade tests will fail until we bump this version. Can we detect the version we're upgrading from and pass it as a parameter here instead?

dcos_installer/backend.py
```diff
@@ -25,6 +25,14 @@ def print_messages(messages):
         log.error('{}: {}'.format(key, error))
 
 
+def validate_gen(config):
+    validate = config.do_validate(include_ssh=False)
+    if len(validate) > 0:
+        for key, error in validate.items():
+            log.error('{}: {}'.format(key, error))
+        return 1
```
amitaekbote: This should return and the do_configure should handle it and return 1 .

dcos_installer/backend.py
```diff
@@ -33,13 +41,26 @@ def do_configure(config_path=CONFIG_PATH):
     """
     config = Config(config_path)
 
-    validate_gen = config.do_validate(include_ssh=False)
-    if len(validate_gen) > 0:
-        for key, error in validate_gen.items():
-            log.error('{}: {}'.format(key, error))
-        return 1
+    validate_gen(config)
```
amitaekbote: Will capture the return and bubble it up if its 1

dcos_installer/cli.py
amitaekbote: it just returns 0 or 1 , i could change this to sys.exit(messages) , that work?

dcos_installer/test_backend.py
amitaekbote: It is split by line, I am using splitlines and look at the last line. As for the name of the script i want it in here incase we mess up with a typo or something for unit test.

test_util/cluster.py
amitaekbote: The unit test would capture it being in the last line, I can use the same thing here.

test_util/installer_api_test.py
amitaekbote: yup still working on this

dcos_installer/cli.py
branden: Sure, and let's rename messages to status.

dcos_installer/test_backend.py
amitaekbote: output.decode('utf-8').splitlines()[-1]

branden: Ah my mistake, I missed the `splitlines()`.

dcos_installer/test_backend.py
amitaekbote: it makes sure we are building the artifacts in the tmpdir, it was needed when I was creating files here, will get rid of it , the create artifacts takes care of it

dcos_installer/upgrade.py
amitaekbote: they might confuse more than help, these were for my debugging

dcos_installer/test_backend.py
```diff
@@ -67,6 +67,35 @@ def test_set_superuser_password(tmpdir):
...
+
+    try:
+        subprocess.check_output(['dcos_installer', '--generate-node-upgrade-script'], cwd=str(tmpdir))
+    except subprocess.CalledProcessError as e:
+        print(e.output)
+        assert e.output.decode('ascii') == "Must provide the version of the cluster upgrading from\n"
```
branden: If the installer doesn't exit with an error, this exception handler won't get called, and the test won't fail. Could catch that case with an `else` block that raises an exception.

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,134 @@
...
+
+    package_list = ' '.join(package['id'] for package in gen_out.cluster_packages.values())
+
+    bash_script = gen.template.parse_str(node_upgrade_template).render({
+        'dcos_image_commit': util.dcos_image_commit,
+        'generation_date': util.template_generation_date,
+        'bootstrap_url': bootstrap_url,
+        'cluster_packages': package_list,
+        'installed_cluster_version': installed_version,
+        'installer_version': installer_version})
```
branden: Let's not have both `installed_version` and `installer_version` together. Those are really easy to mix up. I like `installer_version` as a name, it's as simple and direct as possible. This function uses both `installed_version` and `installed_cluster_version`. Let's pick one name for that and use it everywhere on this PR -- tests and function signatures and CLI help output included. My bias is toward `cluster_version` over `installed_cluster_version` since the latter is verbose, but whatever's clear and consistent and keeps the two versions distinguishable works for me.

test_util/dcos_api_session.py
```diff
@@ -388,3 +388,8 @@ def mesos_sandbox_file(self, slave_id, framework_id, task_id, filename):
         )
         r.raise_for_status()
         return r.text
+
+    def get_version(self):
+        version_metadata = self.get('/dcos-metadata/dcos-version.json')
```
branden: Let's do a `r.raise_for_status()` to be consistent with other methods.

dcos_installer/backend.py
```diff
@@ -25,6 +25,17 @@ def print_messages(messages):
         log.error('{}: {}'.format(key, error))
 
 
+def validate_gen(config):
+    """ Returns True on success
+    """
```
branden: nit: Single-line docstrings can have all their quotes on one line. https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

dcos_installer/upgrade.py
```diff
@@ -0,0 +1,135 @@
...
+
+    hash_string = hash_str(uuid.uuid4().hex)
+    upgrade_script_path = '/upgrade/' + hash_string
```
branden: No need to hash the UUID.

amitaekbote:
@spahl ship-it label?

PR #970, opened 2016-11-18
Update / rewrite Tunnel, TunnelCollection

cmaloney:
- TunnelCollection is replaced with a contextlib.ExitStack, so it has less code while keeping the same behavior, uses more standard code.
 - Tunnel class is replaced with a "contextmanager" function that yields a Tunneled object.

the Tunnel contextmanager function makes it harder to use the tunnels
incorrectly as the tunnel is only accessible inside the contextmanager / when
the contextmanager has been entered. The "Tunneled" class only has the simple
API / behavior people expect to work on things which are tunneled to.

# Issues

# Checklist

### DIFF
--- a/ssh/tunnel.py
+++ b/ssh/tunnel.py
@@ -8,6 +8,7 @@ with Tunnel(*args, **kwargs) as tunnel:
 """
 import logging
 import tempfile
+from contextlib import contextmanager, ExitStack
 from subprocess import check_call, check_output, TimeoutExpired
 
 LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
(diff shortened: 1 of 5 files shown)


cmaloney:
cc: @mellenburg

ssh/ssh_tunnel.py
```diff
@@ -81,62 +43,73 @@ def remote_cmd(self, cmd, timeout=None, stdout=None):
...
-    def close(self):
-        close_tunnel = self.ssh_cmd + ['-p', str(self.port), '-O', 'exit', self.target]
-        logger.debug('Closing SSH Tunnel: ' + ' '.join(close_tunnel))
-        check_call(close_tunnel)
-
 
-class TunnelCollection():
+@contextmanager
+def Tunnel(ssh_user, ssh_key_path, host, port=22):
+    tunnel_socket = tempfile.NamedTemporaryFile(delete=False)
```
mellenburg: Since this is now always a context manager can we delete after we yield?

ssh/tunnel.py
```diff
@@ -81,50 +43,64 @@ def remote_cmd(self, cmd, timeout=None, stdout=None):
...
+        '-oConnectTimeout=10',
+        '-oControlMaster=auto',
+        '-oControlPath=' + tunnel_socket.name,
+        '-oStrictHostKeyChecking=no',
+        '-oUserKnownHostsFile=/dev/null',
+        '-oBatchMode=yes',
+        '-oPasswordAuthentication=no',
+        '-p', str(port)]
+
+    start_tunnel = base_cmd + ['-fnN', '-i', key_path, '-p', str(port), target]
```
cmaloney: the `-p str(port) probably doesn't need to be repeated here`

ssh/tunnel.py
```diff
@@ -81,50 +43,64 @@ def remote_cmd(self, cmd, timeout=None, stdout=None):
...
+    logger.debug('Starting SSH tunnel: ' + ' '.join(start_tunnel))
+    check_call(start_tunnel)
+    logger.debug('SSH Tunnel established!')
+
+    yield Tunnelled(base_cmd, target)
+
+    close_tunnel = base_cmd + ['-p', str(port), '-O', 'exit', target]
```
cmaloney: same here for the `-p`

ssh/tunnel.py
```diff
@@ -81,50 +43,64 @@ def remote_cmd(self, cmd, timeout=None, stdout=None):
...
+    logger.debug('Closing SSH Tunnel: ' + ' '.join(close_tunnel))
+    check_call(close_tunnel)
+
+
+@contextmanager
+def TunnelCollection(user: str, key_path: str, host_names: list):
+    """Convenience collection of Tunnels so that users can keep
+    multiple connections alive with a single self-closing context
+    Args:
+        user: (str) user with access to host
```
cmaloney: nit: drop the types here since they're in the function signature

cmaloney:
@mellenburg ready for another review

test_util/installer_api_test.py
```diff
@@ -10,7 +10,7 @@
 import yaml
 from retrying import retry
 
-from ssh.tunnel import run_scp_cmd, run_ssh_cmd, Tunnel
+from ssh.tunnel import run_scp_cmd, run_ssh_cmd, Tunneled
```
cmaloney: Tunnelled

cmaloney:
@mellenburg ready for another pass

ssh/tunnel.py
```diff
@@ -81,62 +44,76 @@ def remote_cmd(self, cmd, timeout=None, stdout=None):
...
+def tunnel_collection(user: str, key_path: str, host_names: list):
+    """Convenience collection of Tunnels so that users can keep
+    multiple connections alive with a single self-closing context
+    Args:
+        user: user with access to host
+        key_path: local path w/ permissions to user@host
+        host_names: list of locally resolvable hostname:port to tunnel to
+    """
 
-class TunnelCollection():
```
mellenburg: :+1:  I really like the conversion of an extremely specific class to a simple context manager method

test_util/installer_api_test.py
```diff
@@ -21,11 +22,11 @@ def __init__(self):
         self.offline_mode = False
 
     def setup_remote(
-            self, tunnel, installer_path, download_url,
+            self, tunnel: Optional[Tunnelled], installer_path, download_url,
```
mellenburg: Note to self: Do we even use both of these interfaces? Should probably just make this tunnel only.

mellenburg:
Way cleaner implementation, thanks! :ship:

PR #973, opened 2016-11-18
Fix DC/OS DNS Large Query & Spartan TCP handling

sargun:
 - It fixes Spartan TCP proxying. It fixes the issue when someone would send two queries across a TCP DNS question before waiting for a reply

It includes the following:
```
commit b44292fc37acb165d85f996b4965ac769f35cec0
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Wed Nov 16 16:40:43 2016 -0800

    Make sure dig runs against test server

commit 02e026e97ee33f4f6e1faf79bf16f1b2659fbe55
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Wed Nov 16 16:23:34 2016 -0800

    Use test fixture zone

commit c9507d75b04b94dfdcbc5f6bb577a862d01a4ad5
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Wed Nov 16 16:01:46 2016 -0800

    adjust circle.yml to Ubuntu 14.04

commit 6eccc614554459c3e208778ec872d744874ba2d2
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Wed Nov 16 15:57:33 2016 -0800

    Add test for multiple queries

commit 7d681647dec425d35d4741ec6b2f29abd26167a4
Author: Sargun Dhillon <sargun@sargun.me>
Date:   Tue Nov 15 22:15:40 2016 -0800

    Refactor spartan tcp handler
    
    This refactors the Spartan TCP Handler to handle multiple queries
    over one tcp connection. This has the limitation that the queries
    are performed serially.
3c075477e55e:spartan sdhillon$ 
```

# Issues

@bernadinm Do you have the ticket for this?

# Checklist

 - [X] Included a test which will fail if code is reverted but test is not (There is a test in the repo)

- PR: https://github.com/dcos/spartan/pull/24 (test coverage, change log, and results)


packages/spartan/extra/gen_resolvconf.py
```diff
@@ -51,6 +51,7 @@ def check_server(addr):
 
 options timeout:1
 options attempts:3
+options edns0
```
cmaloney: We may list external servers in `/etc/resolv.conf` which we don't control, so this option can't be used.

sargun: I think it's pretty reasonable to expect people's resolvers to support edns0 in this day.

amitaekbote: Whats the worst case when they dont support edns0? Would we need to check for this support and make sure it works before anyone uses this? Doc change?

sargun:
@spahl / @cmaloney Review?

cmaloney:
I still don't think we should do edns0. I suspect by default, in multiple customer environments this would make their default upgrade result in a non-functioning cluster. For that potentially really bad / discovered at cluster install/rollout time (either during an upgrade causing an outage or during first install causing the field team to scramble to cover why the product broke DNS), we aren't gaining a whole lot of benefit as far as I can tell.

"DNS query packets with EDNS0 extension is not compatible with non-EDNS0 DNS servers.  The option must be used only when all the DNS servers listed in nameserver lines are able to handle EDNS0 extension."

sargun:
So, the problem is with $CUSTOMER's environment is that they have very large DNS resource record sets because of a large number of backends. They do about 100 queries/sec. Normally, this is okay, but when EDNS0 is off, and they get these large replies, modern glibc retries with TCP. Now, doing 100 TCP DNS queries/sec makes Spartan / Mesos-DNS a bit unhappy (it seems Mesos-DNS on one master, and 10 agents runs out of file descriptors). 

Can we perhaps make it disabled by default?

cmaloney:
I'd still rather not need two different configurations just in case sometimes. Why can't we make it so Mesos-DNS doesn't send so big of replies that glibc falls back to TCP? Why can't we increase the socket limit some if that's the limiting factor?  Why can't we cache results locally if they're having that many requests per second? Why can't we make Spartan and Mesos-DNS actually work well together in both these cases without any special config. Local requests to spartan are cheap, why can't it work with Mesos-DNS to handle the scenario well? It knows if it's about to talk to Mesos-DNS or not.

Beyond the "should it be default or not" which means it will cause pain to some people always, making it so we have more documentation, more untested permutations of DC/OS which may or may not be functional, doesn't seem like a product improvement to me. It makes the product more complicated to operate, document, maintain and tests. There are more caveats support and customres need to know to actually make the product work.

sargun:
I'd still rather not need two different configurations just in case sometimes. Why can't we make it so Mesos-DNS doesn't send so big of replies that glibc falls back to TCP? 
This violates the DNS standard. I did a PoC here: https://github.com/mesosphere/mesos-dns/pull/485. We would want to make this behaviour configurable, and I think that's the more complicated factor.

Why can't we increase the socket limit some if that's the limiting factor?
I think that going from the current limit (1024) to 100x that limit (100,000k+) seems like it's not going to scale super well.

Why can't we cache results locally if they're having that many requests per second? 
Spartan is not in the business of caching. Caching is an entirely different, complex problem that I'd rather not implement. If we want to do caching, I'd rather have something with the explicit responsibility of caching. 

Why can't we make Spartan and Mesos-DNS actually work well together in both these cases without any special config. Local requests to spartan are cheap, why can't it work with Mesos-DNS to handle the scenario well?
Eventually, we'll run into the same scenario, where is someone does significantly more local TCP queries, they'll knock over Spartan as well.

------

In general, we want to avoid TCP queries. They take up a ton of resources, and make client software slower, even when querying locally. I think the only other reasonable option is to make it so we only return a sample of results, as opposed to the full corpus. This is another configuration option people will have to change.

EDNS0 has been a standard since 1999. It's been supported by BIND since September, 2000. In an attempt to disable EDNS, I was unable to find a server that didn't support it (BIND, DNSMasq, PowerDNS).

sargun:
@cmaloney Can you please provide me a server which does not support EDNS0 to test against? You can test by running the command `dig +edns +notcp +qr @$IP value.testing.express`

If it doesn't work, edns0 is broken.
If the answer doesn't carry the optional pseudosection with `EDNS: version: 0`, it's doing something odd.

sargun:
@cmaloney works?

sargun:
It appears like the resolvers we have in Vagrant might not support EDNS0...

spahl:
We are closing this down for now and will revisite edns0 at another time.

PR #974, opened 2016-11-18
Update cosmos

jsancio:
  - Update Cosmos to support the following new endpoints
    - /package/add - Add a package to the local repository
    - /service/start - Starts a service using Marathon for the given package
  - Change the configuration generator to allow the user to optionally provide permanent store for packages. If the user configure Cosmos with this information it enables the features outlined above.
  - None
- High-Level how things changed
  - See above.

# Issues

[DCOS-8992](https://dcosjira.atlassian.net/browse/DCOS-8992)

# Checklist

 - [x] There are extensive unit and CI tests in the Cosmos repo.

 - This feature is integrated tested in the Cosmos CI builds. See [results](https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcosIo_Cosmos_Ci).

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -6,6 +6,7 @@ import socket
 import textwrap
 from math import floor
 from subprocess import check_output
+from urllib.parse import urlparse
 
 import yaml
 
@@ -65,6 +66,15 @@ def validate_ipv4_addresses(ips: list):
             return None
     invalid_ips = list(filter(lambda ip: try_parse_ip(ip) is None, ips))
     assert not len(invalid_ips), 'Invalid IPv4 addresses in list: {}'.format(', '.join(invalid_ips))
+
+
+def validate_url(url: str):
+    try:
+        urlparse(url)
+    except ValueError as ex:
+        raise AssertionError(
+            "Couldn't parse given value `{}` as an URL".format(url)
+        ) from ex
 
 
 def is_azure_addr(addr: str):
@@ -427,6 +437,65 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
     return 'false' if cluster_docker_registry_url == '' else 'true'
 
 
+def validate_cosmos_config(cosmos_config):
+    """The schema for this configuration is.
+    {
+      "schema": "http://json-schema.org/draft-04/schema#",
+      "type": "object",
+      "properties": {
+        "staged_package_storage_uri": {
+          "type": "string"
+        },
+        "package_storage_uri": {
+          "type": "string"
+        }
+      }
+    }
+    """
+
+    config = validate_json_dictionary(cosmos_config)
+    expects = ['staged_package_storage_uri', 'package_storage_uri']
+    found = list(filter(lambda value: value in config, expects))
+
+    if len(found) == 0:
+        # User didn't specify any configuration; nothing to do
+        pass
+    elif len(found) == 1:
+        # User specified one parameter but not the other; fail
+        raise AssertionError(
+            'cosmos_config must be a dictionary containing both {}, or must '
+            'be left empty. Found only {}'.format(' '.join(expects), found)
+        )
+    else:
+        # User specified both parameters; make sure they are URLs
+        for value in found:
+            validate_url(config[value])
+
+
+def calculate_cosmos_staged_package_storage_uri_flag(cosmos_config):
+    config = validate_json_dictionary(cosmos_config)
+    if 'staged_package_storage_uri' in config:
+        return (
+            '-com.mesosphere.cosmos.stagedPackageStorageUri={}'.format(
+                config['staged_package_storage_uri']
+            )
+        )
+    else:
+        return ''
+
+
+def calculate_cosmos_package_storage_uri_flag(cosmos_config):
+    config = validate_json_dictionary(cosmos_config)
+    if 'package_storage_uri' in config:
+        return (
+            '-com.mesosphere.cosmos.packageStorageUri={}'.format(
+                config['package_storage_uri']
+            )
+        )
+    else:
+        return ''
+
+
 def calculate_set(parameter):
     if parameter == '':
         return 'false'
@@ -478,7 +547,8 @@ entry = {
         lambda cluster_docker_credentials_write_to_etc: validate_true_false(cluster_docker_credentials_write_to_etc),
         lambda cluster_docker_credentials: validate_json_dictionary(cluster_docker_credentials),
         lambda aws_masters_have_public_ip: validate_true_false(aws_masters_have_public_ip),
-        validate_exhibitor_storage_master_discovery
+        validate_exhibitor_storage_master_discovery,
+        validate_cosmos_config
     ],
     'default': {
         'bootstrap_tmp_dir': 'tmp',
@@ -551,7 +621,8 @@ entry = {
         'cluster_docker_credentials_dcos_owned': calculate_docker_credentials_dcos_owned,
         'cluster_docker_credentials_write_to_etc': 'false',
         'cluster_docker_credentials_enabled': 'false',
-        'cluster_docker_credentials': "{}"
+        'cluster_docker_credentials': "{}",
+        'cosmos_config': '{}'
     },
     'must': {
         'custom_auth': 'false',
@@ -583,7 +654,11 @@ entry = {
         'cluster_docker_credentials_path': calculate_cluster_docker_credentials_path,
         'cluster_docker_registry_enabled': calculate_cluster_docker_registry_enabled,
         'has_master_external_loadbalancer':
-            lambda master_external_loadbalancer: calculate_set(master_external_loadbalancer)
+            lambda master_external_loadbalancer: calculate_set(master_external_loadbalancer),
+        'cosmos_staged_package_storage_uri_flag':
+            calculate_cosmos_staged_package_storage_uri_flag,
+        'cosmos_package_storage_uri_flag':
+            calculate_cosmos_package_storage_uri_flag
     },
     'conditional': {
         'master_discovery': {
--- a/packages/cosmos/build
+++ b/packages/cosmos/build
@@ -21,7 +21,13 @@ PermissionsStartOnly=True
 User=dcos_cosmos
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/proxy.env
+EnvironmentFile=/opt/mesosphere/etc/cosmos.env
 ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-cosmos
-ExecStart=/opt/mesosphere/bin/java -Xmx2G -classpath ${PKG_PATH}/usr/cosmos.jar com.simontuffs.onejar.Boot -com.mesosphere.cosmos.dataDir=/var/lib/dcos/cosmos
+ExecStart=/opt/mesosphere/bin/java \\
+    -Xmx2G \\
+    -classpath ${PKG_PATH}/usr/cosmos.jar \\
+    com.simontuffs.onejar.Boot \\
+    \${COSMOS_STAGED_PACKAGE_STORAGE_URI_FLAG} \\
+    \${COSMOS_PACKAGE_STORAGE_URI_FLAG}
 EOF
--- a/packages/dcos-integration-test/extra/test_endpoints.py
+++ b/packages/dcos-integration-test/extra/test_endpoints.py
@@ -4,6 +4,8 @@ import urllib.parse
 import bs4
 from retrying import retry
 
+from pkgpanda.util import load_json
+
 
 def test_if_dcos_ui_is_up(cluster):
     r = cluster.get('/')
@@ -142,6 +144,43 @@ def test_if_we_have_capabilities(cluster):
     )
     assert r.status_code == 200
     assert {'name': 'PACKAGE_MANAGEMENT'} in r.json()['capabilities']
+
+
+def test_cosmos_package_add(cluster):
+    r = cluster.post(
+        '/package/add',
+        headers={
+            'Accept': (
+                'application/vnd.dcos.package.add-response+json;'
+                'charset=utf-8;version=v1'
+            ),
+            'Content-Type': (
+                'application/vnd.dcos.package.add-request+json;'
+                'charset=utf-8;version=v1'
+            )
+        },
+        json={
+            'packageName': 'cassandra',
+            'packageVersion': '1.0.20-3.0.10'
+        }
+    )
+
+    user_config = load_json("/opt/mesosphere/etc/expanded.config.json")
+    if (user_config['cosmos_staged_package_storage_uri_flag'] and
+            user_config['cosmos_package_storage_uri_flag']):
+        # if the config is enabled then Cosmos should accept the request and
+        # return 202
+        assert r.status_code == 202, 'status = {}, content = {}'.format(
+            r.status_code,
+            r.content
+        )
+    else:
+        # if the config is disabled then Cosmos should accept the request and
+        # return Not Implemented 501
+        assert r.status_code == 501, 'status = {}, content = {}'.format(
+            r.status_code,
+            r.content
+        )
 
 
 def test_if_overlay_master_is_up(cluster):
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -383,6 +383,10 @@ package:
         ],
         "enabled": "{{ telemetry_enabled }}"
       }
+  - path: /etc/cosmos.env
+    content: |
+      COSMOS_STAGED_PACKAGE_STORAGE_URI_FLAG={{ cosmos_staged_package_storage_uri_flag }}
+      COSMOS_PACKAGE_STORAGE_URI_FLAG={{ cosmos_package_storage_uri_flag }}
   - path: /etc/adminrouter.env
     content: |
       ADMINROUTER_ACTIVATE_AUTH_MODULE={{ adminrouter_auth_enabled }}
--- a/packages/cosmos/buildinfo.json
+++ b/packages/cosmos/buildinfo.json
@@ -5,8 +5,8 @@
   ],
   "single_source": {
     "kind": "url",
-    "url": "https://downloads.dcos.io/cosmos/0.2.0-4/cosmos-server-0.2.0-4-one-jar.jar",
-    "sha1": "7d659d8133a37a57c03218e3fcb7105551be3068"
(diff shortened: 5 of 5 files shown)


packages/cosmos/build
```diff
@@ -9,6 +9,8 @@ chmod -R o+r "${PKG_PATH}/usr"
 cosmos_service="${PKG_PATH}/dcos.target.wants_master/dcos-cosmos.service"
 mkdir -p $(dirname "$cosmos_service")
 
+source /opt/mesosphere/etc/cosmos.env
```
jsancio: @cmaloney Is this valid?

gen/calc.py
```diff
@@ -421,6 +421,64 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
     return 'false' if cluster_docker_registry_url == '' else 'true'
 
 
+def validate_cosmos_config(cosmos_config):
```
jsancio: What is `cosmos_config`? I am sure how this symbol is getting resolved. I notice an increasing correlation between the name of the parameter and the name of the property in `entry`. Is this how it is resolved?

packages/cosmos/build
spahl: No you don't have /opt/mesosphere/ in the build environment. None of these files are generated or rendered in there.

jsancio: I am open to suggestions around this. Do I source the file in the actual unit and replace `$COSMOS...` with `\$COSMOS...`?

spahl: Correct. You source the file with systemd and escape.

jsancio: Done. Thanks!

gen/calc.py
branden: Yeah, we use `inspect` to find the names of this validator's arguments, and bind them to the values of the config parameters with those same names. https://github.com/dcos/dcos/blob/master/gen/internals.py#L334

branden: Same story for calculate functions, btw. https://github.com/dcos/dcos/blob/master/gen/internals.py#L61

jsancio:
@branden and @spahl. This is ready for review and merging.

gen/calc.py
```diff
@@ -420,6 +420,64 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
...
+          "type": "string"
+        },
+        "package_storage_uri": {
+          "type": "string"
+        }
+      }
+    }
+    """
+
+    config = json.loads(cosmos_config)
```
cmaloney: Please use https://github.com/dcos/dcos/blob/master/gen/calc.py#L187 to load out this dictionary, makes nice errors if the user provides something not json

gen/calc.py
```diff
@@ -420,6 +420,64 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
...
+    if ('staged_package_storage_uri' in config and
```
cmaloney: These get turned into environment variables. What happens if a customer has a space inside their storage uri?

gen/calc.py
jsancio: Done.

gen/calc.py
jsancio: Done. I am validating that it is an URL. That should also validate that it doesn't contain spaces.

jsancio:
@cmaloney Thanks for the review. I addressed your inline comments. Working on getting tests going for the additional configuration option.

cmaloney:
The vagrant/bash failures being seen earlier (And I'm guessing which will reoccur with the latest push) are actual failures caused by changes in this pr https://teamcity.mesosphere.io/viewLog.html?buildId=500151&tab=buildResultsDiv&buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr

All the journal logs are available for the test runs in the "Cleanup" step: https://teamcity.mesosphere.io/viewLog.html?buildId=500151&buildTypeId=DcosIo_Dcos_VagrantIntegrationTests_ProvisioningTestGenconfBashPr&tab=buildLog#_state=10550,10556&focus=10556

jsancio:
@cmaloney Are you saying that this PR is causing the vagrant tests to fail?

cmaloney:
Yes, when the vagrant integration test runs it fails because of changes this PR makes.

jsancio:
I found this:
```
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[7799]: Processed session termination for sessionid: 0x158d6093efc0079
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000d14a0000, 320180224, 0) failed; error='Cannot allocate memory' (errno=12)
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: #
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: # There is insufficient memory for the Java Runtime Environment to continue.
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: # Native memory allocation (mmap) failed to map 320180224 bytes for committing reserved memory.
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: # An error report file with more information is saved as:
[21:35:44][m3] Dec 06 13:33:17 m3.dcos java[13659]: # /tmp/hs_err_pid13659.log
[21:35:44][m3] Dec 06 13:33:17 m3.dcos systemd[1]: dcos-cosmos.service: main process exited, code=exited, status=1/FAILURE
[21:35:44][m3] Dec 06 13:33:18 m3.dcos java[7799]: Expiring session 0x158d6093efc0073, timeout of 10000ms exceeded
[21:35:44][m3] Dec 06 13:33:18 m3.dcos java[7799]: Processed session termination for sessionid: 0x158d6093efc0073
[21:35:44][m3] Dec 06 13:33:18 m3.dcos systemd[1]: Unit dcos-cosmos.service entered failed state.
[21:35:44][m3] Dec 06 13:33:18 m3.dcos systemd[1]: dcos-cosmos.service failed.
```

cmaloney:
Another high level comment: How does this expect to get permissions / credentials to push stuff to the storage uri when a package is being added?

jsancio:
> Another high level comment: How does this expect to get permissions / credentials to push stuff to the storage uri when a package is being added?

We support two object store:

1. If the object store is the file system then the credential will come from the process.
2. If the object store is S3 then we use the standard aws chain for getting credentials: env, java system property, etc. http://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/credentials.html. When we move to providing a secret store we can design a different solution and maybe expose it in the gen config.

For now we are going to recommend users to either use the file system object store or S3 with EC2 roles.

cmaloney:
The standard AWS lookup scheme isn't functional / desired for s3 based on experience accessing s3 for exhibitor storage.

For s3 users need to be able to:

1. in config.yaml provide explicit credentials. The IAM credentials of a host often don't match the credentials needed.
2. in config.yaml specify to load the credentials out of some file in `/etc` that is managed by their credential management system
3. Defer to IAM (Which should be the config for our AWS CF templates)

They have no way of injecting environment variables at runtime into our services (And that is by design). They also can't edit the java system properties.

When you say for filesystem the credentials come from the process do you mean "credentials are just the standard filesystem permissions of the user cosmos runs as"?

jsancio:
@cmaloney The tests pass. The issue was that the Vagrant VM doesn't have much memory and the JVM was running out of memory loading the Jar. I took a look at the Jar and we were including Jars from the AWS SDK that we didn't need. I removed those dependency, now the Jar is small enough for Vagrant. We should look into changing that integration test and increasing the VM memory. cc @karlkfi 

Regarding your credential issue. This is an experimental issue. We will document it as only supporting the file system option. Before we remove the experimental tag we will add support for S3 credential in the `config.yaml`. What do you think? cc @cruhland

karlkfi:
VM mem is entirely up to the dcos-vagrant user. It's configurable in the VagrantConfig.yaml. 

-Karl

cmaloney:
If we've started using quite a bit more memory than we used to, that likely means we need to up the minimum requirements of DC/OS. Currently we use the default VagrantConfig.yaml, and if that can't run DC/OS it means we probably need to update the default to have more memory on the masters.

karlkfi:
The default mem on the master is pretty low (1GB). I can up the default if you tell me what it actually needs. It was already paging rather aggressively at 1GB, but any change needs to be balanced against a reduced space for deploying apps on a standard 16GB dev laptop. So we can't just up it arbitrarily high.

jsancio:
Back to you @cmaloney and @mellenburg. Ready for review. The failures I am seeing seem to be flaky tests, please confirm.

jsancio:
@mellenburg I used the following value in `config.yaml` during my manual tests:
```
cosmos_config:
  staged_package_storage_uri: file:///var/lib/dcos/cosmos/staged-packages
  package_storage_uri: file:///var/lib/dcos/cosmos/packages
```

jsancio:
Success! All of the tests pass. cc @lingmann

gen/calc.py
```diff
@@ -419,6 +429,68 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
...
+          "type": "string"
+        },
+        "package_storage_uri": {
+          "type": "string"
+        }
+      }
+    }
+    """
+    config = validate_json_dictionary(cosmos_config)
+    if ('staged_package_storage_uri' in config and
```
cmaloney: How about something like the following? I think it simplifies the code a bit, makes it harder to get wrong.

```
expected = ['staged_package_storage_uri', 'package_storage_uri']
found = list(filter(lambda param: param in config, expected))
if len(found) == 0:
    return

if len(found) == 1:
    raise AssertionError(
        'cosmos_config must be a dictionary containing both {}, or must be left '
        'empty. Found only {}'.format(' '.join(expected), found))

for parameter in expected:
    validate_url(config[parameter])
```

gen/calc.py
```diff
@@ -419,6 +429,68 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
...
+        )
+    elif ('staged_package_storage_uri' in config and
+          'package_storage_uri' in config):
+        # Let's try to parse the string as an URL
+        validate_url(config['staged_package_storage_uri'])
+        validate_url(config['package_storage_uri'])
+
+
+def calculate_cosmos_staged_package_storage_uri_flag(cosmos_config):
+    if 'staged_package_storage_uri' in cosmos_config:
```
cmaloney: this is doing string searching rather than dictionary lookup. Can we switch this around to

1. load_json(cosmos_config)
2. check if key exists and if so return
3. else return empty

gen/calc.py
```diff
@@ -419,6 +429,68 @@ def calculate_cluster_docker_registry_enabled(cluster_docker_registry_url):
...
+            '-com.mesosphere.cosmos.stagedPackageStorageUri={}'.format(
+                json.loads(cosmos_config)['staged_package_storage_uri']
+            )
+        )
+    else:
+        return ''
+
+
+def calculate_cosmos_package_storage_uri_flag(cosmos_config):
+    if 'package_storage_uri' in cosmos_config:
```
cmaloney: Same as above

packages/adminrouter/buildinfo.json
```diff
@@ -9,7 +9,7 @@
       "adminrouter": {
           "kind": "git",
           "git": "https://github.com/dcos/adminrouter.git",
-          "ref": "5e23b06be6c061b394d2123424675c542072a6f2",
+          "ref": "44bb0b2d119154f9b4ea9f1c3af184c278bb39b5",
```
cmaloney: nit: this is causing adminrouter to include a lot of unrelated changes which aren't tested or talked about in this PR.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,50 @@ def test_if_we_have_capabilities(cluster):
     assert {'name': 'PACKAGE_MANAGEMENT'} in r.json()['capabilities']
 
 
+def test_cosmos_package_add(cluster):
```
cmaloney: @mellenburg is going to need to review these tests. I'm doing a first pass here.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,50 @@ def test_if_we_have_capabilities(cluster):
...
+    def package_add_configured():
+        user_config = load_yaml("/opt/mesosphere/etc/user.config.yaml")
```
cmaloney: The user config isn't what we want to load here. Ideally we want to say from outside the cluster that is running what state the cluster should be in for these conditional things. (See things like dns_search in https://github.com/dcos/dcos/blob/master/test_util/runner.py#L17). 

As a midway point, can for now probably load `/opt/mesosphere/etc/expanded.config.json` which has all the configuration values which were set or calculated. and grab the flags out of there. There might be a helper around for getting that file, talk to @mellenburg

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,50 @@ def test_if_we_have_capabilities(cluster):
...
+        json={
+            'packageName': 'cassandra',
+            'packageVersion': '1.0.20-3.0.10'
+        }
+    )
+
+    if package_add_configured():
+        # if the config is enabled then Cosmos should accept the request and
+        # return 202
+        assert r.status_code == 202
```
cmaloney: `assert r.status_code == 202` ends up producing really hard to debug / track down errors when it fails (you don't get the status code or response body printed which are both usually critical in solving). Please use r.raise_for_status() or do a `assert r.status_code == 202, "status_code: {}, content: {}".format(r.status_code, r.content))`

gen/calc.py
jsancio: Fixed. Changed the algorithm above slightly.

gen/calc.py
jsancio: Oh, wow. Good catch. Fixed.

gen/calc.py
jsancio: Fixed.

packages/adminrouter/buildinfo.json
jsancio: I am open to suggestions here, I am not sure how to fix it. We need the changes in that commit to expose Cosmos correctly. Do you want me to change the AdminRouter in a different PR?

packages/dcos-integration-test/extra/test_endpoints.py
jsancio: I fixed it to load the JSON you enumerated above. I greped the code and didn't find an util function. @mellenburg Let know if I missed it.

packages/dcos-integration-test/extra/test_endpoints.py
jsancio: Good point. Fixed.

jsancio:
Thanks for the review @cmaloney!

@mellenburg, Let @cmaloney and I know if there are any changes you would like me to make.

jsancio:
> dcos-docs PR for adding the new configuration options (https://github.com/dcos/dcos-docs/blob/master/1.9/administration/installing/custom/configuration-parameters.md)

I am working on this. I was going to do this after this PR gets merged but I can do this now.

jsancio:
> dcos-docs PR for adding the new configuration options (https://github.com/dcos/dcos-docs/blob/master/1.9/administration/installing/custom/configuration-parameters.md)

https://github.com/dcos/dcos-docs/pull/787

mellenburg:
@jsancio Here is a job that is running with the additional config injected: https://teamcity.mesosphere.io/viewLog.html?buildId=504633&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2&tab=buildLog

This can be toggled by setting the env var ENABLE_COSMOS_ADVANCED_TESTING in onprem TeamCity integration tests (set to default of false until this PR is merged, then will be enabled on all PRs). You can force this setting by running a custom build on this branch and filling in true in the parameters section

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,49 @@ def test_if_we_have_capabilities(cluster):
     assert {'name': 'PACKAGE_MANAGEMENT'} in r.json()['capabilities']
 
 
+def test_cosmos_package_add(cluster):
+    def package_add_configured():
+        user_config = load_json("/opt/mesosphere/etc/expanded.config.json")
```
mellenburg: Please only use inner functions if they must take advantage of local variables. There are no such variables being used here

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,49 @@ def test_if_we_have_capabilities(cluster):
...
+                'charset=utf-8;version=v1'
+            )
+        },
+        json={
+            'packageName': 'cassandra',
+            'packageVersion': '1.0.20-3.0.10'
+        }
+    )
+
+    if package_add_configured():
```
mellenburg: This function can be removed outright in favor of:
```
user_config = load_json("/opt/mesosphere/etc/expanded.config.json")
if user_config['cosmos_staged_package_storage_uri_flag'] and user_config['cosmos_package_storage_uri_flag']:
     # same code
```

packages/dcos-integration-test/extra/test_endpoints.py
jsancio: Removed.

packages/dcos-integration-test/extra/test_endpoints.py
jsancio: Done.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -144,6 +146,43 @@ def test_if_we_have_capabilities(cluster):
...
+            'packageVersion': '1.0.20-3.0.10'
+        }
+    )
+
+    user_config = load_json("/opt/mesosphere/etc/expanded.config.json")
+    if (user_config['cosmos_staged_package_storage_uri_flag'] and
+            user_config['cosmos_package_storage_uri_flag']):
+        # if the config is enabled then Cosmos should accept the request and
+        # return 202
+        assert r.status_code == 202, 'status = {}, content = {}'.format(
```
orsenthil: Minor comment.

These assert lines could be folded into a single line as the fall under our 120 col limit.

packages/adminrouter/buildinfo.json
adam-mesos: Adminrouter has been bumped. Now this bump only includes 3 commits to AR: debugging-support attempt1 (accidentally reverted by unified-proxy), unified-proxy, and cosmos-endpoint.
Needs a rebase, but you can keep the 44bb SHA

PR #975, opened 2016-11-19
Fix requests wrapper

mellenburg:
I am confused how this even got merged in the first place. Nothing should be working as requests.request.get is not a thing.

### DIFF
--- a/test_util/cluster_api.py
+++ b/test_util/cluster_api.py
@@ -1,5 +1,4 @@
 import copy
-import functools
 import logging
 import os
 from urllib.parse import urlparse
@@ -262,7 +261,8 @@ class ClusterApi(test_util.helpers.ApiClient):
         super().__init__(
             default_host_url=self.dcos_url,
             api_base=None,
-            ca_cert_path=ca_cert_path)
+            ca_cert_path=ca_cert_path,
+            get_node_url=self.get_node_url)
         self.masters = sorted(masters)
         self.public_masters = sorted(public_masters)
         self.slaves = sorted(slaves)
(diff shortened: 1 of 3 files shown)


mellenburg:
I see where this failed now. All error exits from TC were accidentally muted so all errors appeared green :(((

mellenburg:
@spahl @cmaloney 
This is a fix for my bad merge. Please merge this or https://github.com/dcos/dcos/pull/976 ASAP to restore master to working state

test_util/cluster_api.py
```diff
@@ -326,7 +326,7 @@ def request_wrapper(self, request_fn, path, host_url=None, node=None, port=None,
         if node is not None:
             assert host_url is None, 'Cannot set both node and host_url'
             host_url = self.get_node_url(node, port=port)
-        return super().request_wrapper(request_fn, path, host_url=host_url, **kwargs)
```
mellenburg: If I would have thought about this harder I would have realized this causes spawned clients to go this ClusterApi's super, not their super and thereby it has abi_base of ClusterApi's ApiClient and not whatever the client set for itself

mellenburg:
@cmaloney @spahl This is ready for review. This is what should have been merged with PR #950 to get stuff to work

test_util/cluster_api.py
```diff
@@ -327,31 +326,28 @@ def cluster_api_request(self, method, path, host_url=None, node=None, port=None,
...
             ca_cert_path=self.ca_cert_path,
             default_headers=self.default_headers if default_headers is None else default_headers)
+        # give spawned clients the node= reqest option
+        new_client.extra_wrapper = self.extra_wrapper
+        return new_client
 
     @property
     def marathon(self):
-        return test_util.marathon.Marathon(
+        marathon_client = test_util.marathon.Marathon(
```
cmaloney: Also note for that, might be possible to just do one ApiClient, making since making the `get_node_url` a parameter gets rid of the circular reference.

test_util/helpers.py
```diff
@@ -78,19 +78,45 @@ def api_request(self, method, path, host_url=None, **kwargs):
...
+    def wrapper(self, request_fn, path, host_url=None, node=None, port=None, **kwargs):
+        """If node kwarg is given, a URL will be fetched from _get_node_url instead
+        of using the default host url from ApiClient
+        """
+        if node is not None:
+            assert host_url is not None, 'Cannot set both node and host_url'
+            host_url = self._get_node_url(node, port)
+        super().wrapper(request_fn, path, host_url=host_url, **kwargs)
+
+    get = functools.partialmethod(wrapper, requests.get)
```
mellenburg: When we declare methods this way instead of doing the hackish setattr method, we have to copy/paste this into each class that wants to touch 'wrapper'

test_util/helpers.py
```diff
@@ -78,19 +78,45 @@ def api_request(self, method, path, host_url=None, **kwargs):
...
+    patch = functools.partialmethod(wrapper, requests.patch)
+    delete = functools.partialmethod(wrapper, requests.delete)
+
+
+class NodeApiClient(ApiClient):
+    def __init__(self, get_node_url, *args, **kwargs):
+        """ ApiClient variant allowing node awareness into the requests wrapper
+        """
+        self._get_node_url = get_node_url
+        super().__init__(*args, **kwargs)
```
mellenburg: This is where this broke down for me. This will always go to ApiClient's init. In the ClusterMarathon case we need this to call Marathon's init. So,  it seems I would have to go into marathon and add NodeMarathon or something like that.

test_util/cluster_api.py
```diff
@@ -316,42 +316,24 @@ def get_node_url(self, node, port=None):
...
-    get = functools.partialmethod(cluster_api_request, 'get')
-    post = functools.partialmethod(cluster_api_request, 'post')
-    put = functools.partialmethod(cluster_api_request, 'put')
-    delete = functools.partialmethod(cluster_api_request, 'delete')
-    options = functools.partialmethod(cluster_api_request, 'options')
-    head = functools.partialmethod(cluster_api_request, 'head')
-    patch = functools.partialmethod(cluster_api_request, 'patch')
-    delete = functools.partialmethod(cluster_api_request, 'delete')
-
     def get_client(self, path, default_headers=None):
```
cmaloney: nit: Could we just have this be a method of ApiClient? From what I can tell there's nothing cluster specific about it.

test_util/cluster_api.py
cmaloney: note for future readers / this comment is out of date now.

test_util/helpers.py
```diff
@@ -76,7 +80,13 @@ def api_request(self, method, path, host_url=None, **kwargs):
         if self.api_base:
             path = path_join(self.api_base, path)
 
-        request_url = path_join(host_url if host_url else self.default_host_url, path)
+        if node is not None:
+            assert host_url is None, 'Cannot set both node and host_url'
```
cmaloney: nit: Might make sense to include the values of node and host_url in the assert message

test_util/cluster_api.py
mellenburg: Definitely could, it just felt a little weird as it would be unlikely to spin off a client from a client as opposed to a client from a cluster

cmaloney: In this case though we're really just saying "Give me a new client which points to a sub-url of the current one". That operation I think makes sense at the client layer

mellenburg:
@cmaloney Comments address. Should be good to :ship:

PR #987, opened 2016-11-23
Release and installer fixes

branden:
Small fixes extracted from #524.

# Issues

N/A

# Checklist


cmaloney:
it's odd that the tests didn't auto-trigger on this... Just hand-triggered them in teamcity.

cmaloney:
- I don't think the third commit https://github.com/dcos/dcos/pull/987/commits/e4cfc82f2aa76a7e641d5610984e5315218958fe is necessary now, there is no tests directory, things seem to be working now. My guess is there was a circular import before.
 - For the GENCONF_DIR patch, there aren't very many sites changed / doing a quick grep for `genconf` we have the path hardcoded lots and lots of places. Could we carve out a rule like none of the `dcos_intaller` folder will reference it by anything other than the constant and enforce that?

release/__init__.py
```diff
@@ -267,23 +267,36 @@ def get_gen_package_artifact(package_id_str):
         'local_path': package_filename}
 
 
-def make_bootstrap_artifacts(bootstrap_id, variant_name, artifact_prefix):
+def make_bootstrap_artifacts(bootstrap_id, package_ids, variant_name, artifact_prefix):
     bootstrap_filename = "{}.bootstrap.tar.xz".format(bootstrap_id)
+    bootstrap_filename = "{}.bootstrap.tar.xz".format(bootstrap_id)
```
cmaloney: This line is repeated twice

release/__init__.py
```diff
@@ -267,23 +267,36 @@ def get_gen_package_artifact(package_id_str):
...
+    active_filename = "{}.active.json".format(bootstrap_id)
+    active_local_path = artifact_prefix + '/bootstrap/' + active_filename
+    latest_filename = "{}bootstrap.latest".format(pkgpanda.util.variant_prefix(variant_name))
+    latest_complete_filename = "{}complete.latest.json".format(pkgpanda.util.variant_prefix(variant_name))
```
cmaloney: general musing: I'll be happy when all this probably moves either to pathlib or py.path...

release/__init__.py
```diff
@@ -267,23 +267,36 @@ def get_gen_package_artifact(package_id_str):
...
+
+    # Assert that the bootstrap active packages are in the package list.
+    with open(active_local_path) as f:
+        missing_packages = set(json.load(f)) - set(package_ids)
```
cmaloney: A couple things here
1. Why do we have to do json loading here to get out hte data? Why can't we be passed it
2. can we use `pkgpanda.util.load_json` it makes nice error messages when things like the file doesn't exist or can't be parsed well as json
3. Rather than doing the set math here, and then assert on the length, if we just do an `assert set(active_packages) < set(package_ids)` py.test assertion introspection will do things like show a diff between the two sets. Can then leave a comment on the lin above as to what hte assert is doing (Adding an assert message disables py.test assert introspection)

release/__init__.py
```diff
@@ -267,23 +267,36 @@ def get_gen_package_artifact(package_id_str):
...
         'reproducible_path': 'bootstrap/' + bootstrap_filename,
         'local_path': artifact_prefix + '/bootstrap/' + bootstrap_filename
     }
-    active_filename = "{}.active.json".format(bootstrap_id)
     yield {
         'reproducible_path': 'bootstrap/' + active_filename,
-        'local_path': artifact_prefix + '/bootstrap/' + active_filename
+        'local_path': active_local_path
     }
-    latest_filename = "{}bootstrap.latest".format(pkgpanda.util.variant_prefix(variant_name))
```
cmaloney: :+1: thanks for consolidating these to the top

release/__init__.py
branden: 1. The set of packages activated in the bootstrap isn't published/returned as build metadata, only the variant's bootstrap ID and complete set of packages. We could return that metadata from a build, but it would be irrelevant outside of validation.
1. 👍 
1. 👍

dcos_installer/config_util.py
```diff
@@ -55,20 +55,19 @@ def fetch_bootstrap(bootstrap_id):
     filenames = [
         "{}.bootstrap.tar.xz".format(bootstrap_id),
         "{}.active.json".format(bootstrap_id)]
-    dest_dir = "genconf/serve/bootstrap/"
-    container_cache_dir = "artifacts/bootstrap/"
+    container_cache_dir = "artifacts/bootstrap"
```
branden: Removed the trailing `/` here for consistency with `BOOTSTRAP_DIR`.

release/__init__.py
```diff
@@ -735,8 +735,16 @@ def create(self, repository_path, channel, tag):
...
         assert 'bootstrap_dict' in metadata
+        assert 'complete_dict' in metadata
         assert 'commit' in metadata
 
+        # Assert that each variant's bootstrap's active packages are included in its complete package list.
+        for info in metadata['all_completes'].values():
+            bootstrap_active_packages = set(
+                pkgpanda.util.load_json('packages/cache/bootstrap/{}.active.json'.format(info['bootstrap']))
+            )
+            assert bootstrap_active_packages <= set(info['packages'])
```
branden: @cmaloney I moved this assertion outside of `make_bootstrap_artifacts()` because now `dcos_installer.backend.do_aws_cf_configure()` is calling it, and that code has no way of knowing the complete package list since it's not yet included in `cluster_packages.json` (blocked on late binding since it blows up CF template size). I don't think this assertion has any value outside of the release code anyway.

branden:
Ready for review.

release/__init__.py
```diff
@@ -738,6 +738,15 @@ def create(self, repository_path, channel, tag):
         assert 'complete_dict' in metadata
         assert 'commit' in metadata
 
+        # Assert that each variant's bootstrap's active packages are included in its complete package list.
+        # TODO(branden): Make the complete package list available in the installer (for
+        # dcos_installer.backend.do_aws_cf_configure()) and move this assertion to make_bootstrap_artifacts().
+        for info in metadata['all_completes'].values():
+            bootstrap_active_packages = set(
+                pkgpanda.util.load_json('packages/cache/bootstrap/{}.active.json'.format(info['bootstrap']))
```
cmaloney: nit: Will be nice when we can just programatically get back this from `pkgpanda.build_tree` but that's definitely future / out of scope work.

cmaloney:
:ship:

patch reviews:
DRY genconf paths in dcos_installer/
  - LGTM
  - Verified no more `genconf/` paths hardcoded via a `git grep 'genconf/' dcos_installer. Only appears in a couple documentation strings and tests now.
assert complete_dict in build_metadata
  - LGTM
Validate bootstrap active packages are in packages list
  - LGTM
Add completes to metadata in test_make_channel_artifacts()
  - LGTM

PR #988, opened 2016-11-23
Replace isinstance type assertions with python 3.5 type annotations

cmaloney:
The set of things to replace was found with:
`git grep 'assert isinstance'`

The python 3.5 type hints are now expressive enough for the all of these,
through the new "typing" module. The hints will effectively do the same thing
the asserts used to.

Some type assertions still exist, particularly in cases where there are circular
references (referring to a not-defined-yet class), as well as inside gen
validation where we're explicitly trying to walk through the type.

There is a second commit here which converts two error messages which didn't actually error / abort.

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not: N/A, syntactic change / modernizing the existing code. Old tests pass.

### DIFF
--- a/pkgpanda/actions.py
+++ b/pkgpanda/actions.py
@@ -1,4 +1,3 @@
-import collections
 import logging
 import os
 import sys
@@ -29,7 +28,6 @@ def activate_packages(install, repository, package_ids, systemd, block_systemd):
     block_systemd: if systemd, block waiting for systemd services to come up
 
     """
-    assert isinstance(package_ids, collections.Sequence)
     install.activate(repository.load_packages(package_ids))
     if systemd:
         _start_dcos_target(block_systemd)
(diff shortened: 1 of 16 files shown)


cmaloney:
@branden can you review the first commit here: https://github.com/dcos/dcos/pull/988/commits/4f78716f79f9cf0a2fc0f7e9dee2096ada28e7f9? Changes two ERRORs in the action library

pkgpanda/actions.py
```diff
@@ -105,8 +103,9 @@ def add_package_file(repository, package_filename):
     name = os.path.basename(package_filename)
 
     if not name.endswith(filename_suffix):
-        print("ERROR: Can only add package tarballs which have names "
-              "like {{pkg-id}}{}".format(filename_suffix))
+        raise ValidationError(
+            "ERROR: Can only add package tarballs which have names like "
+            "pkg-id--{}".format(filename_suffix))
```
branden: This'll render as `pkg-id--.tar.xz`. Previously it was `{{pkg-id}}.tar.xz`.

orsenthil:
👍 

What does `python3.5 -m mypy <package>/<dir>` give with this?

cmaloney:
conflicts with other things in train 101, will rebase after that train lands.

cmaloney:
@branden fixed
@orsenthil the times I've run mypy so far have not found any bugs / actual errors, even as I removed/fixed a number of the common messages it gave.

jgehrcke:
:heart:

gpaul:
yay for types!

PR #989, opened 2016-11-23
Remove variable pass through for items found locally

mellenburg:
We want to minimize the amount of input needed to run a test and these member variables serve no purpose other than passing through information that can be readily pulled from the config on the host where the test is running:
-provider
-auth_enabled
-dns_search_set

# Issues
https://mesosphere.atlassian.net/browse/DCOS-11977

# Checklist

### DIFF
--- a/packages/dcos-integration-test/extra/cluster_fixture.py
+++ b/packages/dcos-integration-test/extra/cluster_fixture.py
@@ -2,7 +2,6 @@
 ClusterApi object that will be injected into the pytest 'cluster' fixture
 via the make_cluster_fixture() method
 """
-
 from test_util.cluster_api import ClusterApi, get_args_from_env
 from test_util.helpers import CI_AUTH_JSON, DcosUser
 
--- a/packages/dcos-integration-test/extra/test_helpers.py
+++ b/packages/dcos-integration-test/extra/test_helpers.py
@@ -1,0 +1,3 @@
+from pkgpanda.util import load_json
+
+dcos_config = load_json('/opt/mesosphere/etc/expanded.config.json')
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -62,9 +62,7 @@ def get_value(template_parameter):
 
 
 def get_test_config():
-    # Currently tests only pass with auth disabled, TEST_ADD_ENV_DCOS_AUTH_ENABLED=true
-    # can be set in the environment to override this
-    add_env = {'DCOS_AUTH_ENABLED': 'false'}
+    add_env = {}
     prefix = 'TEST_ADD_ENV_'
     for k, v in os.environ.items():
         if k.startswith(prefix):
@@ -184,8 +182,6 @@ def main():
                 master_list=ip_buckets['master'],
                 agent_list=ip_buckets['private'],
                 public_agent_list=ip_buckets['public'],
-                provider='azure',
-                test_dns_search=False,
                 add_env=get_test_config(),
                 pytest_cmd=os.getenv('DCOS_PYTEST_CMD', "py.test -vv -s -rs -m 'not ccm' ") + os.getenv('CI_FLAGS', ''))
         test_successful = True
--- a/test_util/cluster_api.py
+++ b/test_util/cluster_api.py
@@ -23,13 +23,6 @@ def get_args_from_env():
     assert 'PUBLIC_MASTER_HOSTS' in os.environ
     assert 'SLAVE_HOSTS' in os.environ
     assert 'PUBLIC_SLAVE_HOSTS' in os.environ
-    assert 'DNS_SEARCH' in os.environ
-    assert 'DCOS_PROVIDER' in os.environ
-
-    # must be true or false (prevents misspellings)
-    assert os.environ['DNS_SEARCH'] in ['true', 'false']
-
-    assert os.environ['DCOS_PROVIDER'] in ['onprem', 'aws', 'azure']
 
     return {
         'dcos_url': os.environ['DCOS_DNS_ADDRESS'],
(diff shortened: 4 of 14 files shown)


mellenburg:

test_util/helpers.py
```diff
@@ -173,3 +175,10 @@ def remove_file():
     # Attempt to remove the file upon normal interpreter exit.
     atexit.register(remove_file)
     return temp_path
+
+
+def parse_env_file(path):
+    """ converts any file with one key=value assignment per line
+    into a python dict
+    """
+    return {i[0]: '='.join(i[1:]) for i in [j.split('=') for j in load_string(path).split('\n')]}
```
cmaloney: there are comments in the file, and the splits should be limited. If we want ot write this function, will also need to write a unit test for it, I can give some cases which will break this which are fairly common.

mellenburg:
@cmaloney I switched this to a more specific implementation PTAL

packages/dcos-integration-test/extra/cluster_fixture.py
```diff
@@ -3,12 +3,26 @@
...
 from test_util.cluster_api import ClusterApi, get_args_from_env
 from test_util.helpers import DcosUser
 
 
+def get_auth_enabled():
+    """This file is intended to always be utilized from inside a cluster
+    as such, we can just derive the auth state and not bother with
+    passing in a potentially incorrect setting
+    """
+    res = check_output([
```
cmaloney: `res = pkgpanda.util.load_json('/opt/mesosphere/etc/expanded.config.json')['adminrouter_auth_enabled']`?

mellenburg:
@cmaloney I knew there was a reason I implemented it by looking at the env file and not the expanded config :stuck_out_tongue: 
in non-onprem deploys, we get this in the expanded config:
```
"adminrouter_auth_enabled":"{ \"Ref\" : \"OAuthEnabled\" }",
```
whereas that env file is present in all deploys and always has a nice and simple true/false

cmaloney:
The new late binding stuff should fix that injection if I get it righ! :P 

Does this need to land right now (I'm really not a fan of the subprocess for doing this) or could it wait for #1009 (which should be code complete by end of day)

mellenburg:
No hurry, can wait forever on this. Just made refactoring azure marginally nicer (which isnt a priority anyway)

mellenburg:
@cmaloney I upgraded this to include all the parameters that don't belong as a part of the cluster API. Can we put this back to "ready for review"?
@orsenthil PTAL

mellenburg:
Also, @cmaloney I still can't use expanded config for auth state because auth disabled still gets a result that cannot be clearly interpreted: ``` "adminrouter_auth_enabled":"[[[variables('oauthEnabled')]]]",  ```

test_util/cluster_api.py
```diff
@@ -187,7 +177,7 @@ def _wait_for_metronome(self):
 
     def wait_for_dcos(self):
         self._wait_for_adminrouter_up()
-        if self.auth_enabled and self.web_auth_default_user:
+        if self.web_auth_default_user:
```
orsenthil: `if self.web_auth_default_user is not None`:

(Since we use the default of None).

test_util/azure_test_driver.py
```diff
@@ -62,9 +62,7 @@ def get_value(template_parameter):
 
 
 def get_test_config():
-    # Currently tests only pass with auth disabled, TEST_ADD_ENV_DCOS_AUTH_ENABLED=true
-    # can be set in the environment to override this
-    add_env = {'DCOS_AUTH_ENABLED': 'false'}
+    add_env = {}
```
orsenthil: This function could be single-lined.

```
return {k.replace(prefix, ''): v for k, v in os.environ.items() if k.startswith(prefix)}
```

packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -0,0 +1,3 @@
+from pkgpanda.util import load_json
+
+dcos_config = load_json('/opt/mesosphere/etc/expanded.config.json')
```
orsenthil: This brings an external dependency. (Existence of file).
- How do we ensure that the file always exists?
- Is there any reasonable defaults for the test module that can be used in the absence of the file?

mellenburg: -Is it possible to have a master node that is considered "healthy" that doesnt have an expanded config? I do not think so. Given that, we already have made it a requirement that we run tests on masters so I do not see an issue here
-there is no such thing as a default provider and dns_search defaults to None. Could add some default None but then that gets confused with when the actual config is None. I would rather hard error on the config not being present that be ambiguous. Same goes for adminrouter auth state, default is enabled, but I would rather error hard if the util needs to guess. Also, there will be an endpoint to get security type, so we will probably switch to using that at some point.

test_util/azure_test_driver.py
mellenburg: This whole module will be nuked with https://github.com/dcos/dcos/pull/914 so can we just skip the small style changes? Fixing style/idiom inconsistencies in this file is a rabbit hole

orsenthil: Sounds good to me.

lingmann:
Ping @cmaloney and @orsenthil can you both update your reviews on this? Thanks!

packages/dcos-integration-test/extra/test_helpers.py
orsenthil: OK. I don't have any other suggestion here. I see other pieces of code adopting the same approach of reading from `expanded.config.json` on master.

lingmann:
Also @mellenburg can you rebase on master and make sure tests are all green?

lingmann:
@mellenburg the AWS deploy test on this keeps failing... I've tried re-running 3 times now. Can you investigate?

mellenburg:
@lingmann Thanks for pointing that out to me. Fixed now, sorry for the goof up!

lingmann:
This LGTM and I believe that all of @cmaloney's concerns were addressed.

PR #994, opened 2016-11-28
Fixed a non-idiomatic logging statement in the test utils.

greggomann:
This is a follow-up PR from pull/358. It fixes a small logging nit that was not fixed in the previous PR.

### DIFF
--- a/test_util/marathon.py
+++ b/test_util/marathon.py
@@ -237,7 +237,7 @@ class Marathon(ApiClient):
             r = self.get('v2/deployments')
             data = r.json()
             if len(data) > 0:
-                logging.info('Waiting for pod to be deployed %s', repr(data))
+                logging.info('Waiting for pod to be deployed %r', data)
                 return False
             # deployment complete
             r = self.get('v2/pods' + pod_id)


test_util/marathon.py
```diff
@@ -237,7 +237,7 @@ def _wait_for_pod_deployment(pod_id):
             r = self.get('v2/deployments')
             data = r.json()
             if len(data) > 0:
-                logging.info('Waiting for pod to be deployed %s', repr(data))
+                logging.info('Waiting for pod to be deployed %r', data)
```
mellenburg: If we are going for more idiomatic, I would do 'Waiting for pod to be deployed {}'.format(data)

cmaloney: logging does a repr interpretation, passing the extra parameters as introspectable data. The %r is actually more idiomatic in this case (I wish python would move to .format() pretty much everywhere but that is another discussion entirely).

jgehrcke: I have so many opinions here, one of which is: the actual grammar and content of that sentence is probably more important than which formatting syntax to use :-) But hey :ship: and :video_game:

PR #997, opened 2016-11-28
Only remove auth headers if present

mellenburg:
when trying to spin off an authorized session, the auth headers are
cleared but then the resultant session has a user (and no headers yet)
so when it tries to spin off a non-user session to authenticate it tries


cmaloney:
@mellenburg is there a way we can add a unit test or something for this behavior?

mellenburg:
@cmaloney How about this?
```
import requests
from test_util.cluster_api import ClusterApi
from test_util.helpers import DcosUser

def test_make_user_session(monkeypatch):
    monkeypatch.setattr(requests.Response, 'raise_for_status', lambda: pass)
    monkeypatch.setattr(ClusterApi, 'post', lambda *args, **kwargs: {'token': 'abc'. 'dcos-acs-auth-cookie': 'foo'}
    user_1 = DcosUser({'foo': 'bar'})
    user_2 = DcosUser({'baz': 'qux'})
    cluster_1 = ClusterApi(..., default_web_auth_user=user_1)
    cluster_2  = cluster_1.get_user_session(user_2)
```
a lot of patching but should show the pipes work

mellenburg:
confirmed this works in upstream: https://teamcity.mesosphere.io/viewLog.html?buildId=488974&buildTypeId=ClosedSource_Dcos_DcOsEnterpriseCcmDeployTestSecurityPermissive

Just adding a unit test then this should be good to go

mellenburg:
- dnspython was added to setup as previously, there were no unit tests that used the test harness so it was never required for what tox runs
- test_runner was renamed runner as now that test_util is scanned by pytest, any file that begins with the word 'test' should really be a test file

PR #778, opened 2016-09-26
DC/OS Metrics Service

malnick:
### DC/OS Metrics Service
The PR enables the DC/OS metrics service. The service runs on a domain socket on both masters and agents. On masters it exposes host-level metrics and on agents it exposes host, container and metrics gathered from the Mesos Metrics module. 

### Mesos Metrics Module
This PR adds the Mesos metrics module to OSS DC/OS. Along with that addition are metrics module-specific CPP libraries for avro and boost. 

### Integration Tests
The integration test ensures that the metrics service runs and returns host level metrics for each DC/OS role. It ensures that metrics sent to STATSD_UDP_HOST/PORT inside a Mesos deployed container are collected by the metrics service and exposed with the expected values. 

### Summary
```
Packages   100% Files   100% Classes   100% Methods   86% Lines   63% Conditionals   100%
```
### Detailed Coverage
```
github.com/dcos/dcos-metrics/collectors/framework/framework.go:63:	New				100.0%
github.com/dcos/dcos-metrics/collectors/framework/framework.go:74:	Transform			78.9%
github.com/dcos/dcos-metrics/collectors/framework/framework.go:131:	RunFrameworkTCPListener		75.0%
github.com/dcos/dcos-metrics/collectors/framework/framework.go:156:	handleConnection		73.5%
github.com/dcos/dcos-metrics/collectors/framework/framework.go:228:	GetTopic			100.0%
github.com/dcos/dcos-metrics/collectors/framework/framework.go:244:	Read				80.0%
github.com/dcos/dcos-metrics/collectors/framework/record.go:41:		extract				83.9%
github.com/dcos/dcos-metrics/collectors/framework/record.go:108:	createObjectFromRecord		80.0%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:152:	RunPoller			0.0%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:162:	pollMesosAgent			0.0%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:182:	getContainerMetrics		100.0%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:198:	getAgentState			100.0%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:212:	transformContainerMetrics	81.8%
github.com/dcos/dcos-metrics/collectors/mesos/agent/agent.go:248:	getFrameworkInfoByFrameworkID	100.0%
[...]

PR #998, opened 2016-11-29
Add golang 1.7.3 to dcos base build image and use it in our golang projects

malnick:
This PR adds golang 1.7.3 to the DC/OS base build image. This supports consolidating to a single container for building (faster builds) and allowing pkgpanda packages which build many components can use a single container (see dcos-metrics service and module consolidation).

### DIFF
--- a/packages/dcos-oauth/buildinfo.json
+++ b/packages/dcos-oauth/buildinfo.json
@@ -1,5 +1,4 @@
 {
-  "docker": "golang:1.6",
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/dcos/dcos-oauth.git",
--- a/packages/dcos-signal/build
+++ b/packages/dcos-signal/build
@@ -1,5 +1,4 @@
 #!/bin/bash
-export GOPATH=$GOPATH:/pkg
 mkdir -p /pkg/src/github.com/dcos
 # Create the GOPATH for the go tool to work properly
 mv /pkg/src/dcos-signal /pkg/src/github.com/dcos/
--- a/packages/dcos-signal/buildinfo.json
+++ b/packages/dcos-signal/buildinfo.json
@@ -1,5 +1,4 @@
 {
-  "docker": "golang:1.6",
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/dcos/dcos-signal.git",
--- a/pkgpanda/docker/dcos-builder/Dockerfile
+++ b/pkgpanda/docker/dcos-builder/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get -qq update && apt-get -y install \
   git \
   gzip \
   libapr1-dev \
+  libc6-dev \
   libcurl4-openssl-dev \
   libnl-3-dev \
   libnl-genl-3-dev \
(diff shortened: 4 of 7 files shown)


malnick:
@spahl @cmaloney this is ready for review.

pkgpanda/docker/dcos-builder/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get -qq update && apt-get -y install \
   gettext-base \
   git \
   gzip \
+  libc6-dev \
```
cmaloney: please place alphabetically sorted in the list

pkgpanda/docker/dcos-builder/Dockerfile
```diff
@@ -49,4 +50,18 @@ RUN ln -sf /usr/bin/cpp-4.8 /usr/bin/cpp && \
...
+ENV GOLANG_DOWNLOAD_URL https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz
+ENV GOLANG_DOWNLOAD_SHA256 508028aac0654e993564b6e2014bf2d4a9751e3b286661b0b0040046cf18028e
+
+RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz \
+  && echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
+  && tar -C /usr/local -xzf golang.tar.gz \
+  && rm golang.tar.gz
+
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
```
cmaloney: why does `/usr/local/go/bin` need to be in the PATH? Why isn't it just `$GOPATH/bin`? It doesn't look like things are extracted to the GOPATH `/go` at all

malnick: The /usr/local/go/bin path is where go places it's utilities.

alberts: This is from https://github.com/docker-library/golang/blob/4fd5df86eea53623b1009b3621b40a97f9f359e5/1.7/Dockerfile#L22

If I had to guess:

/usr/local/go/bin goes in the PATH for the Go tools itself

The GOPATH stuff is there for downstream projects that want to build packages inside of the Go Docker. Hence all the go-wrapper stuff that was doing magic with /go/src/app:

https://github.com/docker-library/golang/blob/4fd5df86eea53623b1009b3621b40a97f9f359e5/1.7/go-wrapper

Seems like all our packages currently do some variant of export GOPATH=$GOPATH:/pkg. I think they should just do export GOPATH=/pkg and we can probably just not set GOPATH here.

cmaloney:
@malnick can we update at least one of the go packages in this to use the default image instead of `golang` currenty have no check that the go we now bundle actually works.

malnick:
#778 is rebased on this PR @cmaloney - it works.

packages/dcos-oauth/buildinfo.json
```diff
@@ -1,5 +1,4 @@
 {
-  "docker": "golang:1.6",
```
cmaloney: the repo / it's tests don't run docker 1.7 currently, need to change that before we swap this out, make sure the in-repo tests pass under go 1.7

packages/dcos-signal/buildinfo.json
```diff
@@ -1,5 +1,4 @@
 {
-  "docker": "golang:1.6",
```
cmaloney: the repo / it's tests don't run docker 1.7 currently, need to change that before we swap this out, make sure the in-repo tests pass under go 1.7

malnick: Jenkins hard codes 1.7 in our test environment.

packages/dcos-oauth/buildinfo.json
malnick: https://github.com/dcos/dcos-oauth/pull/6

packages/dcos-signal/build
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-export GOPATH=$GOPATH:/pkg
```
alberts: just to confirm, dcos-signal build doesn't need GOPATH set at all?

pkgpanda/docker/dcos-builder/Dockerfile
```diff
@@ -49,4 +50,19 @@ RUN ln -sf /usr/bin/cpp-4.8 /usr/bin/cpp && \
...
+ENV GOLANG_DOWNLOAD_URL https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz
+ENV GOLANG_DOWNLOAD_SHA256 508028aac0654e993564b6e2014bf2d4a9751e3b286661b0b0040046cf18028e
+
+RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz \
+  && echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
+  && tar -C /usr/local -xzf golang.tar.gz \
+  && rm golang.tar.gz
+
+# Set GOPATH to expected pkgpanda package path for DC/OS
+ENV GOPATH /pkg
```
alberts: Wasn't sure if I agreed with this, but since this is the dcos-builder specifically, it's probably fine.

malnick: I thought this was the conversation we had yesterday? We can remove the setting of GOPATH in our build scripts if we set it here, which is what we were resetting in our build scripts explicitly?

PR #1003, opened 2016-11-29
Raise error to exit nicely

mellenburg:
- if do_build_packages fail, we get the real exception in the log stream, but it is buried under an UnboundError exception that occurs because the code tries to go forward after failing when it should have simply stopped 
- Just raise the error as was originally the case

# Checklist

 - [x] Followed relevant code rules [Rules for Packages and Systemd]

### DIFF
--- a/release/__init__.py
+++ b/release/__init__.py
@@ -303,7 +303,8 @@ def make_stable_artifacts(cache_repository_url):
         try:
             all_completes = do_build_packages(cache_repository_url)
         except pkgpanda.build.BuildError as ex:
-            logger.error("Building packages: {}".format(ex))
+            logger.error("Failure building package(s): {}".format(ex))
+            raise
 
     # The installer is a built bootstrap, but not a DC/OS variant. We use
     # iteration over the complete_dict to enumerate all variants a whole lot,
--- a/release/test_release.py
+++ b/release/test_release.py
@@ -8,6 +8,7 @@ import pytest
 import gen.build_deploy.aws
 import release
 import release.storage.aws
+from pkgpanda.build import BuildError
 from pkgpanda.util import variant_prefix, write_json, write_string
 
 
@@ -552,6 +553,10 @@ stable_artifacts_metadata = {
 }
 
 
+def mock_failed_build_packages(_):
+    raise BuildError('This build failed!')
+
+
 # TODO(cmaloney): Add test for do_build_packages returning multiple bootstraps
 # containing overlapping
 def test_make_stable_artifacts(monkeypatch, tmpdir):
@@ -561,6 +566,11 @@ def test_make_stable_artifacts(monkeypatch, tmpdir):
     with tmpdir.as_cwd():
         metadata = release.make_stable_artifacts("http://test")
         assert metadata == stable_artifacts_metadata
+
+    # Check that a BuildError is propogated
+    monkeypatch.setattr("release.do_build_packages", mock_failed_build_packages)
+    with pytest.raises(BuildError):
+        release.make_stable_artifacts("http://test")
 
 
 # NOTE: Implicitly tests all gen.build_deploy do_create functions since it calls them.


orsenthil:
+1. Is there a test coverage for this? Something along `assertRaises`

mellenburg:
@orsenthil There is not a test, but in order to make a test I would mock so much stuff I don't think it would be terribly valuable (as with a lot of failure condition testing). Plus I am not looking to dig into that in this PR. Just want to make this message understandale

release/__init__.py
```diff
@@ -302,8 +302,9 @@ def make_stable_artifacts(cache_repository_url):
     with logger.scope("Building packages"):
         try:
             all_completes = do_build_packages(cache_repository_url)
-        except pkgpanda.build.BuildError as ex:
-            logger.error("Building packages: {}".format(ex))
+        except pkgpanda.build.BuildError:
+            logger.exception("An error was encountered while building packages")
```
cmaloney: It's still nice to have the error message formatted into the log stream if possible here. Sometimes can be hard to find the exception otherwise.

cmaloney:
@mellenburg there is an existing test for make_stable_artifacts: https://github.com/dcos/dcos/blob/master/release/test_release.py#L557

That would be fairly straightforward to do another monkeypatch.setattr to change mock_do_build_packages to throw a BuildError and check that the exception gets propogated with `pytest.raises`

Also, the main PR comment needs to follow the Pull Request Template (https://github.com/dcos/dcos/blob/master/PULL_REQUEST_TEMPLATE.md)

branden:
`logging.error(str(exc))` and `logging.exception()` do different things. The latter includes the traceback:

```
In [1]: import logging

In [2]: try:
   ...:     raise Exception('foo')
   ...: except Exception as exc:
   ...:     logging.error('Got exception: {}'.format(exc))
   ...:
ERROR:root:Got exception: foo

In [3]: try:
   ...:     raise Exception('foo')
   ...: except Exception as exc:
   ...:     logging.exception('Got exception: ')
   ...:
ERROR:root:Got exception:
Traceback (most recent call last):
  File "<ipython-input-3-62006bdffdc0>", line 2, in <module>
    raise Exception('foo')
Exception: foo

In [4]:
```

Both printing the traceback and re-raising the exception is redundant. If an exception handler is re-raising the exception, I'd expect it to log at most the error message, and let an exception handler above it decide what to do with the traceback. Looks like you can get what you need just by adding the `raise`.

cmaloney:
The whole reason this try / except exists is to make the source of these errors and what they mean more obvious / visible where it came from and what it relates to. We catch the build error right there and print the basic message making it easier to read / find. It's why the try/catch was added there. Without it there isn't value. The string of teh exception is different than the traceback.

When other teams see the traceback they say "hey your tools broke come debug for me"
whereas the build error message being printed says  "Your pkgpanda package failed to build for this reason you need to fix". People interact _very_ differently with those.

mellenburg:
@cmaloney So do you want there to be no traceback at all? I.E. 
```
try:
    all_completes = do_build_packages(cache_repository_url)
except pkgpanda.build.BuildError as ex:
    logger.error("An error was encountered while building packages: {}".format(ex))
    sys.exit(1)
```

cmaloney:
found the commit that broke this: https://github.com/dcos/dcos/commit/e49d19650584c6b0347b3bb9bdfe50287a6201ca#diff-f1d07a3e273cef624bac200fe7e81dd6L304

from my perspective print an error line, then re-raise proobably for now. do_build_packages is not setup as a "CLI only" function / not sure if things would unwrap properly with a `sys.exit(1)` for failure. Going back to the old code is well known for now, fixes the bug that was introduced.

https://github.com/dcos/dcos/commit/848a3326108e67e4577e9ae2a4d6398738c3ef42 and https://github.com/dcos/dcos/pull/209 for history of this print getting added. IT's important it's the simplified print.

mellenburg:
@cmaloney All changes addressed.

orsenthil:
ship-it from me.

The raise is cody's preference. I liked `logging.exception` suggestion as it accomplished this in a better way.

orsenthil:
I take back my previous comment. The current change `logging.error` is better (did not think about having two tracebacks in the console)

PR #1006, opened 2016-12-01
Bump to ZK 3.4.9, get rid of zookeeper.out

alberts:
Change ZooKeeper log location (https://mesosphere.atlassian.net/browse/DCOS-4357)

https://zookeeper.apache.org/doc/r3.4.9/releasenotes.html

### DIFF
--- a/packages/exhibitor/buildinfo.json
+++ b/packages/exhibitor/buildinfo.json
@@ -9,8 +9,8 @@
     },
     "zookeeper": {
       "kind": "url_extract",
-      "url": "https://downloads.mesosphere.com/zookeeper/zookeeper-3.4.8.tar.gz",
-      "sha1": "51b61611a329294f75aed82f3a4517a4b6ff116f"
+      "url": "https://downloads.mesosphere.com/zookeeper/zookeeper-3.4.9.tar.gz",
+      "sha1": "0285717bf5ea87a7a36936bf37851d214a32bb99"
     },
     "log4j": {
         "kind": "url",
--- a/packages/exhibitor/extra/zkServer.sh
+++ b/packages/exhibitor/extra/zkServer.sh
@@ -1,0 +1,223 @@
+#!/usr/bin/env bash
+
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


packages/exhibitor/extra/zkServer.sh
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env bash
```
cmaloney: copying and hand-modifying this script isn't very useful. Most of this code we don't maintain and hand-updating the changes and figuring out what the implicit diff from the one bundled with zk 3.4.9 isn't going to be fun. Need to reduce this to the single exec line with arguments in it. Vast majority of this script is never touched.

packages/exhibitor/buildinfo.json
```diff
@@ -9,8 +9,8 @@
     },
     "zookeeper": {
       "kind": "url_extract",
-      "url": "https://downloads.mesosphere.com/zookeeper/zookeeper-3.4.8.tar.gz",
-      "sha1": "51b61611a329294f75aed82f3a4517a4b6ff116f"
+      "url": "https://downloads.mesosphere.com/zookeeper/zookeeper-3.4.9.tar.gz",
```
cmaloney: Also need to update the mesos bundled zookeeper to 3.4.9 if we want to do this. Want to have one version of zookeeper and zookeeper clients if at all possible.

alberts: Looks like this will have to wait for Mesos 1.2.

alberts: https://issues.apache.org/jira/browse/MESOS-6667

packages/exhibitor/extra/zkServer.sh
cmaloney: The only thing which _ever_ calls this script inside DC/OS is exhibitor. How are people doing weird things they might be depending upon in it? They aren't allowed to modify `/opt/mesosphere`. That means they aren't allowed to edit exhibitor, our systemd units, or any of the environment variables exhibitor is sending to this. They also aren't allowed to add jars to the folders this refers to.

spahl:
This will be needed in 1.9 once it uses 1.2 (currently master)

lingmann:
I definitely think we need to address the zookeeper.out logging issue for 1.9. The Microsoft ACS team has received reports about this being a problem for some users. And if master node disks fill up the entire cluster is in jeopardy. Another potential strategy here is to just expand our logrotate policy and do something similar to what we are doing for the mesos logs.

spahl:
Do we need to couple this to the upgrade? If yes then we need this: https://issues.apache.org/jira/browse/MESOS-6667

lingmann:
Parts of this are being superseded by https://github.com/dcos/dcos/pull/1192

gpaul:
Let's not override `zkServer.sh`? I'd rather we not ship our own, especially now that (as @lingmann says) the zookeeper.log issue was is fixed.

PR #1010, opened 2016-12-01
Merge navstar minuteman

sargun:
- This combines the Navstar and Minuteman processes. We're now running both systems in the same BEAM (Erlang VM)
- It also adds an LB disable flag to the config, and there is a test to make sure this disable works

# Checklist

 - [X] Included a test which will fail if code is reverted but test is not

-Changes to make it so we can merge the processes:
https://github.com/dcos/navstar/commit/6300e1b5ccc011d3067686dcdec30286f99c3f36
```
6300e1b5ccc011d3067686dcdec30286f99c3f36 Merge pull request #35 from dcos/merge-navstar-minuteman
31703266d1e48cc0c85ed3d6d095d9aee12a6f27 Merge Minuteman into navstar
b1e6d06dd57b4f3ec7c0ee3966b01da2e0405eb4 Merge pull request #34 from GoelDeepak/netlink
77e8860b59274ece8e4913fd70d74739453c2620 adds gen_netlink dependency
b0219c7f232f7d05f16889344dfdec13d1c9f1c8 Use lashup keys(MatchSpec) to fetch navstar overlay keys
396de3c334ee3861b82c224518e2170ff0e23063 Fixes circle ci issue
3db265251be8a4b5406edd28aa3e5b2c816f9a27 redesign the state machine
c70a95dedebbf000ee13ecdbaf347fb15fb0cadc Adds circle ci tests for navstar overlay netlink
61d71af7afcaf57e67055a88bbe7343c7c761809 Adds reapplying state to reapply the config periodically
a699a3657d736056c6ae45f667ddb9184892912a minor code refactoring
058fb92c59b6c2bc514b1c066df39fee48b0ca93 removes unused code
7fd3a689b4f9f414b86613d19344a3863ae6847d Minor test cleanup and gen_netlink ref update
3e02f681f9831eaefaa7a3f062aec23850f979b2 Moves configurator to netlink
1b14c04c8251d77f0bc132dd47b2824e2babef8c Adds netlink to navstar overlay
cbcfb7c4f4e50015d7127b4bef81db030fc29f3d Merge pull request #33 from GoelDeepak/dev
efdeab7755141c1b413bfe1805098e1dbfb0b7aa increases batching time to 5 secs
56064e7e4a8f89335115fb568004b27aba02eb7a cleanup
d800742f783af6f6015ac62c0af7d7f8e36fd7e0 adds a timer to kill in case configurator hangs
06621269fde261e22120715d36af2ae25aeb66aa put message queue on heap and set the max heap size
e64e1f0c8c4f45a0eee5feb08cb27bfc54d491db adds a pre condition in circle ci adds typespecs
96083f25257ffe7ab73fb14d5f00891bbd0edba2 Only apply delta config
```


spahl:
@mellenburg can you do a first pass as the tests?

sargun:
This PR requires some touch ups before it's ready to merge, but I don't want to do those things until we know we're near a merge window because it stalls out some other development efforts.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,11 +1,25 @@
...
 
 import requests
 import retrying
 
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
+def config():
+    with open("/opt/mesosphere/etc/expanded.config.json", "r") as f:
+        expanded_config = json.load(f)
```
mellenburg: Please use https://github.com/dcos/dcos/blob/master/pkgpanda/util.py#L123

spahl:
@sargun you are in a merge window. Please get this ready to go:-)

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,40 @@ def _ensure_routable():
...
+    def _check_overlay(hostname, port):
+        uri = 'http://{}:{}/overlay-agent/overlay'.format(hostname, port)
+        resp = requests.get(uri).json()
+        overlays = resp['overlays']
+        assert(len(overlays) > 0)
+        for overlay in overlays:
+            assert(overlay['state']['status'] == 'STATUS_OK')
+        return True
+
+    @retrying.retry(wait_fixed=2000,
```
mellenburg: just move this retry decorator on top of the test and dont use this inner function at all

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,40 @@ def _ensure_routable():
...
+                    stop_max_delay=120 * 1000,
+                    retry_on_result=lambda ret: ret is False,
+                    retry_on_exception=lambda x: True)
+    def _ensure_overlay_ok():
+        for i in cluster.masters:
```
mellenburg: in line these for loops please and place the retry decorator on the entire function as its the same functionally

sargun:
@mellenburg So like:
```

@retrying.retry(wait_fixed=2000,
                stop_max_delay=120 * 1000,
                retry_on_result=lambda ret: ret is False,
                retry_on_exception=lambda x: True)
def test_if_overlay_ok(cluster):
    def _check_overlay(hostname, port):
        uri = 'http://{}:{}/overlay-agent/overlay'.format(hostname, port)
        resp = requests.get(uri).json()
        overlays = resp['overlays']
        assert(len(overlays) > 0)
        for overlay in overlays:
            assert(overlay['state']['status'] == 'STATUS_OK')
        return True

    for i in cluster.masters:
        assert(_check_overlay(i, 5050))
    for i in cluster.all_slaves:
        assert(_check_overlay(i, 5051))
```

sargun:
@mellenburg Can I get another review?

sargun:
```
 ####  #    # # #####     # ##### 
#      #    # # #    #    #   #   
 ####  ###### # #    #    #   #   
     # #    # # #####     #   #   
#    # #    # # #         #   #   
 ####  #    # # #         #   #   
```
maybe?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,11 +1,24 @@
...
 
+from pkgpanda.build import load_json
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
+def config():
+    return load_json('/opt/mesosphere/etc/expanded.config.json')
+
+
+skip_if_lb_enabled = pytest.mark.skipif(config()['enable_lb'] == 'true', reason='Load Balancer enabled')
```
orsenthil: My suggestion is to factor out 

```
def lb_enabled(): 
    return config()['enable_lb'] == 'true'
```

And similar for `lb_disabled`

and then you could use the `@pytest.mark.skipif(lb_enabled()` as a decorator for the tests instead of "factoring out the entire skipif object". 

The advantages I see are:

* This logic if is lb_enabled, lb_disabled can be used independent of using it with skipif
* The skipif decorator stays closer to the test as it is commonly used.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,36 @@ def _ensure_routable():
     return _ensure_routable
 
 
+@retrying.retry(wait_fixed=2000,
```
orsenthil: This retries on 
* AssertionErrors too.
* retries the tests too!

I guess, it was unintentional.  You might have to factor out `_check_overlay` outside of this test code and have a retry application only there (as it does the requests.get call).

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -49,7 +93,7 @@ def test_ip_per_container(cluster):
     """Test if we are able to connect to a task with ip-per-container mode
     """
     # Launch the test_server in ip-per-container mode
-
+    test_if_overlay_ok(cluster)
```
mellenburg: Is this supposed to be here? What is the point of rerunning this test?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -63,9 +107,11 @@ def test_ip_per_container(cluster):
         ensure_routable(cmd, service_points)()
 
 
+@skip_if_lb_disabled
 def test_ip_per_container_with_named_vip(cluster):
     """Test if we are able to connect to a task with ip-per-container mode using named vip
     """
+    test_if_overlay_ok(cluster)
```
mellenburg: It looks like you are trying to use this as a setup fixture. The proper form would be:

```
@pytest.fixture
def overlay_ok():
    # do all of your stuff
```
then, in your test:
```
@skip_if_lb_disabled
def test_ip_per_container_with_named_vip(cluster, overlay_ok):
    # do you test code
```
By naming the fixture as your argument in the test function, pytest will ensure that the code is run before this test. Furthermore, when code is to be repeated (and not expected to change) you can do this with pytest.fixture(scope='session') to only run it once
I think this is what you are intending and is the proper way to use setup fixture

packages/dcos-integration-test/extra/test_networking.py
sargun: @orsenthil I agree.

packages/dcos-integration-test/extra/test_networking.py
sargun: This was intentional. Unfortunately, while the cluster is coming up, it can throw assertions.

sargun:
@branden I've been told not to squash commits during the review process because it makes it harder for people to see what changes have been made since the last review. Shall I do differently?

packages/dcos-integration-test/extra/test_networking.py
sargun: So, right now, test_if_overlay_ok acts an implicit barrier to the tests that might rely on the overlay. Sometimes the overlay takes some time to come up. If someone reorders the tests, it could be bad. I don't really want this to be an independent fixture either.

packages/dcos-integration-test/extra/test_networking.py
orsenthil: My point was, you were `retrying` on 

```
for i in cluster.masters:
    assert(_check_overlay(i, 5050))
for i in cluster.all_slaves:
    assert(_check_overlay(i, 5051))
```

Not sure if that was intentional.

branden:
@sargun I'm OK with waiting until people are happy with the changes in the PR before you clean up history.

packages/dcos-integration-test/extra/test_networking.py
orsenthil: Also, while you are at it replace the variable name i in the first to to `master` and the second for loop to `slave`.

sargun: Yeah, the loops are very lightweight, so it's fine to retry the entire loop, rather than just the individual _check_overlay.

sargun:
Okay, I made the change:
* Squashed commit
* Renamed i to master, and slave

I don't think changing the retry logic is a good idea.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,11 +1,24 @@
...
 
 import requests
+
 import retrying
 
+from pkgpanda.build import load_json
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
+def config():
```
mellenburg: This function isnt used except for in lb_enabled, so just delete it and do load_json in lb_enabled

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,36 @@ def _ensure_routable():
     return _ensure_routable
 
 
+@retrying.retry(wait_fixed=2000,
+                stop_max_delay=120 * 1000,
+                retry_on_result=lambda ret: ret is False,
```
mellenburg: There is no return False so this line does nothing; please remove it. Default is to exit once any return is registered

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,36 @@ def _ensure_routable():
...
+                retry_on_exception=lambda x: True)
+def test_if_overlay_ok(cluster):
+    def _check_overlay(hostname, port):
+        uri = 'http://{}:{}/overlay-agent/overlay'.format(hostname, port)
+        resp = requests.get(uri).json()
+        overlays = resp['overlays']
+        assert(len(overlays) > 0)
+        for overlay in overlays:
+            assert(overlay['state']['status'] == 'STATUS_OK')
+        return True
```
mellenburg: This return True can be removed

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,36 @@ def _ensure_routable():
...
+
+    for master in cluster.masters:
+        assert(_check_overlay(master, 5050))
```
mellenburg: This assert can be removed. just run the function

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -20,6 +33,36 @@ def _ensure_routable():
...
+    for slave in cluster.all_slaves:
+        assert(_check_overlay(slave, 5051))
```
mellenburg: This assert should also be removed

PR #1013, opened 2016-12-02
Cleanup interfaces for passing template urls as well as resolve_configuration

cmaloney:
Cleans up two of the interfaces inside of the codebase. This is work that was done as part of getting to #1009 / late binding, pulled out of that PR so it can be reviewed / land independently.

1. Make resolve_configuration operate purely on sources and targets

This makes it have a more straightforward API and moves the responsibility for
building a source out of user arguments (if that is even needed) to the various
callsites (just two of them)

2. Remove magic global passing of azure and aws storage urls

Previously these were passed by globals when we were doing a release and as arguments when we did a advanced / custom configuration template generation. Now they are always passed as arguments which are injected by release/__init__.py in the case of release for the configuration, and via the normal argument means otherwise.

3. Fixup lingering comments from #956 by replacing the namedtuple and dictionary with just returning a tuple which gets expanded when needed.

# Issues

[DCOS-10453](https://mesosphere.atlassian.net/browse/DCOS-10453)

# Checklist

### DIFF
--- a/release/__init__.py
+++ b/release/__init__.py
@@ -407,12 +407,24 @@ def make_channel_artifacts(metadata):
         # Add templates for the default variant.
         # Use keyword args to make not matching ordering a loud error around changes.
         with logger.scope("Creating {} deploy tools".format(module.__name__)):
+            # TODO(cmaloney): Cleanup by just having this make and pass another source.
+            module_specific_variant_arguments = copy.deepcopy(variant_arguments)
+            for arg_dict in module_specific_variant_arguments.values():
+                if module.__name__ == 'gen.build_deploy.aws':
+                    arg_dict['cloudformation_s3_url_full'] = metadata['cloudformation_s3_url_full']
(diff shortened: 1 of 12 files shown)


release/__init__.py
```diff
@@ -409,12 +409,25 @@ def make_channel_artifacts(metadata):
         # Add templates for the default variant.
         # Use keyword args to make not matching ordering a loud error around changes.
         with logger.scope("Creating {} deploy tools".format(module.__name__)):
+            # TODO(cmaloney): Cleanup by just having this make and pass another source.
+            module_specific_variant_arguments = copy.deepcopy(variant_arguments)
+            for arg_dict in module_specific_variant_arguments.values():
+                if module.__name__ == 'gen.build_deploy.aws':
+                    arg_dict['cloudformation_s3_url_full'] = metadata['cloudformation_s3_url_full']
+                    module_specific_variant_arguments
```
branden: This line can be removed.

PR #1014, opened 2016-12-02
yaml.load -> load_yaml for better error messages, simpler code

cmaloney:
Changes from calling yaml.load() directly to using a wrapper `load_yaml` which does a couple things
1. Makes syntax errors include the filename (re-wrap the exception with more context)
2. Make it so not everyone has to do a `with open() as f; yaml.load(f)`
3. Switch from yaml.load -> yaml.safe_load so that we're always keeping to basic yaml types / not arbitrary code.

Found during / relates to Late Binding #1009  

# Checklist

### DIFF
--- a/release/__init__.py
+++ b/release/__init__.py
@@ -19,7 +19,6 @@ from distutils.version import LooseVersion
 from typing import Optional
 
 import pkg_resources
-import yaml
 
 import gen.build_deploy.util as util
 import pkgpanda
@@ -64,8 +63,7 @@ def expand_env_vars(config):
 
 
 def load_config(filename):
-    with open(filename) as config_file:
-        return expand_env_vars(yaml.load(config_file))
+    return expand_env_vars(pkgpanda.util.load_yaml(filename))
 
 
 def strip_locals(data):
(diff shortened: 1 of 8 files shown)


release/__init__.py
```diff
@@ -64,8 +63,7 @@ def expand_env_vars(config):
 
 
 def load_config(filename):
-    with open(filename) as config_file:
-        return expand_env_vars(yaml.load(config_file))
+    return expand_env_vars(pkgpanda.util.load_yaml(filename))
```
orsenthil: `load_yaml` could be imported and used in the module namespace.

cmaloney: pkgpanda.util is/was already imported and this function isn't used much. My general rule is that unless it's used a number of times (3+) then doing the whole name-spacing in the callsite reduces the amount of work it takes a reader to follow the code back to source (Someone reading the code sees a called function from another module). If it's used a lot then there's an advantage gained from typing less by making a local alias (what from import does). In this case the from import would actually be more code.

orsenthil: The way I see it, when you browse the entire code base, usually traversing multiple files at a time, the consistency in function calls helps while reading the code. Also, usually we don't count the imports as lines of code.

PR #1015, opened 2016-12-02
Adjust DC/OS UI build to properly install dependencies

orlandohohmeier:
Add `unsafe-perm` flag to run DC/OS UI install scripts with proper privileges. Without those privileges the installation will fail to write some of the dependencies (e.g. the Marathon RAML files) with the correct user/permissions.

### DIFF
--- a/packages/dcos-ui/build
+++ b/packages/dcos-ui/build
@@ -4,7 +4,10 @@ cd "/pkg/src/dcos-ui"
 
 npm install -g npm@3.9.1
 
-npm install
+# Run npm install with `unsafe-perm` flag to run install scripts with root
+# privileges. Without those privileges the installation will fail to write some
+# of the dependencies with the correct user/permissions.
+npm --unsafe-perm install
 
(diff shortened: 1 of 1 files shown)


orlandohohmeier:
I was wondering if it is possible to move the dcos-ui build script to our `dcos/dcos-ui` repository so that we use the same build process for all build. 

/cc @spahl @cmaloney @rcorral

cmaloney:
@orlandohohmeier there isn't a way to put it into the repository specified in buildinfo.json currently. There is a certain extent to which the build script (and things like systemd units it installs) is important to actually be reviewed by Foundations.

In terms of the building of individual components, it's fine to just have the build script in the repository call some script which is committed inside the individual component which gives you the same effect. build could do something like just call `build_in_docker.sh` which is committed in the dcos-ui repository to enable the most code sharing

orlandohohmeier:
@cmaloney thanks a bunch for the explanation. I like the idea of moving parts of the build into `dcos/dcos-ui` repository and probably go to refactor our build accordingly in preparation of the next release.

PR #1016, opened 2016-12-02
Use requests paradigms

mellenburg:
- ApiClient was added with the intention of making it easier to handle common requests by extending all requests with some default. However, this is exactly what requests.session does (and it does it much better with less cruft)
- Also adds a Url abstration to that allows easy URL composing
- By using session, a simple BaseAuth object can be bound
- With these improvements, ApiClient only takes a Url and an optional get_node_port which is a call back for determining ports for an API that has different ports on different nodes
- ApiClient.api_request still takes the same arguments (HTTP verb, something to add to path, node=) but also accepts arbitrary default url substitutions like scheme='http' or query='last_cache=0' 

# Checklist

 - [x] Included a test which makes sure cookies are handled by get_user_session as the underlying session will automatically propagate cookies now

### DIFF
--- a/packages/dcos-integration-test/extra/cluster_fixture.py
+++ b/packages/dcos-integration-test/extra/cluster_fixture.py
@@ -1,14 +1,14 @@
 """This file has only one function: to provide a correctly configured
-ClusterApi object that will be injected into the pytest 'cluster' fixture
+DcosApiSession object that will be injected into the pytest 'cluster' fixture
 via the make_cluster_fixture() method
 """
-from test_util.cluster_api import ClusterApi, get_args_from_env
-from test_util.helpers import CI_AUTH_JSON, DcosUser
+from test_util.dcos_api_session import DcosApiSession, DcosUser, get_args_from_env
+from test_util.helpers import CI_CREDENTIALS
 
 
 def make_cluster_fixture():
     args = get_args_from_env()
-    args['web_auth_default_user'] = DcosUser(CI_AUTH_JSON)
-    cluster_api = ClusterApi(**args)
+    args['auth_user'] = DcosUser(CI_CREDENTIALS)
+    cluster_api = DcosApiSession(**args)
     cluster_api.wait_for_dcos()
     return cluster_api
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -177,13 +177,13 @@ def test_octarine_srv(cluster, timeout=30):
 
 def test_pkgpanda_api(cluster):
 
-    def get_and_validate_package_ids(node, path):
-        r = cluster.get(node=node, path=path)
+    def get_and_validate_package_ids(path, node):
+        r = cluster.get(path, node=node)
         assert r.status_code == 200
         package_ids = r.json()
         assert isinstance(package_ids, list)
         for package_id in package_ids:
-            r = cluster.get(node=node, path=path + package_id)
+            r = cluster.get(path + package_id, node=node)
             assert r.status_code == 200
             name, version = package_id.split('--')
             assert r.json() == {'id': package_id, 'name': name, 'version': version}
@@ -207,8 +207,8 @@ def test_pkgpanda_api(cluster):
                 assert package == buildinfo_package
 
     for node in cluster.masters + cluster.all_slaves:
-        package_ids = get_and_validate_package_ids(node, '/pkgpanda/repository/')
-        active_package_ids = get_and_validate_package_ids(node, '/pkgpanda/active/')
+        package_ids = get_and_validate_package_ids('pkgpanda/repository/', node)
+        active_package_ids = get_and_validate_package_ids('pkgpanda/active/', node)
 
         assert set(active_package_ids) <= set(package_ids)
         assert_packages_match_active_buildinfo(active_package_ids)
--- a/packages/dcos-integration-test/extra/test_auth.py
+++ b/packages/dcos-integration-test/extra/test_auth.py
@@ -3,11 +3,6 @@ import subprocess
 import pytest
 
 
-@pytest.fixture(scope='module')
-def noauth_cluster(cluster):
-    return cluster.get_user_session(None)
-
-
 def auth_enabled():
     out = subprocess.check_output([
         '/bin/bash', '-c',
@@ -38,7 +33,7 @@ def test_adminrouter_access_control_enforcement(cluster, noauth_cluster):
 
     # Test authentication with auth cookie instead of Authorization header.
     authcookie = {
-        'dcos-acs-auth-cookie': cluster.web_auth_default_user.auth_cookie}
+        'dcos-acs-auth-cookie': cluster.auth_user.auth_cookie}
     r = noauth_cluster.get(
         '/service/marathon/',
         cookies=authcookie)
--- a/test_util/dcos_api_session.py
+++ b/test_util/dcos_api_session.py
@@ -1,17 +1,21 @@
+""" Utilities for interacting with a DC/OS instance via REST API
+
+Most DC/OS deployments will have auth enabled, so this module includes
+DcosUser and DcosAuth to be attached to a DcosApiSession. Additionally,
+it is sometimes necessary to query specific nodes within a DC/OS cluster,
+so there is ARNodeApiClientMixin to allow querying nodes without boilerplate
+to set the correct port and scheme.
+"""
 import copy
 import logging
 import os
-from urllib.parse import urlparse
+from typing import Optional
 
 import requests
 import retrying
 
-import test_util.helpers
 import test_util.marathon
-
-ADMINROUTER_PORT_MAPPING = {
-    'master': {'http': 80, 'https': 443},
-    'agent': {'http': 61001, 'https': 61002}}
+from test_util.helpers import ApiClientSession, Url
 
 
 def get_args_from_env():
@@ -30,11 +34,107 @@ def get_args_from_env():
         'public_masters': os.environ['PUBLIC_MASTER_HOSTS'].split(','),
         'slaves': os.environ['SLAVE_HOSTS'].split(','),
         'public_slaves': os.environ['PUBLIC_SLAVE_HOSTS'].split(','),
-        'default_os_user': os.getenv('DCOS_DEFAULT_OS_USER', 'root'),
-        'ca_cert_path': os.getenv('DCOS_CA_CERT_PATH', None)}
+        'default_os_user': os.getenv('DCOS_DEFAULT_OS_USER', 'root')}
 
 
-class ClusterApi(test_util.helpers.ApiClient):
+class DcosUser:
+    """A lightweight user representation for grabbing the auth info and stashing it"""
+    def __init__(self, credentials: dict):
+        self.credentials = credentials
+        self.auth_token = None
+        self.auth_cookie = None
+
+    @property
+    def auth_header(self):
+        return {'Authorization': 'token={}'.format(self.auth_token)}
+
+
+class DcosAuth(requests.auth.AuthBase):
+    def __init__(self, auth_token: str):
+        self.auth_token = auth_token
+
+    def __call__(self, request):
+        request.headers['Authorization'] = 'token={}'.format(self.auth_token)
+        return request
+
+
+class ARNodeApiClientMixin:
+    def api_request(self, method, path_extension, *, scheme=None, host=None, query=None,
+                    fragment=None, port=None, node=None, **kwargs):
+        """ Communicating with a DC/OS cluster is done by default through Admin Router.
+        Use this Mixin with an ApiClientSession that requires distinguishing between nodes.
+        Admin Router has both a master and agent process and so this wrapper accepts a
+        node argument. node must be a host in self.master or self.all_slaves. If given,
+        the request will be made to the Admin Router endpoint for that node type
+        """
+        assert self.masters
+        assert self.all_slaves
+        if node is not None:
+            assert port is None, 'node is intended to retrieve port; cannot set both simultaneously'
+            assert host is None, 'node is intended to retrieve host; cannot set both simultaneously'
+            if node in self.masters:
+                # Nothing else to do, master Admin Router uses default HTTP (80) and HTTPS (443) ports
+                pass
+            elif node in self.all_slaves:
+                scheme = scheme if scheme is not None else self.default_url.scheme
+                if scheme == 'http':
+                    port = 61001
+                if scheme == 'https':
+                    port = 61002
+            else:
+                raise Exception('Node {} is not recognized within the DC/OS cluster'.format(node))
+            host = node
+        return super().api_request(method, path_extension, scheme=scheme, host=host,
+                                   query=query, fragment=fragment, port=port, **kwargs)
+
+
+class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
+    def __init__(self, dcos_url: str, masters: list, public_masters: list,
+                 slaves: list, public_slaves: list, default_os_user: str,
+                 auth_user: Optional[DcosUser]):
+        """Proxy class for DC/OS clusters.
+
+        Args:
+            dcos_url: address for the DC/OS web UI.
+            masters: list of Mesos master advertised IP addresses.
+            public_masters: list of Mesos master IP addresses routable from
+                the local host.
+            slaves: list of Mesos slave/agent advertised IP addresses.
+            default_os_user: default user that marathon/metronome will launch tasks under
+            auth_user: use this user's auth for all requests
+                Note: user must be authenticated explicitly or call self.wait_for_dcos()
+        """
+        super().__init__(Url.from_string(dcos_url))
+        self.masters = sorted(masters)
+        self.public_masters = sorted(public_masters)
+        self.slaves = sorted(slaves)
+        self.public_slaves = sorted(public_slaves)
+        self.all_slaves = sorted(slaves + public_slaves)
+        self.zk_hostports = ','.join(':'.join([host, '2181']) for host in self.public_masters)
+        self.default_os_user = default_os_user
+        self.auth_user = auth_user
+
+        assert len(self.masters) == len(self.public_masters)
+
+    @retrying.retry(wait_fixed=2000, stop_max_delay=120 * 1000)
+    def _authenticate_default_user(self):
+        """retry default auth user because in some deployments,
+        the auth endpoint might not be routable immediately
+        after Admin Router is up. DcosUser.authenticate()
+        will raise exception if authorization fails
+        """
+        if self.auth_user is None:
+            return
+        logging.info('Attempting authentication')
+        # explicitly use a session with no user authentication for requesting auth headers
+        r = self.post('/acs/api/v1/auth/login', json=self.auth_user.credentials, auth=None)
+        r.raise_for_status()
+        logging.info('Received authentication blob: {}'.format(r.json()))
+        self.auth_user.auth_token = r.json()['token']
+        self.auth_user.auth_cookie = r.cookies['dcos-acs-auth-cookie']
+        logging.info('Authentication successful')
+        # Set requests auth
+        self.session.auth = DcosAuth(self.auth_user.auth_token)
 
     @retrying.retry(wait_fixed=1000,
                     retry_on_result=lambda ret: ret is False,
@@ -177,8 +277,7 @@ class ClusterApi(test_util.helpers.ApiClient):
 
     def wait_for_dcos(self):
         self._wait_for_adminrouter_up()
-        if self.web_auth_default_user is not None:
-            self._authenticate_default_user()
+        self._authenticate_default_user()
         self._wait_for_marathon_up()
         self._wait_for_zk_quorum()
         self._wait_for_slaves_to_join()
@@ -187,117 +286,54 @@ class ClusterApi(test_util.helpers.ApiClient):
         self._wait_for_dcos_history_data()
         self._wait_for_metronome()
 
-    @retrying.retry(wait_fixed=2000, stop_max_delay=120 * 1000)
-    def _authenticate_default_user(self):
-        """retry default auth user because in some deployments,
-        the auth endpoint might not be routable immediately
-        after adminrouter is up. DcosUser.authenticate()
-        will raise exception if authorization fails
+    def copy(self):
+        """ Create a new client session without cookies, with the authentication intact.
         """
-        self.web_auth_default_user.authenticate(self)
-        self.default_headers.update(self.web_auth_default_user.auth_header)
-
-    def __init__(self, dcos_url, masters, public_masters, slaves, public_slaves,
-                 default_os_user, web_auth_default_user=None, ca_cert_path=None):
-        """Proxy class for DC/OS clusters.
-
-        Args:
-            dcos_url: address for the DC/OS web UI.
-            masters: list of Mesos master advertised IP addresses.
-            public_masters: list of Mesos master IP addresses routable from
-                the local host.
-            slaves: list of Mesos slave/agent advertised IP addresses.
-            default_os_user: default user that marathon/metronome will launch tasks under
-            web_auth_default_user: use this user's auth for all requests
-                Note: user must be authenticated explicitly or call self.wait_for_dcos()
-            ca_cert_path: (str) optional path point to the CA cert to make requests against
-        """
-        # URL must include scheme
-        assert dcos_url.startswith('http')
-        parse_result = urlparse(dcos_url)
-        self.scheme = parse_result.scheme
-        self.dns_host = parse_result.netloc.split(':')[0]
-
-        # Make URL never end with /
-        self.dcos_url = dcos_url.rstrip('/')
-
-        super().__init__(
-            default_host_url=self.dcos_url,
-            api_base=None,
-            ca_cert_path=ca_cert_path,
-            get_node_url=self.get_node_url)
-        self.masters = sorted(masters)
-        self.public_masters = sorted(public_masters)
-        self.slaves = sorted(slaves)
-        self.public_slaves = sorted(public_slaves)
-        self.all_slaves = sorted(slaves + public_slaves)
-        self.zk_hostports = ','.join(':'.join([host, '2181']) for host in self.public_masters)
-        self.default_os_user = default_os_user
-        self.web_auth_default_user = web_auth_default_user
-
-        assert len(self.masters) == len(self.public_masters)
+        new = copy.deepcopy(self)
+        new.session.cookies.clear()
+        return new
 
     def get_user_session(self, user):
-        """Returns a copy of self with auth headers set for user
+        """Returns a copy of this client but with auth for user (can be None)
         """
-        new_session = copy.deepcopy(self)
-        # purge old auth headers
-        if self.web_auth_default_user is not None:
-            for k in self.web_auth_default_user.auth_header.keys():
-                if k in new_session.default_headers:
-                    del new_session.default_headers[k]
-        # if user is given then auth and update the headers
-        new_session.web_auth_default_user = user
+        new = self.copy()
+        new.session.auth = None
+        new.auth_user = None
         if user is not None:
-            new_session._authenticate_default_user()
-        return new_session
-
-    def get_node_url(self, node, port=None):
-        """
-        Args:
-            node: (str) the hostname of the node to be requested from, if node=None,
-                then public cluster address (see environment DCOS_DNS_ADDRESS)
-            port: (int) port to be requested at. If port=None, the default port
-                for that given node type will be used
-        Returns:
-            fully-qualified URL string for this API
-        """
-        if node in self.masters:
-            role = 'master'
-        elif node in self.all_slaves:
-            role = 'agent'
-        else:
-            raise Exception('Node {} is not recognized within the DC/OS cluster'.format(node))
-        if port is None:
-            port = ADMINROUTER_PORT_MAPPING[role][self.scheme]
-        # do not explicitly declare default ports
-        if (port == 80 and self.scheme == 'http') or (port == 443 and self.scheme == 'https'):
-            netloc = node
-        else:
-            netloc = '{}:{}'.format(node, port)
-        return '{}://{}'.format(self.scheme, netloc)
+            new.auth_user = user
+            new._authenticate_default_user()
+        return new
 
     @property
     def marathon(self):
-        marathon_client = test_util.marathon.Marathon(
-            default_host_url=self.dcos_url,
+        return test_util.marathon.Marathon(
+            default_url=self.default_url.copy(path='marathon'),
             default_os_user=self.default_os_user,
-            default_headers=self.default_headers,
-            ca_cert_path=self.ca_cert_path,
-            get_node_url=self.get_node_url)
-        return marathon_client
+            session=self.copy().session)
 
     @property
     def metronome(self):
-        return self.get_client('/service/metronome/v1')
+        new = self.copy()
+        new.default_url = self.default_url.copy(path='service/metronome/v1')
+        return new
+
+    @property
+    def health(self):
+        new = self.copy()
+        new.default_url = self.default_url.copy(query='cache=0', path='system/health/v1')
+        return new
 
     @property
     def logs(self):
-        return self.get_client('/system/v1/logs')
+        new = self.copy()
+        new.default_url = self.default_url.copy(path='system/v1/logs')
+        return new
 
     @property
     def metrics(self):
-        return self.get_client('/system/v1/metrics/v0')
+        new = self.copy()
+        new.default_url = self.default_url.copy(path='/system/v1/metrics/v0')
+        return new
 
     def metronome_one_off(self, job_definition, timeout=300, ignore_failures=False):
         """Run a job on metronome and block until it returns success
--- a/test_util/helpers.py
+++ b/test_util/helpers.py
@@ -8,6 +8,8 @@ import os
 import tempfile
 import time
 from collections import namedtuple
+from typing import Union
+from urllib.parse import urlsplit, urlunsplit
 
 import requests
 import retrying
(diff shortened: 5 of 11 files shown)


mellenburg:

test_util/helpers.py
```diff
@@ -23,81 +24,139 @@ def path_join(p1, p2):
...
-        logging.debug('Reqeust kwargs: {}'.format(kwargs))
-        logging.debug('Request headers: {}'.format(headers))
-        return requests.request(method, request_url, headers=headers, **kwargs)
+            assert port is None, 'node is intended to retrieve port; cannot set both simultaneously'
+            assert host is None, 'node is intended to retrieve host; cannot set both simultaneously'
+            assert self._get_node_port is not None, 'get_node_netloc must be specified for this ApiClient!'
+            port = self._get_node_port(node)
+            # do not explicitly declare default ports
+            host = node
+            if (port == 80 and self.url.scheme == 'http') or (port == 443 and self.url.scheme == 'https'):
```
orsenthil: why do we reset these valid ports to None?

test_util/helpers.py
```diff
@@ -189,3 +248,15 @@ def remove_file():
     # Attempt to remove the file upon normal interpreter exit.
     atexit.register(remove_file)
     return temp_path
+
+
+def lazy_property(property_fn):
```
orsenthil: Can  `functools.lru_cache`  be used here? 

https://docs.python.org/3/library/functools.html#functools.lru_cache

Or is `_lazy_prop` completely different?

test_util/helpers.py
mellenburg: That was @cmaloney 's preference. I am agnostic in this matter

test_util/helpers.py
mellenburg: I think that should work. Lemme play around with it... Good suggestion!

test_util/helpers.py
orsenthil: oh no. :(

mellenburg:
@jgehrcke PTAL

packages/dcos-integration-test/extra/cluster_fixture.py
```diff
@@ -21,10 +20,11 @@ def make_cluster_fixture():
     #   "iat": 1460164974
     # }
     auth_json = {'token': 'eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6Ik9UQkVOakZFTWtWQ09VRTRPRVpGTlRNMFJrWXlRa015Tnprd1JrSkVRemRCTWpBM1FqYzVOZyJ9.eyJlbWFpbCI6ImFsYmVydEBiZWtzdGlsLm5ldCIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJpc3MiOiJodHRwczovL2Rjb3MuYXV0aDAuY29tLyIsInN1YiI6Imdvb2dsZS1vYXV0aDJ8MTA5OTY0NDk5MDExMTA4OTA1MDUwIiwiYXVkIjoiM3lGNVRPU3pkbEk0NVExeHNweHplb0dCZTlmTnhtOW0iLCJleHAiOjIwOTA4ODQ5NzQsImlhdCI6MTQ2MDE2NDk3NH0.OxcoJJp06L1z2_41_p65FriEGkPzwFB_0pA9ULCvwvzJ8pJXw9hLbmsx-23aY2f-ydwJ7LSibL9i5NbQSR2riJWTcW4N7tLLCCMeFXKEK4hErN2hyxz71Fl765EjQSO5KD1A-HsOPr3ZZPoGTBjE0-EFtmXkSlHb1T2zd0Z8T5Z2-q96WkFoT6PiEdbrDA-e47LKtRmqsddnPZnp0xmMQdTr2MjpVgvqG7TlRvxDcYc-62rkwQXDNSWsW61FcKfQ-TRIZSf2GS9F9esDF4b5tRtrXcBNaorYa9ql0XAWH5W_ct4ylRNl3vwkYKWa4cmPvOqT5Wlj9Tf0af4lNO40PQ'}  # noqa
-    if 'DCOS_AUTH_JSON_PATH' in os.environ:
```
cmaloney: this is changing the wrapper / caller arguments a bit, have we updated all call sites? How do we keep the CI jobs compatible with old + new config?

test_util/cluster_api.py
```diff
@@ -227,11 +223,10 @@ def _authenticate_default_user(self):
         will raise exception if authorization fails
         """
         self.web_auth_default_user.authenticate(self)
-        self.default_headers.update(self.web_auth_default_user.auth_header)
+        self.session.auth = self.web_auth_default_user.auth
 
     def __init__(self, dcos_url, masters, public_masters, slaves, public_slaves,
-                 dns_search_set, provider, auth_enabled, default_os_user,
-                 web_auth_default_user=None, ca_cert_path=None):
+                 dns_search_set, provider, default_os_user, web_auth_default_user=None):
```
cmaloney: why can we drop ca_cert_path here?

test_util/cluster_api.py
```diff
@@ -271,69 +253,55 @@ def __init__(self, dcos_url, masters, public_masters, slaves, public_slaves,
...
         """
-        new_session = copy.deepcopy(self)
-        # purge old auth headers
-        if self.web_auth_default_user is not None:
-            for k in self.web_auth_default_user.auth_header.keys():
-                if k in new_session.default_headers:
-                    del new_session.default_headers[k]
-        # if user is given then auth and update the headers
-        new_session.web_auth_default_user = user
+        new = self.get_client()
```
cmaloney: why not have get_client take the user stuff? Seems like that would be a reasonable API for get_client() to have (Taking both a path and a user optionally). Then this just becomes a get_client() wrapper (for API compatibility for now).

test_util/helpers.py
cmaloney: This default port stuff I thing is still valuable and relevant. There's usually a reason why code exists and people put time / effort into writing, reviewing, and reasoning about it. Try and think through _why_ someone might have done this and what value it added. If you can't figure it out, then ask. Give some respect / care for the coworkers who did it and what they might have been thinking about, it's not always obvious but it helps relations a lot.

If we didn't do this, we'd generate urls like `http://<cluster_address>:80/foobar` which does not look like how our customers will actually address our services most the time, they will use addresses like `http://<cluster_address>/foobar.`

It also adds more noise that people have to read / comprehend. When I look at something like https://example.com:80/foobar reading through the test log my mind asks "Why is it specifying a port?" "what does that do?" "There must be a reason it's being specified". Really, we shouldn't be specifying it because we don't actually mean / want to. Specifying it adds noise readers / viewers have to parse.

test_util/helpers.py
```diff
@@ -23,81 +24,139 @@ def path_join(p1, p2):
...
+            if (port == 80 and self.url.scheme == 'http') or (port == 443 and self.url.scheme == 'https'):
+                port = None
 
-    def get_client(self, path, default_headers=None):
-        new_client = copy.deepcopy(self)
-        if default_headers is not None:
-            new_client.default_headers.update(default_headers)
-        new_client.api_base = path_join(self.api_base, path) if self.api_base is not None else path
-        return new_client
+        final_path = path_join(path if path else self.url.path, path_ex)
```
cmaloney: Why do we allow changing / editing the base path in calls to this? This also used to take `path` as the argument for "Extra path to add", changing that to path_ex globally seems error prone (Old code which used kwargs would silently just start doing the wrong thing).

test_util/helpers.py
```diff
@@ -189,3 +248,15 @@ def remove_file():
     # Attempt to remove the file upon normal interpreter exit.
     atexit.register(remove_file)
     return temp_path
+
+
+def lazy_property(property_fn):
+    cache_name = '{}_cached'.format(property_fn.__name__)
```
cmaloney: `__<name>_cached` so that we make this a private member plese using python's underscore public / private interface standard

test_util/helpers.py
orsenthil: The original comment was a question based on the thought that this code is in
1. `test_util/helpers.py` which is used internally only in test code.
2. This special case is adding 2-3 lines of code, which can be avoided.

Thanks for sharing the reasons. Acknowledge the preferences.

packages/dcos-integration-test/extra/cluster_fixture.py
mellenburg: This ended up not being used at all. Its only purpose was to allow EE to pass in its authentication to these tests. We refactored EE to inject its cluster_fixture.py directly into this test, so this hacky environment passing can RIP

test_util/helpers.py
cmaloney: @orsenthil we don't want a max size or anything (We want to always cache the lazy properties after first call) like that, and this isn't parameterized like lru_cache works.

test_util/cluster_api.py
mellenburg: because it doesnt need to be defined to create a ApiClient now. The ca_cert_path and verification is a part of the requests.session object and doesnt need to be explicitly handled until something actually wants to use that (which we can do only when security is enabled). The hook had to go here before to allow downstream (EE) to use this without having to change the request pattern with a hack

test_util/helpers.py
mellenburg: The reason this was explicit was to accommodate the Url abstraction. The  idea is that when writing tests, you might want to just enforce a single substitution to any part of the URL used by the default client. Thus, the request wrapper accepts the same paraeters as Url and it would be very confusing to have path mean two different things. Calling this what is really is seems like a good improvement to me

mellenburg: Also, I do change the old code that uses this as kwarrgs, so I know it causes some problem, but its worth it, so that we dont have to say "thats not _really_ the path"

test_util/cluster_api.py
mellenburg: I don't see what we gain by conflating the ApiClient (generic) with the DcosUser abstraction. The latter could be made generic as well, but I don't see the major use for that. I really want to keep ApiClient  as a simple, generic Url and requests.session handler.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -178,12 +178,12 @@ def test_octarine_srv(cluster, timeout=30):
 def test_pkgpanda_api(cluster):
 
     def get_and_validate_package_ids(node, path):
```
orsenthil: Would you prefer refactoring this `get_and_valid_package_ids` signature to read `path, node` because that's how it is being invoked further down and will help greatly in readability.

test_util/helpers.py
```diff
@@ -23,81 +24,139 @@ def path_join(p1, p2):
 
 class DcosUser:
     """A lightweight user representation."""
-    def __init__(self, auth_json):
+    def __init__(self, auth_json: dict):
```
orsenthil: +1 on the type specification of `dict`.

mellenburg:
@orsenthil thanks for the feedback; all comments addressed

orsenthil:
LGTM.

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
+        return clone
 
-        if self.api_base:
-            path = path_join(self.api_base, path)
+    def api_request(self, method, path_ex, scheme=None, host=None, query=None,
+                    fragment=None, path=None, port=None, node=None, **kwargs):
+        """ Direct wrapper for session.request. Returns request.Response
+        Args:
+            method: the HTTP request method to be used
+            path_ex: the extension to the path that is set as the default Url
```
jgehrcke: Can we call this `path_extension`? :)

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
-        If self.ca_cert_path is set, this method will pass it to requests
         Args:
-            method: (str) name of method for requests.request
-            path: see get_url()
-            host_url: override the client's default host URL
-            node: if a get_node_url function is set, node and port will be passed to this function
-                instead of setting a host_url or using the default
-            port: can be used with node with get_node_url is set
-            **kwargs: any optional arguments to be passed to requests.request
+            url: Url object to which requests will be made
```
jgehrcke: So by looking at the code below here it seems to me that the ApiClient will _always_ use this url as full URL, optionally extended with a `path_extension`, and optionally overridden with a different scheme or query etc.

So, it's not really the "Url object to which requests will be made", but more something in the lines of "default URL used by this client, optionally prepended with a path extension or modified upon `client.api_request()` invocation."

Can we call it `default_url`? :)

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
-        headers = copy.copy(self.default_headers)
+        self.url = url
+        self.session = requests.Session()
+        self._get_node_port = get_node_port
 
-        # allow kwarg to override verification so client can be used generically
-        if self.ca_cert_path and 'verify' not in kwargs:
-            kwargs['verify'] = self.ca_cert_path
+    def get_client(self, scheme=None, host=None, path=None, query=None, fragment=None, port=None):
+        """Takes the same arguments as Url.get_url
```
jgehrcke: Why not `get_client(self, url)` with `url` being an object of type `Url`? Then there is only one interface to maintain.

In an unrelated note, it would be great of the docstring could say  more about the purpose of this method. It looks like the fundamental idea is to be able to re-use existing ApiClient objects, but mutate their `url` attribute. Is that the core idea? In that case I don't quite like the name `clone`.

Just a proposal, but what do you think about having an _actual_ `.clone()` method that does what it says and then consider `.url` (or `default_url`) as public interface?

In _using_ code, this could look like this (I find this rather expressive):

```
# Generate an actual clone of this client.
new_client = apiclient.clone()

# Override its url prefix:
new_client.default_url = Url(...)
```

Of course one could also do `new_client = apiclient.clone(new_default_url=...)`, which I also would find expressive enough.

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
+        """
+        clone = copy.deepcopy(self)
+        clone.url = self.url.get_url(scheme=scheme, host=host, path=path, query=query, fragment=fragment, port=port)
+        return clone
 
-        if self.api_base:
-            path = path_join(self.api_base, path)
+    def api_request(self, method, path_ex, scheme=None, host=None, query=None,
+                    fragment=None, path=None, port=None, node=None, **kwargs):
+        """ Direct wrapper for session.request. Returns request.Response
```
jgehrcke: `requests.Response`? (with `s`)

jgehrcke: I am sure you checked this yourself, but for the record: there is no conflict between these many keyword argument names and those consumed by `requests.request` (http://docs.python-requests.org/en/master/api/#requests.request).

That's obviously important, might be worth a comment.

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
+        Args:
+            method: the HTTP request method to be used
+            path_ex: the extension to the path that is set as the default Url
 
+        Keyword Args:
+            All the named keyword args except node are identical to Url.get_url
```
jgehrcke: .... and they are used for... "(partially) overriding the default URL settings"?

As a general note, that's a super flexible, open, powerful interface. I wonder if it's too flexible. Anyway, seems to work, and I like -- as discussed -- that we expose the raw requests interface.

Do we have implicit validation that catches unexpected keyword arguments? (`self.session.request(method, request_url, **kwargs)` further down below probably errors out for unexpected arguments, does it?).

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
-            host_url = host_url if host_url else self.default_host_url
-        request_url = path_join(host_url, path)
-        headers.update(kwargs.pop('headers', {}))
-        logging.info('Request method {}: {}'.format(method, request_url))
-        logging.debug('Reqeust kwargs: {}'.format(kwargs))
-        logging.debug('Request headers: {}'.format(headers))
-        return requests.request(method, request_url, headers=headers, **kwargs)
+            assert port is None, 'node is intended to retrieve port; cannot set both simultaneously'
+            assert host is None, 'node is intended to retrieve host; cannot set both simultaneously'
+            assert self._get_node_port is not None, 'get_node_netloc must be specified for this ApiClient!'
```
jgehrcke: If it helps, there always is `self.__class__.__name__` for not being forced to hard-code class names into strings (easier to maintain).

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
+            port = self._get_node_port(node)
+            # do not explicitly declare default ports
+            host = node
+            if (port == 80 and self.url.scheme == 'http') or (port == 443 and self.url.scheme == 'https'):
+                port = None
```
jgehrcke: I agree this is important. We don't want to have `http://<cluster_address>:80/foo` URLs by default.

Buuuut. If I _want_ to have an ApiClient interaction with such a URL -- how can I achieve this? Just a thought -- no need to answer this w/o having an explicit use case.

test_util/helpers.py
jgehrcke: Here I am now actually lost and do not understand anymore how we're able to precisely control the URL path. Now it kinda seems to me that we can basically override anything everywhere. Re-thinking concepts, renaming variables, updating docstrings is the only way to resolve that confusion.

jgehrcke: > the request wrapper accepts the same paraeters as Url 

Maybe changing _that_ is the key (maybe `api_request()` should not receive `path` argument, but _just_ `path_extension`).

Guarantees only help if they are _simple_ (easy to understand). If an ApiClient object _guarantees_ to keep a certain URL "path prefix" constant, then that's simple. This has semantics that allows for writing solid tests. Then I can keep in mind "a client object actually _represents_ certain URL settings and client state, such as authentication.

If the URL path prefix later on extensible with a suffix, that's also easy (and actually _required_ to make ApiClient usable).

However, if we can later on override this URL prefix in `api_request()`, that's not simple anymore. Then, there are actually two methods for achieving the same thing: set `path` _or_ `path_extension`. Then, the ApiClient object actually represents a little less state that I can rely on. And I wonder why I should inject URL settings in the first place, upon ApiClient construction.

Pretty sure that needs to be consolidated, so that we get closer to the principle of "_one_ way to achieve a given thing."

test_util/helpers.py
mellenburg: @jgehrcke I agree with you, however, I notice in a lot of our code, people tend towards labeling dicts that are derived from or intended to be JSON with _json...

test_util/helpers.py
```diff
@@ -23,81 +24,137 @@ def path_join(p1, p2):
...
         logging.info('Attempting authentication')
         # explicitly use a session with no user authentication for requesting auth headers
-        if cluster.web_auth_default_user:
-            post = cluster.get_user_session(None).post
-        else:
-            post = cluster.post
-        r = post('/acs/api/v1/auth/login', json=self.auth_json)
+        r = cluster.post('/acs/api/v1/auth/login', json=self.auth_json, auth=None)
         r.raise_for_status()
         logging.info('Received authorization blob: {}'.format(r.json()))
```
mellenburg: True. Damn the fools that decided to start using the header called 'Authorization' for authentication everywhere :P

test_util/helpers.py
mellenburg: Requests definitely comes through with raising an error for unexpected kwargs

test_util/helpers.py
mellenburg: As written there will be no way to use ApiClient with node keyword set and port=80, however when not using the node keyword, you can set port=80 or 443 all you desire.

For this reason of a somewhat awkard API, I originally implemented this with a get_node_url function instead of get_node_port. Given that all it was really adding was the port.

I think I should go back to get_node_url as to keep ApiClient as generic as possible

packages/dcos-integration-test/extra/test_auth.py
```diff
@@ -3,16 +3,11 @@
 
 @pytest.fixture(scope='module')
 def auth_cluster(cluster):
-    if not cluster.auth_enabled:
+    if cluster.session.auth is None:
```
mellenburg: I agree. I have long been annoyed by the vagueness of the 'cluster' name. It has survived simply for the reason that I didnt want to deal with the fallout of having to cleanup
```
sed -i 's/cluster/cluster_api_client/g' dcos-integration-tests/*
```
but, given what this refactor is about, this seems to be a particularly appropriate time to tackle this

test_util/cluster_api.py
```diff
@@ -271,69 +253,55 @@ def __init__(self, dcos_url, masters, public_masters, slaves, public_slaves,
...
+        new.session.auth = None
+        new.session.cookies.clear()
```
mellenburg: Clear is required because I think we will get auth rejection if we send requests from a session that has cookies set from a previous user while supplying an Authorization header for the new user

mellenburg:
@jgehrcke RE: using requests.session vs requests.request
I agree the magic of session (e.g. cookie setting) is somewhat worrisome, but the motivation for using session was how clean its interface is. If we don't use session, then we need to pass through headers and ca_cert explicitly. Same goes for whatever session defaults may be used in the future. things like ca_cert and auth are easy enough to track with class member variables, but headers is where it gets really nasty, whereas session has a very nice override mechanism. For example, when not using session we have to manicure headers in a pretty awkward way, whereas with session, we can blindly pass all kwargs along and only do the custom logic we add.

Also, if we tried to do this with PreparedRequest we could remove most of the implicit magic, but then we would be left with a ton of boiler plate for holding onto requests defaults, not to mention a somewhat tedious constructor would be needed.

E.G. I had to add ca_cert_path everywhere in the old implementation despite nothing in this repo using it simply because if I didnt, I would have to make children to do magic to make it work. Now I can remove that kwarg everywhere and in EE when it actually matters, I just have to do:
```
def EE_cluster():
    c = ClusterApiClient()
    c.ca_cert_path = ca_cert_path
    return c
```
and everything will work beautifully with no code addition 

TL;DR Yes, session has a little more magic than I would like, but the interface slimming is super valuable. I would rather add a little bit of extra logic to explicitly remove magic from session than make my own interface with its whole slew of testing and naming issues

packages/dcos-integration-test/extra/test_auth.py
jgehrcke: :rocket:

test_util/cluster_api.py
jgehrcke: Yeah, reuse/mixing/side effects among sessions are to be conceptually prevented. Can we just create a fresh session object here?

jgehrcke: A little background nevertheless: a well thought-through (and probably stable) implementation detail in admin router is that the `Authorization` header always takes precedence over the authentication cookie. It's however not a thing that the integration tests should rely on.

Which brings me to a _really_ important point here: 

The _only_ means of T

jgehrcke:
@mellenburg

> I agree the magic of session (e.g. cookie setting) is somewhat worrisome, but the motivation for using session was how clean its interface is [...]

:+1: for that trade-off.

>  I would rather add a little bit of extra logic to explicitly remove magic from session than make my own interface with its whole slew of testing and naming issues

I am willing to help us go down this road.. :) The first thing that comes to my mind here is, as you also say, the implicit handling of cookies. In general, we would love to take control of every single HTTP request in a well-readable, explicit form. We do generally not wish to see side effects among HTTP requests. Especially we do not want an HTTP response have a side effect on a subsequent request w/o _explicitly_ writing that down in the test code. However, cookies do exactly that.

For instance, I'd love to be really sure that the authentication in tests is by default controlled by _only_ the `Authorization` header in a request. Don't want to see an authentication cookie injected somewhere by accident! Of course we also have to have tests that test the effect of the authentication cookie (and we do have those), but that must be explicitly readable from the test code.

I think we should, by default, disarm the cookie management part of the `requests.Session` object used in the ApiClient.

In the `requests.Session` response generator this function is used to extract cookies from the response(s):

    extract_cookies_to_jar(prepared_request._cookies, req, resp.raw)

(https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L159)

`extract_cookies_to_jar` in turn reads cookies from the response via the `extract_cookies()` method of the jar:

    jar.extract_cookies(res, req)

(https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L135)

The jar is a `RequestsCookieJar` which is API-compatible with `http.cookielib.CookieJar` in the Python 3 stdlib. It inherits `extract_cookies()` from there (https://github.com/kennethreitz/requests/blob/master/requests/cookies.py#L174).

`http.cookielib.CookieJar.extract_cookies(res, req)` is documented here: https://docs.python.org/3.5/library/http.cookiejar.html#http.cookiejar.CookieJar.extract_cookies

That method is meant to modify the jar, it returns `None`. So, by turning _this_ method into a noop we can easily create an API-compatible cookie jar that simply discards cookies incoming via `extract_cookies()`.

Minimal working example (I find the result kinda cute):

```
import requests
from requests.cookies import RequestsCookieJar


class DiscardCookieJar(RequestsCookieJar):

    def extract_cookies(self, req, resp):
        ...


class DiscardCookiesSession(requests.Session):

    def __init__(self):
        super().__init__()
        self.cookies = DiscardCookieJar()


with requests.Session() as s:
    s.get('http://httpbin.org/cookies/set?foo=bar')
    print('cookies in %r: %s' % (s, s.cookies))


with DiscardCookiesSession() as s:
    s.get('http://httpbin.org/cookies/set?foo=bar')
    print('cookies in %r: %s' % (s, s.cookies))
```

Output:
```
cookies in <requests.sessions.Session object at 0x7f6353c59c50>: <RequestsCookieJar[<Cookie foo=bar for httpbin.org/>]>
cookies in <__main__.DiscardCookiesSession object at 0x7f635f329908>: <DiscardCookieJar[]>
```

test_util/helpers.py
jgehrcke: Well, the ambiguity of `auth` is more worrisome to me than the `_json` suffix. So, I'd prefer `credentials`. If you then prefer `credentials_json`, well -- ok.

mellenburg:
@orsenthil @jgehrcke I did another pass on this, PTAL

test_util/helpers.py
```diff
@@ -28,107 +30,136 @@
...
-class DcosUser:
-    """A lightweight user representation."""
-    def __init__(self, auth_json):
-        self.auth_json = auth_json
-        self.auth_header = {}
-        self.auth_token = None
-        self.auth_cookie = None
+class Url:
+    """URL abstraction to allow convenient substitution of URL anatomy
+    without having to copy and disect the entire original URL
```
orsenthil: dissect?

test_util/helpers.py
```diff
@@ -28,107 +30,136 @@
...
-    get = functools.partialmethod(api_request, 'get')
-    post = functools.partialmethod(api_request, 'post')
-    put = functools.partialmethod(api_request, 'put')
-    delete = functools.partialmethod(api_request, 'delete')
-    options = functools.partialmethod(api_request, 'options')
-    head = functools.partialmethod(api_request, 'head')
-    patch = functools.partialmethod(api_request, 'patch')
+        return self.session.request(method, request_url, **kwargs)
+
+    @functools.wraps(api_request)
```
orsenthil: 👍 

These are easier to think and reason about.

test_util/test_upgrade_vpc.py
```diff
@@ -144,14 +144,14 @@ def main():
 
     master_list = [h.private_ip for h in cluster.masters]
 
-    cluster_api = ClusterApi(
+    cluster_api = DcosApiSession(
         'http://{ip}'.format(ip=cluster.masters[0].public_ip),
         master_list,
         master_list,
         [h.private_ip for h in cluster.agents],
         [h.private_ip for h in cluster.public_agents],
```
orsenthil: this corresponds to 'slaves' that is typed as 'str'. I guess, that should be list in DcosApiSession.

test_util/test_usage.py
```diff
@@ -25,19 +25,44 @@ def trivial_env(monkeypatch):
...
+    assert cluster_2.session.auth.auth_token == 'bar'
+    # check cleared cookie
+    assert cluster_2.session.cookies.get('dcos-acs-auth-cookie') is None
+    # make no auth session from use session
+    cluster_none = cluster_2.get_user_session(None)
+    assert cluster_none.session.auth is None
+    assert len(cluster_none.session.cookies.items()) == 0
+
+
+class MyTestClient(ApiClientSession):
```
orsenthil: Is the purpose of this test is to exercise 'super().api_request`?

1. This relies on external network. is it a good idea to do this without the network requirement? ( Just run a simple HTTP server class in the thread).

2. The point of 404 and then 200 is not obvious at all. Why do this in a test?

test_util/test_usage.py
```diff
@@ -25,19 +25,44 @@ def trivial_env(monkeypatch):
...
+    """Test wrapper that will retry the request with a deleted path
+    when the first request returns 404
+    """
+    def api_request(self, method, path_extension, **kwargs):
+        r = super().api_request(method, path_extension, **kwargs)
+        if r.status_code == 404:
+            r = super().api_request(method, '', **kwargs)
+        return r
+
+
```
orsenthil: Also, do we need more tests to exercise the various requests method that are wrapped with api_session?
For .e.g, after mocking ApiClientSessions's api_request, once the `get', `post` and any of the other methods are called, it can be verified that `api_request` is in-turn called with proper arguments (especially, the HTTP verb).

test_util/helpers.py
```diff
@@ -28,107 +30,136 @@
...
+
         Args:
-            method: (str) name of method for requests.request
-            path: see get_url()
-            host_url: override the client's default host URL
-            node: if a get_node_url function is set, node and port will be passed to this function
-                instead of setting a host_url or using the default
-            port: can be used with node with get_node_url is set
-            **kwargs: any optional arguments to be passed to requests.request
+            method: the HTTP request method to be used
```
orsenthil: suggestion: `method: the HTTP verb`.

test_util/dcos_api_session.py
```diff
@@ -30,11 +30,111 @@ def get_args_from_env():
...
+
+    def authenticate(self, dcos_url: str):
+        logging.info('Attempting authentication')
+        # explicitly use a session with no user authentication for requesting auth headers
+        r = requests.post(path_join(dcos_url, '/acs/api/v1/auth/login'), json=self.credentials, auth=None)
+        r.raise_for_status()
+        logging.info('Received authentication blob: {}'.format(r.json()))
+        self.auth_token = r.json()['token']
+        self.auth_cookie = r.cookies['dcos-acs-auth-cookie']
+        logging.info('Authentication successful')
```
mellenburg: getting a 200 back from the post is actually the indicator that authentication was successful. In the off-chance that we get a successful error code, but a broken payload, I wait until we have what is necessary to auth other requests until formally logging that Authentication is successful. Actual prepping of requests is done by attaching DcosAuth to DcosApiSession.session

test_util/dcos_api_session.py
```diff
@@ -30,11 +30,111 @@ def get_args_from_env():
...
+class DcosAuth(requests.auth.AuthBase):
+    def __init__(self, auth_token: str):
+        self.auth_token = auth_token
+
+    def __call__(self, request):
+        request.headers['Authorization'] = 'token={}'.format(self.auth_token)
+        return request
+
+
+class ARNodeApiClientMixin:
```
mellenburg: >   the specific scenario of sending node argument to api_request in addition to host and port
That is not accurate. This is for when you are supplying node INSTEAD of host and port

> Can this be just extracted to a function call and api_request calls when the node argument (let's call it admin_router is not None, and we extract the updated host and port from it).

Once again, I don't think you read this right, the whole point of this extra node argument is that user's want to do something like this:
```
for node in cluster.all_slaves + cluster.masters:
    cluster.get('my/master/or/slave/endpoint', node=node)
```
those member variables are not provided with any knowledge of what port admin router is running on. If the user needs to supply the port then this whole thing is pointless

The reason this is a mixin is a is because we may want to create more specialized clients that need to know what a node is when spun off of a DcosApiSession.. For example, consider the test_util.marathon module, if we want to make a similar thing for health or logs (e.g. dump these logs when the test fails) then how do we teach that Api what a node is? Do we lock it into being a child of DcosApiSession? No, because it should be a sibling. Then we can do this:
```
class HealthClient(ApiClientSession): pass
# in reality there would be a lot of helpers and stuff instead of pass

# can be instantiated with only a url to health and not know anything about DC/OS
health = HealthClient(Url('http://dcos.dcos/system/health'))
```
```
from test_util.health import HealthClient

class HealthNodeApiClientSession(ARNodeApiClientMixin, HealthClient): pass
# Literally all the code you need to make the node-aware HealthClient

class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
     # ...

    @lazy_property
    def health(self):
         return HealthNodeApiClientSession(self.default_url.copy(path='system/health'))

dcos_api_session = DcosApiSession(*my_api_args)
dcos_api_session.health.get('some/endpoint', node=cluster.slave[0])
```
I hope this makes the value of the Mixin logic clear. A separate, independent health utility can be written free of all the junk in DcosApiSession and still get the benefits of knowing about how to make authenticated communication without knowing that agent admin router runs on different ports.

Furthermore, this sets up a way for anyone to easily add additional arguments or change a Api's signature without having the redefine and entire heirarchy of inter-dependent helper classes. You can imagein a MesosNodeClient mixin that automatically sets the correct port for master vs agent, or perhaps one could make a mixin that checks if a certain field is in the response body and then hard error if not, etc...

The great thing about mixins like that is you can arbitrarily combine as many as you want because of how python's super works in calling siblings before parents

> The assertions done in this Mixin really belong to either ApiClientSession or DcosApiSession. If point 1. is addressed, then this is automatically taken care.
They do not belong in ApiClientSession because ApiClientSession should have node idea what a 'node' is to stay generic. They do not belong in DcosApiSession because of the above reason we need to make this a mixin

test_util/dcos_api_session.py
```diff
@@ -30,11 +30,111 @@ def get_args_from_env():
...
+            default_os_user: default user that marathon/metronome will launch tasks under
+            auth_user: use this user's auth for all requests
+                Note: user must be authenticated explicitly or call self.wait_for_dcos()
+        """
+        super().__init__(Url.from_string(dcos_url))
+        self.masters = sorted(masters)
+        self.public_masters = sorted(public_masters)
+        self.slaves = sorted(slaves)
+        self.public_slaves = sorted(public_slaves)
+        self.all_slaves = sorted(slaves + public_slaves)
```
mellenburg: Good catch; thanks. I guess python 3.5 doesnt hard error on the typing strings?

test_util/test_usage.py
mellenburg: Yes. This is important because as this was written before with functools.partialmethod, the class was unwrappable as it would map ApiClientSession.api_request instead of self.api_request to self.get (or whatever http verb). Although, I ended up using mixins instead of directly wrapping, so I should add the mixin case here if I want to be thorough
1. Good point. Will do!
2. The point of the 404 and 200 is just to exercise that complicated logic could be used in the wrapper. Specifically that a 'wrong' request can be made, inspected, modified, and re-requested without the user of the new ApiClientSession caring. I admit this is a bit convoluted

test_util/test_usage.py
mellenburg: I dont see much value in making sure all the verbs work, as its really just testing that I wrote the class correctly in the first place. That being said, its not a hard test to write, so no harm

test_util/dcos_api_session.py
orsenthil: Got it.

> I wait until we have what is necessary to auth other requests until formally logging that Authentication is successful.

What would have helped to get this logic while reading the first time, would be a `try-expect` block that that signifies everything we want in the try block with authentication successful as the last statement. The except block would catch, `KeyError` and `ResponseException` and show the `str` of that exception.
Minor improvement that I can think of. If you think, it is a good idea, we can do this next when visiting this code again.

test_util/dcos_api_session.py
orsenthil: > That is not accurate. This is for when you are supplying node INSTEAD of host and port

What led me to my original comment was the API signature varying only with extra arg *node*

```
class ARNodeApiClientMixin:
    def api_request(self, method, path_extension, scheme=None, host=None, query=None,
                    fragment=None, port=None, node=None, **kwargs):
```

vs

```
class ApiClientSession:
    def api_request(self, method, path_extension, scheme=None, host=None, query=None,
                    fragment=None, port=None, **kwargs):
```

Agree to the arguments made in favor of Mixin here. 

Question: If we want to a "Node" specific mixin, is a good idea to change the API signature to

```
class ARNodeApiClientMixin:
    def api_request(self, method, path_extension, node=None, scheme=None, query=None,
                    fragment=None, **kwargs):
```
?

* That is, remove the host, port as a keyword arg in the API itself.
* That will eliminate the assertion errors of having all of node, host, port in the current method.

The rest of the method can stay the same.

What do you think?

(Also, irrespective of this, I think, this method can greatly benefit from keyword-only argument, like 
```
def api_request(self, method, path_extension, *, scheme=None, host=None, query=None,
                    fragment=None, port=None, node=None, **kwargs):
```
That will force providing correct keywords and as keyword args only by the caller.

mellenburg: We cannot change the api signature to exclude port and host as keyword args because plenty of code relies on using those. This intention of this is NOT to be "You can ONLY query via a node/host string". Rather, this is to allow the _option_ of specifying node instead of host/port

Good suggestion on the keyword-only specifier. Will do!

test_util/dcos_api_session.py
```diff
@@ -30,11 +34,111 @@ def get_args_from_env():
...
+                raise Exception('Node {} is not recognized within the DC/OS cluster'.format(node))
+            host = node
+        return super().api_request(method, path_extension, scheme=scheme, host=host,
+                                   query=query, fragment=fragment, port=port, **kwargs)
+
+
+class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
+    def __init__(self, dcos_url: str, masters: list, public_masters: list,
+                 slaves: list, public_slaves: list, default_os_user: str,
+                 auth_user=Optional[DcosUser]):
```
mellenburg: https://docs.python.org/3/library/typing.html#typing.Optional

> Optional[X] is equivalent to Union[X, None].

if we did auth_user: DcosUser = None, that would be invalid as the default could never work (NoneType is not DcosUser type)

test_util/test_upgrade_vpc.py
```diff
@@ -144,14 +144,14 @@ def main():
...
         "root",             # default_os_user
-        web_auth_default_user=DcosUser(CI_AUTH_JSON),
+        web_auth_default_user=DcosUser(CI_CREDENTIALS),
         ca_cert_path=None)
```
mellenburg: Good catch, thanks! Will give that file a look over.

test_util/dcos_api_session.py
orsenthil: For

> plenty of code relies on using those.

If this is only due to the effort, it could be a separate PR, if we think that having segregated args (node / (host,port)) make it better.

> The intention of this is NOT to be "You can ONLY query via a node/host string". Rather, this is to allow the option of specifying node instead of host/port

Yeah, the option is provided, but isn't "You can ONLY query via a node/host string"  realized by the code in the mixin?

```
        if node is not None:
            assert port is None, 'node is intended to retrieve port; cannot set both simultaneously'
            assert host is None, 'node is intended to retrieve host; cannot set both simultaneously'
```

test_util/dcos_api_session.py
orsenthil: Actually, None can be assigned to any object. 
This short snippet demonstrates the problem with this code. (foo is my recommendation, bar is the current approach).
```
$ cat a2.py
from typing import Optional

class DcosUser:
    def __init__(self, username):
        self.username = username
    def __str__(self):
        return self.username

class SomethingElse: pass

def foo(o:DcosUser = None):
    print(o)

foo()   # Valid
foo(None) # Valid
foo(DcosUser("s")) # Valid
foo(SomethingElse()) # Invalid


def bar(o=Optional[DcosUser]):
    print(o)

bar() # valid
bar(None) # valid. But is not using Optional, because it is setting to None
bar(DcosUser("s")) # valid
bar(SomethingElse()) # valid because it is setting to a different object. I don't think we want this.

$ mypy a2.py
a2.py:17: error: Argument 1 to "foo" has incompatible type "SomethingElse"; expected "DcosUser"
```

In addition, the original comment was pointing to this scenario. We use something like.

```
        if self.auth_user is not None:
```
In the way `self.auth` set, unless we specifically set the argument from the caller to `None` the the value is always not None.
```
>>> auth_user = Optional[object]
>>> auth_user is None
False
```

mellenburg: I see now. Thanks for clarifying that! I did not realize that caveat of Optional. Now that I read this again, I have no idea why auth_user should even be an optional argument. Ill just make it required arg and that further simplifies this issue

test_util/dcos_api_session.py
mellenburg: No, It is not realized by the code in the Mixin. The mixin implements that "you MAY query via node" because node is by default None, thus meaning the mixin only does anything if the user explicitly gives node=. Otherwise, this mixin is invisible to the user.

Also, changing code else where to not use host/port is not an effort problem but an API problem. If you remove host/port setting, then you won't be able to query VIPs using auth because they are not going to be nodes. We need to query nodes and VIPs with this client. I do not understand why you would want to have two separate clients where the only difference is one may optionally substitute parts of the default url and the other one optionally substitutes parts of the default url, but has a special lookup that allows not needing to set host and port

Those functions are completely inclusive. Those are both necessary tasks in many spots in the integration tests. Why do you feel they should be exclusive? What is the use case/gain?

orsenthil: > Why do you feel they should be exclusive? What is the use case/gain?

Well, I was coming from a API design perspective that **if node arg requirement** is exclusive to (host, port) then we can probably encode it in API signature itself. 

But then realized a limitation with my suggestion. We cannot have same named method 'api_request' with variable arguments in python. I realized this as I tried to code the idea that I had.

```
class ApiMixin:
    def api(self, a, b, c):
        h,p = c
        super().api(a, b, h, p)

class Base:
    def api(self, a, b, h, p):
        print(a, b, h, p)

class Combo(ApiMixin, Base): pass

c = Combo()
c.api(1, 2, (3, 4)) # calls ApiMixin
c.api(1, 2, 3, 4) # Does not work. I cannot really reach Base.api this way.
```

Please ignore the above suggestion.

mellenburg:
This PR should be merged at the same time as #1096

test_util/dcos_api_session.py
jgehrcke: That was a good read :)

test_util/dcos_api_session.py
```diff
@@ -30,11 +34,111 @@ def get_args_from_env():
...
+
+
+class ARNodeApiClientMixin:
+    def api_request(self, method, path_extension, *, scheme=None, host=None, query=None,
+                    fragment=None, port=None, node=None, **kwargs):
+        """ Communicating with a DC/OS cluster is done by default through Adminrouter.
+        Use this Mixin with an ApiClientSession that requires distinguishing between nodes.
+        Adminrouter has both a master and agent process and so this wrapper accepts a
+        node argument. node must be a host in self.master or self.all_slaves. If given,
+        the request will be made to the adminrouter endpoint for that node type
```
jgehrcke: Just for spelling/naming consistency, can we do a search/replace and have `Admin Router` where we currently have `Adminrouter` or `adminrouter` in the comments/docstrings?

test_util/dcos_api_session.py
```diff
@@ -187,117 +290,57 @@ def wait_for_dcos(self):
...
-        self.public_masters = sorted(public_masters)
-        self.slaves = sorted(slaves)
-        self.public_slaves = sorted(public_slaves)
-        self.all_slaves = sorted(slaves + public_slaves)
-        self.zk_hostports = ','.join(':'.join([host, '2181']) for host in self.public_masters)
-        self.default_os_user = default_os_user
-        self.web_auth_default_user = web_auth_default_user
-
-        assert len(self.masters) == len(self.public_masters)
+        new = copy.deepcopy(self)
```
jgehrcke: :+1:

test_util/helpers.py
```diff
@@ -28,107 +30,136 @@
...
+CI_CREDENTIALS = {'token': 'eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6Ik9UQkVOakZFTWtWQ09VRTRPRVpGTlRNMFJrWXlRa015Tnprd1JrSkVRemRCTWpBM1FqYzVOZyJ9.eyJlbWFpbCI6ImFsYmVydEBiZWtzdGlsLm5ldCIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJpc3MiOiJodHRwczovL2Rjb3MuYXV0aDAuY29tLyIsInN1YiI6Imdvb2dsZS1vYXV0aDJ8MTA5OTY0NDk5MDExMTA4OTA1MDUwIiwiYXVkIjoiM3lGNVRPU3pkbEk0NVExeHNweHplb0dCZTlmTnhtOW0iLCJleHAiOjIwOTA4ODQ5NzQsImlhdCI6MTQ2MDE2NDk3NH0.OxcoJJp06L1z2_41_p65FriEGkPzwFB_0pA9ULCvwvzJ8pJXw9hLbmsx-23aY2f-ydwJ7LSibL9i5NbQSR2riJWTcW4N7tLLCCMeFXKEK4hErN2hyxz71Fl765EjQSO5KD1A-HsOPr3ZZPoGTBjE0-EFtmXkSlHb1T2zd0Z8T5Z2-q96WkFoT6PiEdbrDA-e47LKtRmqsddnPZnp0xmMQdTr2MjpVgvqG7TlRvxDcYc-62rkwQXDNSWsW61FcKfQ-TRIZSf2GS9F9esDF4b5tRtrXcBNaorYa9ql0XAWH5W_ct4ylRNl3vwkYKWa4cmPvOqT5Wlj9Tf0af4lNO40PQ'}     # noqa
```
jgehrcke: Can we add a comment about where this token actually came from?

jgehrcke:
Hey Michael, I gave this another look. Thanks for sitting this through :). We should really start integration this now, as the changes add a lot of value -- we can then iterate on the whole concept in smaller chunks when the need arises.

One thing I am not sure about: from the point of view where I would write some new tests now -- how easy is it to emit _non-authenticated_ requests to the cluster? 

And, before Christmas we talked about worrisome implicit state sharing across HTTP requests through request's cookie handling. I proposed:

> I think we should, by default, disarm the cookie management part of the requests.Session object used in the ApiClient.

(original comment plus code proposal: https://github.com/dcos/dcos/pull/1016#issuecomment-267330911)

Do we think we need that?

test_util/helpers.py
mellenburg: @jgehrcke  directly above this ( lines 22 through 31) there is a comment describing this auth token. Is that insufficient information or is it just not clear that it relates to this bit of data?

mellenburg:
@jgehrcke 

1. how easy is it to emit non-authenticated requests? Quite easy. This PR is meant to be paired with a renaming PR. See: https://github.com/mesosphere/dcos-enterprise/blob/mellenburg/rename_cluster/packages/dcos-integration-test/extra/conftest.py#L37

So, currently you can call "noauth_cluster" fixture in some tests, but we will just roll out a global (and more comprehensible) 'noauth_api_session' that will make requests without any auth persistence

2. Thanks for reraising the point on killing the cookie persistence between test cases. I went ahead and added the following tidbit that will take care of that issue:
```
 r = self.session.request(method, request_url, **kwargs)
self.session.cookies.clear()
return r
```
Thus, the user can get the cookies in the response, but the underlying session will never auto-persist cookies from a given request

test_util/helpers.py
jgehrcke: Haha, thanks for healing a blind person. Perfect!

jgehrcke:
> but we will just roll out a global (and more comprehensible) 'noauth_api_session' that will make requests without any auth

:+1: 

> Thus, the user can get the cookies in the response, but the underlying session will never auto-persist cookies from a given request

Perfect, thanks.

I suppose we should ship this now.

PR #1028, opened 2016-12-06
Make CalculatorError.__init__() chain arg optional

branden:
This restores a default value removed in de76c26a844c4f11e4794861f59bb177f32479e7.

# Issues

[DCOS-11894](https://mesosphere.atlassian.net/browse/DCOS-11894)

# Checklist

### DIFF
--- a/gen/internals.py
+++ b/gen/internals.py
@@ -3,7 +3,7 @@ import inspect
 import logging
 from contextlib import contextmanager
 from functools import partial, partialmethod
-from typing import Callable, List, Optional, Tuple, Union
+from typing import Callable, List, Tuple, Union
 
 from gen.exceptions import ValidationError
 from pkgpanda.build import hash_checkout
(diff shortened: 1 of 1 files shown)


gen/internals.py
```diff
@@ -275,7 +275,7 @@ def make_id(self):
 # NOTE: This exception should never escape the Resolver
 class CalculatorError(Exception):
 
-    def __init__(self, message: str, chain: Optional[list]):
```
orsenthil: The change could simply be:

chain = None

Ref: https://www.python.org/dev/peps/pep-0484/#union-types
```
An optional type is also automatically assumed when the default value is None , for example:

def handle_employee(e: Employee = None): ...
This is equivalent to:

def handle_employee(e: Optional[Employee] = None) -> None: ...
```

orsenthil:
LGTM.

PR #1000, opened 2016-11-29
Add an anonymous module to Mesos that modifies logging behavior.

kaysoky:
The prior change to logging (`tee`-ing stderr to a file) resulted in some incorrect agent recovery behavior.  When the agent exited or crashed (for any reason), there was a chance that the `tee` process would stay behind (presumably because the agent's stdout/stderr was leaked/inherited by a subprocess) indefinitely.  This led to the agent being removed and tasks being lost.

## This commit: 
* Removes the `tee_stderr` wrapper.
* Adds an anonymous module instead, which implements a `glog` `LogSink` that writes to a file.  Because this is a module, it is guaranteed to die if the master/agent dies too.
  * Updates the `mesos-modules` package.  https://github.com/dcos/dcos-mesos-modules/pull/3

## Checklist
* [x] ~Add an agent recovery integration test:~ This requires tests to have SSH access, rather than interacting with exposed endpoints like all the other tests.  As such, this test is left out for now.
  1) Launch a simple sleep
  2) Tell the agent to exit (directly, not via systemd)
  3) Check for agent recovery and task existence
* [x] Make sure the `LogSink` plays well with the `logrotate` service.  (Probably a manual test)
* [x] Merge the `dcos-mesos-modules` PR too.

PR #1031, opened 2016-12-07
Debugging API changes in Mesos

adam-mesos:
This change bumps Mesos (and mesos-modules to match) to include debugging fixes.
Also bumping AdminRouter to pickup some debugging support.
But most significant is a dcos-integration test of the new Mesos Debug API calls.

# Issues

[DCOS-9574](https://mesosphere.atlassian.net/browse/DCOS-9574)

# Checklist

# Component updates:
Mesos

 - [X] Change Log from last: 
https://github.com/mesosphere/mesos/compare/dcos-mesos-master-30d6edb...mesosphere:dcos-mesos-master-e99ea9c
(mostly tests/fixes for debugging)
 - [x] Test Results: https://mesosphere.slack.com/archives/core-ci/p1481694252000143 
(3 known cmake build errors, 1 consistently failing test soon-to-be-fixed, and several known flaky tests)
 - [ ] Code Coverage: N/A

Mesos-modules

 - [X] Change Log from last: https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae 
(fix build for Mesos interface changes; log-sink module for #1000 and upgrade-fix slipped in too)
 - [ ] Test Results: N/A
 - [ ] Code Coverage: N/A

Adminrouter

 - [X] Change Log from last: https://github.com/dcos/adminrouter/compare/ae8d4d8328599d67e7ca8d9246229ff1420e792c...d44fd04d38db485145dd8c0fd839cc44c648c675 
("Disable buffering to mesos agent API and use HTTP/1.1" for debugging support)
 - [ ] Test Results: N/A
 - [ ] Code Coverage: N/A

### DIFF
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -1,112 +1,135 @@
+import json
 import logging
 import uuid
 
 import requests
 
 from test_util.marathon import get_test_app
+from test_util.recordio import Decoder, Encoder
+
+
+# Wrapper to post, log, and validate return status
+def post(url, headers, json=None, data=None, stream=False):
+    r = requests.post(url, headers=headers, json=json, data=data, stream=stream)
+    logging.info(
+        'Got %s with POST request to %s with headers %s and json data %s.',
+        r.status_code,
+        url,
+        headers,
+        json
+    )
+    assert r.status_code == 200
+    return r
+
+
+# Creates and yields the initial ATTACH_CONTAINER_INPUT message, then a data message,
+# then an empty data chunk to indicate end-of-stream.
+def input_streamer(nested_container_id):
+    encoder = Encoder(lambda s: bytes(json.dumps(s, ensure_ascii=False), "UTF-8"))
+    message = {
+        'type': 'ATTACH_CONTAINER_INPUT',
+        'attach_container_input': {
+            'type': 'CONTAINER_ID',
+            'container_id': nested_container_id}}
+    yield encoder.encode(message)
+
+    message['attach_container_input'] = {
+        'type': 'PROCESS_IO',
+        'process_io': {
+            'type': 'DATA',
+            'data': {'type': 'STDIN', 'data': 'meow'}}}
+    yield encoder.encode(message)
+
+    # Place an empty string to indicate EOF to the server and push
+    # 'None' to our queue to indicate that we are done processing input.
+    message['attach_container_input']['process_io']['data']['data'] = ''
+    yield encoder.encode(message)
 
 
 def test_if_marathon_app_can_be_debugged(cluster):
-    def post(url, data):
-        r = requests.post(url, json=data)
-        logging.info(
-            'Got %s with POST request to %s with %s. Response: \n%s',
-            r.status_code,
-            url,
-            data,
-            r.text
-        )
-        assert r.status_code == 200
-
-    def find_container_id(state, app_id):
-        if 'frameworks' in state:
-            for framework in state['frameworks']:
-                # TODO: Skip anything that's not Marathon
-                if 'tasks' in framework:
-                    for task in framework['tasks']:
-                        if 'id' in task and app_id in task['id']:
-                            if 'statuses' in task and 'container_status' in task['statuses'][0]:
-                                if 'container_id' in task['statuses'][0]['container_status']:
-                                    container_status = task['statuses'][0]['container_status']
-                                    if 'container_id' in container_status:
-                                        if 'value' in container_status['container_id']:
-                                            return container_status['container_id']['value']
-
-    def find_agent_id(state, app_id):
-        if 'frameworks' in state:
-            for framework in state['frameworks']:
-                # TODO: Skip anything that's not Marathon
-                if 'tasks' in framework:
-                    for task in framework['tasks']:
-                        if 'id' in task and app_id in task['id']:
-                            if 'slave_id' in task:
-                                return task['slave_id']
-
-    def find_agent_hostname(state, agent_id):
-        if 'slaves' in state:
-            for agent in state['slaves']:
-                if 'id' in agent and agent['id'] == agent_id and 'hostname' in agent:
-                    return agent['hostname']
-
     # Launch a basic marathon app (no image), so we can debug into it!
+    # Cannot use deploy_and_cleanup because we must attach to a running app/task/container.
     app, test_uuid = get_test_app()
-    test_app_id = 'integration-test-{}'.format(test_uuid)
+    app_id = 'integration-test-{}'.format(test_uuid)
     cluster.marathon.deploy_app(app)
 
-    # Find the agent_id and container_id from master state
+    # Fetch the mesos master state once the task is running
     master_state_url = 'http://{}:{}/state'.format(cluster.masters[0], 5050)
     r = requests.get(master_state_url)
-    logging.info('Got %s with request for %s. Response: \n%s', r.status_code, master_state_url, r.text)
+    logging.debug('Got %s with request for %s. Response: \n%s', r.status_code, master_state_url, r.text)
     assert r.status_code == 200
     state = r.json()
 
-    container_id = find_container_id(state, test_app_id)
-    agent_id = find_agent_id(state, test_app_id)
-    agent_hostname = find_agent_hostname(state, agent_id)
+    # Find the agent_id and container_id from master state
+    container_id = None
+    agent_id = None
+    for framework in state['frameworks']:
+        for task in framework['tasks']:
+            if app_id in task['id']:
+                container_id = task['statuses'][0]['container_status']['container_id']['value']
+                agent_id = task['slave_id']
+    if container_id is None:
+        raise Exception('Container ID not found for instance of app_id {}'.format(app_id))
+    if agent_id is None:
+        raise Exception('Agent ID not found for instance of app_id {}'.format(app_id))
+
+    # Find hostname and URL from agent_id
+    agent_hostname = None
+    for agent in state['slaves']:
+        if agent['id'] == agent_id:
+            agent_hostname = agent['hostname']
+    if agent_hostname is None:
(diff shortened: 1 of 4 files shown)


adam-mesos:
I've fixed all my TODOs and the CI is green, so I'd say this is ready for review. Maybe @mellenburg and @klueska can tell me how bad my python is.

test_util/recordio.py
```diff
@@ -0,0 +1,155 @@
...
+
+        for c in data:
+            if self.state == self.HEADER:
+                if c != ord('\n'):
+                    self.buffer += bytes([c])
+                    continue
+
+                try:
+                    self.length = int(self.buffer.decode("UTF-8"))
+                except Exception as exception:
```
mellenburg: As this commit is written now, this try/catch/raise DCOSException is never really utilized so when this does get hit, everything will come to a halt anyway, thus, just leave it out.

However, if you envision this being used elsewhere, I could get the idea of a failed state/custom exception. In that case, just make in this module:
```
class RecordIoDecoderException(Exception):
    pass
```
or call it RecordIoCodecError if you want to use one error that also covers the encoder. In either case, those names will be more usable than "DCOSExpcetion" when this has nothing to do with DC/OS on the surface

test_util/recordio.py
```diff
@@ -0,0 +1,155 @@
...
+        s = self.serialize(message)
+
+        if not isinstance(s, bytes):
+            raise DCOSException("Calling 'serialize(message)' must"
+                                " return a 'bytes' object")
+
+        return bytes(str(len(s)) + "\n", "UTF-8") + s
+
+
+class Decoder(object):
```
mellenburg: As above, no need to inherit from object in python3

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,140 @@
...
+            for agent in state['slaves']:
+                if 'id' in agent and agent['id'] == agent_id and 'hostname' in agent:
+                    return agent['hostname']
+
+    # Creates and yields the initial ATTACH_CONTAINER_INPUT message, then a data message,
+    # then an empty data chunk to indicate end-of-stream.
+    def _input_streamer(encoder, nested_container_id, input_data):
+        message = {'type': 'ATTACH_CONTAINER_INPUT', 'attach_container_input': {'type': 'CONTAINER_ID'}}
+        message['attach_container_input']['container_id'] = nested_container_id
+        yield encoder.encode(message)
```
mellenburg: Does this need to be a generator or a function at all? Could we not just construct this payload before the post? If this must be a method, then I would remove the input_data and encoder arguments and just hardcode them within this method as they are only called once in a totally static manner

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Place an empty string to indicate EOF to the server and push
+        # 'None' to our queue to indicate that we are done processing input.
+        message['attach_container_input']['process_io']['data']['data'] = ''
+        yield encoder.encode(message)
+
+    # Launch a basic marathon app (no image), so we can debug into it!
+    app, test_uuid = get_test_app()
+    test_app_id = 'integration-test-{}'.format(test_uuid)
+    cluster.marathon.deploy_app(app)
```
adam-mesos: Unfortunately, `deploy_and_cleanup` destroys the app before I can get a hold of the container to debug into it. I need something that only cleans up after I'm done with it, since I need a live running task for the duration of my test. Something like a test Teardown?

packages/dcos-integration-test/extra/test_mesos.py
adam-mesos: Yeah, I just need an iterable, so an array should work

test_util/recordio.py
adam-mesos: Dropped it in favor of `Exception`

packages/dcos-integration-test/extra/test_mesos.py
adam-mesos: Hmm, didn't work on first attempt, but maybe you have some tips now that the rest of it's cleaned up.

test_util/errors.py
```diff
@@ -0,0 +1,91 @@
+import abc
```
adam-mesos: Agreed. I just copied the recordio file from dcos-cli and didn't bother to strip out the unnecessary errors.py dependency. It's gone now.

adam-mesos:
Rebased against dcos/dcos master, bumped to latest Mesos master (and modules to match), and cleaned up the integration test according to @mellenburg 's feedback. Two things I wasn't able to address:
1. My first attempt at inlining `input_streamer` into an iterable array failed, so I just pulled it up into a single-parameter top-level function. Maybe python streaming requests (not responses) expects a generator?
2. I can't use `marathon.deploy_and_cleanup` because it destroys the app as soon as it verifies that it's healthy, which means that by the time I query for running tasks/containers to debug into, the app+task has already been killed.

Ready for review again!

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+
+    # Find the agent_id and container_id from master state
+    container_id = None
+    agent_id = None
+    for framework in state['frameworks']:
+        for task in framework['tasks']:
+            if app_id in task['id']:
+                container_id = task['statuses'][0]['container_status']['container_id']['value']
+                agent_id = task['slave_id']
+    if container_id is None:
```
mellenburg: All of these
```
if foo is None:
    raise Exception(my_message)
```
are suboptimal. When using pytest, always try to use plain asserts. The following is identical:
```
assert foo is None, my_message
```

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+        raise Exception('Container ID not found for instance of app_id {}'.format(app_id))
+    if agent_id is None:
```
mellenburg: See above on assert pattern

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+        raise Exception('Agent ID not found for instance of app_id {}'.format(app_id))
+
+    # Find hostname and URL from agent_id
+    agent_hostname = None
+    for agent in state['slaves']:
+        if agent['id'] == agent_id:
+            agent_hostname = agent['hostname']
+    if agent_hostname is None:
```
mellenburg: See above on assert pattern

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+        'Connection': 'keep-alive',
+        'Transfer-Encoding': 'chunked'
+    }
+    post(agent_v1_url, input_headers, data=input_streamer(nested_container_id))
+
+    # Verify the streamed output from the launch session
+    decoder = Decoder(lambda s: json.loads(s.decode("UTF-8")))
+    for chunk in launch_output.iter_content():
+        for r in decoder.decode(chunk):
+            if r['type'] == 'DATA':
```
mellenburg: This test can pass if there are no records that are of the data type. Should we add a check so that this looper will fail if this if statement is never triggered?

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+                logging.debug('Extracted data chunk: %s', r['data'])
+                assert r['data']['data'] == 'meow', 'Output did not match expected'
+
+    # Verify the message from the attached output stream
+    for chunk in attached_output.iter_content():
+        for r in decoder.decode(chunk):
+            if r['type'] == 'DATA':
```
mellenburg: See above on if we need an assertion to check if this if block is never hit

packages/adminrouter/buildinfo.json
```diff
@@ -9,7 +9,7 @@
       "adminrouter": {
           "kind": "git",
           "git": "https://github.com/dcos/adminrouter.git",
-          "ref": "ae8d4d8328599d67e7ca8d9246229ff1420e792c",
+          "ref": "d44fd04d38db485145dd8c0fd839cc44c648c675",
```
alberts: rebase and drop this change -- adminrouter has been bumped to latest already

test_util/recordio.py
```diff
@@ -0,0 +1,150 @@
...
+                except Exception as exception:
+                    self.state = self.FAILED
+                    raise Exception("Failed to decode length '{buffer}': {error}"
+                                    .format(buffer=self.buffer, error=exception))
+
+                self.buffer = bytes("", "UTF-8")
+                self.state = self.RECORD
+
+                # Note that for 0 length records, we immediately decode.
+                if self.length <= 0:
```
branden: Let's assert length is nonnegative after we decode it, to fail earlier on bad encodings and simplify length comparisons.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -0,0 +1,135 @@
...
+    message['attach_container_input']['process_io']['data']['data'] = ''
+    yield encoder.encode(message)
+
+
+def test_if_marathon_app_can_be_debugged(cluster):
+    # Launch a basic marathon app (no image), so we can debug into it!
+    # Cannot use deploy_and_cleanup because we must attach to a running app/task/container.
+    app, test_uuid = get_test_app()
+    app_id = 'integration-test-{}'.format(test_uuid)
+    cluster.marathon.deploy_app(app)
```
adam-mesos: Magic.

adam-mesos:
Addressed all comments from @mellenburg and @branden and removed the obsolete AdminRouter bump.
Ready for review (and hopefully approval) once the CI turns all green again

PR #1032, opened 2016-12-07
Validate that if master_discovery is non-static exhibitor_storage_backend is non-static

cmaloney:
Having static master_discovery but non-static exhibitor_storage_backend is legal,
although doesn't provide a lot of value. There isn't a particular reason to outlaw it however.

Having master_discovery master_http_loadbalancer but exhibitor_storage_backend isn't legal since it
results in configuration conflicts.

# Issues

[DCOS-11827](https://mesosphere.atlassian.net/browse/DCOS-11827)

# Checklist

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -426,6 +426,15 @@ def calculate_set(parameter):
         return 'true'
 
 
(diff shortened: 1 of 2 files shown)


amitaekbote:
🚢

PR #984, opened 2016-11-22
Upgrade Testing With apps and task.

orsenthil:
Uses the integration test infrastructure to add a application and task.
* Verifies task is running after upgrade.
* Verifies that task is not restarted during upgrade.

# Issues

[DCOS-11453](https://mesosphere.atlassian.net/browse/DCOS-11453)

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

If this is a component/package update, include

 - [ ] Change Log from last: <link>
 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

PR #1036, opened 2016-12-08
General fixes in the test code.

orsenthil:
* These are generic fixes in the test code that I found when working on #984 
* Presenting them separately so that other one can stay focussed on functionality.

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -363,6 +363,7 @@ CF_OS_SSH_INFO = {
     'coreos': SSH_INFO['coreos']
 }
 
+
 OS_AMIS = {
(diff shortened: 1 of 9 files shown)


test_util/aws.py
```diff
@@ -363,7 +363,6 @@ def get_vpc_host_ips(self):
     'coreos': SSH_INFO['coreos']
 }
 
-
```
branden: This newline should stay. `OS_AMIS` has no relation to the above `SSH_INFO` dicts.

mellenburg:
Thanks for cleaning those up :ship:

PR #1039, opened 2016-12-08
[WIP] Move make_cluster_fixture to test_util

orsenthil:
The cluster fixture serves well outside the package.


amitaekbote:
@orsenthil is this still work in progress?

mellenburg:
:x:  Strong nak against this. That file is in that specific place so it can be copied over/overridden in place in EE

orsenthil:
@mellenburg  - I got your attention upfront!

* What if I want to use the functionality of `make_cluster_fixture` in a test code that is not bundled as package ? For e.g. in an integration test. ( I have a strong use case for this, that is the reason I was testing this change upfront). [1]

I thought the reason moving `make_cluster_fixture` to `test_util` makes sense because.

* It is a helper function that provides a cluster, and could right belong to `test_util`.
* The pytest context creation is actually in `conftest.py`

By having this inside the `dcos-integration-test` package, I cannot really use this from anywhere else.

I have no idea about the ee dependency. But cannot it be solved after addressing these?
I am primarily interested in [1] above.

mellenburg:
Here is the EE magic:
https://github.com/mesosphere/dcos-enterprise/blob/master/packages/dcos-integration-test/build#L15
The issue of plugging one fixture into both EE and Open tests has been very painful and this hack-ish direct inject method has been the best so far.

cluster_fixture.py is not intended to be used anywhere else on purpose (it is not intended as a utility). What's more, it should only be used to create a cluster fixture as the name states (which your code is not doing with it)

Looking at your intended usage of this, there are two things to note:
- Please don't set environment variables within a test to pass parameters. This can cause other stuff to fail in odd manners
- This would be the better pattern to use
```
from test_util.cluster_api import ClusterApi, get_args_from_env
from test_util.helpers
args = get_args_from_env()
# make custom changes to the env gathered args
args['web_auth_default_user'] = DcosUser(auth_json)
cluster_api = ClusterApi(**args)
cluster_api.wait_for_dcos()
```

It seems like what you really want isnt the cluster_fixture.py file, but rather the hard-coded CI user and the ability to launch a marathon app. For the CI user, I would support moving that to a simple test_util helper get_ci_user() and then launch cluster_api as shown above.

orsenthil:
@mellenburg 

I agree with not moving these files, creating a module as suggested.

Since I am interested in the auth_token, I'd like refactor `auth_json_token` to a common helper and  use it both places.

(Subtext)

*  I was open to change the name of `cluster_fixture.py` to appropriate one if required. The pattern of using ClusterApi directly with environment almost seemed like duplicating the code.

@amitaekbote  - closing this. No longer WIP.

orsenthil:
The topics discussed in this thread are addressed here: https://github.com/dcos/dcos/pull/984 the discussion can continue over there.

PR #1042, opened 2016-12-08
Added POSIX rlimits support to Mesos.

jieyu:
Added 'posix/rlimits' to default Mesos isolator list.

This isolator enables the POSIX rlimits support in Mesos. It allows
frameworks to change the default resource limits (both hard and soft)
for their containers, similar to Docker `--ulimit` flag.

Some data services like Elastic Search or Cassandra typically require
a higher than normal resource limit for their containers, and will do
a check during startup to see if those limits are set properly. This
support in Mesos will unblock them running on DC/OS.

This isolator only affects the launch path of the container. If no
framework is using this feature, adding this isolator to the isolator
list will be a no-op.

# Issues

[MESOS-6402](https://issues.apache.org/jira/browse/MESOS-6402)

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -362,7 +362,7 @@ def calculate_config_yaml(user_arguments):
 
 def calculate_mesos_isolation(enable_gpu_isolation):
     isolators = ('cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,'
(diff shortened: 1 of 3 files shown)


jieyu:
Close this PR in favor of https://github.com/dcos/dcos/pull/1048 that has an integration test.

PR #1045, opened 2016-12-08
enable dcos-log service on all hosts

darkonie:
 - Stop writing STDOUT and STDERR to sandbox
 - Start writing STDOUT and STDERR to sd journal
 - Add integration tests

This PR includes:
 - https://github.com/dcos/dcos/pull/877 write container logs to sd journal
# Checklist

 - [x] Change Log from last: [add sse ping](https://github.com/dcos/dcos-log/pull/27)
 - [x] Test Results: [build & test](https://jenkins.mesosphere.com/service/jenkins/view/DCOS/job/dcos-log%20pull%20requests/64/console)
 - [x] Code Coverage: [build & coverage](https://jenkins.mesosphere.com/service/jenkins/view/DCOS/job/dcos-log%20pull%20requests/64/console)


darkonie:
@cmaloney ready for code review

amitaekbote:
@darkonie thanks for adding tests! can you update the checklist in the template to reflect that?

spahl:
@darkonie can you fill out the pullrequest template above so that we can get this ready to merge?

darkonie:
@spahl done

spahl:
The mesos commits have been merged so you should be able to rebase this now.

darkonie:
@spahl rebased and commits logically squashed

spahl:
@darkonie I updated your template. All it need is a link to your unit tests in jenkins. The link to these integration tests here exist already:-)

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+
+    pod_id = 'integration-test-pod-logs-{}'.format(test_uuid)
+
+    pod_definition = {
+        'id': '/{}'.format(pod_id),
+        'scaling': {'kind': 'fixed', 'instances': 1},
+        'containers': [
+            {
+                'name': 'sleep1',
+                'exec': {'command': {'shell': 'echo STDOUT_LOG;echo STDERR_LOG >&2;sleep 10000'}},
```
mellenburg: Having apps (even trivial ones) without health checks makes for weird deploy conditions. I think we should update this to be like https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_applications.py#L80

The bigger problem here is that the deploy_pod method in test_util marathon should be health check aware which it currently isnt.

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+            executor_id = task['executor_id']
+            if not executor_id:
+                executor_id = task['id']
+            slave_id = task['slave_id']
+
+            statuses = task.get('statuses')
+            assert isinstance(statuses, list), 'Invalid field `statuses`. Got {}'.format(state_response_json)
+            assert len(statuses) == 1, 'Must have only one status TASK_RUNNING. Got {}'.format(state_response_json)
+            status = statuses[0]
+            container_status = status.get('container_status')
```
mellenburg: can we just make these assert status['container_status']? They way this is written now, I cannot tell if container_status isnt present or it is actually equal to None when it fails. If you do it with the single simple assert, a key error will tell you its not present, and the assertion error will tell you the value is empty. Also, in python '' will evaluate as boolean False like None so its best to use 'foo is not None' if an empty string is unacceptable

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+            assert container_status
+            container_id_field = container_status.get('container_id')
+            assert container_id_field
```
mellenburg: see above on simplified dict val assertion

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+
+            # traverse nested container_id fields
+            container_ids = []
+            while True:
+                value = container_id_field.get('value')
+                assert value
```
mellenburg: See above on this dict val assert

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+                container_ids.append(value)
+
+                if 'parent' not in container_id_field:
+                    break
+
+                container_id_field = container_id_field['parent']
```
mellenburg: I this would be easier to read as:
```
container_ids = [container_id_field['value']]
while 'parent' in container_id_field:
    container_id_field = container_id_field['parent']
    container_ids.append[container_id_field['value']]
```

darkonie: parent is not guaranteed to be there

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,145 @@ def test_log_proxy(cluster):
...
+        task_stdout = task_stdout_response.content.decode('utf-8', 'ignore')
+        assert 'STDOUT_LOG' in task_stdout, 'Missing `STDOUT_LOG` in response. Got {}'.format(task_stdout)
+
+        task_stderr_response = cluster.get('{}?filter=STREAM:STDERR'.format(url))
+        check_response_ok(task_stderr_response, {})
+        task_stderr = task_stderr_response.content.decode('utf-8', 'ignore')
+        assert 'STDERR_LOG' in task_stderr, 'Missing `STDERR_LOG` in response. Got {}'.format(task_stderr)
+
+
+def get_task_url(cluster, task_name, stream=False):
```
orsenthil: Would you prefer moving this function up, above line 78? ( The style varies here. but I have found dependencies (function definition) coming before invocation easier to navigate).

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -57,7 +58,7 @@ def test_stream(cluster):
         lines = response.iter_lines()
         sse_id = next(lines)
         assert sse_id, 'First line must be id. Got {}'.format(sse_id)
-        data = next(lines).decode()
+        data = next(lines).decode('utf-8', 'ignore')
```
darkonie: this won't work with `stream=True` because `response.text.iter_lines` will block
i think `.decode` is the only option we can use here

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: done

darkonie:
@mellenburg even though i use the health check with pods, i still see a lot of failures, so i assume this isn't related to the flakiness we see.

darkonie:
this is green now

packages/dcos-integration-test/extra/test_dcos_log.py
mellenburg: I know parent might not be there on the first iteration and thats why I started that code segment with
```
container_ids = [container_id_field['value']]
```
I am pretty sure my 4 lines are functionally identical to what you have here....

darkonie: got it! it's definitely improvement in readability! thanks

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,143 @@ def test_log_proxy(cluster):
     for slave_id in slaves_ids:
         response = cluster.get('/system/v1/agent/{}/logs/v1/range/?skip_prev=10'.format(slave_id))
         check_response_ok(response, {'Content-Type': 'text/plain'})
-        lines = filter(lambda x: x != '', response.content.decode().split('\n'))
+        lines = filter(lambda x: x != '', response.text.split('\n'))
         assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
```
orsenthil: This line will have a merge-conflict.

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,143 @@ def test_log_proxy(cluster):
...
+            assert status['container_status'], 'Invalid field `container_status`. Got {}'.format(state_response_json)
+            container_status = status['container_status']
+            assert container_status['container_id'], 'Invalid field `container_id`. Got {}'.format(state_response_json)
+            container_id_field = container_status['container_id']
+
+            # traverse nested container_id fields
+            container_ids = [container_id_field['value']]
+            while 'parent' in container_id_field:
+                container_id_field = container_id_field['parent']
+                container_ids.append([container_id_field['value']])
```
mellenburg: This should not be appending a list.

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -71,5 +72,143 @@ def test_log_proxy(cluster):
...
+
+    # validate all required fields
+    assert slave_id, 'Missing slave_id'
+    assert framework_id, 'Missing framework_id'
+    assert executor_id, 'Missing executor_id'
+    assert container_id, 'Missing container_id'
+
+    endpoint_type = 'range'
+    if stream:
+        endpoint_type = 'stream'
```
orsenthil: inline this?

darkonie: inline how? can you please provide a snippet of code

packages/dcos-integration-test/extra/test_dcos_log.py
darkonie: why? i don't see any conflict here

darkonie:
@mellenburg there was a race condition in the test. `dcos-log` service flushes the logs every second. The test does not wait at all and there is a small window when we can hit the service and the entry is not there. Fixed with https://github.com/dcos/dcos/pull/1045/files#diff-0983fe52e111e591e8e619b14979ce14R126

PR #1047, opened 2016-12-08
Make oauth_available always derived, even for cloud templates, optionally from oauth_enabled

cmaloney:
This allows downstreams to overwrite the value for oauth_available if they would like for the cloud templates. They used to be able to do this because of how dictionary overwriting worked, but with the use of "Sources" it stopped being possible. This changes the functionality back to how it was originally designed to work.

# Issues

[DCOS-11894](https://dcosjira.atlassian.net/browse/DCOS-11894)

# Checklist

 - [ ] Included a test which will fail if code is reverted but test is not. Hard to integration test at the moment (no existing infrastructure). Hand tested with a downstream taking advantage of it.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -261,7 +261,14 @@ def validate_dcos_overlay_network(dcos_overlay_network):
 
 
 def calculate_oauth_available(oauth_enabled):
-    return oauth_enabled
+    if oauth_enabled in ["[[[variables('oauthEnabled')]]]", '{ "Ref" : "OAuthEnabled" }']:
+        return 'true'
+    elif oauth_enabled == 'false':
+        return 'false'
+    elif oauth_enabled == 'true':
+        return 'true'
+    else:
+        raise AssertionError("Invaild value for oauth_enabled: {}".format(oauth_enabled))
 
 
 def validate_num_masters(num_masters):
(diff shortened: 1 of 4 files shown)


amitaekbote:
:ship:

nit: can you link to the pr which changed the behavior for more context?

gen/calc.py
```diff
@@ -261,7 +261,14 @@ def validate_dcos_overlay_network(dcos_overlay_network):
 
 
 def calculate_oauth_available(oauth_enabled):
-    return oauth_enabled
+    if oauth_enabled in ["[[[variables('oauthEnabled')]]]", '{ "Ref" : "OAuthEnabled" }']:
```
branden: We're duplicating these values outside their context (AWS, Azure). Would it make sense to define them in their respective modules, import them into `gen.calc`, and refer to them by name?

cmaloney: These are also in this file a bit earlier, so I'd like to defer this question until later (https://github.com/dcos/dcos/blob/master/gen/calc.py#L220). This just copy/pasted that set of possible values.

I also think if `gen.calc` imported `gen.build_deploy.aws` / `gen.build_deploy.azure` there would be likely circular import issues.

The need for this also goes away entirely with #1009 since we no longer call calculate or validate functions on variables which are late.

PR #977, opened 2016-11-19
Bump Mesos SHA to recent master

adam-mesos:
This Mesos bump includes all the commits in the official Mesos 1.1.0 tag, the fixes for [MESOS-6465](https://issues.apache.org/jira/browse/MESOS-6465) (task_id -> container_id mapping in state.json), as well as "code-complete" for Debugging (I/O switchboard, AttachContainerInput/Output and LNCSession, authz, etc.).

# Component checklist:
 - [X] Change Log from dcos `master`: https://github.com/mesosphere/mesos/compare/dcos-mesos-master-35ce5042...mesosphere:dcos-mesos-master-30d6edb
 - [x] Test Results: https://mesosphere.slack.com/messages/core-ci/ 
    - OSS Mesos: https://teamcity.mesosphere.io/viewLog.html?buildId=502519&tab=report_project134_Results
 - [ ] Code Coverage: N/A, but coverity runs on Apache Mesos regularly

# Highlights from changelog:
```
30d6edb Fixed container-logger subprocess environment setup.
8dcef9c Added Mesos logo to the list of files installed for the web UI.
a567345 Made the IO Switchboard wait for the reader to signal before exiting.
57d986c Used agent environment variables for I/O switchboard.
77b16d8 Added authorization support when attaching to container input/output.
be45c94 Added support for 'ATTACH_CONTAINER_INPUT' call to the io switchboard.
fd142a0 Supported TTY in I/O switchboard.
0e0fc46 Added API handler for LAUNCH_NESTED_CONTAINER_SESSION.
dcf0f1c Added support to handle ATTACH_CONTAINER_OUPUT in the io switchbaord.
19296e0 Added authorization actions for Nested Container and Debug API.
7e472c3 Added the v1 \`ATTACH_CONTAINER_INPUT\` call handler on the agent.
d536a13 Added API handler for ATTACH_CONTAINER_OUTPUT.
9d73fe7 Added a level of indirection for logger through an IO Switchboard.
d4ba90f Added new parameter for user to ContainerLogger's prepare function.
a030970 Added ContainerStatus.container_id to JSON endpoints.
```

# Issues blocked by this:
[DCOS-9722](https://mesosphere.atlassian.net/browse/DCOS-9722) Update UI to use Logging API instead of Files API
[DCOS-11245](https://mesosphere.atlassian.net/browse/DCOS-11245) `dcos task log` should work with logging API & nested containers (pods)
[...]

PR #1048, opened 2016-12-09
Added POSIX rlimits support to Mesos

bbannier:
Added 'posix/rlimits' to default Mesos isolator list.

This isolator enables the POSIX rlimits support in Mesos. It allows
frameworks to change the default resource limits (both hard and soft)
for their containers, similar to Docker --ulimit flag.

Some data services like Elastic Search or Cassandra typically require
a higher than normal resource limit for their containers, and will do
a check during startup to see if those limits are set properly. This
support in Mesos will unblock them running on DC/OS.

This isolator only affects the launch path of the container. If no
framework is using this feature, adding this isolator to the isolator
list will be a no-op.

# Issues

[MESOS-6402](https://issues.apache.org/jira/browse/MESOS-6402)

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

### DIFF
--- a/packages/dcos-integration-test/extra/test_rlimit.py
+++ b/packages/dcos-integration-test/extra/test_rlimit.py
@@ -1,0 +1,33 @@
+import subprocess
+import uuid
+
+
+def test_if_rlimits_can_be_used(cluster):
+    """This test verifies that rlimits can be used.
+
+    Since marathon does not support rlimits yet we use `mesos-execute` as
+    scheduler. We run a job for which we specify an unlimited `RLIMIT_CORE`.
+    The job then examines the actual limit, and returns a failure if another
+    value is found."""
+
+    name = 'test-rlimits-{}'.format(uuid.uuid4().hex)
+
+    argv = [
+        '/opt/mesosphere/bin/mesos-execute',
+        '--rlimits={"rlimits": [{"type":"RLMT_CORE"}]}',
+        '--master=leader.mesos:5050',
+        '--name={}'.format(name),
+        '--command=ulimit -c | grep -q unlimited',
+        '--shell=true',
+        '--env={"LC_ALL":"C"}']
+
+    output = subprocess.check_output(
+        argv,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True)
+
+    expected_output = \
+        "Received status update TASK_FINISHED for task '{name}'".format(
+            name=name)
+
+    assert expected_output in output
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -369,7 +369,7 @@ def calculate_config_yaml(user_arguments):
 
 def calculate_mesos_isolation(enable_gpu_isolation):
     isolators = ('cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,'
-                 'docker/runtime,docker/volume,volume/sandbox_path,'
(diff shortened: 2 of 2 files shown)


packages/dcos-integration-test/extra/test_rlimit.py
```diff
@@ -0,0 +1,27 @@
+import subprocess
+import uuid
+
+
+def test_if_rlimits_can_be_used(cluster):
+    '''This test verifies that rlimits can be used.
```
orsenthil: You could prefer double quotes `"` for this docstring style comments. I see that more common in our code base.

packages/dcos-integration-test/extra/test_rlimit.py
```diff
@@ -0,0 +1,27 @@
...
+        '--command=ulimit -c | grep -q unlimited',
+        '--shell=true',
+        '--env={"LC_ALL":"C"}']
+
+    output = subprocess.check_output(
+        argv,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True)
+
+    assert output.count('TASK_FINISHED')
```
orsenthil: Do we know how to many times "TASK_FINISHED" comes? 

* (Simply doing `assert "TASK_FINISHED" in output` is equivalent to the above `.count`. 
* Doing a regex search for the pattern associated with TASK_FINISHED might be "tighter"

bbannier: It appears exactly once. I went for a "tighter" search pattern nevertheless.

packages/dcos-integration-test/extra/test_rlimit.py
```diff
@@ -0,0 +1,33 @@
...
+        '--shell=true',
+        '--env={"LC_ALL":"C"}']
+
+    output = subprocess.check_output(
+        argv,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True)
+
+    expected_output = \
+        "Received status update TASK_FINISHED for task '{name}'".format(
```
orsenthil: The `name=name` could come up in line 30.   

(FYI: We check for 120 col rule and in fact the entire line 29 fits into that. The current indentation at line 29 looks better IMO)

bbannier:
Hi @lingmann, I rebased on top of #977 which includes the changelog information.

spahl:
this needs to be rebased on master

spahl:
cc @adam-mesos

bbannier:
@lingmann: Re:your request, as this is now rebased on top of `master` this PR doesn't include component updates anymore.

spahl:
@bbannier something is causing this not to build and fail code-quality checks now.

spahl:
This looks good to me I'm waiting for tests to pass for official approval.

adam-mesos:
All green now! :green_heart:

packages/dcos-integration-test/extra/test_rlimit.py
```diff
@@ -0,0 +1,33 @@
...
+            name=name)
+
+    assert expected_output in output
```
mellenburg: It is now clear to me how this tests what is described in the docstring:

> The job then examines the actual limit, and returns a failure if another value is found

This just makes sure that TASK_FINISHED was returned somewhere in the output. When a task finishes with an nonzero exit code, does it produce a different string? From a layperson's view, it would follow that a non-zero exit would still get a "TASK_FINISHED" message

PR #1049, opened 2016-12-09
Fixed upgrade issue (1.8->1.9) with overlay-modules.

asridharan:
* Allows overlay upgrades from 1.8 -> 1.9
* Allows any agent configuration change (disabling mesos or docker network) to be allowed during upgrades
*Earlier the test cases were assuming that the configuration applied by the Master will always be the configuration given to the network configuration of the agent. This is obviously not true during upgrades since the overlay configuration for a specific agent is retrieved from the replicated log.
  We are therefore changing the test to learn the overlay configuration being applied by the Master, and ensure that the configuration is reflected on the agent end-point.
# Issues

[DCOS-557](https://dcosjira.atlassian.net/browse/DCOS-557)
...

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

### DIFF
--- a/packages/dcos-integration-test/extra/test_endpoints.py
+++ b/packages/dcos-integration-test/extra/test_endpoints.py
@@ -167,18 +167,28 @@ def test_if_overlay_master_is_up(cluster):
 
 
 def test_if_overlay_master_agent_is_up(cluster):
-    r = cluster.get('/mesos/overlay-agent/overlay')
-    assert r.ok, "status_code: {}, content: {}".format(r.status_code, r.content)
+    master_response = cluster.get('/mesos/overlay-master/state')
+    assert master_response.ok,\
+        "status_code: {}, content: {}".format(master_response.status_code, master_response.content)
+
+    master_overlay_json = master_response.json()
+
+    agent_response = cluster.get('/mesos/overlay-agent/overlay')
+    assert agent_response.ok, "status_code: {}, content: {}".format(agent_response.status_code, agent_response.content)
 
     # Make sure the `dcos` overlay has been configured.
-    json = r.json()
+    agent_overlay_json = agent_response.json()
 
-    assert 'overlays' in json
-    assert len(json['overlays']) == 1
+    assert 'ip' in agent_overlay_json
+    agent_ip = agent_overlay_json['ip']
 
+    assert 'overlays' in agent_overlay_json
+    assert len(agent_overlay_json['overlays']) == 1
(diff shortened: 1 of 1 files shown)


asridharan:
@jieyu @kaysoky @cmaloney this is ready for review.

Ran the upgrade integration tests on this PR to verify that it passes.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -191,6 +209,28 @@ def test_if_overlay_master_agent_is_up(cluster):
     except KeyError as ex:
         raise AssertionError("Could not find key 'subnet':" + str(ex)) from ex
 
+    # Pop mesos and docker bridge if they have been configured on the
```
orsenthil: One high level question.  I was trying to understand the exact assertion as an integration test here.

If we all we care about is *agent_overlay_agent_response* should contain either `mesos_bridge` or `docker_bridge`  is it better to assert that straight instead of this `KeyError and assert pattern` ?

orsenthil: It's alright. I realize it was decided to use the pattern of `pop-out` values and do a `dict` compare at the end.

amitaekbote:
@mellenburg can you take a look at this?

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -167,18 +167,37 @@ def test_if_overlay_master_is_up(cluster):
...
     # Make sure the `dcos` overlay has been configured.
-    json = r.json()
+    agent_overlay_json = agent_response.json()
+
+    assert 'ip' in agent_overlay_json
+    agent_ip = agent_overlay_json['ip']
 
-    assert 'overlays' in json
-    assert len(json['overlays']) == 1
+    # Get the Mesos and Docker bridge configuration for this agent from
```
orsenthil: @asridharan  - can you move these lines 185 to 192 to line number 212. (It will be much easier to read and follow after this movement).

Everything else looks good to me. So Ship-it from me.

orsenthil:
🚢   

LGTM.

lingmann:
Hi @asridharan what is the issue you were intending to link to? https://dcosjira.atlassian.net/browse/557 is giving me a broken link...

asridharan:
@lingmann sorry typo in link. Just updated the link.

PR #1053, opened 2016-12-13
minor improvements in docs-integration-test.

orsenthil:
- Cosmetic Improvements throughout the module.

@mellenburg : Should be easy to review!

/cc
@cmaloney , @lingmann and @spahl  : This should be ready to go.

### DIFF
--- a/packages/dcos-integration-test/extra/test_dcos_log.py
+++ b/packages/dcos-integration-test/extra/test_dcos_log.py
@@ -4,7 +4,7 @@ import requests
 
 
 def validate_json_entry(entry: dict):
-    required_fields = set(['fields', 'cursor', 'monotonic_timestamp', 'realtime_timestamp'])
+    required_fields = {'fields', 'cursor', 'monotonic_timestamp', 'realtime_timestamp'}
 
     assert set(entry.keys()) <= required_fields, (
         "Entry didn't have all required fields. Entry fields: {}, required fields:{}".format(entry, required_fields))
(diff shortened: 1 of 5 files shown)


packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -32,7 +32,7 @@ def test_log_text(cluster):
 
         # expect 10 lines
         lines = filter(lambda x: x != '', response.content.decode().split('\n'))
-        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(list(lines)), lines)
```
branden: This isn't correct either, the second `len(list(lines))` will always evaluate to `0` since the `lines` generator has been consumed by the first `list()` in the assertion. Should amend the above line to be `lines = list(filter(...))` and then the two `len()`s in this line will work.

packages/dcos-integration-test/extra/test_dcos_log.py
```diff
@@ -72,4 +72,4 @@ def test_log_proxy(cluster):
         response = cluster.get('/system/v1/agent/{}/logs/v1/range/?skip_prev=10'.format(slave_id))
         check_response_ok(response, {'Content-Type': 'text/plain'})
         lines = filter(lambda x: x != '', response.content.decode().split('\n'))
-        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(lines), lines)
+        assert len(list(lines)) == 10, 'Expect 10 log entries. Got {}. All lines {}'.format(len(list(lines)), lines)
```
branden: Same as above.

packages/dcos-integration-test/extra/test_dcos_log.py
orsenthil: Yeah! 👍

orsenthil:
@branden - thanks for catching this. Fixed.
@spahl  - updated, thank you!

PR #1058, opened 2016-12-14
Use /var/lib/dcos/mesos for metrics module state path

spahl:
- fixed the state directory of the mesos volume to go to the standard path

# Issues

n/a

# Checklist

 - [ ] Change Log from last: n/a
 - [ ] Test Results: n/a
 - [ ] Code Coverage: n/a

### DIFF
--- a/packages/dcos-metrics/extra/metrics.json
+++ b/packages/dcos-metrics/extra/metrics.json
@@ -16,7 +16,7 @@
             {"key": "output_collector_chunk_size_datapoints", "value": "100"},
             {"key": "output_collector_chunk_timeout_seconds", "value": "10"},
             {"key": "state_path_dir",
-             "value": "/var/run/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
+             "value": "/run/dcos/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
         ]
     }]
   }]


alberts:
@spahl apparently this is about 1 kB of state that doesn't need to be persisted. I'd probably recommend that we put this under /run

spahl:
ok let's move this to /run

packages/dcos-metrics/extra/metrics.json
```diff
@@ -16,7 +16,7 @@
             {"key": "output_collector_chunk_size_datapoints", "value": "100"},
             {"key": "output_collector_chunk_timeout_seconds", "value": "10"},
             {"key": "state_path_dir",
-             "value": "/var/run/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
+             "value": "/var/lib/dcos/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
```
spahl: this module creates it.

packages/dcos-metrics/extra/metrics.json
```diff
@@ -16,7 +16,7 @@
             {"key": "output_collector_chunk_size_datapoints", "value": "100"},
             {"key": "output_collector_chunk_timeout_seconds", "value": "10"},
             {"key": "state_path_dir",
-             "value": "/var/run/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
+             "value": "/run/dcos/mesos/isolators/com_mesosphere_MetricsIsolatorModule/"}
```
orsenthil: Question.

Don't have to create this path before using?
I see at-least metronome doing this ( https://github.com/dcos/dcos/blob/master/packages/metronome/build#L27)

PR #1063, opened 2016-12-14
cluster-id for metrics agents

alberts:
Write /var/lib/dcos/cluster-id on agents for dcos-metrics service to use

# Issues

N/A

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not: checked in systemd unit

### DIFF
--- a/packages/bootstrap/extra/dcos_internal_utils/bootstrap.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/bootstrap.py
@@ -37,13 +37,17 @@ class Bootstrapper(object):
     def __exit__(self, type, value, tb):
         self.close()
 
-    def cluster_id(self, path):
+    def cluster_id(self, path, readonly=False):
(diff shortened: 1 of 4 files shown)


malnick:
LGTM, thanks @alberts

orsenthil:
The code looks good to me. But I am not confident on the logic. It requires `zk` knowledge and expertise. So, I will differ it to someone else.

jgehrcke:
Looks good, let's test this.

A small thing: the log msg emission in `_consensus()` is not currently adjusted to that readonly case: If `value` is `None`, it's currently silent: https://github.com/alberts/dcos/blob/3ce277b91bf1433c1e2c9208e69fb4a57da8a9e2/packages/bootstrap/extra/dcos_internal_utils/bootstrap.py#L80

We should maybe add a log statement after the read (some might say we should in addition short-circuit and `return` after successful create, but that's not for today).

PR #1064, opened 2016-12-14
Train 111

spahl:
#1031
#974
#1048
#1045


adam-mesos:
LGTM, just note that when bumping mesos-modules, we're now also building the logsink anonymous module from #1000 but it is not actually configured to be loaded. Presumably we'll rebase #1000 and get it in a future train.

spahl:
yeah we'll take care of #1000 later.

PR #1066, opened 2016-12-15
Late binding config

branden:
## High Level Description

Depends on https://github.com/dcos/dcos/pull/1138.

This PR introduces late-bound config values as a first-class concept within the gen library.

Immutable configuration is an important part of DC/OS's design, but there are some scenarios where a cluster has config values that can't be known until after the cluster has been launched. We call these values "late-bound" because they're bound after cluster config has been generated. For example, we want to configure the name of a DC/OS cluster launched from a CloudFormation template with its `StackName`, which doesn't exist until after the stack has been created from the template. We accomplish this by defining `cluster_name` in the config source for CloudFormation templates to be the variable reference `{ "Ref" : "AWS::StackName" }`, which is then included in the cluster template and resolved by CloudFormation when it parses the template and launches the cluster.

We currently bind late values by adding all DC/OS config content to the cloud config within CloudFormation and Azure cluster templates, which causes the template variable references within that config to be resolved by the cloud service that handles the template. No further config generation happens during node bootstrap. This approach has some limitations: We can't derive other config values from late-bound values, and cloud templates are unnecessarily large since they contain all cluster config.

This PR defines late-bound config values explicitly within config sources, allowing gen to treat them differently from other config values. When gen builds config from a source that includes late-bound values, it does a couple new things: it builds a late config package for config files that depend on the late-bound config values (with placeholders in its content for those values), and it adds a file `/etc/mesosphere/setup-flags/late-config.yaml` to the cloud config which contains a mapping from the late-bound config value names to the references that were defined in config sources. Since that file is defined in cloud config, it's included in cluster templates and the references within it are resolved when the template is parsed. Then when `pkgpanda setup` runs as part of bootstrap on each cluster node, Pkgpanda checks for the existence of `late-config.yaml`. If the file is present, Pkgpanda downloads the late config package, interpolates the bound values from `late-config.yaml` into the placeholders left in the late config package by gen, and then renders the complete package and adds it to the local Pkgpanda repository for later activation.

The impact of this change is apparent when you compare templates generated from this PR and master. It shrinks the simple single-master CloudFormation template by about 4 kB:

```
$ curl -s https://downloads.dcos.io/dcos/testing/master/cloudformation/single-master.cloudformation.json | wc -c
   48022
$ curl -s https://downloads.dcos.io/dcos/testing/pull/1066/cloudformation/single-master.cloudformation.json | wc -c
   43807
$
```

### Deferred Work

These missing pieces will be added in followup PRs.

* Factor out duplicated split-by-token logic #1163
* Convert all config packages from `.tar.xz` to `.dcos_config` YAML (no need for two different config package formats)
* Resume config calculation during node bootstrap (for config values derived from late-bound values)
 
## Related Issues

## Checklist for all PR's


gen/__init__.py
```diff
@@ -306,20 +305,67 @@ def do_gen_package(config, package_filename):
...
+            return
+
+        # Yield the string preceding the placeholder, if any.
+        if placeholder_start[0] > 0:
+            yield string[:placeholder_start[0]]
+
+        # Find the end of the placeholder.
+        placeholder_end = _next_substring(string, gen.internals.LATE_BIND_PLACEHOLDER_END, placeholder_start[1])
+        if not placeholder_end:
+            raise Exception("Bad late config file: Can't find end of placeholder")
```
cmaloney: would be good to include the reaminder of the string here so when this error comes up can see where in the string that was used as input the error is.

gen/__init__.py
```diff
@@ -306,20 +305,67 @@ def do_gen_package(config, package_filename):
...
+
+        # Extract the name between the start and end and yield its value.
+        name = string[placeholder_start[1]:placeholder_end[0]]
+        if name not in late_values:
+            raise Exception('Bad late config file: Found placeholder for unknown value "{}"'.format(name))
```
cmaloney: same as above

gen/__init__.py
```diff
@@ -306,20 +305,67 @@ def do_gen_package(config, package_filename):
...
+        yield from _interpolate_values(string[placeholder_end[1]:])
+
+    return ''.join(_interpolate_values(content))
+
+
+def resolve_late_package(config, late_values):
+
+    return {
+        'package': [
+            {k: render_late_content(v, late_values) if k == 'content' else v for k, v in fileinfo.items()}
```
cmaloney: Should probably check there are no instances of the LATE BIND placeholder in any of the other fields of the file just to keep things from possibly ending very badly with errors down the line.

gen/__init__.py
```diff
@@ -306,20 +305,67 @@ def do_gen_package(config, package_filename):
     log.info("Package filename: %s", package_filename)
 
 
-def extract_files_with_path(start_files, paths):
+def render_late_content(content, late_values):
```
cmaloney: talked about this in slack, but this is doing the same thing as https://github.com/dcos/dcos/blob/master/gen/build_deploy/azure.py#L23 and https://github.com/dcos/dcos/blob/master/gen/build_deploy/aws.py#L230, would be cool to make them all use one function which is tested and more readable in the future.

gen/__init__.py
```diff
@@ -306,20 +305,67 @@ def do_gen_package(config, package_filename):
...
+            {k: render_late_content(v, late_values) if k == 'content' else v for k, v in fileinfo.items()}
+            for fileinfo in config['package']
+        ]
+    }
+
+
+def extract_files_containing_late_variables(start_files):
     found_files = []
-    found_file_paths = []
     left_files = []
```
cmaloney: nit / bug from me: Would be nice to check LATE_BLIND_PLACEHOLD_START never is in any other key or value of the file_info dictionary and error early here. I forsee someone trying to late bind part of a path and it breaking badly.

gen/__init__.py
```diff
@@ -492,38 +570,56 @@ def generate(
...
+    cluster_package_info = {}
+
+    # Prepare late binding config, if any.
+    late_package = build_late_package(late_files, argument_dict['config_id'], argument_dict['provider'])
+    if late_variables:
+        # Render the late binding package. This package will be downloaded onto
+        # each cluster node during bootstrap and rendered into the final config
+        # using the values from the late config file.
+        late_package_id = PackageId(late_package['name'])
+        late_package_filename = make_package_filename(late_package_id, '.config')
```
cmaloney: Can we do something like `.dcos_config` so it's dcos specific?

gen/__init__.py
```diff
@@ -492,38 +570,56 @@ def generate(
...
+            'owner': 'root',
+            # TODO(cmaloney): don't prettyprint to save bytes.
+            # NOTE: Use yaml here simply to make avoiding painful escaping and
+            # unescaping easier.
+            'content': render_yaml({
+                'late_bound_package_id': late_package['name'],
+                'bound_values': late_variables
+            })})
+
+    # Render the rest of the packages.
```
cmaloney: do we still need these .tar.xz packages? Would be really nice to just ship all packages as yaml files (And would significantly speed up `tox -e py35-unittests release/test_release.py::make_channel_artifacts` among others

branden:
@cmaloney The `AssertionError` was happening because `dcos-config.yaml` has a switch on `oauth_available`, which was inferred to be late-bound due to its dependency on the explicitly late `oauth_enabled`, and of course there's no case for the placeholder value. This exposes the limitation that we can't switch on late-bound arguments unless we turn the late config into a real template and resolve it during bootstrap. I hacked around this by adding `oauth_available` to the sources for AWS and Azure, but we'll probably want a better solution in the long term.

gen/__init__.py
```diff
@@ -569,7 +569,12 @@ def generate(
     # of all arguments would want to include itself).
     # Explicitly / manaully setup so that it'll fit where we want it.
     # TODO(cmaloney): Make this late-bound by gen.internals
-    argument_dict['expanded_config'] = textwrap.indent(json_prettyprint(argument_dict), prefix='  ' * 3)
+    argument_dict['expanded_config'] = textwrap.indent(
```
cmaloney: Seems acceptable for now. Practically the late config being yaml rather than json is also due to escaping issues. It's tech debt which we can reasonably push off for now.

gen/calc.py
```diff
@@ -288,9 +276,7 @@ def validate_dcos_overlay_network(dcos_overlay_network):
 
 
 def calculate_oauth_available(oauth_enabled):
```
cmaloney: nit: this could just be `return oauth_enabled` now, optionally with a `validate_true_false`

gen/internals.py
```diff
@@ -602,7 +668,7 @@ def _calculate_target(self, target):
             if resolvable.is_error:
                 continue
 
-            assert resolvable.is_resolved, "uhhh: {}".format(resolvable)
+            assert resolvable.is_resolved or resolvable.is_late, "uhhh: {}".format(resolvable)
```
cmaloney: Should update my comment from `uhhh` to something more indicative what went wrong "Internal logic error: Variable was not resolved to a value or found to be late, but somehow we got out of the variable value calculation logic without an error."

gen/__init__.py
branden: I'm working on a followup PR that cleans up this bit of tech debt.

branden:
Ready for review! @cmaloney @amitaekbote @orsenthil

gen/__init__.py
```diff
@@ -306,20 +305,82 @@ def do_gen_package(config, package_filename):
     log.info("Package filename: %s", package_filename)
 
 
-def extract_files_with_path(start_files, paths):
+def render_late_content(content, late_values):
+    # TODO(branden): Unit tests
+
+    def _next_substring(string, substring, start=0):
```
mesosphere-mergebot: In this module, consider replacing the variable name string with something else, even `_string` alright.
There is a module that `string` that this variable can shadow. If we ever happen to use it, better to use the module as `string` than import it as `_string`.

Changing this variable name could be better.

orsenthil: In this module, consider replacing the variable name string with something else, even _string alright.
There is a module that string that this variable can shadow. If we ever happen to use it, better to use the module as string than import it as _string.

Changing this variable name could be better.

gen/internals.py
```diff
@@ -292,9 +324,10 @@ class Resolvable:
     - Keeps track of what caused the resolvable to enter existence (it's cause)
     """
 
-    UNRESOLVED = 'unresolved'
     ERROR = 'error'
+    LATE = 'late'
```
orsenthil: More of a todo: consider if enum would be suitable here.
1) https://docs.python.org/3/library/enum.html
2) https://github.com/mesosphere/mergebot/blob/master/mergebot/mergebot/interfaces.py#L13

gen/internals.py
```diff
@@ -543,6 +608,9 @@ def _ensure_finalized(self, resolvable):
         with self._stack_layer(resolvable.name):
             try:
                 resolvable.finalize_value(*self._calculate(resolvable))
+            except LateBoundException as ex:
```
orsenthil: `ex` is unused, can be killed.

gen/internals.py
branden: https://github.com/dcos/dcos/pull/1066/commits/b9ee7edb64d39991ea19d4c2c580d7a1c32e4287

gen/__init__.py
```diff
@@ -305,20 +305,82 @@ def do_gen_package(config, package_filename):
...
+        placeholder_end = _next_substring(string_, gen.internals.LATE_BIND_PLACEHOLDER_END, placeholder_start[1])
+        if not placeholder_end:
+            log.debug("Can't find end of placeholder in string: {}".format(repr(string_)))
+            raise Exception("Bad late config file: Can't find end of placeholder")
+
+        # Extract the name between the start and end and yield its value.
+        name = string_[placeholder_start[1]:placeholder_end[0]]
+        if name not in late_values:
+            log.debug('Found placeholder for unknown value "{}" in string: {}'.format(name, repr(string_)))
+            raise Exception('Bad late config file: Found placeholder for unknown value "{}"'.format(name))
```
orsenthil: Should this be a `LateboundException` or something else like `ConfigError` ?

gen/internals.py
```diff
@@ -284,6 +309,13 @@ class SkipError(Exception):
     pass
 
 
+# has a deterministic chunk before and after the variable name so we can deterministically get out
+# the name
+LATE_BIND_PLACEHOLDER_START = 'LATE_BIND_PLACEHOLDER_START_'
```
orsenthil: `PLACEHOLDER` usually means they are substituted /replaced. I guess, it is technically right. It is more than just a placeholder. It's a `MARKER` or `DELIMITER`.  Would those terms be more preferable instead of placeholder?

gen/internals.py
orsenthil: Cool. Thank you!

gen/internals.py
```diff
@@ -484,6 +527,20 @@ def all_conditions_met(setter):
...
+        # master_external_loadbalancer late bound, but still have `has_master_external_loadbalancer`
+        # bound regularly which is critical for proper operation.
+        def has_no_late_parameters(setter) -> bool:
+            for parameter in setter.parameters:
+                self._ensure_finalized(self._arguments[parameter])
+                if self._arguments[parameter].is_late:
+                    return False
+            return True
+
+        if len(feasible) > 1:
```
orsenthil: minor: Just `len(feasible)` should be enough or even `feasible`.

branden: That'd be true if it was comparing against 0, not 1.

orsenthil: ah! I see. ok.

gen/__init__.py
branden: It's not a recoverable error, so we wouldn't be catching it anyway.

gen/internals.py
orsenthil: Sounds ok to me. I don't have a strong stance.

I was getting biased from some templating stuff calling `{{` symbols  as delims rather than placeholders.  
http://jinja.pocoo.org/docs/2.9/templates/

> There are a few kinds of delimiters. The default Jinja delimiters are configured as follows:

branden: Syntactically these are delimiters, and semantically (which we know because this isn't a general-purpose template language) they're the start and end of late bind placeholders.

gen/__init__.py
orsenthil: ok. Would `ConfigError` be better or is it not applicable? If applicable, I think, the type signifies meaning along with the information that is given in the string.

branden: Just not applicable. This would be a serious error that requires debugging what went wrong between build and cluster bootstrap, and the error may be only incidentally related to config generation.

orsenthil: I see, what you mean.

Sounds good to me. I don't have a good case for a simple `Exception` being better than a specific one. The only reason seems, we have it our code base. 

LGTM.

orsenthil:
the changes look good to me. 👍

gen/__init__.py
```diff
@@ -306,20 +305,82 @@ def do_gen_package(config, package_filename):
     log.info("Package filename: %s", package_filename)
 
 
-def extract_files_with_path(start_files, paths):
+def render_late_content(content, late_values):
+    # TODO(branden): Unit tests
```
amitaekbote: Remove this todo.

gen/__init__.py
```diff
@@ -464,15 +543,42 @@ def generate(
...
+                    continue
+
+                if setter.name not in resolver.late:
+                    continue
+
+                # Skip late vars that aren't referenced by config.
+                if not resolver.arguments[setter.name].is_finalized:
+                    continue
+
+                # Validate a late variable should only have one source.
```
amitaekbote: Would something like a cluster name have multiple sources? Also this seems to be in a loop for each source so why do we need this check?

pkgpanda/build/__init__.py
```diff
@@ -410,47 +409,6 @@ def get_docker_id(docker_name):
     return check_output(["docker", "inspect", "-f", "{{ .Id }}", docker_name]).decode('utf-8').strip()
 
 
-def hash_str(s: str):
-    hasher = hashlib.sha1()
```
amitaekbote: Is there a reason these functions moved to util? was it part of "it makes more sense there" or required?

gen/__init__.py
branden: It's removed when unit tests are added at https://github.com/dcos/dcos/pull/1163.

gen/__init__.py
branden: It wouldn't, but this is one of the checks that enforces that. Without this check we could silently allow multiple late expressions for a variable in sources, by selecting whichever gets iterated through last.

pkgpanda/build/__init__.py
branden: To prevent an import cycle. See https://github.com/dcos/dcos/pull/1138/commits/e4facff15584ab6d8ddfa0adda06c60ff359e6ca

PR #1067, opened 2016-12-15
De-Duplicate Logs

malnick:
Use env for logs in order to deduplicate package usage across DC/OS variants. 

# Issues

[DCOS-<number>](https://dcosjira.atlassian.net/browse/DCOS-<number>)
...

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -338,6 +338,9 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log.env
+    content: |
+      DCOS_LOG_CONFIG_PATH=/opt/mesosphere/etc/dcos-log-config.json
   - path: /etc/dcos-log-config.json
     content: |
       {
--- a/packages/dcos-log/build
+++ b/packages/dcos-log/build
@@ -20,37 +20,9 @@ mkdir -p $(dirname $master_service)
 mkdir -p $(dirname $slave_service)
 mkdir -p $(dirname $slave_public_service)
 
-cat <<EOF | tee $master_service
-[Unit]
-Description=DC/OS Log service master: DC/OS logging service
-[Service]
-PermissionsStartOnly=True
-EnvironmentFile=/opt/mesosphere/environment
-Restart=always
-StartLimitInterval=0
-RestartSec=5
-User=dcos_log
-ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log-master
-ExecStart=/opt/mesosphere/bin/dcos-log -config /opt/mesosphere/etc/dcos-log-config.json
-EOF
-
-cat <<EOF | tee $slave_service $slave_public_service
-[Unit]
-Description=DC/OS Log service agent: DC/OS logging service
-[Service]
-PermissionsStartOnly=True
-EnvironmentFile=/opt/mesosphere/environment
-Restart=always
-StartLimitInterval=0
-RestartSec=5
-User=dcos_log
-ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log-agent
-ExecStart=/opt/mesosphere/bin/dcos-log -config /opt/mesosphere/etc/dcos-log-config.json
-EOF
-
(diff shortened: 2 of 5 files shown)


gen/dcos-config.yaml
```diff
@@ -338,6 +338,9 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log.env
+    content: |
+      DCOS_LOG_CONFIG_PATH=/etc/dcos-log-config.yaml
```
alberts: yaml or json? which is it? :)

packages/dcos-log/build
```diff
@@ -40,12 +42,14 @@ Description=DC/OS Log service agent: DC/OS logging service
...
 EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=/opt/mesosphere/etc/dcos-log.env
+EnvironmentFile=-/opt/mesosphere/etc/dcos-log-extra.env
 Restart=always
 StartLimitInterval=0
 RestartSec=5
 User=dcos_log
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log-agent
-ExecStart=/opt/mesosphere/bin/dcos-log -config /opt/mesosphere/etc/dcos-log-config.json
+ExecStart=/opt/mesosphere/bin/dcos-log -config $DCOS_LOG_CONFIG_PATH
```
alberts: ${ }

Basic environment variable substitution is supported. Use "${FOO}" as part of a word, or as a word of its own, on the command line, in which case it will be replaced by the value of the environment variable including all whitespace it contains, resulting in a single argument. Use "$FOO" as a separate word on the command line, in which case it will be replaced by the value of the environment variable split at whitespace, resulting in zero or more arguments. For this type of expansion, quotes are respected when splitting into words, and afterwards removed.

https://www.freedesktop.org/software/systemd/man/systemd.service.html

packages/dcos-log/build
```diff
@@ -26,12 +26,14 @@ Description=DC/OS Log service master: DC/OS logging service
...
 EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=/opt/mesosphere/etc/dcos-log.env
+EnvironmentFile=-/opt/mesosphere/etc/dcos-log-extra.env
 Restart=always
 StartLimitInterval=0
 RestartSec=5
 User=dcos_log
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log-master
-ExecStart=/opt/mesosphere/bin/dcos-log -config /opt/mesosphere/etc/dcos-log-config.json
+ExecStart=/opt/mesosphere/bin/dcos-log -config $DCOS_LOG_CONFIG_PATH
```
alberts: ${...}

packages/dcos-log/extra/dcos-log-agent.service
```diff
@@ -0,0 +1,14 @@
...
+EnvironmentFile=/opt/mesosphere/environment
+EnvironmentFile=/opt/mesosphere/etc/dcos-log.env
+EnvironmentFile=-/opt/mesosphere/etc/dcos-log-extra.env
+Restart=always
+StartLimitInterval=0
+RestartSec=5
+User=dcos_log
+ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-log-agent
+ExecStart=/opt/mesosphere/bin/dcos-log -config ${DCOS_LOG_CONFIG_PATH} 
+
```
alberts: let's drop these stray lines at the end of the units

gen/dcos-config.yaml
```diff
@@ -338,6 +338,9 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
+  - path: /etc/dcos-log.env
+    content: |
+      DCOS_LOG_CONFIG_PATH=/etc/dcos-log-config.json
```
alberts: /opt/mesosphere/etc.,,,

darkonie: this has to be `/opt/mesosphere/etc/dcos-log-config.json`

PR #1075, opened 2016-12-19
Avoid tempfile cannot be deleted error

mellenburg:
Test logs currently generate this very unhelpful message that makes debugging much harder

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not - N/A logging improvement

### DIFF
--- a/ssh/tunnel.py
+++ b/ssh/tunnel.py
@@ -2,9 +2,9 @@
 Module for creating persistent SSH connections for use with synchronous
 commands. Typically, tunnels should be invoked as a context manager to
 ensure proper cleanup. E.G.:
-with Tunnel(*args, **kwargs) as tunnel:
-    tunnel.write_to_remote('/usr/local/usrpath/testfile.txt', 'test_file.txt')
-    tunnel.remote_cmd(['cat', 'test_file.txt'])
+with tunnel(*args, **kwargs) as my_tunnel:
+    my_tunnel.write_to_remote('/usr/local/usrpath/testfile.txt', 'test_file.txt')
+    my_tunnel.remote_cmd(['cat', 'test_file.txt'])
 """
 import logging
 import tempfile
@@ -56,7 +56,8 @@ class Tunnelled():
 
 @contextmanager
 def tunnel(user: str, key_path: str, host: str, port: int=22):
-    tunnel_socket = tempfile.NamedTemporaryFile()
+    # This will be cleaned up by SSH when the tunnel is closed
+    tunnel_socket = tempfile.NamedTemporaryFile(delete=False)
     target = user + '@' + host
 
     base_cmd = [


mellenburg:
@lingmann this is ready for review. I checked the logs and this red herring message is now gone. @orsenthil please take a look

mellenburg:
@branden Small change to clean up logs PTAL

orsenthil:
LGTM.

ssh/tunnel.py
```diff
@@ -81,6 +82,9 @@ def tunnel(user: str, key_path: str, host: str, port: int=22):
     logger.debug('Closing SSH Tunnel: ' + ' '.join(close_tunnel))
     check_call(close_tunnel)
 
+    if os.path.exists(tunnel_socket.name):
+        os.remove(tunnel_socket.name)
```
lingmann: Why is this wrapped in a conditional, is there a possibility that `tunnel_socket` will not exist?

branden: The `os.remove()` you added at the end won't get called if an exception is raised before it's reached, so this creates possible execution paths where the tempfile doesn't get cleaned up. I'd suggest creating the tempfile within another context manager that's responsible for checking if it's present before attempting to delete it.

orsenthil: @lingmann , notice the tempfile `/teamcity/temp/buildTmp/tmp0ys0e50q` here, which is the NamedTemperoryFile (tunnel_socket.name) that is used with SSH tunnel command `-oControlPath`.

The theory is, when ssh exit happens, the ControlPath is cleaned up too. that's why the current Exception (harmless) is observed.


```
[11:13:37]	[Step 1/1] [2016-12-20 19:13:37,181|ssh.tunnel|DEBUG]: Closing SSH Tunnel: /usr/bin/ssh -oConnectTimeout=10 -oControlMaster=auto -oControlPath=/teamcity/temp/buildTmp/tmp0ys0e50q -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oBatchMode=yes -oPasswordAuthentication=no -p 22 -O exit core@35.156.217.233
[11:13:37]	[Step 1/1] Exit request sent.
[11:13:37]	[Step 1/1] Exception ignored in: <bound method _TemporaryFileCloser.__del__ of <tempfile._TemporaryFileCloser object at 0x7fccadf82438>>
[11:13:37]	[Step 1/1] Traceback (most recent call last):
[11:13:37]	[Step 1/1]   File "/usr/lib/python3.5/tempfile.py", line 589, in __del__
[11:13:37]	[Step 1/1]     self.close()
[11:13:37]	[Step 1/1]   File "/usr/lib/python3.5/tempfile.py", line 585, in close
[11:13:37]	[Step 1/1]     unlink(self.name)
[11:13:37]	[Step 1/1] FileNotFoundError: [Errno 2] No such file or directory: '/teamcity/temp/buildTmp/tmp0ys0e50q'
```

By checking and cleaning, I guess, it is just being proactive to verify if the file that we asked Tempfile module to not clean (delete=False), exists and if so, we clean it.

mellenburg: Definitely agree to add a comment.

@branden Frankly, this file doenst even need to be deleted and I would rather completely remove this bit and leave it at NamedTemporaryFile(delete=False). Creating an entire context manager just to cover the case where we might have an early exception and then leave this harmless file seems like cruft in my opinion

branden: A comment explaining why we're not cleaning up the file, and getting rid of this `os.remove()`, works for me. 

Ideally the code that's responsible for creating the file ought to be deleting it. If we could create a tmp filename and pass that to SSH and let it own the lifecycle of the file, that'd be ideal. We could append a random string to `tempfile.gettempdir()` and let SSH own the file entirely.

mellenburg:
@lingmann @branden unnecessary logic removed and explanatory comment added

PR #1076, opened 2016-12-20
Bump dcos-metrics to 1.0.0-rc4

rji:
Bump dcos-metrics to 1.0.0-rc4 to resolve an issue fetching the framework principal.

# Issues

N/A

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

If this is a component/package update, include

 - [x] Change Log from last: https://github.com/dcos/dcos-metrics/compare/1.0.0-rc3...1.0.0-rc4
 - [x] Test Results: https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-metrics/job/public-dcos-metrics-master/72/
 - [x] Code Coverage: https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-metrics/job/public-dcos-metrics-master/72/cobertura/

PR #1077, opened 2016-12-20
Train 113

lingmann:
#989 
#1076

### DIFF
--- a/packages/dcos-integration-test/extra/test_helpers.py
+++ b/packages/dcos-integration-test/extra/test_helpers.py
@@ -1,0 +1,3 @@
+from pkgpanda.util import load_json
+
+dcos_config = load_json('/opt/mesosphere/etc/expanded.config.json')
--- a/test_util/cluster.py
+++ b/test_util/cluster.py
@@ -71,7 +71,6 @@ class Cluster:
         self,
         ssh_info,
         ssh_key_path,
-        provider,
         masters,
         agents,
         public_agents,
@@ -81,7 +80,6 @@ class Cluster:
         gen.calc.validate_num_masters(str(len(masters)))
 
         self.ssher = Ssher(ssh_info.user, ssh_info.home_dir, ssh_key_path)
-        self.provider = provider
         self.masters = masters
         self.agents = agents
         self.public_agents = public_agents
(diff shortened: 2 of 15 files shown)


packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -0,0 +1,3 @@
+from pkgpanda.util import load_json
+
+dcos_config = load_json('/opt/mesosphere/etc/expanded.config.json')
```
adam-mesos: Should we name this differently, so it doesn't conflict with the existing `dcos_config` parameter in the dcos-enterprise downstream?

test_util/cluster.py
```diff
@@ -256,9 +252,6 @@ def install_dcos(
     install_prereqs=True,
     install_prereqs_only=False,
 ):
-    if cluster.provider != 'onprem':
-        raise NotImplementedError('Install is only supported for onprem clusters')
```
adam-mesos: Will this not be hit for cloud installs?

test_util/cluster.py
```diff
@@ -91,10 +89,7 @@ def __init__(
             'All cluster hosts require a private IP. hosts: {}'.format(repr(self.hosts))
         )
 
-        if self.provider == 'onprem':
-            self.check_ssh(self.hosts)
```
adam-mesos: No longer needed? Why did we check_ssh for all `hosts` onprem instead of just `masters` in the first place?

packages/dcos-integration-test/extra/test_helpers.py
spahl: We'll change downstream for this.

lingmann:
@mellenburg can you address @adam-mesos's questions above?

test_util/cluster.py
mellenburg: There is no such thing as an onprem install from cloudformation, so this message will never get hit. Documentation should be improved to start that cloudformation is really "DCOS provided via cloudformation template" as opposed to the incorrect interpretation of DC/OS installed on-prem using a cloudformation VPC (which is the from_vpc method)

test_util/cluster.py
mellenburg: All hosts are checked because this class is made specifically for facilitating upgrades in onprem deploys, which requires lots of SSH tidbits. In theory this could be done with a daisy chain through a master, but at the moment that would require a bit too much effort or would increase flakiness. Anyway, this check was preserved below in the from_vpc method. The reason it was added there is it was really just required because AWS will say a VPC CF is ready to go before SSH will accept connections, so you need to make sure all hosts are ready. This could also be done in from_hosts, but I would rather we just make this so that its a check we only do for upgrades or for specific provisioning where we know its required

packages/dcos-integration-test/extra/test_helpers.py
mellenburg: Note: there is no conflict downstream as these files will live in totally different namespaces, but I agree with changing this to expanded_config and will get that in one of the future cleanups

PR #1078, opened 2016-12-20
flake8 plugin for dcos.

orsenthil:

* Provides ability to add custom rule checks.

# Issues

[DCOS-12408](https://mesosphere.atlassian.net/browse/DCOS-12408)
...

# Checklist

* The code style test will fail with the introduction of this plugin. Proof that the checking with rules works.
```
./packages/dcos-integration-test/extra/test_3dt.py:27:9: D001 Invalid assertion of type assert <symbol>.ok found.
```

# Example Failure

* https://teamcity.mesosphere.io/viewLog.html?buildId=517381

### DIFF
--- a/flake8_dcos_lint/check_rules.py
+++ b/flake8_dcos_lint/check_rules.py
@@ -1,0 +1,10 @@
+import re
+
+from collections import namedtuple
+
+CheckRegex = namedtuple("CheckRegex", "regex code reason")
+
+
+regex_rules = [
+    CheckRegex(re.compile(r'assert\s(\w+\.ok)$'), "D001", "Invalid assertion of type assert <symbol>.ok found.")
+]
--- a/flake8_dcos_lint/check_rules.py
+++ b/flake8_dcos_lint/check_rules.py
@@ -6,5 +6,6 @@ CheckRegex = namedtuple("CheckRegex", "regex code reason")
 
 
 regex_rules = [
-    CheckRegex(re.compile(r'assert\s(\w+\.ok)$'), "D001", "Invalid assertion of type assert <symbol>.ok found.")
+    # Prefer asserting actual == expected, instead of asserting on .ok on a result object.
+    CheckRegex(re.compile(r'assert\s(\w+\.ok)$'), "D001", "Assertion of type assert <symbol>.ok detected.")
 ]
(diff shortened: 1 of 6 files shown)


flake8_dcos_lint/check_rules.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+from collections import namedtuple
+
+CheckRegex = namedtuple("CheckRegex", "regex code reason")
+
+
+regex_rules = [
+    CheckRegex(re.compile(r'assert\s(\w+\.ok)$'), "D001", "Invalid assertion of type assert <symbol>.ok found.")
```
branden: Can we add a comment here with the rationale for this rule? It'll at least be something to point to when someone tries to add an `ok` attribute to their class and runs into this.

orsenthil: So, the rationale is probably this. (I had to figure out myself as I got the rule from hearsay)


```
    @property
    def ok(self):
        try:
            self.raise_for_status()
        except HTTPError:
            return False
        return True
```

* This hides the status_code and it is better to verify the status-code than a boolean that hides the status_code.
* It's not documented in the API (http://docs.python-requests.org/en/master/api/#requests.Response) 

I agree a short-comment, that states the reason for the rule will be a better and will help with buy-in.

orsenthil:
@mellenburg 👍  Will point to this (http://flake8.pycqa.org/en/latest/plugin-development/index.html) 

As a flake8 extension, and it get's rolled up into flake8 command by default. After registering this extension, just running `flake8` will catch the new kind of errors that we look for in the plugins.

orsenthil:
@branden and @mellenburg , please take a look again.

flake8_dcos_lint/check_rules.py
branden: I think it'd be worth putting this entire rationale in the comment for posterity. The current comment leaves out the context that this is about the `requests` API, for example. Anyone who runs into this error and tries to dig into it would probably appreciate the full explanation, and it's something that we can use if we ever revisit this rule or wonder why it's there.

orsenthil: I've addressed this comment. thanks!

lingmann:
@branden please take another look

PR #1080, opened 2016-12-20
Upgrade Spartan to have cleaner logging

sargun:
There were two issues that are cleaned up in this PR
1. Every time we had a TCP connection idle out (and not close cleanly), we printed a warning to the console
2. It bumps the FD limit for Spartan

# Checklist

Component change:
https://github.com/dcos/spartan/commit/baefef9e613ed489d208bf273f8bdfdf8efef7b5

### DIFF
--- a/packages/spartan/buildinfo.json
+++ b/packages/spartan/buildinfo.json
@@ -4,7 +4,7 @@
     "spartan": {
       "kind": "git",
       "git": "https://github.com/dcos/spartan.git",
-      "ref": "92320ea8c802eff8d06d0b02139f162581b353cd",
+      "ref": "baefef9e613ed489d208bf273f8bdfdf8efef7b5",
       "ref_origin": "master"
     }
   }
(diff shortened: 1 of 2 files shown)


orsenthil:
@sargun - It will be helpful to a test case that  either

1) Fails w/o this change and succeeds after change. Is this applicable here?
2) Integration test case that asserts the changes that included. (The `LimitNOFILE=32000:32000` part especially.)

mellenburg:
@orsenthil 
1. This is a logging change to cleanup, not to specifically catch something, so I think it would be a bit beyond the pail to make this assert that newly formatted messages appear
2. LimitNOFILE is a systemd property. All of our other tests trust that systemd does what it says it will, so this seems like an odd point to start checking that it works as intended

sargun:
This is more of a performance thing. Spartan hits its fd limit early, and this can limit the number of simultaneous TCP queries it can do.

orsenthil:
Ok. sounds good on the reasoning shared above.  LGTM too.

jgehrcke:
> Spartan hits its fd limit early, and this can limit the number of simultaneous TCP queries it can do.

Under which conditions? Do we have a JIRA ticket that documents these findings?

I understand that we don't add a `LimitNOFILE=32000:32000` test here in this PR. But I think we should document the requirement that lead to this configuration change, and link to this documentation in the unit file. It's useful documentation even if it's as arbitrary as 'Required for a 1000 node cluster, otherwise ...'.

PR #1085, opened 2016-12-21
Network staging

sargun:
Do not merge this

We need to do this because we need to test a train.


sargun:
Here is the performance report, CC: @spahl, @orsenthil @mellenburg @lingmann @branden 
https://modeanalytics.com/mesosphere/reports/e4d624f6d9af/runs/cfcd856393e5

PR #1079, opened 2016-12-20
dcos-log: remove bootstrap

darkonie:
no bootstrap required anymore for dcos-logs in upstream

High level description including:
 - Bump dcos-log
 - Remove the bootstrap from dcos-log

# Checklist

 - [ ] Included a test which will fail if code is reverted but test is not
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

If this is a component/package update, include

 - [x] Change Log from last: https://github.com/dcos/dcos-log/pull/28 https://github.com/dcos/dcos-go/pull/21
 - [x] Test Results: https://jenkins.mesosphere.com/service/jenkins/job/dcos-log%20pull%20requests/71/
 - [x] Code Coverage: https://jenkins.mesosphere.com/service/jenkins/job/dcos-log%20pull%20requests/71/

PR #1087, opened 2016-12-22
Train 114

lingmann:
#1080 
#1010 
#1079 
#1078 
#1075


lingmann:
cc: @spahl and @adam-mesos

test_util/run-all
```diff
@@ -68,6 +68,7 @@ exhibitor_zk_hosts: 192.168.65.50:2181
 exhibitor_zk_path: /zk-shared
 master_discovery: static
 check_time: 'false'
+enable_lb: 'false'
```
adam-mesos: @sargun Shouldn't we test both true and false? It defaults to true, but we're only testing false here?

mellenburg:
@adam-mesos if default is true for enable_lb, then the AWS, onprem (CLI and API), and Azure tests are covering it. if anything, one of those should also be set to false

test_util/run-all
mellenburg: This file only covers vagrant test, so it is being tested true else where

jgehrcke:
I have looked at the 5 pull requests linked here and they LGTM!

Nit: I'd love to see a comment in the code that contains or links to background information on `LimitNOFILE` in Spartan's unit file.

PR #1089, opened 2016-12-22
Reduce 3dt test verbosity

mellenburg:
-way too much unused information is being pumped out by default
-attempts to limit data dump to failure scenarios only
# Checklist

### DIFF
--- a/packages/dcos-integration-test/extra/test_3dt.py
+++ b/packages/dcos-integration-test/extra/test_3dt.py
@@ -27,7 +27,6 @@ def make_3dt_request(cluster):
         assert response.status_code == 200
         try:
             json_response = response.json()
-            logging.info('Response: {}'.format(json_response))
         except ValueError:
             logging.exception('Could not deserialize response contents:{}'.format(response.content.decode()))
             raise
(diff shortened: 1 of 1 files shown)


mellenburg:
@orsenthil I think I might cull some more log statements that make these logs so huge that they are unusable. Specifically:
-Marathon response dumps when everything is fine. Would rather only dump on failure. https://teamcity.mesosphere.io/viewLog.html?buildId=521320&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_ProvisioningTestAwsDe_2&tab=buildLog#_focus=1465
-3dt validating messages for units and nodes and 3dt unit response messages. I fail to see any purpose in getting this data when it conforms to expectation

What do you think? Are there other ones that should be removed?

orsenthil:
@mellenburg , I am open to that. 
If there are way too many log.infos that displayed in the console, than another idea to consider will be to control them with a "--verbose" switch. (I felt a need for controlling the verbosity of logging in  tests via switch/config once).

mellenburg:
@darkonie PTAL. As it stands, the 3dt tests produce more output than the rest of the test combined, most of it repetitive data with little insight. Due to this, the logs take forever to load and are near impossible to visually inspect. Ideally, we would make a utility in conftest so at the end of any given test run with failure we pump out the wealth the system health info to disk and store as an artifact

mellenburg:
@lingmann This is ready for review

darkonie:
🚢

PR #1088, opened 2016-12-22
 Launch prereqs

mellenburg:
- no longer pass ssh_key_path, but rather the key contents
  ssh/tunnel.py now handles creating a temp key file with the
  correct permissions (and cleans it up)
- safer passing of AWS parameters by casting to str
- no longer need to write to remote host in order to test;
  now, everything is injected as a command
- removes unused code in ssh/tunnel.py
- removes unused code in test_util/installer_api_test.py
- no longer pass add_env separately in internals

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not (N/A this PR removes cruft)
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #1086, opened 2016-12-21
Use Marathon 1.4.0-RC4.

aquamatthias:
High level description including:

- Bug fix release for Marathon. We discovered a serious migration issue in RC3.

# Issues

- Fixes https://github.com/mesosphere/marathon/issues/4873 - Tasks with configured Marathon HealthChecks fail HealthChecks after migration to 1.4
- Fixes https://github.com/mesosphere/marathon/issues/4842 - by also writing deprecated pods to the proto.
- Fixes https://github.com/mesosphere/marathon/issues/4828 - Use async InstanceTracker methods in SchedulerActor
- Fixes https://github.com/mesosphere/marathon/issues/4882 - by renaming KillSelection enum values
- Fixes https://github.com/mesosphere/marathon/issues/4890 - Quote some vars and make sed more explicit 
- Fixes https://github.com/mesosphere/marathon/issues/4872 - Disallow usage of Command Checks on Pods until mesos supports them
- Fixes https://github.com/mesosphere/marathon/issues/4863 - Inform the rate limiter for pods and apps
- Fixes https://github.com/mesosphere/marathon/issues/4818 - Update kill selection in App as well.
- Fixes https://github.com/mesosphere/marathon/issues/4877 - fix various bugs in our RAML specification. add omitEmpty to some fields for cleaner output.

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

If this is a component/package update, include

 - [x] Change Log from last: https://github.com/mesosphere/marathon/releases/tag/v1.4.0-RC4
 - [x] Test Results: https://teamcity.mesosphere.io/viewType.html?buildTypeId=Oss_Marathon_Packages&branch_Oss_Marathon=v1.4.0-RC4&tab=buildTypeStatusDiv
 - [ ] Code Coverage: we have the data, but they are currently not published

PR #1083, opened 2016-12-20
Grab latest fix to package describe and install

jsancio:
High level description including:
 - What does this change enable
 - What bugs does this change fix
 - High-Level how things changed

# Issues

[DCOS-12411](https://mesosphere.atlassian.net/browse/DCOS-12411)

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not:
  1. https://github.com/dcos/cosmos/commit/da39604cdacbc4f79c84f397ec7fc0d14b83c47c#diff-c51a7c12e1cf4f27093f6f1caf7b54b4
  1. https://teamcity.mesosphere.io/viewQueued.html?itemId=520272
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

If this is a component/package update, include

- [x] Change Log from last:
 -  Return postInstallNotes instead of uninstall notes ([#560](https://github.com/dcos/cosmos/pull/560)) 
 - Make sure that we set None to false for selected and framework ([#559](https://github.com/dcos/cosmos/pull/559))
 - Add ServiceAlreadyStarted error ([#557](https://github.com/dcos/cosmos/pull/557))
 - Fix package ordering ([#558](https://github.com/dcos/cosmos/pull/558))
 - Remove appId field from ServiceStartRequest ([#555](https://github.com/dcos/cosmos/pull/555))
 - Integration tests for /service/start ([#552](https://github.com/dcos/cosmos/pull/552))
- [x] Test Results: Cosmos integration tests pass.
- [ ] Code Coverage:

PR #1093, opened 2016-12-23
Train 115

lingmann:
#1086 
#1088 
#1083


lingmann:
cc: @adam-mesos

lingmann:
@orsenthil can I get a second review on #1088 so it can land with this train? thanks! cc: @mellenburg

PR #1103, opened 2017-01-04
Factor out common / duplicate code

cmaloney:
- Factors out three different bits of duplicate code CodeClimate found (most the others it found (https://codeclimate.com/github/dcos/dcos/issues) I don't think are worth fixing at the moment or make code much more complicated to fix. These are low hanging fruit.

# Checklist

 - [x] Included a test which will fail if code is reverted but test is not N/A, simply de-duplicating existing code

### DIFF
--- a/dcos_installer/prettyprint.py
+++ b/dcos_installer/prettyprint.py
@@ -47,13 +47,12 @@ class PrettyPrint():
         self.success_hosts = [ip for ip in self.success_hosts if ip not in self.fail_hosts]
         return failed_data, success_data
 
-    def print_data(self):
-        print_header('OUTPUT FOR {}'.format(self.stage_name))
-        if len(self.failed_data) > 0:
-            for host in self.failed_data:
+    def _print_host_set(self, status, hosts):
+        if len(hosts) > 0:
+            for host in hosts:
                 for ip, data in host.items():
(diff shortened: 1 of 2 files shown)


dcos_installer/exhibitor_migrate.py
```diff
@@ -127,7 +118,7 @@ def perform():
             print("Migration already complete")
             sys.exit(0)
         else:
-            assert_can_migrate(config)
+            validate_config(config)
```
lingmann: So the expected values are different between these two functions (e.g. `validate_config()` uses `LOG_INDEX_DIR_GOAL` whereas `assert_can_migrate()` uses `LOG_INDEX_DIR_START`). Seems like different functionality... but if `validate_config()` is what we actually want then we should also clean up `LOG_INDEX_DIR_START` and friends since they don't seem to be used anymore.

cmaloney: Good catch. I've pulled out this commit / code. Could switch to having "start" and goal arrays which get indexed into which would reduce the code here a lot, but this script is temporary / going away, so easier just to leave it out for now.

lingmann:
@amitaekbote this is ready for the next train

PR #1106, opened 2017-01-05
DCOS-12673/fix_azure_flaky

orsenthil:
Attempts to fix certain infrastructure flakiness with Azure.

The main change is this.

```
@retry(retry_on_exception=retry_if_assertion_error, stop_max_attempt_number=45)
```

Rest of the changes are only cursory improvements in the code that I touched.

Issues

* [DCOS-12673](https://mesosphere.atlassian.net/browse/DCOS-12673)

Checklist

[ ] Included a test which will fail if code is reverted but test is not
[ ] Followed relevant code rules Rules for Packages and Systemd

### DIFF
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -8,9 +8,8 @@ import uuid
 import azure.common.credentials
 from azure.mgmt.network import NetworkManagementClient
 from azure.mgmt.resource.resources import ResourceManagementClient
-from azure.mgmt.resource.resources.models import (DeploymentMode,
-                                                  DeploymentProperties,
-                                                  ResourceGroup, TemplateLink)
+from azure.mgmt.resource.resources.models import DeploymentMode, DeploymentProperties, ResourceGroup, TemplateLink
+
 from retrying import retry
 
 import pkgpanda.util
@@ -19,10 +18,10 @@ from test_util.runner import integration_test
 
 
 def validate_env():
-    '''
+    """
     The following environment variables must be set to ensure the template can be deployed successfully and that
     parameters we rely on from a testing perspective are explicitly set and not relying on defaults which may change.
-    '''
+    """
     required = [
         'AZURE_PARAM_agentEndpointDNSNamePrefix',
         'AZURE_PARAM_linuxAdminUsername',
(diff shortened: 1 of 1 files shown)


orsenthil:
This is ready for review @lingmann , @branden , and @cmaloney .

test_util/azure_test_driver.py
```diff
@@ -124,14 +120,17 @@ def main():
         print("Creating template deployment ...")
         deploy_poller = rmc.deployments.create_or_update(group_name, deployment_name, deployment_properties)
 
-        # Stop after 45 attempts (each one takes up to one minute)
-        @retry(stop_max_attempt_number=45)
+        def retry_if_assertion_error(exception):
+            """Return True if we should retry (in this case when it's an AssertionError), False otherwise"""
+            return isinstance(exception, AssertionError)
+
+        @retry(retry_on_exception=retry_if_assertion_error, stop_max_attempt_number=45)
```
BenWhitehead: How frequently will the retry be attempted? I see it's limited to 45 attempts, but not sure on the retry interval.

orsenthil: The retry interval is defined by the thread wait time and assertion in the below.

```
             res = deploy_poller.result(timeout=60)
             ...
             assert deploy_poller.done(), "Not done deploying."		    
```

So, after every 60 seconds and assertion is made if the task is done and this is done for 45 times.

* Previously, the 45 times retry was done for any failure.
* Now, this retry of task is done only only carried out for `AssertionError`, any other Infrastructure issues like `CloudError` will fail fast.

Hope this helps.

BenWhitehead: got it, I figured the `timeout=60` meant the timeout for the rpc to azure had to return within 60 seconds.

Thanks for the clarification.

amitaekbote: Was the flakiness caused by AssertionError and the fact that we didnt retry on that particular failure?

orsenthil: Actually, retrying on AssertionError only is the right thing to do. The way the code is written, AssertionError is part of the logic for checking if the task is done and a retry is a while loop to for waiting until the task is complete. 

The unfortunate scenario when CloudError on Azure happens and it kills the task, we should not be retrying for 45 times, the clouderror must have killed the long running task and we should bail out then and exit with a failure.  This our fix (fail fast) is the fix the flaky azure tests which are not in our control.

HTH.

amitaekbote: makes sense, lgtm

PR #1112, opened 2017-01-06
Add parameters to test_upgrade_vpc.py to all overrides

BenWhitehead:
Two new environment variables have been added that allow providing a path to a config.yaml that defines override properties for the install of the cluster and the upgrade of the cluster.

* CONFIG_YAML_OVERRIDE_INSTALL
* CONFIG_YAML_OVERRIDE_UPGRADE

### Issues
[DCOS-9717](https://mesosphere.atlassian.net/browse/DCOS-9717)

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -27,6 +27,16 @@ DCOS_PYTEST_CMD: string(default='py.test -rs -vv ' + os.getenv('CI_FLAGS', ''))
     The complete py.test command used to run integration tests. If provided,
     CI_FLAGS is ignored.
 
+CONFIG_YAML_OVERRIDE_INSTALL: file path(default=None)
+    If provided the file specified will be loaded as a config.yaml and all properties specified
+    in the file will override any previously defined values.
+    This value will be used for the initial install of the cluster.
+
+CONFIG_YAML_OVERRIDE_UPGRADE: file path(default=None)
+    If provided the file specified will be loaded as a config.yaml and all properties specified
+    in the file will override any previously defined values.
+    This value will be used when upgrading the cluster.
+
 """
 import collections
 import datetime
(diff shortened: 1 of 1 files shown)


orsenthil:
Interesting!. That was a simple change.

branden:
Changes LGTM. Let's squash the two commits, or give the second one a better message.

test_util/test_upgrade_vpc.py
```diff
@@ -115,6 +115,9 @@ def main():
     stable_installer_url = os.environ['STABLE_INSTALLER_URL']
     installer_url = os.environ['INSTALLER_URL']
 
+    config_yaml_override_install = os.getenv('CONFIG_YAML_OVERRIDE_INSTALL')
+    config_yaml_override_upgrade = os.getenv('CONFIG_YAML_OVERRIDE_UPGRADE')
```
lingmann: I would suggest removing the "override" term from these variables. We have other environment variables which are also overriding defaults (see beginning of this file), but we don't explicitly call out that fact in the variable name.

lingmann:
@branden can you add your review using the github review feature? makes it easier to quickly scan the status.

test_util/test_upgrade_vpc.py
branden: There's a default config that these configs get merged with, so it might be helpful to someone who's debugging a failed job if there's a hint somewhere that these aren't the final configs that get used. No opinion on whether `OVERRIDE` is the right word to communicate that.

BenWhitehead: according to [`DcosApiInstall::genconf`](https://github.com/BenWhitehead/dcos/blob/9afb4fc00bc5190d4257c768901cba34f94f280f/test_util/installer_api_test.py#L143) the `add_config_path` arg represents a yaml that will override any previously specified values. I think it's important to keep the override in the name so that people know it's the highest priority of any passed in value(s).

BenWhitehead:
@lingmann I've added the comments at the top of the file.

@branden I will squash after the review is complete.

test_util/test_upgrade_vpc.py
lingmann: That seems reasonable, thanks for the clarification.

PR #1113, opened 2017-01-09
test for l4lb

aeyakovenko:

adds a minuteman test to the test suite that verifies that l4lb is actually doing load balancing

automates manual testing of minuteman

 - High-Level how things changed

new minuteman test added

# Issues

[DCOS-12382](https://mesosphere.atlassian.net/browse/DCOS-12382)
...

# Checklist

 - [ ] Test Results: <link>
 - [ ] Code Coverage: <link>

### DIFF
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -1,4 +1,7 @@
 import logging
+
+import threading
+
 from subprocess import check_output
 
 import pytest
@@ -8,6 +11,7 @@ import requests
 import retrying
 
 from pkgpanda.build import load_json
+
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
         with dcos_api_session.marathon.deploy_and_cleanup(proxy_app) as service_points:
             cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://foo.marathon.l4lb.thisdcos.directory:6000/ping'
             ensure_routable(cmd, service_points)()
+
+
+def geturl(tries, url):
+    rv = None
+    while tries > 0:
+        try:
+            rs = requests.get(url)
+            if rs.status_code == 200:
+                tries = 0
+                rv = rs.json()
+        except Exception as e:
+            logging.info("test_l4lb failed to connect to geturl {}: {}".format(url, e))
+        tries = tries - 1
+    return rv
+
+
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
+def test_l4lb(cluster):
+    """Test l4lb is load balancing between all the backends
+    """
+    numapps = 5
+    numthreads = numapps * 4
+    tries = 100
+    apps = []
+    rvs = []
+    try:
+        for _ in range(0, numapps):
+            origin_app, origin_uuid = get_test_app()
+            origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '/l4lbtest:5000'}
+            apps.append(origin_app)
+            sp = cluster.marathon.deploy_app(origin_app)
+            geturl(tries, 'http://{}:{}/ping'.format(sp[0].host, sp[0].port))
+            geturl(tries, 'http://l4lbtest.marathon.l4lb.thisdcos.directory:5000/ping')
+
+        def thread_request():
+            rvs.append(geturl(tries, 'http://l4lbtest.marathon.l4lb.thisdcos.directory:5000/test_uuid'))
+
+        threads = [threading.Thread(target=thread_request) for i in range(0, numthreads)]
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+    except Exception as e:
+        logging.info("test_l4lb exception: {}".format(e))
+    for a in apps:
+        cluster.marathon.destroy_app(a['id'])
+    tids = [a['id'].split('-')[2] for a in apps]
+    rids = [r['test_uuid'] for r in rvs]
+    assert(set(tids) == set(rids))


tox.ini
```diff
@@ -52,7 +52,7 @@ deps =
   dnspython
   pytest
   webtest
-  webtest-aiohttp
+  webtest-aiohttp==1.1.0
```
aeyakovenko: https://github.com/dcos/dcos/pull/1114

adammesos:
Marking as WIP for now. Please rebase, fix the conflict, fill out the description (from the new PR template if you like), and assign a couple of reviewers.

aeyakovenko:
who would be a good reviewer for this?

adammesos:
Considering that it's a "networking test", I'd suggest:
1. Somebody from the networking team, to verify that you're testing the correct functionality
2. Somebody familiar with our test infrastructure ( @mellenburg ), or at least somebody who knows python
It also looks like your new test is failing, so you'll need to fix that

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+    numthreads = numapps * 4
+    tries = 100
+    apps = []
+    rvs = []
+    try:
+        for _ in range(0, numapps):
+            origin_app, origin_uuid = get_test_app()
+            origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '/l4lbtest:5000'}
+            apps.append(origin_app)
+            sp = cluster.marathon.deploy_app(origin_app)
```
mellenburg: In general, marathon.deploy_app is not to be used because if your test fails you will be leaving state around in the cluster for other tests to choke on. I see that you added ant exception handler to do some cleanup with, but its still a faulty method. Rather, use marathon.deploy_and_cleanup . Given that you want to launch and cleanup many apps, you should combine that with an ExitStack: https://docs.python.org/3.5/library/contextlib.html#contextlib.ExitStack

E.G.:
```
with ExitStack() as stack:
    for i in range(num_apps):
        sp = stack.enter_context(cluster.deploy_and_cleanup(get_test_app()[0])
```
which is going to be far more robust in cleanup as it will attempt to cleanup other apps even if some of the preceding app cleanups fail

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+def test_l4lb(cluster):
+    """Test l4lb is load balancing between all the backends
+    """
+    numapps = 5
+    numthreads = numapps * 4
+    tries = 100
+    apps = []
+    rvs = []
+    try:
+        for _ in range(0, numapps):
```
mellenburg: This should be range(numapps). The zero serves no purpose

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+            t.start()
+        for t in threads:
+            t.join()
+    except Exception as e:
+        logging.info("test_l4lb exception: {}".format(e))
+    for a in apps:
+        cluster.marathon.destroy_app(a['id'])
+    tids = [a['id'].split('-')[2] for a in apps]
+    rids = [r['test_uuid'] for r in rvs]
+    assert(set(tids) == set(rids))
```
mellenburg: Tests all about readability; would you mind expanding out these variable names (tids and rids) to something meaningful and/or maybe add a comment on what is actually being checked?

Also, you are comparing these two sets. Does distribution of load balancing between the test apps not matter? E.G. this test will pass if we get 1 response from 4 apps and then 16 responses from 1 app. Additionally, the list comprehensions above will fail if r is ever None (which happens if all tries fails), however it will produce an ugly error message. Rather, can we add a clean assertion for when the tread cannot talk to the app?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+    apps = []
+    rvs = []
+    try:
+        for _ in range(0, numapps):
+            origin_app, origin_uuid = get_test_app()
+            origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '/l4lbtest:5000'}
+            apps.append(origin_app)
+            sp = cluster.marathon.deploy_app(origin_app)
+            geturl(tries, 'http://{}:{}/ping'.format(sp[0].host, sp[0].port))
+            geturl(tries, 'http://l4lbtest.marathon.l4lb.thisdcos.directory:5000/ping')
```
mellenburg: Neither of these geturl's have any sort of assert nor is their return used. geturl has catch-them-all exception handling, so these lines have no meaning. Its impossible to know if one or both of these failed

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+
+        def thread_request():
+            rvs.append(geturl(tries, 'http://l4lbtest.marathon.l4lb.thisdcos.directory:5000/test_uuid'))
```
mellenburg: Is this safe? It seems odd to provide a regular old python list and then spawn a bunch of threads to start appending into it whenever they please. Usually what I see done for getting the response from many threads is using a pool, or having threads dump their results into specific indices in the list, or using a threading Queue

I don't know what the case is, but if python doesn't implement some kind of thread locking on its own, this could fall over

aeyakovenko: python lists are thread safe

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -114,3 +118,52 @@ def test_ip_per_container_with_named_vip(dcos_api_session):
...
+
+        threads = [threading.Thread(target=thread_request) for i in range(0, numthreads)]
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+    except Exception as e:
```
aeyakovenko: we only care if we got the expected uuid's from the load balanced apps.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: it just verifies that the server is reachable at the time.  So VIPs are up and something is listening to them.  We don't really care what it returns as long as the network is up.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: yea, distribution doesn't matter.
+1 on variable names!
why would `r` ever be None?

aeyakovenko:
@adammesos can you add @nlsun  as a reviewer?

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: +1 on the ExitStack
that's why i catch all exceptions and then clean up the list of apps.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: it just waits until the VIPs are up.  If both fail, then getting the uuid is going to fail as well.

mellenburg: Neither of these will fail, they will just return None and that return does nothing here, so you wont find out about that failure til a bit later

packages/dcos-integration-test/extra/test_networking.py
mellenburg: Look at geturl again; if you raise an exception on each try, you will run out of tries and just return None. Given this and the issue above with you expecting geturl to fail if you cannot talk to the app, you should either assert there that rv cannot be None at return time, but that will get weird if you put it in a thread, so I guess you would need to check at the top level that no None's were returned

PR #1120, opened 2017-01-10
Fix upgrade integration test

orsenthil:

* Drain Deployments before upgrade.

This is discussed in this ticket: https://mesosphere.atlassian.net/browse/DCOS-12746

### DIFF
--- a/test_util/marathon.py
+++ b/test_util/marathon.py
@@ -160,7 +160,7 @@ class Marathon(ApiClient):
         @retrying.retry(wait_fixed=1000, stop_max_delay=timeout * 1000,
                         retry_on_result=lambda ret: ret is None,
                         retry_on_exception=lambda x: False)
-        def _pool_for_marathon_app(app_id):
+        def _poll_marathon_for_app_deployment(app_id):
             Endpoint = collections.namedtuple("Endpoint", ["host", "port", "ip"])
             # Some of the counters need to be explicitly enabled now and/or in
             # future versions of Marathon:
@@ -200,10 +200,38 @@ class Marathon(ApiClient):
                 return None
 
         try:
-            return _pool_for_marathon_app(app_definition['id'])
+            return _poll_marathon_for_app_deployment(app_definition['id'])
         except retrying.RetryError:
             raise Exception("Application deployment failed - operation was not "
                             "completed in {} seconds.".format(timeout))
+
+    def ensure_deployments_complete(self, timeout=120):
+        """
+        This method ensures that, there are no pending deployments
+
+        :return: True if all deployments are completed within time out. Raises an exception otherwise.
+        """
+
+        @retrying.retry(wait_fixed=1000, stop_max_delay=120 * 1000, retry_on_exception=lambda x: False)
+        def _get_deployments_json():
+            r = self.get('v2/deployments')
+            r.raise_for_status()
+            return r.json()
+
+        def retry_on_assertion_error(exception):
+            return isinstance(exception, AssertionError)
+
+        @retrying.retry(retry_on_exception=retry_on_assertion_error,
+                        stop_max_attempt_number=10,
+                        wait_fixed=timeout * 1000)
+        def ensure_deployment_is_finished():
+            deployments_json = _get_deployments_json()
+            assert not deployments_json, "No deployment should be happening."
+
+        try:
+            ensure_deployment_is_finished()
+        except retrying.RetryError:
+            raise Exception("Deployments were not completed within {timeout} seconds".format(timeout=timeout))
 
     def deploy_pod(self, pod_definition):
         """Deploy a pod to marathon
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -54,6 +54,7 @@ from test_util.cluster_api import ClusterApi
 from test_util.helpers import CI_AUTH_JSON, DcosUser
 from test_util.marathon import TEST_APP_NAME_FMT
 
+
 logging.basicConfig(format='[%(asctime)s|%(name)s|%(levelname)s]: %(message)s', level=logging.DEBUG)
 log = logging.getLogger(__name__)
 
@@ -93,10 +94,12 @@ def get_test_app():
 def get_task_info(apps, tasks):
     idx = 0    # We have a single app and a task for this test.
 
+    logging.info("v2/apps json: {apps}".format(apps=apps))
+    logging.info("v2/tasks json: {tasks}".format(tasks=tasks))
+
     try:
         tasks_state = tasks["tasks"][idx]["state"]
-        health_check_interval = \
-            apps["apps"][idx]["healthChecks"][idx]["intervalSeconds"]
+        health_check_interval = apps["apps"][idx]["healthChecks"][idx]["intervalSeconds"]
         task_id = tasks["tasks"][idx]["id"]
         last_success = tasks["tasks"][idx]["healthCheckResults"][idx]["lastSuccess"]
     except IndexError as e:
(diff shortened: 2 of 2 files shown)


branden:
Blocking until the deploy is complete would be a handy feature for `Marathon.deploy_app()`. Arguably that's what we should have instead of its `check_health` flag.

orsenthil:
@branden, I think, there is a subtle difference between deploy_app  + health check verification vs. verifying deployment completion via `v2/deployments`.  While the former is applicable to an individual app whereas later works on all apps together.

If we do `/v2/deployments` for every deploy_app, it means we are serializing deployments, and I don't think that is our intention or reflection of the customer scenario.

Instead, I think, we should do a `v2/deployments` verification after all the apps are installed (deploy_apps) are called.

I have one question, when we are installing 10 instances of an app, when one of them is ready, can /heath-check return 200? If that is the case, than above argument holds even stronger.

branden:
That makes sense. I still think this feature would be useful as a part of `test_util.marathon.Marathon`, especially if it reflects common user behavior. Maybe a new method?

orsenthil:
@branden , I have addressed your comments. Please take a look again.

This can see additional review from @BenWhitehead 

Thanks!

BenWhitehead:
The deployment will take the health check, and total instance count into consideration when determining if the deployment has completed. IIRC the deployment will exist as long as the number of runnings healthy instances isn't what is defined in the app json.

It may be worth adding an additional method to Marathon along the lines of `deploy_add_and_wait_for_deployment` if we want to add this blocking behavior for an individual app creation.

orsenthil:
@BenWhitehead , acknowledge. Since parts to add that `deploy_add_and_wait_for_deployment` where already available and we were only checking with a single app. I have added that method  and used it in this PR.

This will address @branden 's initial suggestion in a direct way. 

Please take a look again. If there are upgrade flakiness beyond these changes, I will raise that against marathon team.

test_util/marathon.py
```diff
@@ -200,11 +200,46 @@ def _pool_for_marathon_app(app_id):
...
+            deployments_json = _get_deployments_json()
+            logging.info("v2/deployments json: {deployments_json}".format(deployments_json=deployments_json))
+            assert not deployments_json, "No deployment should be happening."
+
+        try:
+            ensure_deployment_is_finished()
+        except retrying.RetryError:
+            raise Exception("Deployments were not completed within {timeout} seconds".format(timeout=timeout))
+
+    def deploy_add_and_wait_for_deployment(self, app_definition, timeout=120):
```
BenWhitehead: typo: `s/deploy_add/deploy_app/`

orsenthil: That was a copy pasta from your previous comment! :)

BenWhitehead: haha, sorry about that.

test_util/marathon.py
```diff
@@ -200,11 +200,46 @@ def _pool_for_marathon_app(app_id):
...
 
+    def ensure_deployments_complete(self, timeout=600):
+        """
+        This method ensures that, there are no pending deployments
+
+        :return: True if all deployments are completed within time out. Raises an exception otherwise.
+        """
+
+        @retrying.retry(wait_fixed=1000, stop_max_delay=120 * 1000, retry_on_exception=lambda x: False)
+        def _get_deployments_json():
```
branden: Why the retry here? We should be able to get deployments reliably, for the same reason we POST apps without retrying.

test_util/marathon.py
branden: * Why the different default timeout here? 120 vs 600 in `ensure_deployments_complete()`.
* I don't think deploy-and-block makes sense as a separate method. That we can have different deploy timeouts to get mismatched is a smell. What if we want to set the `deploy_app()` timeout as well? What if we want to block on the deploy, _and_ allow for tasks to fail during it? In order to get all our options we'd have to tediously duplicate kwargs and pass them through to one of the two methods we're calling here, and make sure kwarg names for the two methods don't overlap and introduce an ambiguity. If we need to provide deploy-and-block behavior, I think it'd make more sense as an optional `deploy_timeout` kwarg for `deploy_app()` that defaults to `None`. If it's `None` we don't block, otherwise we time out at `deploy_timeout` seconds.

test_util/marathon.py
```diff
@@ -200,11 +200,46 @@ def _pool_for_marathon_app(app_id):
...
+            deployments_json = _get_deployments_json()
+            logging.info("v2/deployments json: {deployments_json}".format(deployments_json=deployments_json))
+            assert not deployments_json, "No deployment should be happening."
+
+        try:
+            ensure_deployment_is_finished()
+        except retrying.RetryError:
+            raise Exception("Deployments were not completed within {timeout} seconds".format(timeout=timeout))
+
+    def deploy_app_and_wait_for_deployment(self, app_definition, timeout=120):
```
branden: * Why the different default timeout here? 120 vs 600 in `ensure_deployments_complete()`.
* I don't think deploy-and-block makes sense as a separate method. That we can have different deploy timeouts to get mismatched is a smell. What if we want to set the `deploy_app()` timeout as well? What if we want to block on the deploy, _and_ allow for tasks to fail during it? In order to get all our options we'd have to tediously duplicate kwargs and pass them through to one of the two methods we're calling here, and make sure kwarg names for the two methods don't overlap and introduce an ambiguity. If we need to provide deploy-and-block behavior, I think it'd make more sense as an optional `deploy_timeout` kwarg for `deploy_app()` that defaults to `None`. If it's `None` we don't block, otherwise we time out at `deploy_timeout` seconds.

orsenthil: 1. On `ensure_deployments_complete`. The reason for 10-minute timeout vs 2-minute timeout was that individual apps have a 2-minute default timeout, whereas this method should be called after a number of apps have been deployed together and thus will need longer time to ensure that all apps have been deployed. 

2. Since `deploy_app_and_wait_for_deployment` is for a single app,  it is using the same timeout as `deploy-app` (120 secs).  I ensured that part.

> What if we want to block on the deploy, and allow for tasks to fail during it?

I imagine that will be a deployment failure. Isn't it? 

> If we need to provide deploy-and-block behavior, I think it'd make more sense as an optional deploy_timeout

Hmm. This seems to be introducing a special case of `deployment` verification under timeout flag.

I agree with the different questions on kwargs that you raised, I faced the same problem and tried to make applicable choices with respect to deployment while reusing the existing API.

If the usage of this new API method is not intuitive enough, I think, I would go back to the previous state of this PR.

That is

* Keep ensure_deployments_complete as a separate method (which already is), but with the timeout as 120. Let the client customize it.
* Let the client call `ensure_deployments_complete` after calling the `deploy_app` method. It's easy to know the intention while reading the code too.

Is that alright with you and @BenWhitehead  ?

branden: That works for me. 👍

orsenthil:
@branden , refactored this (and removed the deploy-and-block method) as we discussed. Please take a look again. thanks!

branden:
Looks good! I'll approve once tests pass.

test_util/marathon.py
```diff
@@ -200,11 +200,40 @@ def _pool_for_marathon_app(app_id):
...
+
+        def retry_on_assertion_error(exception):
+            return isinstance(exception, AssertionError)
+
+        @retrying.retry(retry_on_exception=retry_on_assertion_error,
+                        stop_max_attempt_number=10,
+                        wait_fixed=timeout * 1000)
+        def ensure_deployment_is_finished():
+            deployments_json = _get_deployments_json()
+            logging.info("v2/deployments json: {deployments_json}".format(deployments_json=deployments_json))
```
branden: Just noticed this -- can we remove this log line? I'm biased against adding verbose logging to utility functions/methods unless they have obvious value, since they add so much clutter to test logs.

orsenthil: Ok. Sounds good to me.

adam-mesos:
Is this actually ready to merge, or do we need to fix the (new?) upgrade-onprem integration test?

orsenthil:
Yes, this is ready to merge. The failure of the upgrade-onprem is the problem that marathon team would like to fix. We demonstrate those flaky failures with a valid test in this code

PR #1121, opened 2017-01-10
Set LIBPROCESS_PORT for marathon and metronome

zen-dog:
Set LIBPROCESS_PORT for marathon (15101) and metronome (15102) in the dcos-config.yaml explicitly. Update the admin-port documentation.

# Issues
[DCOS-12706](https://mesosphere.atlassian.net/browse/DCOS-12706)

### DIFF
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -38,6 +38,8 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 8181: dcos-exhibitor
  - 9990: dcos-cosmos
  - 15055: dcos-history-service
+ - 15101: dcos-marathon libprocess
+ - 15201: dcos-metronome libprocess
 
 ### UDP
 
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -587,6 +587,7 @@ package:
   - path: /etc_master/marathon
     content: |
       MARATHON_ZK=zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181/marathon
+      LIBPROCESS_PORT=15101
 {% switch dcos_overlay_enable %}
 {% case "true" %}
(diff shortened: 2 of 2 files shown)


docs/admin-ports.md
```diff
@@ -38,6 +38,8 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 8181: dcos-exhibitor
  - 9990: dcos-cosmos
  - 15055: dcos-history-service
+ - 15101: dcos-marathon libprocess
+ - 15201: dcos-metronome libprocess
```
lingmann: What is the reasoning behind the port gaps here and not going with 15056 and 15057, or 9991 and 9992, etc? Before approving I'd like to get @spahl input on whether or not these ports line up with our longer term plans of port allocation. This also needs a PR against https://github.com/dcos/dcos-docs

zen-dog: No particular reason for that. Keeping the gaps in case there are more services that feet that range better. I can change them to numbers you feel are better suitable.

adammesos: Why not follow `8080: dcos-marathon` with 8081 for marathon libprocess?
Does Metronome have a TCP port for its own API, not listed here?

zen-dog: Marathon uses 8080 (http) and 8443 (https) by default. For metronome these are 9090 and 9443. I thought giving libmesos it's "own range" so it's distinct from others.

lingmann: @spahl and I chatted and we are ok leaving these ports as is, there is not much value at this point in trying to use a contiguous range since master ports are already all over the place. I did verify that 15101 and 15201 are unassigned in the IANA services and port registry.

lingmann:
code-quality failure should go away when rebased on master

zen-dog:
Rebased

zen-dog:
Created dcos/docs [PR](https://github.com/dcos/dcos-docs/pull/887)

PR #1123, opened 2017-01-10
Don't expose --uninstall flag when the operation is not complete.

orsenthil:
# Issues

* [DCOS-12968](https://mesosphere.atlassian.net/browse/DCOS-12968)

# Checklist

 - [x] Reverted but test that will fail without this feature.

# Local Testing

```
$ wget https://s3-us-west-2.amazonaws.com/downloads.dcos.io/dcos/testing/pull/1123/commit/48cb7792006fc47aaf43ab66f8245be659fa4036/dcos_generate_config.sh
$ chmod +x dcos_generate_config.sh
$ ./dcos_generate_config.sh --help
DC/OS Install and Configuration Utility

optional arguments:
  -h, --help            show this help message and exit
  --hash-password [password]
                        Hash a password and print the results to copy into a
                        config.yaml.
  --set-superuser-password [password]
                        Hash the given password and store it as the superuser
                        password in config.yaml
  -v, --verbose         Verbose log output (DEBUG).
  --offline             Do not install preflight prerequisites on CentOS7,
                        RHEL7 in web mode
  --cli-telemetry-disabled
                        Disable the CLI telemetry gathering for SegmentIO
  --validate-config     Validate the configuration for executing --genconf and
                        deploy arguments in config.yaml
  --version             Print the DC/OS version
  --genconf             Execute the configuration generation (genconf).
  --web                 Run the web interface
  --aws-cloudformation  Generate AWS Advanced AWS CloudFormation templates
                        using the provided config
  --install-prereqs     Execute the preflight checks on a series of nodes.
  --postflight          Execute postflight checks on a series of nodes.
  --preflight           Execute the preflight checks on a series of nodes.
  --deploy              Execute a deploy.
```

PR #1118, opened 2017-01-10
Add zkEnv.sh to /bin

branden:
`zkCli.sh` expects `zkEnv.sh` to be in the same directory.

```
$ dcos-shell /opt/mesosphere/packages/exhibitor--9b89c1f1654d9f2addfb8dfaa138053d6c6a5eb7/usr/zookeeper/bin/zkCli.sh ls >/dev/null
$ echo $?
0
$ dcos-shell /opt/mesosphere/bin/zkCli.sh ls >/dev/null
/opt/mesosphere/bin/zkCli.sh: line 36: /opt/mesosphere/bin/zkEnv.sh: No
such file or directory
/opt/mesosphere/bin/zkCli.sh: line 39: : command not found
$ echo $?
127
$
```

# Issues
https://mesosphere.atlassian.net/browse/DCOS-9775

# Checklist

 - [ ] Included a test which will fail if code is reverted but test is not
 - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #1114, opened 2017-01-09
Pinning the version for webtest-aiohttp

orsenthil:
Need to pin the webtest-aiohttp as it is using a different features from a higher version aiohttp than us.

The reasons are explained here: [DCOS-13081](https://mesosphere.atlassian.net/browse/DCOS-13081)

*Tested Change locally*

```
# tox -e py35-unittests
GLOB sdist-make: /dcos/setup.py
py35-unittests create: /dcos/.tox/py35-unittests
py35-unittests installdeps: dnspython, pytest, webtest, webtest-aiohttp==1.1.0
py35-unittests inst: /dcos/.tox/dist/dcos_image-0.1.zip
py35-unittests installed: aiohttp==0.22.5,analytics-python==1.2.6,azure-common==1.1.4,azure-mgmt-network==0.30.0rc4,azure-mgmt-nspkg==1.0.0,azure-mgmt-resource==0.30.0rc4,azure-nspkg==1.0.0,azure-storage==0.32.0,beautifulsoup4==4.5.3,boto3==1.4.3,botocore==1.4.93,certifi==2016.9.26,chardet==2.3.0,click==6.7,coloredlogs==5.2,dcos-image==0.1,dnspython==1.15.0,docopt==0.6.2,docutils==0.13.1,enum34==1.1.6,Flask==0.12,Flask-Compress==1.4.0,humanfriendly==2.2,isodate==0.5.4,itsdangerous==0.24,Jinja2==2.9.3,jmespath==0.9.0,keyring==9.1,MarkupSafe==0.23,msrest==0.4.0,msrestazure==0.4.1,multidict==1.2.2,oauthlib==2.0.1,passlib==1.7.0,py==1.4.32,PyInstaller==3.2,pytest==3.0.5,python-dateutil==2.6.0,PyYAML==3.12,requests==2.10.0,requests-oauthlib==0.7.0,retrying==1.3.3,s3transfer==0.1.10,six==1.10.0,teamcity-messages==1.21,waitress==1.0.1,WebOb==1.7.0,WebTest==2.0.24,webtest-aiohttp==1.1.0,Werkzeug==0.11.15
py35-unittests runtests: PYTHONHASHSEED='982528175'
py35-unittests runtests: commands[0] | py.test --basetemp=/dcos/.tox/py35-unittests/tmp
========================================================================================================= test session starts =========================================================================================================
[...]

PR #1124, opened 2017-01-11
Train 117

lingmann:
Train merge:
#1106 Fix azure flaky tests
#1114 Pinning the version for webtest-aiohttp
#1123 Don't expose the --uninstall flag since it is not supported or documented and not functional
#1112 Add parameters to test_upgrade_vpc.py to all overrides
#1118 Add zkEnv.sh to /bin

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -322,6 +322,8 @@ exit $RETCODE"""
     return result
 
 
+# TODO: DCOS-250 (skumaran@mesosphere.com)- Create an comprehensive DC/OS uninstall strategy.
+# This routine is currently unused and unexposed.
 @asyncio.coroutine
 def uninstall_dcos(config, block=False, state_json_dir=None, async_delegate=None, options=None):
     targets = get_full_nodes_list(config)
--- a/dcos_installer/cli.py
+++ b/dcos_installer/cli.py
@@ -83,21 +83,6 @@ def run_loop(action, options):
     return exitcode
 
 
-def tall_enough_to_ride():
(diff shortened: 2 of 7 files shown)


lingmann:
@adam-mesos I'm not sure if that is really a sustainable approach here to always have an enterprise bump for each train without automatic tooling... last time I tried a bump of enterprise we had other issues that blocked merging basic fixes in open (such as all code-quality builds are currently failing.) @spahl thoughts?

adam-mesos:
> last time I tried a bump of enterprise we had other issues
That's exactly the problem I'm trying to avoid. I can't (easily) test any of my OSS changes (e.g. Mesos bump) in EE with the current state of things, and piling more commits on top of dcos/dcos will only make the eventual upstream bump even nastier.
If every OSS train checked the upstream bump health, we could avoid all these broken windows.
And I'm 100% in favor of automated tooling for upstream bumps. I'd prioritize that ASAP, personally.
As it stands now, it seems like upstream bump is healthy for ~1 week, then broken for ~1 month, then fixed for 1 week, etc. This drastically slows down development of our core $$ product.

lingmann:
Summarizing RL discussion had earlier today with @spahl @adam-mesos: we are going to always bump downstream with open trains, even though it is going to slow us down in short term. Train conductors will own making sure this happens. We also agreed to merge this train without the downstream bump, since we are going to prioritize getting downstream building again today.

lingmann:
Since I created this train I will make sure all of the issues with downstream are triaged / prioritized / merged today as well.

PR #1125, opened 2017-01-11
Updated the Mesos package readme.

greggomann:
This patch updates the Mesos package readme to include the latest versions of the cherry-pick commits that must be applied to open source Mesos before building in DC/OS.

This is a non-functional change.


lingmann:
@adam-mesos can you review? non-functional change, should be a quick one.

PR #1130, opened 2017-01-13
Explicitly set JAVA_HOME for exhibitor gradle build

BenWhitehead:
Fixes a bug where exhibitor was building with the jdk present in the base docker container, rather than the defined java pkgpanda package.

### DIFF
--- a/packages/exhibitor/build
+++ b/packages/exhibitor/build
@@ -12,6 +12,9 @@ ln -s "$PKG_PATH/usr/zookeeper/bin/zkCli.sh" "$PKG_PATH/bin/zkCli.sh"
 # zkCli.sh expects zkEnv.sh to be in the same directory.
 ln -s "$PKG_PATH/usr/zookeeper/bin/zkEnv.sh" "$PKG_PATH/bin/zkEnv.sh"
 
+# Explicitly set JAVA_HOME to our java package
+export JAVA_HOME="$(find /opt/mesosphere/packages -type d -name java--*)/usr/java"
+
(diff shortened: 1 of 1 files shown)


packages/exhibitor/build
```diff
@@ -12,6 +12,9 @@ ln -s "$PKG_PATH/usr/zookeeper/bin/zkCli.sh" "$PKG_PATH/bin/zkCli.sh"
 # zkCli.sh expects zkEnv.sh to be in the same directory.
 ln -s "$PKG_PATH/usr/zookeeper/bin/zkEnv.sh" "$PKG_PATH/bin/zkEnv.sh"
 
+# Explicitly set JAVA_HOME to our java package
+export JAVA_HOME="$(find /opt/mesosphere/packages -type d -name java--*)/usr/java"
```
adam-mesos: Why not use `/opt/mesosphere/active/java/usr/java`? Or is that symlink not setup yet at this time in the build?

adam-mesos:
Who else would be good reviewers for this? Let's try to get 2 ShipIts ASAP for this. Thanks!

packages/exhibitor/build
BenWhitehead: IIRC packages aren't activated in the build so the symlink doesn't exist in the build container.

BenWhitehead:
@lingmann who else do you want to review this? Or do you want to do it yourself?

orsenthil:
I'll review this, since I know the background.  This is a 🚢 !

PR #1134, opened 2017-01-17
cosmos: restrict admin and http ports to localhost

gpaul:
The Cosmos service currently runs with default options for its admin port (`0.0.0.0:9990`) and its http port (`0.0.0.0:7070`.) We should restrict Cosmos to listen on the loopback interface only.

This change prevents bypassing Admin Router via direct access to Cosmos.

# Issues

[DCOS-613](https://dcosjira.atlassian.net/browse/DCOS-613)

# Checklist

### DIFF
--- a/packages/dcos-integration-test/extra/test_endpoints.py
+++ b/packages/dcos-integration-test/extra/test_endpoints.py
@@ -2,6 +2,7 @@ import ipaddress
 import urllib.parse
 
 import bs4
+from requests.exceptions import ConnectionError
 from retrying import retry
 
 from pkgpanda.util import load_json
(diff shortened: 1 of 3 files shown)


gpaul:
I could not find any prior art for a test of this type but would gladly add one - suggestions welcome.

packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -311,3 +312,30 @@ def test_if_overlay_master_agent_is_up(dcos_api_session):
...
+    except ConnectionError:
+        pass
+    else:
+        raise AssertionError("Cosmos reachable from non-loopback interface")
+
+    # One should be able to connect to the cosmos HTTP and admin ports at
+    # 127.0.0.1:7070 and 127.0.0.1:9990.
+    # Getting HTTP error codes shows that we made it all the way to
+    # cosmos which is exactly what we're testing.
+    r = cluster.get('/', host_url="http://127.0.0.1:7070")
```
alberts: @gpaul this will pass against the Enteprise build too, right?

gpaul:
code-quality build currently broken due to pycodestyle bug: https://github.com/PyCQA/pycodestyle/issues/617

packages/dcos-integration-test/extra/test_endpoints.py
gpaul: Yes - once it's twin lands in enterprise.

amitaekbote:
cosmos test seems to be failing in enterprise - https://github.com/mesosphere/dcos-enterprise/pull/548 

this will not be merged with today's train

amitaekbote:
@jsancio  @gpaul can you look into it?

adam-mesos:
Closing in favor of #1198 which explicitly specifies scheme=http

PR #1138, opened 2017-01-18
Assorted tooling fixes

branden:
## High Level Description

This PR contains fixes for small issues I discovered while working on #1066.

## Related Issues

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Covered by existing tests.

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -356,7 +356,7 @@ class Repository:
         filename = os.path.join(path, "pkginfo.json")
         try:
             pkginfo = load_json(filename)
-        except FileNotFoundError as ex:
+        except OSError as ex:
             raise PackageError("No / unreadable pkginfo.json in {0}: {1}".format(id, ex.strerror)) from ex
 
         if not isinstance(pkginfo, dict):
--- a/dcos_installer/config.py
+++ b/dcos_installer/config.py
@@ -80,6 +80,12 @@ class Config():
                 "available configuration options. You can also use the GUI web installer (--web),"
                 "which provides a guided configuration and installation for simple "
                 "deployments.".format(self.config_path)) from ex
+        except OSError as ex:
+            raise NoConfigError(
+                "Failed to open config file at {}: {}. See the DC/OS documentation to learn "
+                "how to create a config file. You can also use the GUI web installer (--web), "
+                "which provides a guided configuration and installation for simple "
+                "deployments.".format(self.config_path, ex)) from ex
         except YamlParseError as ex:
             raise NoConfigError("Unable to load configuration file. {}".format(ex)) from ex
 
--- a/dcos_installer/config.py
+++ b/dcos_installer/config.py
@@ -77,7 +77,7 @@ class Config():
         except FileNotFoundError as ex:
             raise NoConfigError(
                 "No config file found at {}. See the DC/OS documentation for the "
-                "available configuration options. You can also use the GUI web installer (--web),"
+                "available configuration options. You can also use the GUI web installer (--web), "
                 "which provides a guided configuration and installation for simple "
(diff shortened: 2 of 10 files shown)


pkgpanda/__init__.py
```diff
@@ -356,7 +356,7 @@ def load(self, id: str):
         filename = os.path.join(path, "pkginfo.json")
         try:
             pkginfo = load_json(filename)
-        except FileNotFoundError as ex:
+        except OSError as ex:
```
amitaekbote: What error did we miss to use OS instead of FileNotFound?

branden: NotADirectoryError

orsenthil: The `FileNotFoundError` captures just the file not found, while `OSError` captures the file with bad permissions to read.  I don't know if if the latter with occur in practice.  The status quo might be better here IMO. Also, I see all placed were we use load_json in code base looking for `FileNotFoundError` only.

orsenthil: @branden , the `NotADirectoryError` shouldn't  have been raised here.

At line 353, we have the explicit check for  the directory presence: `if not os.path.exists(path):`

branden: The path can exist but not be a directory. https://docs.python.org/3/library/exceptions.html#NotADirectoryError
> Raised when a directory operation (such as `os.listdir()`) is requested on something which is not a directory. Corresponds to errno `ENOTDIR`.

orsenthil: @branden , sure, if that happened to be the case here, then, I think, we should catch it earlier.

1) Catching specific exception is `FileNotFoundError` is better than `OSError` because we are looking for specific error.

2) If a `NotADirectoryError` really got raised here, it could be caught in the `self.package_path` or  you could change the statement  `if not os.path.exists(path)` to `if not os.path.isdir(path):`

branden: 1. We aren't looking for a specific error here. We're trying to raise an exception with a meaningful message to aid debugging in the event that a `pkginfo.json` can't be read, for whatever reason. This is why we include the exception's error message in the `PackageError` that we raise -- we aren't certain what the cause is until runtime.
1. This would make the `PackageNotFound` exception that gets raised ambiguous. If we catch the error here, we get a more detailed error with a clearer cause.

orsenthil: > We aren't looking for a specific error here

I meant, in the generic sense. For e.g, one could argue that using `except Exception`  is perfectly alright here because we are going to raise `PackageError` immediately afterward.   But, in general, it is agreed upon by many that `except Exception` is a bad idea and looking for focussed exception is better.  Some folks think that the type of Exception being caught helps while reading or understanding the code.  

> we aren't certain what the cause is until runtime.

Please note that there are multiple places of this load_json and exception in the scenario. So, that needs to be kept in mind, (Like line 779 and line 519 `build/__init__.py`  (This is an important point to consider because we introduce an inconsistency here).

On 2. I meant "earlier" that this stanza, like ensuring that path + id is definitely a path.  I agree that printing the error message gives better error info on the exact reason for the failure.

If the inconsistency point is address, I am fine with this change.

lingmann: @branden were you able to address @orsenthil's feedback above / is this ready for another review?

branden: Haven't addressed this feedback yet.

branden: @orsenthil I found other a few other spots where we were catching `FileNotFoundError` with the expectation that it covers all cases where a file doesn't exist on the filesystem and expanded them to cover `OSError`. 88447c58ec8be4e4cc7f1503f28490ad3756de90

PR #1139, opened 2017-01-18
Document missing exhibitor ports

karlkfi:
Not sure if these are missing ports or weren't documented because they don't need to be exposed or aren't actually used...

This list, dcos.io, and docs.mesosphere.com all say different things. So it's hard to tell which is correct.

2888 & 3888 are used here:
https://github.com/dcos/dcos/blob/master/packages/exhibitor/extra/start_exhibitor.py

Feel free to correct me...

### DIFF
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -31,6 +31,8 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 1050: dcos-3dt
  - 1801: dcos-oauth
  - 2181: dcos-exhibitor
+ - 2888: dcos-exhibitor
+ - 3888: dcos-exhibitor
  - 5050: dcos-mesos-master
  - 7070: dcos-cosmos
  - 8080: dcos-marathon


adam-mesos:
@alberts ?

karlkfi:
Sounds like `2181` is zookeeper's listening port, `8181` is exhibitor's listening port, and `2888` and `3888` are used by the default configuration for bootstrap (i think).

karlkfi:
Can't find either port at runtime:

```
$ netstat -tulpn | grep LISTEN | grep tcp | tr -s ' ' | cut -d ' ' -f4 | rev | cut -d ':' -f1 | rev | sort -n | uniq
(No info could be read for "-p": geteuid()=501 but you should be root.)
22
25
53
80
443
1050
2181
5050
7070
8080
8101
8123
8181
9000
9942
9990
15055
15101
15201
43478
61053
61420
62053
62080
62501
62502
```

PR #1142, opened 2017-01-18
Better Error Reporting for Upgrade Integration Test.

orsenthil:
* Better Error reporting for upgrade failure.

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -178,10 +178,10 @@ def main():
 
     cluster_api.marathon.ensure_deployments_complete()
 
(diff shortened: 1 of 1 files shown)


branden:
I already have a refactor in progress for verifying upgrade test workloads in #1105. Would it be quicker if you left feedback on that PR for better error reporting?

orsenthil:
@branden . thanks. I have commented it there. If you think, it is a good idea, please include something similar in your PR itself. I will close this.

My motivation behind this PR was - I was feeling uncomfortable seeing "bad error messages" in the upgrade tests, when I thought something like this

> AssertionError: v2/apps was determined empty after upgrade!

Would be better.

PR #1146, opened 2017-01-18
Add dcos-shell to invocation of pkgpanda uninstall during upgrade test

BenWhitehead:
### Issues
* [DCOS-9717](https://mesosphere.atlassian.net/browse/DCOS-9717)

### DIFF
--- a/test_util/cluster.py
+++ b/test_util/cluster.py
@@ -320,7 +320,7 @@ def upgrade_dcos(cluster, installer_url, add_config_path=None):
         tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
 
         # Remove the old DC/OS.
-        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
+        tunnel.remote_cmd(['sudo', '/opt/mesosphere/bin/dcos-shell', 'pkgpanda', 'uninstall'])
         tunnel.remote_cmd(['sudo', 'rm', '-rf', '/opt/mesosphere', '/etc/mesosphere'])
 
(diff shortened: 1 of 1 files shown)


test_util/cluster.py
```diff
@@ -320,7 +320,7 @@ def upgrade_host(tunnel, role, bootstrap_url):
         tunnel.remote_cmd(curl_cmd + ['--remote-name', bootstrap_url + '/dcos_install.sh'])
 
         # Remove the old DC/OS.
-        tunnel.remote_cmd(['sudo', '-i', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
+        tunnel.remote_cmd(['sudo', '-i', 'dcos-shell', '/opt/mesosphere/bin/pkgpanda', 'uninstall'])
```
branden: Thinking about how we'd document this in upgrade docs (which this should match), I think we want `sudo /opt/mesosphere/bin/dcos-shell pkgpanda uninstall`.

The `-i` was only there to get the environment right for `pkgpanda`, so we should be able to drop it now that we have `dcos-shell`. We should invoke it explicitly as `/opt/mesosphere/bin/dcos-shell`, since we're not assuming that `/opt/mesosphere/bin` is in `$PATH`.  `dcos-shell` _does_ add `/opt/mesosphere/bin` to `$PATH`, so we can drop the path prefix from the `pkgpanda` invocation.

BenWhitehead: I've updated to your recommendation in a new commit. Builds kicking off now.

BenWhitehead:
@branden I've rebased after previous all green, can you please review?

BenWhitehead:
Awesome, thanks!

BenWhitehead:
I don't think this is actually needed anymore.

PR #1153, opened 2017-01-21
Disable streaming for pytest

mellenburg:
Streaming, when combined with the teamcity plugin
causes some output to be randomly unparseable, thus
leading to some tests being ignored and unreported

### DIFF
--- a/test_util/azure_test_driver.py
+++ b/test_util/azure_test_driver.py
@@ -181,7 +181,7 @@ def main():
                 agent_list=ip_buckets['private'],
                 public_agent_list=ip_buckets['public'],
                 test_cmd=os.getenv(
-                    'DCOS_PYTEST_CMD', get_add_env() + " py.test -vv -s -rs -m 'not ccm' ") + os.getenv('CI_FLAGS', ''))
+                    'DCOS_PYTEST_CMD', get_add_env() + " py.test -vv -rs -m 'not ccm' ") + os.getenv('CI_FLAGS', ''))
         test_successful = True
     except Exception as ex:
         traceback.print_exc()
--- a/test_util/run-all
+++ b/test_util/run-all
@@ -95,7 +95,7 @@ export PUBLIC_SLAVE_HOSTS=192.168.65.60
 export TEAMCITY_VERSION="${TEAMCITY_VERSION:-}"
 export DCOS_DEFAULT_OS_USER=root
 cd /opt/mesosphere/active/dcos-integration-test
-/opt/mesosphere/bin/dcos-shell py.test -vv -s -rs -m "not ccm" ${CI_FLAGS:-}
+/opt/mesosphere/bin/dcos-shell py.test -vv -rs -m "not ccm" ${CI_FLAGS:-}
 EOF
 chmod +x test_wrapper.sh
 echo "Running integration test"
--- a/test_util/test_aws_cf.py
+++ b/test_util/test_aws_cf.py
@@ -98,7 +98,7 @@ def check_environment():
     for k, v in os.environ.items():
         if k.startswith(prefix):
             add_env.append(k.replace(prefix, '') + '=' + v)
-    options.test_cmd = os.getenv('DCOS_PYTEST_CMD', ' '.join(add_env) + ' py.test -vv -s -rs ' + options.ci_flags)
+    options.test_cmd = os.getenv('DCOS_PYTEST_CMD', ' '.join(add_env) + ' py.test -vv -rs ' + options.ci_flags)
     return options
 
 
--- a/test_util/test_aws_vpc.py
+++ b/test_util/test_aws_vpc.py
@@ -139,7 +139,7 @@ def check_environment():
     for k, v in os.environ.items():
         if k.startswith(prefix):
             add_env.append(k.replace(prefix, '') + '=' + v)
-    options.test_cmd = os.getenv('DCOS_PYTEST_CMD', ' '.join(add_env) + ' py.test -vv -s -rs ' + options.ci_flags)
+    options.test_cmd = os.getenv('DCOS_PYTEST_CMD', ' '.join(add_env) + ' py.test -vv -rs ' + options.ci_flags)
     return options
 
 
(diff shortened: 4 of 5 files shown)


mellenburg:
@adam-mesos @lingmann This ready for review. We should try to get this in ASAP so we kill the problem of failing test cases not showing up in our PR checks!

orsenthil:
```-s                    shortcut for --capture=no.```

I guess, it was not really  about streaming.

If we wanted some capture behavior,

```pytest --capture=sys```  could be tried again later.

**Comment by mellenburg** [comment]
@orsenthil Dont quite understand your comment. --capture=no is the same as 'enable streaming'. When pytest is not capturing output, it directly streams STDOUT to your own STDOUT (and out SSH tunnel is already streaming that from remote)

**Comment by orsenthil** [comment]
@mellenburg , It was confusion I had with the terms.

Without any context, 'capture=false' and 'streams' are two different concepts and I confused it.
With the context of py.test behavior, they seem to be same.

**Comment by lingmann** [comment]
So, it's unclear to me why the `-s` flag was originally set and it would be nice to have a better understanding before we make this change... here are two examples of a build, one using [py.test -vv -rs --teamcity](https://teamcity.mesosphere.io/viewLog.html?buildId=538246&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp&tab=buildLog#_state=35&focus=329) (this PR) and one on master using [py.test -vv -s -rs --teamcity](https://teamcity.mesosphere.io/viewLog.html?buildId=539161&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp&tab=buildLog#_state=37,339,1713&focus=329).

If py.test is configured to capture stdout/stderr (this PR) Teamcity appears to be properly aggregating the individual test results (you can click + to expand), but there are still spots where it [misses some of the TC annotation](https://teamcity.mesosphere.io/viewLog.html?buildId=538246&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmInstallTestAp&tab=buildLog#_state=35,339,340,348,355,369,397,686&focus=796). In the other case where py.test does not capture stdout/stderr the individual tests are **not** being aggregated in the build log. In **both** cases, TC was able to pick up the total number of tests passed in the overview section of the job report.

According to the [py.test documentation](http://doc.pytest.org/en/latest/capture.html), we probably want py.test to capture and manage stdout/stderr... otherwise I suspect the outputs of py.test and the tests themselves will be merged in unpredictable ways and TC will be unable to parse properly. However, with that said... I'm still concerned that even with this PR some TC annotation is being missed (at least two cases in the example above). So how do we even know that we can rely on TC to catch py.test failures reliably?

@BenWhitehead would appreciate any thoughts you might have on this.

**Comment by mellenburg** [comment]
@lingmann The -s flag was originally set because when -s is *not* set, you get no output until the test is over. Thus, if you have cluster where one node doesnt come up and you start running the tests (without -s), you will only see the following in the teamcity logs:
```
test_composition.test_cluster_is_up (57s)
```
and you will never get anything else util the test times out, the pytest process is abruptly killed and then you get *none* of the captured pytest output. If, on the other hand, you have that same case and *do* have -s set, then you will see:
```
	[test setup] [Test Error Output]
[2017-01-21 03:16:36,657] INFO: Request method get: http://10.10.0.187/
[2017-01-21 03:16:36,664] INFO: Nginx is UP!
[2017-01-21 03:16:36,664] INFO: Attempting authentication
[2017-01-21 03:16:36,664] INFO: Request method post: http://10.10.0.187/acs/api/v1/auth/login
[2017-01-21 03:16:37,496] INFO: Received authorization blob: {'token': 'eyJhbGciOiJIUzI1NiIsImtpZCI6InNlY3JldCIsInR5cCI6IkpXVCJ9.eyJhdWQiOiIzeUY1VE9TemRsSTQ1UTF4c3B4emVvR0JlOWZOeG05bSIsImVtYWlsIjoiYWxiZXJ0QGJla3N0aWwubmV0IiwiZW1haWxfdmVyaWZpZWQiOnRydWUsImV4cCI6Mi4wOTA4ODQ5NzRlKzA5LCJpYXQiOjEuNDYwMTY0OTc0ZSswOSwiaXNzIjoiaHR0cHM6Ly9kY29zLmF1dGgwLmNvbS8iLCJzdWIiOiJnb29nbGUtb2F1dGgyfDEwOTk2NDQ5OTAxMTEwODkwNTA1MCIsInVpZCI6ImFsYmVydEBiZWtzdGlsLm5ldCJ9.8SQ6Jadz9q4WMaReQv1OUiFGVKMqBo92_-cjHygOEdQ'}
[2017-01-21 03:16:37,496] INFO: Authentication successful
[2017-01-21 03:16:37,496] INFO: Request method get: http://10.10.0.187/marathon/ui/
[2017-01-21 03:16:37,520] INFO: Marathon is probably up
[2017-01-21 03:16:37,520] INFO: Request method get: http://10.10.0.187/exhibitor/exhibitor/v1/cluster/status
[2017-01-21 03:16:37,528] INFO: Exhibitor cluster status: [{'isLeader': True, 'description': 'serving', 'code': 3, 'hostname': '10.10.0.187'}]
[2017-01-21 03:16:37,528] INFO: Request method get: http://10.10.0.187/mesos/master/slaves
[2017-01-21 03:16:37,548] INFO: Inufficient (2 >= ['10.10.0.149', '10.10.0.53', '10.10.0.54']) number of slaves have joined the cluster. Continuing to wait...
[2017-01-21 03:17:37,548] INFO: Inufficient (2 >= ['10.10.0.149', '10.10.0.53', '10.10.0.54']) number of slaves have joined the cluster. Continuing to wait...
[2017-01-21 03:18:37,548] INFO: Inufficient (2 >= ['10.10.0.149', '10.10.0.53', '10.10.0.54']) number of slaves have joined the cluster. Continuing to wait...
```
because py.test is now pumping out the logging straight to the terminal instead. Thus, you can now look in the CI log and see that your cluster failed to come up after provisioning because a agent was missing. Without -s, you have nothing.

So, it seemed like a clean gain to add -s and get all the output as it was generated rather than hoping that the test would finish cleanly enough for py.test to serve us up the logs. Unfortunately, when -s is added, the teamcity plugin stops working and the log output has the teamcity symbols, but for some reason, they are nonfunctional and the logs are no longer collapsed. I am not certain that adding -s makes the test detection worse (as you clearly point out, teamcity messes up in either case), but I feel like I have seen more failed tests not get picked up since adding -s, so I figured it would be best to play it safe and revert it

PR #1154, opened 2017-01-21
Refactor test_upgrade_vpc.py by moving all test logic into a class

BenWhitehead:
Refactor test_upgrade_vpc.py by moving all test logic into a class

* The new class VpcClusterUpdateTest now provides an abstraction from running the test and configuring the test. This will allow the test to be instantiated with a different cluster api when ran against and enterprise cluster.
* Add teamcity logging messages for each of the major scopes run in this test

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -42,62 +42,32 @@ import logging
 import os
 import pprint
 import sys
+import traceback
 import uuid
-from contextlib import contextmanager
+from typing import Callable, List
 
 import retrying
+from teamcity.messages import TeamcityServiceMessages
 
 import test_util.aws
 import test_util.cluster
-from pkgpanda.util import load_string
+from pkgpanda.util import load_string, logger
 from test_util.dcos_api_session import DcosApiSession, DcosUser
 from test_util.helpers import CI_CREDENTIALS, marathon_app_id_to_mesos_dns_subdomain, random_id
 
 
 logging.basicConfig(format='[%(asctime)s|%(name)s|%(levelname)s]: %(message)s', level=logging.DEBUG)
-log = logging.getLogger(__name__)
 
 
 TEST_APP_NAME_FMT = 'upgrade-{}'
 
 
-@contextmanager
-def cluster_workload(cluster_api):
-    """Start a cluster workload on entry and verify its health on exit."""
-
-    def app_task_ids(app_id):
-        """Return a list of Mesos task IDs for app_id's running tasks."""
-        assert app_id.startswith('/')
-        response = cluster_api.marathon.get('/v2/apps' + app_id + '/tasks')
-        response.raise_for_status()
-        tasks = response.json()['tasks']
-        return [task['id'] for task in tasks]
-
-    def parse_dns_log(dns_log_content):
-        """Return a list of (timestamp, status) tuples from dns_log_content."""
-        dns_log = [line.strip().split(' ') for line in dns_log_content.strip().split('\n')]
-        if any(len(entry) != 2 or entry[1] not in ['SUCCESS', 'FAILURE'] for entry in dns_log):
-            message = 'Malformed DNS log.'
-            log.debug(message + ' DNS log content:\n' + dns_log_content)
-            raise Exception(message)
-        return dns_log
-
-    @retrying.retry(
-        wait_fixed=(1 * 1000),
-        stop_max_delay=(120 * 1000),
-        retry_on_result=lambda x: not x)
-    def wait_for_dns(hostname):
-        """Return True if Mesos-DNS has at least one entry for hostname."""
-        hosts = cluster_api.get('/mesos_dns/v1/hosts/' + hostname).json()
-        return any(h['host'] != '' and h['ip'] != '' for h in hosts)
-
-    test_id = uuid.uuid4().hex
-
+def create_marathon_healthcheck_app(app_id: str) -> dict:
     # HTTP healthcheck app to make sure tasks are reachable during the upgrade.
     # If a task fails its healthcheck, Marathon will terminate it and we'll
     # notice it was killed when we check tasks on exit.
-    healthcheck_app = {
-        "id": '/' + TEST_APP_NAME_FMT.format('healthcheck-' + test_id),
+    return {
+        "id": '/' + app_id,
         "cmd": "python3 -m http.server 8080",
         "cpus": 0.5,
         "mem": 32.0,
@@ -125,11 +95,13 @@ def cluster_workload(cluster_api):
         ],
     }
 
+
+def create_marathon_dns_app(app_id: str, healthcheck_app_id: str) -> dict:
     # DNS resolution app to make sure DNS is available during the upgrade.
     # Periodically resolves the healthcheck app's domain name and logs whether
     # it succeeded to a file in the Mesos sandbox.
-    dns_app = {
-        "id": '/' + TEST_APP_NAME_FMT.format('dns-' + test_id),
+    return {
+        "id": '/' + app_id,
         "cmd": """
 while true
 do
@@ -144,7 +116,7 @@ do
 done
 """,
         "env": {
-            'RESOLVE_NAME': marathon_app_id_to_mesos_dns_subdomain(healthcheck_app['id']) + '.marathon.mesos',
+            'RESOLVE_NAME': marathon_app_id_to_mesos_dns_subdomain(healthcheck_app_id) + '.marathon.mesos',
             'DNS_LOG_FILENAME': 'dns_resolve_log.txt',
             'INTERVAL_SECONDS': '1',
             'TIMEOUT_SECONDS': '1',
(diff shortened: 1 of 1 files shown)


BenWhitehead:
@orsenthil @branden Ready for Review

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
-    dns_failure_times = [entry[0] for entry in dns_log if entry[1] != 'SUCCESS']
-    if len(dns_failure_times) > 0:
-        message = 'Failed to resolve Marathon app hostname {} at least once.'.format(dns_app['env']['RESOLVE_NAME'])
-        log.debug(message + ' Hostname failed to resolve at these times:\n' + '\n'.join(dns_failure_times))
-        # Skip raising an exception on DNS failure so that other tests can run.
-        # DNS failure is being tracked at
-        # https://mesosphere.atlassian.net/browse/DCOS-13426.
-        #raise Exception(message)  # noqa
+
+class VpcClusterUpdateTestDcosApiSessionFactory:
```
orsenthil: One of `Update` or `Upgrade`. `Upgrade` is alright as we call the other function `VpcClusterUpgradeTest`

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+        self.installer_url = installer_url
+        self.aws_region = aws_region
+        self.aws_access_key_id = aws_access_key_id
+        self.aws_secret_access_key = aws_secret_access_key
+        self.ssh_key = ssh_key
+        self.default_os_user = default_os_user
+        self.config_yaml_override_install = config_yaml_override_install
+        self.config_yaml_override_upgrade = config_yaml_override_upgrade
+
+        self.teamcity_msg = TeamcityServiceMessages()
```
orsenthil: Right now, we have two ways to interact with Teamcity messages here. `pkgpanda.util.logger` and this one.

Having the TC logging done from a single place, ideally using the helper from `pkgpanda.util.` might be better IMO.  (Please vet this idea).

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+    def parse_dns_log(self, dns_log_content):
+        """Return a list of (timestamp, status) tuples from dns_log_content."""
+        dns_log = [line.strip().split(' ') for line in dns_log_content.strip().split('\n')]
+        if any(len(entry) != 2 or entry[1] not in ['SUCCESS', 'FAILURE'] for entry in dns_log):
+            message = 'Malformed DNS log.'
+            self.log.debug(message + ' DNS log content:\n' + dns_log_content)
+            raise Exception(message)
+        return dns_log
+
+    def log_test(self, test_name: str, call: Callable[[], None]) -> None:
```
orsenthil: This can be a decorator that can wrap the methods that we want to have the teamcity log. (Also, can this be from `pkgpanda.util` helper?  Same comment as the one made above.

If these are at the single place, probably more tests can see value of this by just using the helper.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+        try:
+            self.teamcity_msg.testStarted(test_name)
+            call()
+        except Exception:
+            # we want this except to be broad so that we can keep any exception from taking
```
orsenthil: `Exception`

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+            self.tasks_start = {app_id: sorted(self.app_task_ids(dcos_api, app_id)) for app_id in self.test_app_ids}
+            self.log.debug('Test app tasks at start:\n' + pprint.pformat(self.tasks_start))
+
+            for app in test_apps:
+                assert app['instances'] == len(self.tasks_start[app['id']])
+
+    def verify_apps_state(self, dcos_api: DcosApiSession, dns_app: dict):
+        with logger.scope("verify apps state"):
+
+            # nested methods here so we can "close" over external state
```
orsenthil: I don't see the `close over` here. need help.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+
+            master_list = [h.private_ip for h in cluster.masters]
+
+            dcos_api_install = self.dcos_api_session_factory_install.apply(
+                'http://{ip}'.format(ip=cluster.masters[0].public_ip),
+                master_list,
+                master_list,
+                [h.private_ip for h in cluster.agents],
+                [h.private_ip for h in cluster.public_agents],
+                self.default_os_user)             # default_os_user
```
orsenthil: The comment could be killed. The variable name label is descriptive.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+        # the client is being created to interact with. This client is specifically for the cluster after the
+        # upgrade has taken place, and can account for any possible settings that may change for the client under
+        # the hood when it probes the cluster.
+        dcos_api_upgrade = self.dcos_api_session_factory_upgrade.apply(
+            'http://{ip}'.format(ip=cluster.masters[0].public_ip),
+            master_list,
+            master_list,
+            [h.private_ip for h in cluster.agents],
+            [h.private_ip for h in cluster.public_agents],
+            self.default_os_user)             # default_os_user
```
orsenthil: Kill comment on `default_os_user`

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+
+        dcos_api_upgrade.wait_for_dcos()  # here we wait for DC/OS to be "up" so that we can auth this new client
+
+        self.verify_apps_state(dcos_api_upgrade, dns_app)
+
+        with logger.scope("run integration tests"):
+            # copied from test_util/test_aws_cf.py:96
```
orsenthil: What's the intention of this comment? Possible refactor in future? If not, then you could omit this.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+            vpc.delete()
+        else:
+            self.log.info("Test failed! VPC cluster will remain available for "
+                          "debugging for 2 hour after instantiation.")
+
+        return result
 
 
 def main():
+    class DAF(VpcClusterUpdateTestDcosApiSessionFactory):
```
orsenthil: This class can be outside of the main method. There is one advantage in keeping the class in the scope of the module. Suppose, you want to write a mock test for this class, It is way easier to mock it from the scope of the module than if the class is with the scope of the function.

Also, what is DAF?  DcosApiSessionFactory?

test_util/test_upgrade_vpc.py
BenWhitehead: The pkgpanda MessageLogger doesn't currently have methods for logging test info. And since pkgpanda is not a test library I feel like it shouldn't have them added to it.

test_util/test_upgrade_vpc.py
BenWhitehead: I'll look into the decorator option.

I don't really want to centralize this too much, since I would prefer that we get to the point that instead of doing this we are able to write pytests that are aware of the lifecycle of a cluster being upgraded.

test_util/test_upgrade_vpc.py
BenWhitehead: Yeah, possible future. I think @mellenburg was saying that this should be able to be cleaned up when we start using dcos-launch.

test_util/test_upgrade_vpc.py
orsenthil: +1. I was thinking as why `pkgpanda` as a namespace is coming to any teamcity based interaction, even if it only has to do with logging.  I think, logging itself should be separate and should be common (but overridable) for build and test. If you think along these lines and agree, we could take this up later.

test_util/test_upgrade_vpc.py
orsenthil: Ok, sgtm.

test_util/test_upgrade_vpc.py
BenWhitehead: The `dcos_api` and the `dns_app` passed to `verify_apps_state` is closed over by these functions that are then passed off to `log_test` and then executed.

test_util/test_upgrade_vpc.py
BenWhitehead: The defined factory is just a handle to be able to defer construction of the client, I don't think there is anything in particular to ever want to test in this implementation.

I'll rename `DAF` to `DASF` to match with `DcosApiSessionFactory`.

test_util/test_upgrade_vpc.py
BenWhitehead: I'm almost always in favor of consistent and clear lines of separation.

test_util/test_upgrade_vpc.py
orsenthil: (That makes a great skit dialog :) )

test_util/test_upgrade_vpc.py
orsenthil: I see!  I think, this may not require as a separate comment. It is usually taken for granted that nested functions are utilized only for accessing objects from the immediate outer scope.

test_util/test_upgrade_vpc.py
mellenburg: @BenWhitehead pkgpanda is already used all throughout the test library, so I find it a bit more confusing to add two interfaces. Maybe we should take those super generic utils out of there and make a top-level util for stuff to share

BenWhitehead: True, pkgpanda is used throughout the test library already. But the test library isn't used by pkgpanda for prod code. That is why I don't think we want to push test specific code back up into pkgpanda.

mellenburg: Ah, thats a fair point for not adding this to pkgpanda

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,277 @@ def wait_for_dns(hostname):
...
+            vpc.delete()
+        else:
+            self.log.info("Test failed! VPC cluster will remain available for "
+                          "debugging for 2 hour after instantiation.")
+
+        return result
 
 
 def main():
+    class DASF(VpcClusterUpgradeTestDcosApiSessionFactory):
```
orsenthil: Hi @BenWhitehead , my preference here was for a full name and not abbreviation.  Even if the fullname is a only a partial match with the intention. 

For moving it to outer scope, as a practice, I think, it might be preferable to define classes in module scope as much as possible and have defining class/functions within the function only when neccessary (like the closure concept you illustrated above).

orsenthil:
my only comment left. If addressed/ (rejected), I am fine with the change and will give the approval.

test_util/test_upgrade_vpc.py
BenWhitehead: rename and class pulled out of main pushed

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,279 @@ def wait_for_dns(hostname):
...
+        wait_fixed=(1 * 1000),
+        stop_max_delay=(120 * 1000),
+        retry_on_result=lambda x: not x)
+    def wait_for_dns(dcos_api, hostname):
+        """Return True if Mesos-DNS has at least one entry for hostname."""
+        hosts = dcos_api.get('/mesos_dns/v1/hosts/' + hostname).json()
+        return any(h['host'] != '' and h['ip'] != '' for h in hosts)
+
+    def setup_cluster_workload(self, dcos_api: DcosApiSession, healthcheck_app: dict, dns_app: dict):
+        """Start a cluster workload on entry and verify its health on exit."""
```
branden: The verification part of this docstring is no longer true. The method's name describes it pretty well, so we could just get rid of the docstring.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,279 @@ def wait_for_dns(hostname):
...
+
+class DcosApiSessionFactory(VpcClusterUpgradeTestDcosApiSessionFactory):
+    def apply(self, dcos_url: str, masters: List[str], public_masters: List[str], slaves: List[str],
+              public_slaves: List[str], default_os_user: str) -> DcosApiSession:
+        return DcosApiSession(dcos_url, masters, public_masters, slaves, public_slaves,
+                              default_os_user, DcosUser(CI_CREDENTIALS))
 
 
 def main():
+
```
branden: Is this added newline a leftover from something?

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,279 @@ def wait_for_dns(hostname):
...
+                if len(dns_failure_times) > 0:
+                    message = 'Failed to resolve Marathon app hostname {} at least once.'.format(
+                        dns_app['env']['RESOLVE_NAME'])
+                    err_msg = message + ' Hostname failed to resolve at these times:\n' + '\n'.join(dns_failure_times)
+                    self.log.debug(err_msg)
+                    self.teamcity_msg.testFailed("test_upgrade_vpc.test_app_dns_survive_upgrade", details=err_msg)
+                    # Skip raising an exception on DNS failure so that other tests can run.
+                    # DNS failure is being tracked at
+                    # https://mesosphere.atlassian.net/browse/DCOS-13426.
+                    # raise Exception(message)  # noqa
```
branden: Since we're now failing the test in TeamCity, we can get rid of the above comments.

test_util/test_upgrade_vpc.py
```diff
@@ -159,125 +132,279 @@ def wait_for_dns(hostname):
...
+    log = logging.getLogger(__name__)
+
+    def __init__(self,
+                 num_masters: int, num_agents: int, num_public_agents: int,
+                 stable_installer_url: str, installer_url: str,
+                 aws_region: str, aws_access_key_id: str, aws_secret_access_key: str,
+                 ssh_key: str, default_os_user: str,
+                 config_yaml_override_install: str, config_yaml_override_upgrade: str,
+                 dcos_api_session_factory_install: VpcClusterUpgradeTestDcosApiSessionFactory,
+                 dcos_api_session_factory_upgrade: VpcClusterUpgradeTestDcosApiSessionFactory):
```
orsenthil: When I looked at it during review, I felt Factory was ok. Folks do factory in python [twisted.internet.protocol.Factory](http://twistedmatrix.com/documents/current/api/twisted.internet.protocol.Factory.html), [sqlalchemy example](https://github.com/zzzeek/sqlalchemy/blob/772374735da27df1ddb907f4a0f5085b46dbe82b/test/base/test_events.py#L813) and [stdlib example](https://github.com/python/cpython/blob/c30098c8c6014f3340a369a31df9c74bdbacc269/Lib/symtable.py#L16) 

The choice of using a factory vs not using a factory is to be done based on if want to the action in `Class.__init__` or a method that is called by the object later ( like `apply` here).

This http://softwareengineering.stackexchange.com/a/253256/16639 is a good argument for factory in this case. 

I'd like hear @BenWhitehead thought on his choice.

test_util/test_upgrade_vpc.py
```diff
@@ -42,62 +42,33 @@
 import os
 import pprint
 import sys
+import traceback
 import uuid
-from contextlib import contextmanager
+
+from typing import Callable, List
```
BenWhitehead: When I try to move it up with the above group flake8 gives me the following error:
```
./test_util/test_upgrade_vpc.py:47:1: I100 Import statements are in the wrong order.import uuid should be before from typing
```

test_util/test_upgrade_vpc.py
BenWhitehead: I specifically want this to be a factory. The reason being (possible over enforcement from static typing mindset) that I want the thing that is used to create the `DcosApiSession` to not be able to pass an `instanceof` check, whereas the thing returned by `apply` must pass.

test_util/test_upgrade_vpc.py
BenWhitehead: will remove

test_util/test_upgrade_vpc.py
BenWhitehead: will remove

test_util/test_upgrade_vpc.py
BenWhitehead: will remove

test_util/test_upgrade_vpc.py
branden: My proposal should meet that requirement:

```
In [1]: from test_util.dcos_api_session import DcosApiSession

In [2]: isinstance(DcosApiSession, DcosApiSession)
Out[2]: False

In [3]: isinstance(DcosApiSession(dcos_url='foo', masters=[], public_masters=[], slaves=[], public_slaves=[], default_os_user='', auth
   ...: _user=''), DcosApiSession)
Out[3]: True

In [4]:
```

BenWhitehead: What would the use sight look like then? I don't want to loose the "Give me a thing that when I invoke it will give me a `DcosApiSession`" that I get with the factory class right now (admittedly this is my python ignorance) the whole point of introducing the factory is that user is able to provide the path, but the test controls the time when creating the client.

branden: You'd just pass the class like it were the factory object (which it basically is). It'd look like this:

```
test = VpcClusterUpgradeTest(..., dcos_api_session_factory_install=DcosApiSession, dcos_api_session_factory_upgrade=DcosApiSession, ...)
```

And the class could call its factory attribute with the args to `DcosApiSession.__init__()` to get an instance:

```
dcos_api_install = self.dcos_api_session_factory_install(*args)
```

branden: Chatted with @BenWhitehead offline, we agreed that this class makes it easier for downstream repos to override factory behavior while maintaining type safety. 👍

PR #1163, opened 2017-01-24
WIP: Split by token

branden:
## High Level Description

Depends on https://github.com/dcos/dcos/pull/1066.

This PR cleans up some tech debt introduced with https://github.com/dcos/dcos/pull/1066 by consolidating the logic for finding named tokens within a string.

## Related Issues

## Checklist for all PR's


branden:
Ready for review @amitaekbote @orsenthil

gen/__init__.py
```diff
@@ -306,46 +306,23 @@ def do_gen_package(config, package_filename):
...
-        yield late_values[name]
-
-        # Interpolate values into the rest of the string.
-        yield from _interpolate_values(string_[placeholder_end[1]:])
-
-    return ''.join(_interpolate_values(content))
+
+    def _dereference_placeholders(parts):
+        for part, is_placeholder in parts:
+            if is_placeholder:
```
orsenthil: The nested if in line 313 can be rolled up to line 312 with an `and` statement.

pkgpanda/util.py
```diff
@@ -413,3 +414,74 @@ def hash_checkout(item):
         return hash_list(list(item))
     else:
         raise NotImplementedError("{} of type {}".format(item, type(item)))
+
+
+def split_by_token(token_prefix, token_suffix, string_, strip_token_decoration=False):
```
orsenthil: I feel this is great improvement. 👍 

Some suggestions, and please see if you would like to consider them. (Also, feel free state a) if you agree, but it is not not worth it at this at this point, or b) it is a bad idea).


* Currently, defining token_prefix and token_suffix left to client. So, I can use the split_by_token like this.

```
split_by_token("foo", "bar", "foo something bar")
```
And I'd get the same string back.

* I am not sure if that flexibility is desirable. It can give way to negative test cases like above.

* Instead, if we define a well known tokens.

```
BLOCK_NAME_START = "{"
BLOCK_NAME_END = "}"

block_token = (BLOCK_NAME_START, BLOCK_NAME_END)

and split_by_token(block_token, "{something}")

```

We are restricting it to well-known-tokens that we define.  We can raise ValueError when a new kind of unknown token is encountered.

branden: We can't define a universal token format because this function has to detect references in AWS and Azure templates, whose syntax we don't control. See usage examples for the different token decoration we need to catch.

gen/__init__.py
branden: Not without losing the `yield late_values[part]` -- this block is doing more than an error check. If I rolled that up I'd need to add an `elif is_placeholder and part in late_values` to make sure we actually dereference placeholders, which just complicates the logic.

orsenthil: I missed seeing the Exception at the indented level. 
See the reason now.

pkgpanda/util.py
orsenthil: I yield to the current design.

My suggestion was not a new format. Basically keeping all the formats matched-up at a single place called format_types

```
format-types = {
'braces': ["{", "}"],
'double_square_brackets: ["[[", "]]"],
}
```

So, that when matching you only give the format type `split_by_token("braces", "{something}")` the module takes care of the rest.   This can prevent the possibility of a bad example that I gave in the previous comment.

Not a critical change request, so ok to drop it.

PR #1169, opened 2017-01-26
Update navstar to fixed ipvs

aeyakovenko:
## High Level Description

Fixes a bug in minuteman routes that prevents VIPs from working in bridge mode.

## Related Issues

  - [DCOS-13228](https://mesosphere.atlassian.net/browse/DCOS-13228) TRACKING: dcosjira-13228, VIPs are not working in bridge mode

## Checklist for all PR's

Added networking tests to cover every permunation of VIP

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review):
https://github.com/dcos/minuteman/compare/452b57b07d08def9a4586b6be046943785440529...fe299aa5c9e7956407643f2284c91fd2c52ac955
https://github.com/dcos/navstar/compare/4ef753e175a93adc597827aa28a836e43b8ccf3d...439b52afd04cd0614d52b9c80c18835fcf916cf2
https://circleci.com/gh/dcos/minuteman/424?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
https://codecov.io/gh/dcos/minuteman/commit/47b47a162da51285724c7a80b29dc5d8cc07db5f
___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -16,7 +16,7 @@ from test_util.marathon import get_test_app, get_test_app_in_docker
 
 log = logging.getLogger(__name__)
 timeout = 500
-maxthreads = 15
+maxthreads = 8
 
 
 def lb_enabled():
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -1,14 +1,22 @@
+import contextlib
+import json
 import logging
+import random
+import threading
+from collections import deque
 from subprocess import check_output
 
 import pytest
-
 import requests
-
 import retrying
 
 from pkgpanda.build import load_json
 from test_util.marathon import get_test_app, get_test_app_in_docker
+
+
+log = logging.getLogger(__name__)
+timeout = 500
+maxthreads = 15
 
 
 def lb_enabled():
@@ -16,22 +24,147 @@ def lb_enabled():
     return config['enable_lb'] == 'true'
 
 
-def ensure_routable(cmd, service_points, timeout=120):
-    @retrying.retry(wait_fixed=2000,
-                    stop_max_delay=timeout * 1000,
-                    retry_on_result=lambda ret: ret is False,
-                    retry_on_exception=lambda x: True)
-    def _ensure_routable():
-        proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
-        logging.info('Sending {} data: {}'.format(proxy_uri, cmd))
-        r = requests.post(proxy_uri, data=cmd)
-        logging.info('Requests Response: %s', repr(r.json()))
-        assert r.json()['status'] == 0
-    return _ensure_routable
+@retrying.retry(wait_fixed=2000,
+                stop_max_delay=timeout * 1000,
+                retry_on_result=lambda ret: ret is False,
+                retry_on_exception=lambda x: True)
+def ensure_routable(cmd, service_points):
+    proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
+    log.info('Sending {} data: {}'.format(proxy_uri, cmd))
+    r = requests.post(proxy_uri, data=cmd)
+    log.info('Requests Response: %s', repr(r.json()))
+    assert r.json()['status'] == 0
+    return json.loads(r.json()['output'])
+
+
+class VipTest:
+    def __init__(self, num, vip, vipaddr, samehost, vipnet, proxynet):
+        self.vip = vip.format(num, 7000 + num)
+        self.vipaddr = vipaddr.format(num, 7000 + num)
+        self.samehost = samehost
+        self.vipnet = vipnet
+        self.proxynet = proxynet
+
+    def __str__(self):
+        return ("VipTest(vip={},vipaddr={},samehost={},"
+                "vipnet={},proxynet={})").format(self.vip, self.vipaddr, self.samehost, self.vipnet, self.proxynet)
+
+    def log(self, s, lvl=logging.DEBUG):
+        m = 'VIP_TEST {} {}'.format(s, self)
+        log.log(lvl, m)
+
+
+def vip_test(dcos_api_session, r):
+    r.log('START')
+    agents = list(dcos_api_session.slaves)
+    # make sure we can reproduce
+    random.seed(r.vip)
+    random.shuffle(agents)
+    host1 = agents[0]
+    host2 = agents[0]
+    if not r.samehost:
+        host2 = agents[1]
+    log.debug('host1 is is: {}'.format(host1))
+    log.debug('host2 is is: {}'.format(host2))
+    origin_app, app_uuid = get_test_app_in_docker()
+    origin_app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': r.vip}
+    origin_app['container']['docker']['network'] = r.vipnet
+    origin_app['mem'] = 0
+    origin_app['cpu'] = 0
+    if r.vipnet == 'USER':
+        origin_app['ipAddress'] = {'networkName': 'dcos'}
+    if r.vipnet == 'HOST':
+        origin_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                            '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        origin_app['container']['docker']['portMappings'][0]['hostPort'] = 0
+        origin_app['container']['docker']['portMappings'][0]['containerPort'] = 0
+        origin_app['container']['docker']['portMappings'][0]['labels'] = {}
+        origin_app['portDefinitions'] = [{'labels': {'VIP_0': r.vip}}]
+
+    origin_app['constraints'] = [['hostname', 'CLUSTER', host1]]
+
+    proxy_app = get_test_app_in_docker()[0]
+    proxy_app['container']['docker']['network'] = r.proxynet
+    proxy_app['constraints'] = [['hostname', 'CLUSTER', host2]]
+    proxy_app['mem'] = 0
+    proxy_app['cpu'] = 0
+    if r.proxynet == 'USER':
+        proxy_app['ipAddress'] = {'networkName': 'dcos'}
+    if r.proxynet == 'HOST':
+        proxy_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                           '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        proxy_app['container']['docker']['portMappings'][0]['hostPort'] = 0
+        proxy_app['container']['docker']['portMappings'][0]['containerPort'] = 0
+
+    returned_uuid = None
+
+    with contextlib.ExitStack() as stack:
+        stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app, timeout=timeout))
+        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app, timeout=timeout))
+        cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(r.vipaddr)
+        returned_uuid = ensure_routable(cmd, sp)
+        log.debug('returned_uuid is: {}'.format(returned_uuid))
+    assert returned_uuid is not None
+    assert returned_uuid['test_uuid'] == app_uuid
+    r.log('PASSED')
+
+
+@pytest.fixture
+def reduce_logging():
+    start_log_level = logging.getLogger('test_util.marathon').getEffectiveLevel()
+    # gotta go up to warning to mute it as its currently at info
+    logging.getLogger('test_util.marathon').setLevel(logging.WARNING)
+    yield
+    logging.getLogger('test_util.marathon').setLevel(start_log_level)
+
+
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
+def test_vip(dcos_api_session, reduce_logging):
+    """Test every permutation of VIP
+    """
+    addrs = [['1.1.1.{}:{}', '1.1.1.{}:{}'],
+             ['/namedvip{}:{}', 'namedvip{}.marathon.l4lb.thisdcos.directory:{}']]
+    samehosts = [True, False]
+    vipnets = ['USER', 'BRIDGE', 'HOST']
+    proxynets = ['USER', 'BRIDGE', 'HOST']
+    permutations = [[vi, va, sh, vn, pn]
+                    for [vi, va] in addrs
+                    for sh in samehosts
+                    for vn in vipnets
+                    for pn in proxynets]
+    tests = [VipTest(i, vi, va, sh, vn, pn) for i, [vi, va, sh, vn, pn] in enumerate(permutations)]
+    threads = []
+    # deque is thread safe
+    failed_tests = deque(tests)
+    passed_tests = deque()
+    skipped_tests = deque()
+    for r in tests:
+        if not r.samehost and len(dcos_api_session.slaves) == 1:
+            failed_tests.remove(r)
+            skipped_tests.append(r)
+            continue
+
+        def run(test):
+            vip_test(dcos_api_session, test)
+            failed_tests.remove(test)
+            passed_tests.append(test)
+
+        threads.append(threading.Thread(target=run, args=([r])))
+    groups = [threads[i:i + maxthreads] for i in range(0, len(threads), maxthreads)]
+    for gs in groups:
+        for t in gs:
+            t.start()
+        for t in gs:
+            t.join()
+    [r.log('PASSED', lvl=logging.INFO) for r in passed_tests]
+    [r.log('SKIPPED', lvl=logging.INFO) for r in skipped_tests]
+    [r.log('FAILED', lvl=logging.INFO) for r in failed_tests]
+    log.info('VIP_TEST num agents: {}'.format(len(dcos_api_session.slaves)))
+    assert len(failed_tests) == 0
 
 
 @retrying.retry(wait_fixed=2000,
-                stop_max_delay=120 * 1000,
+                stop_max_delay=timeout * 1000,
                 retry_on_exception=lambda x: True)
 def test_if_overlay_ok(dcos_api_session):
     def _check_overlay(hostname, port):
@@ -55,33 +188,6 @@ def test_if_minuteman_disabled(dcos_api_session):
     assert str(data).find('9999') == -1
 
 
-@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
-def test_if_minuteman_routes_to_vip(dcos_api_session):
-    """Test if we are able to connect to a task with a vip using minuteman.
-    """
-    origin_app, origin_uuid = get_test_app()
-    origin_app['portDefinitions'][0]['labels'] = {'VIP_0': '1.2.3.4:5000'}
-    with dcos_api_session.marathon.deploy_and_cleanup(origin_app):
-        proxy_app, proxy_uuid = get_test_app()
-        with dcos_api_session.marathon.deploy_and_cleanup(proxy_app) as service_points:
-            cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://1.2.3.4:5000/ping'
-            ensure_routable(cmd, service_points)()
-
-
-@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
-def test_if_minuteman_routes_to_named_vip(dcos_api_session):
-    """Test if we are able to connect to a task with a named vip using minuteman.
-    """
-
-    origin_app, origin_uuid = get_test_app()
-    origin_app['portDefinitions'][0]['labels'] = {'VIP_0': 'foo:5000'}
-    with dcos_api_session.marathon.deploy_and_cleanup(origin_app):
-        proxy_app, proxy_uuid = get_test_app()
-        with dcos_api_session.marathon.deploy_and_cleanup(proxy_app) as service_points:
-            cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://foo.marathon.l4lb.thisdcos.directory:5000/ping'
-            ensure_routable(cmd, service_points)()
-
-
 def test_ip_per_container(dcos_api_session):
     """Test if we are able to connect to a task with ip-per-container mode
     """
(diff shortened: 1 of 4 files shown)


asridharan:
@aeyakovenko looks like code quality issues. Might want to run flake on that file in a container before submitting.

aeyakovenko:
@asridharan can you add @GoelDeepak and @mellenburg as reviewers?

asridharan:
@aeyakovenko once this ready for review just remove the [WIP] tag and ping anyone you want to review this.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,120 @@ def _ensure_routable():
...
+        returned_uuid = ensure_routable(cmd, sp)()
+        logging.info('returned_uuid is: {}'.format(returned_uuid))
+    assert(returned_uuid is not None)
+    assert(returned_uuid['test_uuid'] == app_uuid)
+    logging.info('VIP_TEST PASSED {}'.format(teststr))
+    return app_uuid
+
+
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
+def test_vip(dcos_api_session):
```
orsenthil: This should see some overhaul.

1) Keep `addrs` and one of `vipnets` and `proxynets` (call it something common).
2) Have only two external for-loops to track. Kill the outer for-loop to iterate over samehosts  [True, False]. See if this can folded inside the code.
3) call the variable names failed, passed, skipped something with a more meaningful identifer (easy).
4) See if using of `num` can be avoided. (Encode the ports within).

5) Ideally, if this is constructing the configurations of the ip-addresses, just construct them early and separately as a structure by itself into a list and then iterate over it to add to the threads. ?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -16,7 +19,7 @@ def lb_enabled():
     return config['enable_lb'] == 'true'
 
 
-def ensure_routable(cmd, service_points, timeout=120):
+def ensure_routable(cmd, service_points, timeout=180):
```
mellenburg: This function should be fixed up. It makes no sense to return a callable function in this usage context. Rather, get ride of _ensure_routable entirely and just put this decorator at the top level. Thus, you just need to call ensure_routable(cmd, service_points) to evaluate this and nothing will be lost

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+                                                                           proxynet)
+    logging.info('VIP_TEST START {}'.format(teststr))
+    agents = list(dcos_api_session.slaves)
+    # make sure we can reproduce
+    random.seed(vip)
+    random.shuffle(agents)
+    host1 = agents[0]
+    host2 = agents[0]
+    if not samehost:
+        host2 = agents[1]
```
mellenburg: This test will likely fail for any cluster that has more than 2 private agents. Either this needs to be written in a way that takes the service points from the app deployment to figure out where to ping or this whole test needs to be skipped when we have more than 2 private agents. I would vote for the former. Any reason that's not possible?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+    return app_uuid
+
+
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
+def test_vip(dcos_api_session):
+    """Test every permutation of VIP
+    """
+    num = 0
+    addrs = [['1.1.1.{}:{}', '1.1.1.{}:{}'],
+             ['/namedvip{}:{}', 'namedvip{}.marathon.l4lb.thisdcos.directory:{}']]
```
mellenburg: @aeyakovenko I noticed you changed this from parametrize. That was the preferred way of configuring something like this. Did changing away from that fix that app cleanup issue? Or is app cleanup still a problem?

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: @mellenburg can you explain why it will fail?  it should run with more than two private agents, and distribute across all the private agents randomly.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: you mean the parameter fixtures?  There is no way to run all those tests in parallel.  If run sequentially they take 40 minutes :).

mellenburg: Ah, yes. Of course.

The idea of using a fixture to start a thread pool like you describe won't really work within the pytest framework because what you would end up having (if I understand this correctly) would be:
```
@pytest.fixture
def my_thread_control():
    # generate all your threads and start them
    yield
    # join your threads and check them
```
and all of your asserting (i.e. the test cases) will be done after joining threads. Anything you define in a test case won't be able to get back into a fixture.

However, I think as we get more and more tests, this issue of parallelization will begin concern more people. I am thinking of building a plugin where you could mark tests like so:
```
@parallel(scope='networking_test')
def test_my_network(some_setup_fixture):
    pass

@parallel(scope='networking_test')
def test_my_network_2(some_setup_fixture):
    pass
```
this would require changing up the runtest protocol, but pytest is really open about this, so I dont think it will be that difficult.

I think for now we should probably just stick to making the huge lot of threads in a single test as you do now, so to keep this focused on the test instead of pytest madness

packages/dcos-integration-test/extra/test_networking.py
mellenburg: Oops, brainfart. No problem here. My bad

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+    logging.info('host1 is is: {}'.format(host1))
+    logging.info('host2 is is: {}'.format(host2))
+    origin_app, app_uuid = get_test_app_in_docker()
+    origin_app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': vip}
+    origin_app['container']['docker']['network'] = vipnet
+    if vipnet == 'USER':
+        origin_app['ipAddress'] = {'networkName': 'dcos'}
+    if vipnet == 'HOST':
+        origin_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '
+        origin_app['cmd'] += '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py '
```
mellenburg: Can we remove this string-append to a key-value syntax? Please just use:
```
origin_app['cmd'] = 'wow, this command is really really really '  \
    'long. I don't think I will be able to fit it on one line, so I guess ' \
    'its time to use the escape char'
```

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+        origin_app['cmd'] += '$PORT0'
+        origin_app['container']['docker']['portMappings'][0]['hostPort'] = 0
+        origin_app['container']['docker']['portMappings'][0]['containerPort'] = 0
+        origin_app['container']['docker']['portMappings'][0]['labels'] = {}
+        origin_app['portDefinitions'] = [{'labels': {'VIP_0': vip}}]
+
+    origin_app['constraints'] = [['hostname', 'CLUSTER', host1]]
+    origin_app['constraints'] = [['hostname', 'CLUSTER', host1]]
```
mellenburg: repeated line; remove

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+
+    proxy_app = get_test_app_in_docker()[0]
+    proxy_app['container']['docker']['network'] = proxynet
+    proxy_app['constraints'] = [['hostname', 'CLUSTER', host2]]
+    if proxynet == 'USER':
+        proxy_app['ipAddress'] = {'networkName': 'dcos'}
+    if proxynet == 'HOST':
+        proxy_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '
+        proxy_app['cmd'] += '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py '
+        proxy_app['cmd'] += '$PORT0'
```
mellenburg: see above on formatting at once instead of appending

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -27,9 +30,123 @@ def _ensure_routable():
...
+    returned_uuid = None
+
+    with contextlib.ExitStack() as stack:
+        stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app))
+        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app))
+        cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(vipaddr)
+        returned_uuid = ensure_routable(cmd, sp)()
+        logging.info('returned_uuid is: {}'.format(returned_uuid))
+    assert(returned_uuid is not None)
+    assert(returned_uuid['test_uuid'] == app_uuid)
```
mellenburg: parentheses are not required for assert and are omitted everywhere else in testing

test_util/test_upgrade_vpc.py
```diff
@@ -211,7 +211,7 @@ def wait_for_dns(hostname):
         # Skip raising an exception on DNS failure so that other tests can run.
         # DNS failure is being tracked at
         # https://mesosphere.atlassian.net/browse/DCOS-13426.
-        #raise Exception(message)  # noqa
+        # raise Exception(message)  # noqa
```
mellenburg: Can we remove that silly # noqa You did the right syntax fix here, but it wasnt picked up before due to that # noqa

branden: The missing space here was deliberate -- I wanted this to look like commented-out code, not a block comment. That's what the `# noqa` is there to protect.

orsenthil: wow. interesting. :)

mellenburg: Oh, I see now. Yeah, that's fine

aeyakovenko: code check fails without it.  should i revert?

mellenburg: Apparently your python linter doesnt respect ```# noqa``` which prevents any linting errors from that line

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -16,18 +19,131 @@ def lb_enabled():
...
-                    stop_max_delay=timeout * 1000,
-                    retry_on_result=lambda ret: ret is False,
-                    retry_on_exception=lambda x: True)
-    def _ensure_routable():
-        proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
-        logging.info('Sending {} data: {}'.format(proxy_uri, cmd))
-        r = requests.post(proxy_uri, data=cmd)
-        logging.info('Requests Response: %s', repr(r.json()))
-        assert r.json()['status'] == 0
-    return _ensure_routable
```
mellenburg: Much better. Thanks.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -16,18 +19,131 @@ def lb_enabled():
...
+
+    with contextlib.ExitStack() as stack:
+        stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app))
+        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app))
+        cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(r.vipaddr)
+        returned_uuid = ensure_routable(cmd, sp)
+        logging.info('returned_uuid is: {}'.format(returned_uuid))
+    assert returned_uuid is not None
+    assert returned_uuid['test_uuid'] == app_uuid
+    r.log('PASSED')
```
mellenburg: This log statement is redundant with the pass statement below; please remove

test_util/test_upgrade_vpc.py
aeyakovenko: so revert?

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: it makes watching the build.log live more exciting!

asridharan:
@aeyakovenko any reason why the AWS and Azure tests are failing?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,156 @@
...
-                    retry_on_exception=lambda x: True)
-    def _ensure_routable():
-        proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
-        logging.info('Sending {} data: {}'.format(proxy_uri, cmd))
-        r = requests.post(proxy_uri, data=cmd)
-        logging.info('Requests Response: %s', repr(r.json()))
-        assert r.json()['status'] == 0
-    return _ensure_routable
+@retrying.retry(wait_fixed=2000,
+                stop_max_delay=TIMEOUT * 1000,
```
mellenburg: Timeout has changed from 2 minutes to 8 minutes and 20 seconds. This seems like a pretty big change. If we need to bump it, maybe 5 minutes (300) would be a more suitable value?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -96,21 +188,58 @@ def test_ip_per_container(dcos_api_session):
...
+
+
+@retrying.retry(wait_fixed=2000,
+                stop_max_delay=100 * 2000,
+                retry_on_exception=lambda x: True)
+def geturl(url):
+    rs = requests.get(url)
+    assert rs.status_code == 200
+    r = rs.json()
+    logging.info('geturl {} -> {}'.format(url, r))
```
mellenburg: make this into a debug statement

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,156 @@
...
+    agents = list(dcos_api_session.slaves)
+    # make sure we can reproduce
+    random.seed(r.vip)
+    random.shuffle(agents)
+    host1 = agents[0]
+    host2 = agents[0]
+    if not r.samehost:
+        host2 = agents[1]
+    logging.info('host1 is is: {}'.format(host1))
+    logging.info('host2 is is: {}'.format(host2))
```
mellenburg: debug statements please

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,156 @@
...
-        logging.info('Requests Response: %s', repr(r.json()))
-        assert r.json()['status'] == 0
-    return _ensure_routable
+@retrying.retry(wait_fixed=2000,
+                stop_max_delay=TIMEOUT * 1000,
+                retry_on_result=lambda ret: ret is False,
+                retry_on_exception=lambda x: True)
+def ensure_routable(cmd, service_points):
+    proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
+    logging.info('Sending {} data: {}'.format(proxy_uri, cmd))
```
mellenburg: debug statement please

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,156 @@
...
+    r = requests.post(proxy_uri, data=cmd)
+    logging.info('Requests Response: %s', repr(r.json()))
```
mellenburg: debug statement please

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,156 @@
...
 import requests
-
 import retrying
 
 from pkgpanda.build import load_json
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
+TIMEOUT = 500
+
```
mellenburg: This test adds some non-trivial log output on top of what was already a pretty decent amount of logging. Could we please add a module level logger?
```
```
And then migrate any statements from logging to use that log. This way we can crank up the verbosity of just this test for easy debugging

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,161 @@
...
 import requests
-
 import retrying
 
 from pkgpanda.build import load_json
 from test_util.marathon import get_test_app, get_test_app_in_docker
 
 
+log = logging.getLogger(__name__)
+log.setLevel(logging.INFO)
```
mellenburg: Don't leave this log.setLevel(logging.INFO) in this commit. We want this logger to inherit the root logging level (which is already at INFO)

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,165 @@
...
+        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app, timeout=timeout))
+        cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(r.vipaddr)
+        returned_uuid = ensure_routable(cmd, sp)
+        log.debug('returned_uuid is: {}'.format(returned_uuid))
+    assert returned_uuid is not None
+    assert returned_uuid['test_uuid'] == app_uuid
+    r.log('PASSED')
+
+
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
```
mellenburg: Skipping a fixture doesnt make too much sense, you want this decorator around test_vip I think

packages/navstar/extra/dcos-navstar.service
```diff
@@ -21,4 +21,7 @@ ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-navstar
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-minuteman
 ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/navstar/mnesia
 ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/navstar/lashup
+ExecStartPre=/usr/bin/env modprobe dummy
+ExecStartPre=-/usr/bin/env ip link add minuteman type dummy
```
lingmann: The number of ExecStartPre's here and now the addition of a `-` (fails are ignored) on this line suggest that we need to turn all of this into a proper startup script with error handling.

packages/navstar/extra/dcos-navstar.service
```diff
@@ -21,4 +21,7 @@ ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-navstar
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-minuteman
 ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/navstar/mnesia
 ExecStartPre=/usr/bin/mkdir -p /var/lib/dcos/navstar/lashup
+ExecStartPre=/usr/bin/env modprobe dummy
```
asridharan: Why do we need this dummy interface? I thought setting up the local routing table was enough?

test_util/marathon.py
```diff
@@ -11,6 +11,7 @@
 
 TEST_APP_NAME_FMT = 'integration-test-{}'
 REQUIRED_HEADERS = {'Accept': 'application/json, text/plain, */*'}
+log = logging.getLogger(__name__)
```
asridharan: Why are we changing the logger for the python tests? Just wondering why its part of this PR?

packages/navstar/extra/dcos-navstar.service
asridharan: +1 on @lingmann comment on converting this to script. We should just have one `ExecStartPre` point it to a shell and handle all failure conditions in the shell. Would help us debug errors a lot better.

aeyakovenko: @asridharan @lingmann filed a jira for this, https://mesosphere.atlassian.net/browse/DCOS-13698
i think it can be handled in a separate pull request.  we don't have a "way to do this" yet afaik.  so it will take a little bit of thought to implement

test_util/marathon.py
mellenburg: This is the idiom we are moving towards in all test infrastructure and allows this VIP test in particular to dynamically control logging in an idiomatic way

packages/navstar/extra/dcos-navstar.service
aeyakovenko: we need it to mark the routes as `minuteman`, otherwise we can't reconcile

packages/navstar/extra/dcos-navstar.service
alberts: +1 on taking care of this in a follow-up PR

lingmann:
@aeyakovenko looks good. Ready for the train as soon as tests go green.

lingmann:
@asridharan have your issues been addressed / can you update your PR?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,170 @@
...
+            failed_tests.remove(test)
+            passed_tests.append(test)
+
+        threads.append(threading.Thread(target=run, args=([r])))
+    groups = [threads[i:i + maxthreads] for i in range(0, len(threads), maxthreads)]
+    for gs in groups:
+        for t in gs:
+            t.start()
+        for t in gs:
+            t.join()
```
jgehrcke: For next time :-) https://docs.python.org/dev/library/concurrent.futures.html#threadpoolexecutor

asridharan:
LGTM !! 🚢 
The only concern I have is with the number of `ExecPre` commands we are setting up. But I guess we can handle them in a separate PR.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -1,37 +1,170 @@
...
+    def __str__(self):
+        return ("VipTest(vip={},vipaddr={},samehost={},"
+                "vipnet={},proxynet={})").format(self.vip, self.vipaddr, self.samehost, self.vipnet, self.proxynet)
+
+    def log(self, s, lvl=logging.DEBUG):
+        m = 'VIP_TEST {} {}'.format(s, self)
+        log.log(lvl, m)
+
+
+def vip_test(dcos_api_session, r):
```
asridharan: Should have pointed this out earlier, this test is very `DockerContainerizer` specific. Would be a pain to change it if you want to replicate it for UCR. Can we modify it so that the same vip_test can be invoked with a paramter to test `DockerContainerizer` vs UCR?

asridharan: @aeyakovenko sorry for the preemptive 🚢 wanted to check if the above is going to be a problem?

aeyakovenko: already have a jira and PR for UCR :) https://github.com/dcos/dcos/pull/1224

asridharan: Looks like the refactor will be done in another PR. so 🚢

orsenthil: [minor comments]:
* Please rename `r` to something meaningful 🔴 

* Also, please use a rough guideline of about 5 lines for the function. If you go by this, this function can see some love.

asridharan:
@lingmann is it possible to indicate here which train this pull request is going to catch? I see Train 128 being mentioned here but that doesn't seem to include these commits?

orsenthil:
@mesosphere-mergebot bump-ee

alberts:
@lingmann bump-ee looks green now. let's :train2: ?

lingmann:
Yay! Nice work @aeyakovenko.

PR #1170, opened 2017-01-26
fix systemd units

gpaul:
## High Level Description

What features does this change enable? 
This fixes the `dcos-spartan.service` file so the LimitNOFILE directive works with older versions of systemd. It also adds tests that check all service and socket files in `/etc/systemd/system/*` for sanity. 

What bugs does this change fix? 
We set `LimitNOFILE=32000:32000` in `dcos-spartan.service` but on Centos 7.2.1511, which runs systemd 219, this has no effect as systemd does not understand that format. This functionality was only introduced by https://github.com/systemd/systemd/pull/1994

## Related Issues

  - [DCOS-623](https://dcosjira.atlassian.net/browse/DCOS-623) spartan file limit config is wrong
  - [DCOS-13460](https://mesosphere.atlassian.net/browse/DCOS-13460) TRACKING DCOS-623: spartan file limit config is wrong
  - [DCOS-629](https://dcosjira.atlassian.net/browse/DCOS-629) dcos-navstar.service file has env var in WorkingDirectory=
  - [DCOS-630](https://dcosjira.atlassian.net/browse/DCOS-630) extract dcos-pkgpanda-api.service ExecStart= directive into a shell script
  - [DCOS-631](https://dcosjira.atlassian.net/browse/DCOS-631) dcos-adminrouter.service calls /usr/bin/kill instead of /bin/kill

## Checklist for all PR's

___

Reviewers should be:


packages/dcos-integration-test/extra/test_units.py
```diff
@@ -0,0 +1,34 @@
+import subprocess
+
+import pytest
+
+
+def test_verify_units():
+    """Test that all systemd units are valid."""
```
alberts: @gpaul can you confirm that this catches the broken LimitNOFILE= directive? We were thinking that what we really want (maybe in addition to this) is checks on /proc/<pid>/limits. This is admittedly a bit tricky, since e.g. Spartan ends up with 3 processes running in its unit (maybe we can fix that too).

packages/navstar/extra/dcos-navstar.service
```diff
@@ -5,7 +5,7 @@ Description=Navstar: A distributed systems & network overlay orchestration engin
 Restart=always
 StartLimitInterval=0
 RestartSec=5
-WorkingDirectory=${PKG_PATH}/navstar
+WorkingDirectory=/opt/mesosphere/active/navstar/navstar
```
alberts: Don't think we want to do this. The build script should be taking care of replacing PKG_PATH here. There are other examples in the repo.

gpaul: There is prior art for this in https://github.com/dcos/dcos/blob/master/packages/spartan/extra/dcos-spartan.service

packages/dcos-integration-test/extra/test_units.py
gpaul: Yes - this would have catched the limitNOFILE= issue. I confirmed on a centos 7.2 system.

gpaul:
The added tests will fail if the `LimitNOFILE=` change is reverted - but only if they run on a system running an old enough version of systemd. We should make sure to run our tests on CentOS 7.2 machines too, despite CentOS 7.3 being available.

packages/adminrouter/extra/dcos-adminrouter.service
```diff
@@ -23,6 +23,6 @@ ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id
 # Wait for auth backend to seed the secret key
 ExecStartPre=/usr/bin/curl --fail -sS -o /dev/null 127.0.0.1:8101/acs/api/v1/groups
 ExecStart=$PKG_PATH/nginx/sbin/nginx -c $PKG_PATH/nginx/conf/nginx.master.conf
-ExecReload=/usr/bin/kill -HUP $MAINPID
+ExecReload=/bin/kill -HUP $MAINPID
```
lingmann: Two things:
1. We need to confirm that `/bin/kill` exists on all the distros we currently deploy against (CoreOS, Ubuntu, CentOS)... I think adding a simple test for it's existence in our standard test suite would be sufficient. I would be surprised if it's not... but I've been surprised in this area before.
2. Systemd [ExecReload docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecReload=) strongly recommends to wait synchronously for the reload to complete and not just rely on `kill -HUP`.

packages/dcos-integration-test/extra/test_units.py
```diff
@@ -0,0 +1,52 @@
...
+                    # third-party software ships bad unit files, so we
+                    # explicitly check that 'dcos-' is present on a
+                    # line before checking if it is valid.
+                    if "dcos-" not in line:
+                        return True
+                    # The TasksMax directive exists in newer versions of systemd
+                    # where it is important to set. As we want to support multiple
+                    # versions of systemd our tests must ignore errors that
+                    # complain that it is an unknown directive.
+                    ignore_new_directives = ["TasksMax"]
```
lingmann: Could you conditionally ignore these only when an old version of systemd is detected? I worry that we'll miss all testing for any directive that is placed in `ignore_new_directives`.

packages/navstar/build
```diff
@@ -22,7 +22,7 @@ mkdir -p $(dirname $service)
 
 navstar_service=${PKG_PATH}/dcos.target.wants/dcos-navstar.service
 mkdir -p $(dirname $navstar_service)
-cp /pkg/extra/dcos-navstar.service "$navstar_service"
+envsubst '$PKG_PATH' < "/pkg/extra/dcos-navstar.service" > "${navstar_service}"
```
lingmann: I believe `envsubst '${PKG_PATH}' ...` is more explicit per the [SHELL-FORMAT docs](https://www.gnu.org/software/gettext/manual/gettext.html#sh_002dformat) Also, what was the reason for this change? Was WorkingDirectory not being set correctly before?

packages/pkgpanda-api/build
```diff
@@ -12,4 +12,9 @@ cp /pkg/extra/dcos-pkgpanda-api.socket $systemd_socket
 
 systemd_service=$PKG_PATH/dcos.target.wants/dcos-pkgpanda-api.service
 mkdir -p $(dirname $systemd_service)
-cp /pkg/extra/dcos-pkgpanda-api.service $systemd_service
+envsubst '$PKG_PATH' < /pkg/extra/dcos-pkgpanda-api.service > "$systemd_service"
```
lingmann: I believe `envsubst '${PKG_PATH}' ...` is more explicit per the [SHELL-FORMAT docs](https://www.gnu.org/software/gettext/manual/gettext.html#sh_002dformat)

spahl:
Do we need anything besides the first two commits in this PR? They seem unrelated (which cause confusion in backporting)

spahl:
Ok I understand now, it was failing on azure (ubuntu...)

packages/adminrouter/extra/dcos-adminrouter.service
gpaul: 1. Sure - do we have a list of officially supported distros and versions? I'll pull some dockers down and confirm that /bin/kill exists in all of them.
2. Thanks for pointing that out - I didn't know they suggest using a synchronous reload option if possible. In this case, the official `nginx -s reload` command just sends SIGHUP to the server, too.

packages/dcos-integration-test/extra/test_units.py
gpaul: This would be great but it's tricky. The first idea that jumps to mind is to use the systemd version reported by `systemctl --version`. Unfortunately that reports `v219` in both CentOS 7.2 and 7.3, even though 7.3 has several useful features backported to it that 7.2's systemd does not have.

I get the feeling that to distinguish we'd have to use something like `systemd-analyze verify` and run into a chicken and egg problem.

If we shipped man pages one could grep those for mention of the ignored directives. Can you think of something similar, perhaps?

packages/navstar/build
gpaul: Thanks! Will change.

Regarding WorkingDirectory - yes, systemd doesn't do environment variable expansion for the WorkingDirectory directive.

gpaul: The sections on ExecStart, etc. specifically mention when they support environment variable substitution: https://www.freedesktop.org/software/systemd/man/systemd.service.html

packages/pkgpanda-api/build
gpaul: Thanks - I'll fix this

packages/dcos-integration-test/extra/test_units.py
gpaul: Perhaps I should check that the line should specifically complain about "unknown directive" instead of ignoring any line containing an ignored directive.

packages/adminrouter/extra/dcos-adminrouter.service
gpaul: 1. I checked the following distributions and confirmed that `/bin/kill` is present.

- ubuntu 16.04
- ubuntu 14.04
- ubuntu 12.04

- centos 7.3.1611
- centos 7.2.1511

- coreos 1298.1.0  (Alpha, Jan 22, 2017)
- coreos 1235.8.0 (Stable, Jan 31, 2017)
- coreos 522.6.0 (Stable, Jan 28, 2015)

- debian 8

Earlier debians seem to not ship whatever package provides /bin/kill in them by default.

packages/dcos-integration-test/extra/test_units.py
gpaul: This tightening of the check has been added in the latest commit.

gpaul:
@lingmann PTAL

lingmann:
Thanks @gpaul LGTM

spahl:
Correct. Please create an change in the downstream repo to verify that systemd analyze does not find issues in the units there.

gpaul:
The following PR tests these changes downstream:
https://github.com/mesosphere/dcos-enterprise/pull/571

lingmann:
@gpaul I made a few commented on the corresponding enterprise PR.

gpaul:
Thanks - I've addressed the downstream adminrouter config changes in a separate targeted PR:
https://github.com/mesosphere/dcos-enterprise/pull/574

That PR needs to land downstream before/along with this PR.

gpaul:
@lingmann PTAL

spahl:
@lingmann all your comments seems adressed upstream and downstream. I moved it to shipit, let us know if you see any more issues.

PR #1187, opened 2017-01-30
standardize on zookeeper connection string

gpaul:
## High Level Description

Spartan supports resolving `zk-1.zk, zk-2.zk, ..., zk-5.zk` to the IPs of the running zookeeper instances. When there are fewer instances than 5, the extra hostnames resolve to those servers that are present.

This allows services to run even if their local zookeeper instance is down.

This is nothing new. This patch just does some house cleaning by replacing one or two places where the old style was still used.

## Related Issues

  - [DCOS-626](https://dcosjira.atlassian.net/browse/DCOS-626) Standardize zookeeper address string

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: there are already several tests that will fail if the connection string functionality fails.

___

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -370,7 +370,7 @@ package:
       MESOS_SLAVE_REMOVAL_RATE_LIMIT=1/20mins
       MESOS_OFFER_TIMEOUT=2mins
       MESOS_WORK_DIR=/var/lib/dcos/mesos/master
-      MESOS_ZK=zk://127.0.0.1:2181/mesos
+      MESOS_ZK=zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181/mesos
       MESOS_WEIGHTS={{ weights }}
       MESOS_QUORUM={{ master_quorum }}
       MESOS_HOSTNAME_LOOKUP=false
--- a/test_util/dcos_api_session.py
+++ b/test_util/dcos_api_session.py
@@ -110,7 +110,7 @@ class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
         self.slaves = sorted(slaves)
         self.public_slaves = sorted(public_slaves)
         self.all_slaves = sorted(slaves + public_slaves)
-        self.zk_hostports = ','.join(':'.join([host, '2181']) for host in self.public_masters)
+        self.zk_hostports = "zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181"
         self.default_os_user = default_os_user
         self.auth_user = auth_user
 
--- a/test_util/dcos_api_session.py
+++ b/test_util/dcos_api_session.py
@@ -110,7 +110,6 @@ class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
         self.slaves = sorted(slaves)
         self.public_slaves = sorted(public_slaves)
         self.all_slaves = sorted(slaves + public_slaves)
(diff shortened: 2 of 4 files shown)


gpaul:
The only remaining instance of {{127.0.0.1:2181}} that I can find is {{dcos-oauth}}, which does not currently support multiple zookeeper addresses in its connection string.

I've created a ticket for adding this.
https://dcosjira.atlassian.net/browse/DCOS-627

gpaul:
@mellenburg @alberts please review

test_util/dcos_api_session.py
```diff
@@ -110,7 +110,7 @@ def __init__(self, dcos_url: str, masters: list, public_masters: list,
         self.slaves = sorted(slaves)
         self.public_slaves = sorted(public_slaves)
         self.all_slaves = sorted(slaves + public_slaves)
-        self.zk_hostports = ','.join(':'.join([host, '2181']) for host in self.public_masters)
+        self.zk_hostports = "zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181"
```
mellenburg: In an effort to make this class less God-like, can we completely remove this as a member variable and just define it in the one place that it is used? 
https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_composition.py

This will also need an update:
https://github.com/mesosphere/dcos-enterprise/blob/master/packages/dcos-integration-test/extra/test_clusterstate.py#L23

gpaul: Sure, good idea

gpaul: PTAL. I've updated the accompanying enterprise PR accordingly here https://github.com/mesosphere/dcos-enterprise/pull/502/files#diff-99053e08b3a194291f27808e1dabbcaaR23

gen/dcos-config.yaml
```diff
@@ -370,7 +370,7 @@ package:
       MESOS_SLAVE_REMOVAL_RATE_LIMIT=1/20mins
       MESOS_OFFER_TIMEOUT=2mins
       MESOS_WORK_DIR=/var/lib/dcos/mesos/master
-      MESOS_ZK=zk://127.0.0.1:2181/mesos
+      MESOS_ZK=zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181/mesos
```
mellenburg: Oops, didnt catch this in the last review, should this go up to zk-9?
https://github.com/dcos/dcos/blob/master/gen/calc.py#L302

gpaul: Nope

> Spartan also enables high availability of Zookeepers. You can always use the addresses zk-1.zk, zk-2.zk, zk-3.zk, zk-4.zk, zk-5.zk. If there are fewer than 5 Zookeepers, we will point multiple records at a single Zookeeper.

https://github.com/dcos/spartan/blob/master/README.md

mellenburg: But that only mentions if we have less than 5 ZKs. This doesn't imply that zk 6 through 9 will ever be used in a 9 cluster config...

gpaul: Yep - it only goes up to 5.

jgehrcke:
:ship:!

I am curious about a thing, though:

> This allows services to run even if their local zookeeper instance is down.

But not if Spartan fails. What fails more often, the local ZK or Spartan?

gpaul:
@jgehrcke excellent point - the very scientific approach of taking a quick look at the Open and Enterprise Jira's suggests that spartan is more stable. I find the following argument to be more convincing: we have a team of people who are comfortable improving spartan and lack such a team for zookeeper (and exhibitor.)

PR #1190, opened 2017-01-31
Add dcos-launch support for AWS advanced clusters

mellenburg:
In order to support launching Advanced (aka Zen) AWS templates, DC/OS Launch needed to be aware of which class representation from test_util.aws. Also, the Zen templates do not natively support turnkey deployment, so this PR also adds a helper to support exactly this case

- Clusters are determined to be either simple or zen when they are fetched via checking for the presence of resources that are exclusive between the two
- The only interface difference for a user deploying a zen stack will be in the case where the user did not follow the docs and create network resources in setup, in which case, (s)he can add a field to the config called ```zen_helper: true```
- Cleans up existing AWS test code to have more sensible classes, generators, and properties
- Simplifies key handling by removing confusing method of declaring it with ssh_info-key_name
- Dependent upon shipped PR: https://github.com/dcos/dcos/pull/1205
## Related Issues

## Checklist for all PR's

### DIFF
--- a/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
+++ b/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
@@ -1,52 +1,74 @@
+import logging
 import os
-from functools import partial
 
 import pytest
+import retrying
 
 import test_util.aws
 import test_util.helpers
 from test_util.helpers import retry_boto_rate_limits
 
+log = logging.getLogger(__name__)
+
 ENV_FLAG = 'ENABLE_RESILIENCY_TESTING'
 
-pytestmark = pytest.mark.skipif(
+skip_if_resiliency_not_enabled = pytest.mark.skipif(
     ENV_FLAG not in os.environ or os.environ[ENV_FLAG] != 'true',
     reason='Must explicitly enable resiliency testing with {}'.format(ENV_FLAG))
 
+skip_if_not_aws_stack = pytest.mark.skipif(
+    'AWS_STACK_NAME' not in os.environ,
+    reason='Must use an AWS Cloudformation run this test!')
+
+pytestmark = [skip_if_resiliency_not_enabled, skip_if_not_aws_stack]
+
 
 @pytest.fixture(scope='session')
-def dcos_launchpad(dcos_api_session):
-    """Interface for direct integration to dcos_launchpad hardware
-    Currently only supports AWS CF with AWS VPC coming soon
-    """
-    if 'AWS_STACK_NAME' not in os.environ:
-        # TODO(mellenburg): update when advanced templates are merged
-        pytest.skip('Must use a AWS Cloudformation to run test')
-    stack_name = os.environ['AWS_STACK_NAME']
+def boto_wrapper(dcos_api_session):
     aws_region = os.environ['AWS_REGION']
     aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
     aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
-    bw = test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
-    return test_util.aws.DcosCfSimple(stack_name, bw)
+    return test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
+
+
+@pytest.fixture(scope='session')
+def dcos_stack(boto_wrapper):
+    """ Works with either Zen or Simple cloud formations
+    """
+    stack = test_util.aws.fetch_stack(os.environ['AWS_STACK_NAME'], boto_wrapper)
+    if isinstance(stack, test_util.aws.VpcCfStack):
+        pytest.skip('Onprem Vpc not currently supported')
+    return stack
 
 
 @pytest.mark.last
-def test_agent_failure(dcos_launchpad, dcos_api_session, vip_apps):
-    # make sure the app works before starting
+def test_agent_failure(dcos_stack, boto_wrapper, dcos_api_session, vip_apps):
+    # Accessing AWS Resource objects will trigger a client describe call.
+    # As such, any method that touches AWS APIs must be wrapped to avoid
+    # CI collapse when rate limits are inevitably reached
     @retry_boto_rate_limits
-    def get_running_agents(group_name):
-        return [i for i in dcos_launchpad.get_auto_scaling_instances(group_name)
-                if i.state['Name'] == 'running']
+    def get_running_instances(instance_iter):
+        return [i for i in instance_iter if i.state['Name'] == 'running']
 
+    @retry_boto_rate_limits
+    def get_instance_ids(instance_iter):
+        return [i.instance_id for i in instance_iter]
+
+    @retry_boto_rate_limits
+    def get_private_ips(instance_iter):
+        return sorted([i.private_ip_address for i in get_running_instances(instance_iter)])
+
+    # make sure the app works before starting
     test_util.helpers.wait_for_pong(vip_apps[0][1], 120)
     test_util.helpers.wait_for_pong(vip_apps[1][1], 10)
-    agents = [i.instance_id for i in get_running_agents('PublicSlaveServerGroup') +
-              get_running_agents('SlaveServerGroup')]
+    agent_ids = get_instance_ids(
+        get_running_instances(dcos_stack.public_agent_instances) +
+        get_running_instances(dcos_stack.private_agent_instances))
 
     # Agents are in auto-scaling groups, so they will automatically be replaced
-    dcos_launchpad.boto_wrapper.client('ec2').terminate_instances(InstanceIds=agents)
-    waiter = dcos_launchpad.boto_wrapper.client('ec2').get_waiter('instance_terminated')
-    retry_boto_rate_limits(waiter.wait)(InstanceIds=agents)
+    boto_wrapper.client('ec2').terminate_instances(InstanceIds=agent_ids)
+    waiter = boto_wrapper.client('ec2').get_waiter('instance_terminated')
+    retry_boto_rate_limits(waiter.wait)(InstanceIds=agent_ids)
 
     # Tell mesos the machines are "down" and not coming up so things get rescheduled.
     down_hosts = [{'hostname': slave, 'ip': slave} for slave in dcos_api_session.all_slaves]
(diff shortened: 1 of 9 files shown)


mellenburg:
@orsenthil @amitaekbote PTAL

test_util/aws.py
```diff
@@ -142,32 +216,22 @@ def get_parameter(self, param):
...
-                    AutoScalingGroupNames=[self.stack.Resource(logical_id).physical_resource_id])
-                ['AutoScalingGroups'] for i in asg['Instances']]
-
-    def get_hosts_cached(self, group_name, refresh=False):
-        if refresh or group_name not in self._host_cache:
-            host_list = instances_to_hosts(self.get_auto_scaling_instances(group_name))
-            self._host_cache[group_name] = host_list
-            return host_list
-        return self._host_cache[group_name]
+    def _cached(self, key, value_fn, refresh=False):
```
orsenthil: `invalidate` instead of `refresh` ? (more common term) 
If you agree to this, other usages of 'refresh` below might change too.

test_util/aws.py
```diff
@@ -179,77 +243,59 @@ def create(cls, stack_name, template_url, public_agents, private_agents,
...
+    def delete_exhibitor_s3_bucket(self):
+        """ A non-empty S3 bucket cannot be deleted, so check to
+        see if it should be emptied first. If its non-empty, but
+        has more than one item, error out as the bucket is perhaps
+        not an exhibitor bucket and the user should be alerted
+        """
+        bucket = self.boto_wrapper.resource('s3').Bucket(
+            self.stack.Resource('ExhibitorS3Bucket').physical_resource_id)
+        log.info('Starting bucket {} deletion'.format(bucket))
+        all_objects = list(bucket.objects.all())
```
orsenthil: minor note. In python3, iterators support len.  They dont need to be converted to list. Please see if this is applicable here.

test_util/aws.py
```diff
@@ -71,12 +95,62 @@ def create_stack(self, name, template_url, parameters, deploy_timeout=60):
...
+
+    def delete_internet_gateway(self, gateway_id):
+        self.client('ec2').delete_internet_gateway(InternetGatewayId=gateway_id)
+
+    def delete_vpc(self, vpc_id):
+        self.client('ec2').delete_vpc(VpcId=vpc_id)
+
+    @retry_boto_rate_limits
+    def get_auto_scaling_instances(self, asg_name):
+        """ Get instances in ASG with logical_id. If logical_id is None, all ASGs will be used
```
mellenburg: Good catch. This was a copy paste doc string that needs serious updating

test_util/aws.py
mellenburg: Good to know, but in this specific case we actually want to get the resources in the list to potentially delete, so we do want a list

test_util/aws.py
mellenburg: Actually, what I decided when I slept on this is that this whole caching thing needs to go. It makes the code rather hard to maintain and follow, it has an ugly API, it messes with inheritance, and doesnt even work well for the zen stack case. Furthermore, I looked in the code and no where does anyone call these methods in a successive manner that would actually warrant caching this data silently.

Thus, its un-used code and it can go. Its super easy for the caller of this code to implement its own caching. If the caller for whatever reason cannot do this, then we should address this with a more comprehensive approach when that time comes

orsenthil: sounds good to me.

test_util/aws.py
```diff
@@ -33,6 +47,16 @@ def instances_to_hosts(instances):
     return [Host(i.private_ip_address, i.public_ip_address) for i in instances]
 
 
+def fetch_stack(stack_name, boto_wrapper):
+    stack = boto_wrapper.resource('cloudformation').Stack(stack_name)
+    for resource in stack.resource_summaries.all():
+        if resource.logical_resource_id == 'MasterStack':
+            return DcosZenCfStack(stack_name, boto_wrapper)
+        if resource.logical_resource_id == 'MasterServerGroup':
```
mellenburg: I do not follow your question. These terms are dictated by our own AWS templates. In 'simple deploys' we have a single stack with everything jammed in. This means there must be a master auto scaling group (no guarantee about agents in any scenario). If this is a zen stack, then the autoscaling groups are segregated in their own private stacks. If the arbitrary logical ID's that we put in our templates change, then this will break.

orsenthil: What's the relationship between `MasterStack` and `DcosZenCfStack` and correspondingly, what's the relationship between `MasterServerGroup` and `DcosCfStack` ?

mellenburg:
@orsenthil Thanks for the last pass over, this is ready for a proper review now
@amitaekbote PTAL

test_util/launch.py
```diff
@@ -61,40 +68,103 @@ def delete(self, info):
...
             'stack_name': config['stack_name'],
             'provider': {
                 'region': config['provider_info']['region'],
                 'access_key_id': config['provider_info']['access_key_id'],
                 'secret_access_key': config['provider_info']['secret_access_key']},
-            'ssh': ssh_info}
+            'ssh': ssh_info,
+            'temp_resources': temp_resources}
+
+    def provide_network_prereqs(self, config):
```
orsenthil: Please see if this method can be refactored into smaller chunks.

* These flags, False, False, False, False and conditional actions on top of it works, but hinders readability IHMO.
* Instead, see if features can be composed on top of others. For e.g., I'd check for param["ParameterKey"] for "Vpc" very late and immediately, add the `vpc_id`.
* I'd follow the similar approach for other flags.

test_util/launch.py
```diff
@@ -106,46 +176,61 @@ def describe(self, info):
...
+        ---
+        parameters:
+          - ParameterKey: KeyName
+            ParameterValue: my_key_name
+        ### Result: nothing generated, testing not possible
 
-        This method will take the config dict, determine if a key name is provided, if necessary, generate
-        the key and amend the config, and return the ssh_info for the cluster info JSON
+        ---
+        ### Result: key generated, testing not possible
```
orsenthil: These `###` comments could come on top instead of bottom.

test_util/launch.py
```diff
@@ -154,58 +239,31 @@ def ssh_from_config(self, config):
...
+        return ssh_info, temp_resources
 
     def get_stack(self, info):
-        """Returns the correct class interface depending how the AWS CF is configured
-        NOTE: only supports Simple Cloudformation currently
-        """
         try:
-            return DcosCfSimple(info['stack_name'], self.boto_wrapper)
+            return fetch_stack(info['stack_name'], self.boto_wrapper)
         except Exception as ex:
```
orsenthil: Not your change. But specific Exception?

test_util/test_launch.py
```diff
@@ -13,6 +13,15 @@
 
 MOCK_SSH_KEY_DATA = 'ssh_key_data'
 MOCK_KEY_NAME = 'my_key_name'
+MOCK_VPC_ID = 'vpc-foo-bar'
+MOCK_SUBNET_ID = 'subnet-foo-bar'
+MOCK_GATEWAY_ID = 'gateway-foo-bar'
+
+
+def magic_fn(output):
```
orsenthil: Better function name please. Could be called as `stub`.

test_util/launch.py
```diff
@@ -61,40 +68,103 @@ def delete(self, info):
     def ssh_from_config(self, info):
         raise NotImplementedError()
 
-    def test(self, info):
-        raise NotImplementedError()
+    def test(self, info, test_cmd):
```
mellenburg: This is an abstract class because there will be an AzureLauncher, AwsVpcLauncher, OnpremLaunhcer, and probably a GceVpcLauncher which must all use this same API. Creating a provider generic abstract API was one of the driving design decisions for all of dcos-launch

test_util/launch.py
```diff
@@ -61,40 +68,103 @@ def delete(self, info):
...
-        # parameter to config['parameters']
-        self.boto_wrapper.create_stack(
-            config['stack_name'], config['template_url'], config['parameters'])
+        ssh_info, temp_resources = self.ssh_from_config(config)
+        if config.get('zen_helper', False):
+            temp_resources.update(self.provide_network_prereqs(config))
+        try:
+            self.boto_wrapper.create_stack(
+                config['stack_name'], config['template_url'], config['parameters'])
+        except Exception as ex:
```
mellenburg: No, any exception. If the launch fails for *any* reason, we want to delete all the resources that were implicitly created if we can

test_util/launch.py
mellenburg: 4 conditional flags would be pretty nasty looking. Let's talk this one over in person

test_util/launch.py
mellenburg: Sure, if you find that to be more readable

test_util/launch.py
mellenburg: We want to catch all and recast as a LauncherError here so that the error it self is readable on the command line

mellenburg:
@orsenthil This PR now has the following update:
- Zen resource marginally restructured to allow for deleting a partially provisioned cluster
- small patch to avoid ugly os_string error
- specific error catching inside test_util.aws.py test_util.launch.py is still top level and I want to keep botocore specific details in test_util/aws.py
- made stack information to be stack_id based instead of stack_name so that we can try deleting clusters that failed at deleting (once a stack is 'inactive' it must be accessed via id and NOT name)

amitaekbote:
🚢

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -9,44 +9,56 @@
...
     aws_region = os.environ['AWS_REGION']
     aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
     aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
-    bw = test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
-    return test_util.aws.DcosCfSimple(stack_name, bw)
+    return test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
+
+
+@pytest.fixture(scope='session')
+def dcos_stack(boto_wrapper):
```
mellenburg: I was thinking the term dcos_stack was exclusive to AWS cloudformation as the other providers do not use the term stack, but best to be explicit (although its pretty explicit elsewhere in this file)

PR #1198, opened 2017-02-01
cosmos: restrict admin and http ports to localhost + test fix

alberts:
The Cosmos service currently runs with default options for its admin port (`0.0.0.0:9990`) and its http port (`0.0.0.0:7070`.) We should restrict Cosmos to listen on the loopback interface only.

This change prevents bypassing Admin Router via direct access to Cosmos.

# Issues

[DCOS-613](https://dcosjira.atlassian.net/browse/DCOS-613)

# Checklist

Supercedes https://github.com/dcos/dcos/pull/1134

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -467,7 +467,7 @@ package:
           "http://leader.mesos:1050/system/health/v1/report?cache=0"
         ],
         "cosmos_urls": [
-          "http://leader.mesos:7070/package/list"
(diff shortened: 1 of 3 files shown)


amitaekbote:
might be worth it to test downstream with this change

orsenthil:
Looks like i have to remove #1134 from train-125 and add this one.

Need more approvals from @spahl and @gpaul .

orsenthil:
The tests are green. Waiting for approvals from @spahl , @gpaul .

PR #1202, opened 2017-02-02
Make delete volumes resilient to rate limiting

mellenburg:
It is an on going battle with AWS to find every part of code that fails due to API rate limiting and then patching with this decorator duct tape


mellenburg:
@orsenthil @lingmann This is a quick small fix to improve test stability. Can we get it on the next train?

PR #1204, opened 2017-02-02
exhibitor: write JNA libraries to somewhere that isn't noexec

gpaul:
## High Level Description

By default the JNA tmpdir is set to `/tmp`. This conflicts with some standard security policies that dictate that `/tmp` should be mounted as `noexec`.

@alberts's first thought is to use `/run`, but that is mounted `noexec` by default on Ubuntu 16.04 LTS, too.

We should set the JNA tmpdir to somewhere we control: `/var/lib/dcos/exhibitor/tmp` seems like a good option.

## Related Issues

  - [DCOS-634](https://dcosjira.atlassian.net/browse/DCOS-634) zookepeer should not write JNA libraries to /tmp
  - [DCOS-11056](https://mesosphere.atlassian.net/browse/DCOS-11056) The requirment of mounting /tmp without `noexec` flag conflicts with security policies that some customers have

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a config change - if it doesn't work zookeeper doesn't start and we'll notice.

___

### DIFF
--- a/packages/exhibitor/extra/dcos-exhibitor.service
+++ b/packages/exhibitor/extra/dcos-exhibitor.service
@@ -12,6 +12,7 @@ MountFlags=private
 RuntimeDirectory=dcos_exhibitor
 # On some systems /tmp is mounted as noexec. Make zookeeper write its JNA
 # libraries to a path we control instead. See DCOS-11056
+Environment=JNA_TMP_DIR=/var/lib/dcos/exhibitor/tmp
 Environment=SERVER_JVMFLAGS=-Djna.tmpdir=/var/lib/dcos/exhibitor/tmp
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
--- a/packages/exhibitor/extra/dcos-exhibitor.service
+++ b/packages/exhibitor/extra/dcos-exhibitor.service
@@ -10,6 +10,8 @@ RestartSec=5
 # Run in new mount namespace to create custom resolv.conf.
 MountFlags=private
 RuntimeDirectory=dcos_exhibitor
+# Make ZooKeeper not write JNA libraries to /tmp, which can be noexec
+Environment=SERVER_JVMFLAGS=-Djna.tmpdir=/run/dcos_exhibitor
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
 EnvironmentFile=/opt/mesosphere/etc/exhibitor
--- a/packages/exhibitor/extra/dcos-exhibitor.service
+++ b/packages/exhibitor/extra/dcos-exhibitor.service
@@ -10,8 +10,9 @@ RestartSec=5
 # Run in new mount namespace to create custom resolv.conf.
 MountFlags=private
 RuntimeDirectory=dcos_exhibitor
-# Make ZooKeeper not write JNA libraries to /tmp, which can be noexec
-Environment=SERVER_JVMFLAGS=-Djna.tmpdir=/run/dcos_exhibitor
+# On some systems /tmp is mounted as noexec. Make zookeeper write its JNA
+# libraries to a path we control instead. See DCOS-11056
+Environment=SERVER_JVMFLAGS=-Djna.tmpdir=/var/lib/dcos/exhibitor/tmp
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dns_config
 EnvironmentFile=/opt/mesosphere/etc/exhibitor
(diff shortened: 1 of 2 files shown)


gpaul:
@alberts, @lingmann please review

packages/exhibitor/extra/dcos-exhibitor.service
```diff
@@ -10,6 +10,9 @@ RestartSec=5
 # Run in new mount namespace to create custom resolv.conf.
 MountFlags=private
 RuntimeDirectory=dcos_exhibitor
+# On some systems /tmp is mounted as noexec. Make zookeeper write its JNA
+# libraries to a path we control instead. See DCOS-11056
+Environment=SERVER_JVMFLAGS=-Djna.tmpdir=/var/lib/dcos/exhibitor/tmp
```
BenWhitehead: Wouldn't it make more sense to write this value into the existing `/opt/mesosphere/etc/exhibitor` env file?

packages/exhibitor/extra/start_exhibitor.py
```diff
@@ -49,6 +54,7 @@ def invoke_detect_ip():
     '-Djava.util.prefs.userRoot=/var/lib/dcos/exhibitor/',
     '-Duser.home=/var/lib/dcos/exhibitor/',
     '-Duser.dir=/var/lib/dcos/exhibitor/',
+    '-Djna.tmpdir=%s' % jna_tmpdir,
```
BenWhitehead: Why do we need the `SERVER_JVMFLAGS` above if we're setting the value here? Should we just have a single environment variable for the value and then use it everywhere?

gpaul: `SERVER_JVMFLAGS` is zookeeper-specific. I think we should keep that in the service file as the reader can immediately grep for that string and understand that this gets read by the zookeeper start script.

`-Djna.tmpdir` is specific to exhibitor.

It seems clear that we need to specify both values separately. Whether we need to extract the path into a separate variable is another matter. I don't think it would add much. I'm a "refactor on third use" kind of guy.

packages/exhibitor/extra/dcos-exhibitor.service
gpaul: I'm would be happy to do that. Have a quick look at https://github.com/dcos/dcos/blob/master/gen/dcos-config.yaml#L747. 

There are several options for moving this env var to that section, none of them are clean. The cleanest is probably extracting 

```
  - path: /etc/exhibitor
    content: |
```
from the start of the individual cases:

```
  - path: /etc/exhibitor
    content: |
      SERVER_JVMFLAGS=...
{% switch exhibitor_storage_backend %}
{% case "agent_only_group_no_exhibitor" %}
...
{% endswitch %}
```
But that doesn't quite work as the  `agent_only_group_no_exhibitor` case is mixed in there, too.

I don't see any way to add this env var to that file that doesn't involve copy-pasting it to all the different cases.

In general I would also prefer setting it in this file, but that file is being generated in a weird way.

What do you think?

BenWhitehead: If it's prohibitively difficult to move into the env file it's probably okay here, though I think it'd be good to get an opinion of someone with more experience with the systemd unit files and this sort of usage. Perhaps @lingmann can provide an opinion?

packages/exhibitor/extra/start_exhibitor.py
BenWhitehead: I should have worded my question differently.

ZK will automatically look for that environment variable when it's start script runs, however the value of that variable is being set to the exact same value we're using for exhibitor. 

I could envision something like this in the unit or environment file
```bash
JNA_TMP_DIR=/var/lib/dcos/exhibitor/tmp
SERVER_JVMFLAGS="-Djna.tmpdir=$JNA_TMP_DIR"
```
Then in `start_exhibitor.py`
```python
'-Djna.tmpdir=%s' % os.environ['JNA_TMP_DIR']
```

Doing it this way keeps all the grep points, with a single definition that if (when?) has to be updated touches everything.

gpaul: @BenWhitehead Ah! I follow. That's a great suggestion, I'll update the PR.

gpaul: Sadly systemd doesn't support this. the `Environment=` directive does not perform environment variable substitution.

BenWhitehead: Ah bummer, sorry for accidentally sending you on a wild goose chase.

gpaul: I learned a bunch of new stuff :+1:

adam-mesos:
Merged in train 128 - #1225

PR #1205, opened 2017-02-02
Refactor AWS Zen stack resource model

mellenburg:
- Adds fetch stack helper that can automatically return the correct AWS Stack representation
- Remove most boto/aws - level logic and concentrate it entirely inside test_util/aws.py
- Refactors AwsCfAdvanced into AwsZenCfStack and turn each of the resource stacks within zen into its own class
- Note: all of the @classmethod create's in this file are essentially bits of dcos-launch and will be removed when we switch infrastructure to use dcos-launch

## Related Issues

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: PR is a refactor, nothing added and nothing removed. Just reworked internals and removed dead code

### DIFF
--- a/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
+++ b/packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
@@ -1,52 +1,74 @@
+import logging
 import os
-from functools import partial
 
 import pytest
+import retrying
 
 import test_util.aws
 import test_util.helpers
 from test_util.helpers import retry_boto_rate_limits
 
+log = logging.getLogger(__name__)
+
 ENV_FLAG = 'ENABLE_RESILIENCY_TESTING'
 
-pytestmark = pytest.mark.skipif(
+skip_if_resiliency_not_enabled = pytest.mark.skipif(
     ENV_FLAG not in os.environ or os.environ[ENV_FLAG] != 'true',
     reason='Must explicitly enable resiliency testing with {}'.format(ENV_FLAG))
 
+skip_if_not_aws_stack = pytest.mark.skipif(
+    'AWS_STACK_NAME' not in os.environ,
+    reason='Must use an AWS Cloudformation run this test!')
+
+pytestmark = [skip_if_resiliency_not_enabled, skip_if_not_aws_stack]
+
 
 @pytest.fixture(scope='session')
-def dcos_launchpad(dcos_api_session):
-    """Interface for direct integration to dcos_launchpad hardware
-    Currently only supports AWS CF with AWS VPC coming soon
-    """
-    if 'AWS_STACK_NAME' not in os.environ:
-        # TODO(mellenburg): update when advanced templates are merged
-        pytest.skip('Must use a AWS Cloudformation to run test')
-    stack_name = os.environ['AWS_STACK_NAME']
+def boto_wrapper(dcos_api_session):
     aws_region = os.environ['AWS_REGION']
     aws_access_key_id = os.environ['AWS_ACCESS_KEY_ID']
     aws_secret_access_key = os.environ['AWS_SECRET_ACCESS_KEY']
-    bw = test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
-    return test_util.aws.DcosCfSimple(stack_name, bw)
+    return test_util.aws.BotoWrapper(aws_region, aws_access_key_id, aws_secret_access_key)
+
+
+@pytest.fixture(scope='session')
+def dcos_stack(boto_wrapper):
+    """ Works with either Zen or Simple cloud formations
+    """
+    stack = test_util.aws.fetch_stack(os.environ['AWS_STACK_NAME'], boto_wrapper)
+    if isinstance(stack, test_util.aws.VpcCfStack):
+        pytest.skip('Onprem Vpc not currently supported')
+    return stack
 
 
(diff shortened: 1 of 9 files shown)


packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -59,15 +78,18 @@ def get_running_agents(group_name):
...
+        len(dcos_api_session.slaves),
+        600)
 
     # Reset the dcos_api_session to have the replacement agents
-    dcos_api_session.slaves = sorted([agent.private_ip_address for agent in
-                                      get_running_agents('SlaveServerGroup')])
-    dcos_api_session.public_slaves = sorted([agent.private_ip_address for agent in
-                                             get_running_agents('PublicSlaveServerGroup')])
+    dcos_api_session.slaves = get_private_ips(dcos_stack.private_agent_instances)
+    dcos_api_session.public_slaves = get_private_ips(dcos_stack.public_agent_instances)
```
amitaekbote: s/public_slaves/public_agents

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -59,15 +78,18 @@ def get_running_agents(group_name):
...
+        partial(list, dcos_stack.private_agent_instances),
+        len(dcos_api_session.slaves),
+        600)
 
     # Reset the dcos_api_session to have the replacement agents
-    dcos_api_session.slaves = sorted([agent.private_ip_address for agent in
-                                      get_running_agents('SlaveServerGroup')])
-    dcos_api_session.public_slaves = sorted([agent.private_ip_address for agent in
-                                             get_running_agents('PublicSlaveServerGroup')])
+    dcos_api_session.slaves = get_private_ips(dcos_stack.private_agent_instances)
```
amitaekbote: s/slaves/agents

mellenburg: That's touching dcos_api_session and outside of the scope of this PR in my opinion. If you change this, then you need to change  a few dozen other things throughout the codebase. I figured I would clean that up in the course of doing https://mesosphere.atlassian.net/browse/DCOS-13629 but I would also be fine if you wanted to file a JIRA for resolving that independently (the string change on its own should be easy, just tedious to find everywhere its used)

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
mellenburg: See above. Will not change in this PR as it will add another 30 files or so of diff which would be a lot of noise.

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
amitaekbote: Sure if you do it as part of another clean up that would work, I will approve this.

mellenburg:
Hmm, the tests on this keep flaking out because a public agent group doesnt come up. I.E. public instance never sends the mesos up singal and the template deployment times out.  S3 is probably having a bad day and is killing dcos-download.service

packages/dcos-integration-test/extra/resiliency/test_aws_cf.py
```diff
@@ -1,52 +1,74 @@
...
 
 
 @pytest.mark.last
-def test_agent_failure(dcos_launchpad, dcos_api_session, vip_apps):
-    # make sure the app works before starting
+def test_agent_failure(dcos_stack, boto_wrapper, dcos_api_session, vip_apps):
+    # Accessing AWS Resource objects will trigger a client describe call.
+    # As such, any method that touches AWS APIs must be wrapped to avoid
+    # CI collapse when rate limits are inevitably reached
+    @retry_boto_rate_limits
```
orsenthil: Do we really need `retry` for these instance id lookup methods?

I see that retry is properly handled in places where there is directly interaction with aws api.

```
    @retry_boto_rate_limits
    def get_auto_scaling_instances(self, asg_physical_resource_id):
```

And IMO, that should be enough.

* Also, another note. It may/may not be applicable here. If we are doing a retry on a function that `yields`, essentially that's is not a retry, that is going to the next member of the container.

*  Here, and in many of the methods that we are doing `retry_boto_rate_limits` on "instance values" that have already been fetched and constructed. This is difficult to reason/grasp.  🔴

mellenburg: Sadly, yes, we do need to have this. This  statement is incorrect:

> Here, and in many of the methods that we are doing retry_boto_rate_limits on "instance values" that have already been fetched and constructed.

With the boto3 API, a ec2 Resource is just a physical resource ID bound to a client callback and a filter. I.E. all of the properties are lazy. Thus, its not actually until you go and access instance_id that boto goes and does the client call to describe instances.

What this means is that any access of a boto variable should be wrapped with the rate limiting retry

mellenburg: Also, unless I misread my implementation, I shouldnt be having that issue with yield. Rather, it will retry to walk all the way through the iterator without encountering an error.

orsenthil: Yeah, yield is not a problem here.

Thanks for the explanation.

PR #1208, opened 2017-02-03
Add log-level control for dcos-launch

mellenburg:
- suppresses select libraries to prevent obfuscation
- adds truncated options for existing long-name options
- moves CLI code to its own module

### DIFF
--- a/ssh/tunnel.py
+++ b/ssh/tunnel.py
@@ -16,8 +16,6 @@ from typing import Optional
 
 from pkgpanda.util import write_string
 
-LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
-logging.basicConfig(format=LOGGING_FORMAT, level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
 
--- a/gen/build_deploy/bash/dcos-launch.spec
+++ b/gen/build_deploy/bash/dcos-launch.spec
@@ -4,7 +4,7 @@
 #   - data must be decalared explicitly if not a .py file
 #   - Building will suck up the local SSL .so and package it
 #     with the final exe. Ensure build system has OpenSSL 1.0.2g or greater
-a = Analysis(['test_util/launch.py'],
+a = Analysis(['test_util/launch_cli.py'],
              hiddenimports=['html.parser'],
              datas=[('gen/ip-detect/*.sh', 'gen/ip-detect/')])
 pyz = PYZ(a.pure, a.zipped_data, cipher=None)
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
             'pkgpanda=pkgpanda.cli:main',
             'mkpanda=pkgpanda.build.cli:main',
             'dcos_installer=dcos_installer.cli:main',
-            'dcos-launch=test_util.launch:main',
+            'dcos-launch=test_util.launch_cli:main',
             'dcos-exhibitor-migrate-status=dcos_installer.exhibitor_migrate:status',
             'dcos-exhibitor-migrate-perform=dcos_installer.exhibitor_migrate:perform',
         ],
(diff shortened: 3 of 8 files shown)


mellenburg:
@orsenthil @amitaekbote PTAL

mellenburg:
@lingmann This PR is mostly limited to moving code to smaller files and allowing better debugging visibility when using dcos-launch PTAL

ssh/tunnel.py
```diff
@@ -16,8 +16,6 @@
 
 from pkgpanda.util import write_string
 
-LOGGING_FORMAT = '[%(asctime)s|%(name)s|%(levelname)s]: %(message)s'
```
orsenthil: If there is code which is invoking these modules `tunnel.py`, `test_util/aws.py` **outside** of dcos-launch, then we are effectively loosing the formatting now. Isn't it?  

Is it good idea to have a common formatted logger?

mellenburg: Yes. It is a very good idea to have a common log configuration

The only files that should be setting the logging format are ones that:
- Really specifically need their own format (I cannot see where this would happen curently)
- Files that format and setup the logging for the current session.

Furthermore, when a module chooses to format logging, it should do so at runtime, because when you do it at import time like this, you have no idea if this format will actually be used anywhere. This module specifically gets its logger from logging.getLogger() so its clear that it should inherit the format from whoever is using it.

So, this PR doesnt force user to have a common logger. Rather it allows a user to have a modular logger where settings can be inherited and configured dynamically or per module. Having a common logger would be adding a new file to this module called 'logging.py' which defines how logging should be setup globally and then every file then imports it.

mellenburg: Also, as to the point of callers losing their log format. The only callers of this code are test_*.py files in this folder, dcos-launch, and integration tests. In integration tests, logging format is handled by conftest.py, the test_*.py scripts have their own format explicitly set, and this PR does that for dcos-launch

orsenthil: Ok. I see the rationale. 

Yup, having it in dcos-launch, and other * invokers* is good for this change.

There is one advantage is having a logger at a single place where the modules (even the clients/invokers) can include. 

* Enforcing consistency in the log format.  
* Changing backend. for .e.g we want to publish all logs to a single place. If every client imported a common logger, only that will need change.

mellenburg: Changing the backend can actually be done from the invoker. Python logging handlers are essentially that (e.g. adding a file handler and disabling a stream handler can be done after import before everything runs). Enforcing consistency might be necessary one day, but for now I like the simplicity of calling 'log = logging.getLogger(__name__)'

orsenthil: If we have 5 invokers, we  don't want to add this backend in everyone of them. It might helpful to just provide the "logging infrastructure" that new modules or clients can just use.


I think, my bias is coming from seeing this 'pattern' in many frameworks
For e.g sqlalchemy goes a step further and defines logging configuration classes:

https://github.com/zzzeek/sqlalchemy/blob/64e72b58fd74db0c5c87407c824d08ad699e5e8e/lib/sqlalchemy/log.py#L51-L51


* This PR is an improvement, so +1 to that. Keeping the common logger in mind might be a good idea to implement when we feel the need for it.

mellenburg: You have silently glossed over the core problem :stuck_out_tongue: 

> if we have 5 invokers, ...

We should not have five different scripts that have totally custom ways of doing the exact same things. All of the test_* files will be purged soon (except upgrade_vpc).

Why we do not want to go to that pattern right now also has something to do with the fact that we do not have a consistent module in test_util, rather we have two core functions: provisioning (launch.py, aws.py, azure.py, etc) and API harness for a provisioned DC/OS (dcos_api_session.py, helpers.py, marathon.py). Each of those cases has distinct needs. Once these live in more self-centered environment, the centralized logging would be a good goal

PR #1214, opened 2017-02-07
DCOS Launch Configs

mellenburg:
## High Level Description
The current configuration used in dcos-launch was the simplest for reaching the most basic use case, but now we need to define one config format that will cover the following cases:
- Simple AWS
- Advanced AWS
- Onprem on AWS
- Azure Template
- AWS bare linux cluster
- Onprem from previously launched bare linux cluster
Considered but not implemented:
- GCE bare linux cluster
- Onprem on GCE

## Related Issues
https://jira.mesosphere.com/browse/DCOS-13683
https://mesosphere.atlassian.net/browse/DCOS-13300
https://mesosphere.atlassian.net/browse/DCOS-13299

## Checklist for all PR's


test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+- [Onprem Install on AWS Linux Cluster](aws-onprem.yaml)
+- [Onprem Install on Previously Provisioned Linux Cluster](onprem-from-info.yaml)
+- [Azure Template Deployment](azure.yaml)
+
+## Keywords and Definitions
+
+### Required for Deployment
+* `deployment_type`: Can be one of `template`, `onprem`, or `linux-cluster`.
+  * `template` refers to either an AWS cloudformation template or an Azure Resource Manager deployment template. Use of this options will require setting `deployment_name`, `template_url`, `provider`, and `provider_info`
+  * `linux_cluster` will used hard-coded templates to deploy a cluster of linux machines in a shared network on the provider of choice (currently only supports AWS). `deployment_name` and `provider` must be set. See `linux_cluster` below for options. Only `aws` provider is currently supported for this deploy option
```
BenWhitehead: Is this what would be set to install in a VPC? or is that `onprem`?

mellenburg: This is what would be set to deploy a plain VPC or to deploy onprem within a VPC. The differnce would be in 'deployment_type' which would be 'linux_cluster' for the plain vpc and 'onprem' in the onprem-on-vpc case

branden: Should `linux` be in the name of this setting? Looks like we could use this to deploy a VPC from an AMI installed with any OS, as long as there was one defined. Mesos has experimental support for Windows, so who knows.

mellenburg: I like `linux` as its a bit cleaner, but when you fully write it out, you get:
{code}
---
deployment_type: onprem
provider: aws
linux:
   instance_type: t2.micro
   os_name: cent-os-7
{code}

So then it gets awkward that you have linux-> instance_type because instances don't really exist in the linux namespace.

IMO the real issue here is I just can't think of a good word to describe "a group of machines, all of which have the same OS installed and are provided by some cloud service". `cloud_cluster` ? And then maybe something like:
{code}
---
deployment_type: cloud_cluster
provider: aws
cloud_cluster:
  linux: cent-os-7
  host_count: 4
```
which would be nice because then we have the freedom to put ```ami:``` under there or even ```windows: ```

What do you think?

Comment (branden) [inline test_util/launch_configs/README.md]: Yeah a great name doesn't jump out at me either. `cloud_cluster` works for me. Why not stick with `os_name` for the OS name in that last example?

Comment (mellenburg) [inline test_util/launch_configs/README.md]: No good reason now that I think about it. Sticking with ```os_name:``` sounds good (and we can still have that be optionally swapped with ```image_id:``` if we later decide that updating the internal OS -> cloud image mapping to be too restrictive

File: test_util/launch_configs/aws-cf.yaml [hunk 2]
```diff
@@ -0,0 +1,18 @@
+---
+deployment_type: template
+deployment_name: This-is-a-test-stack
+template_url: https://s3.amazonaws.com/downloads.dcos.io/dcos/testing/master/cloudformation/single-master.cloudformation.json
+provider: aws
+provider_info:
+    region: us-west-2
+    access_key_id: ALSAFLAJ4L20922KF20TL294
+    secret_access_key: ADWOFIAOFWJTPW3924JR
+key_helper: true
```
orsenthil: Since `key_helper` is provider specific information, I would assume it will be nested in the provider_info.

test_util/launch_configs/aws-linux-cluster.yaml
```diff
@@ -0,0 +1,13 @@
+---
+deployment_name: dcos-stack-that-was-provisioned-from-vpc
+deployment_type: linux_cluster
+provider: aws
+provider_info:
+    region: us-west-2
+    access_key_id: ALSAFLAJ4L20922KF20TL294
+    secret_access_key: ADWOFIAOFWJTPW3924JR
+key_helper: true
+linux_cluster:
```
orsenthil: If I have 5 t2.micro cent-os-7 provisioned in my AWS account How will this pick 3 out of 5?

The confusing part for me is `dcos-stack-that-was-provisioned-from-vpc` vs providing the details on `os-name`, `instance-type` and `instance_count` below.

Also, what is assumed of the ssh user and password that is required to login to the cluster?

test_util/launch_configs/aws-zen-cf.yaml
```diff
@@ -0,0 +1,19 @@
...
+deployment_type: template
+deployment_name: This-is-a-test-stack
+template_url: https://s3.amazonaws.com/downloads.dcos.io/dcos/testing/master/cloudformation/coreos-zen-1.json
+provider: aws
+provider_info:
+    region: us-west-2
+    access_key_id: ALSAFLAJ4L20922KF20TL294
+    secret_access_key: ADWOFIAOFWJTPW3924JR
+key_helper: true
+zen_helper: true
```
mellenburg: I was putting it at top-level because its a dcos-launch option, and not a native provider option, but at the end of the day, this is a provider API wrapper, so your argument seems good to me

test_util/launch_configs/aws-zen-cf.yaml
```diff
@@ -0,0 +1,19 @@
...
+dcos_parameters:
+    AdminLocation: 0.0.0.0/0
+    PublicAgentInstanceCount: 2
+    PrivateAgentInstanceCount: 1
+    PublicAgentInstanceType: m4.xlarge
+ssh:
+    user: core
+    private_key_path: /tmp/aawpei3
```
mellenburg: But if you call it `private_key` that implies that this could be the key contents themselves, right?

How about `private_key_file_path` ?

test_util/launch_configs/aws-linux-cluster.yaml
```diff
@@ -0,0 +1,13 @@
+---
+deployment_name: dcos-stack-that-was-provisioned-from-vpc
```
mellenburg: This was an accidentally copy-pasted name; it should be `linux-cluster-made-by-dcos-launch` or something. This is *not* a configuration for dcos

orsenthil:
Here is my high level comment.

I was trying to visualize the deployment type and provider and tried to do the combinations. Please see my questions  based on how I was thinking.

```
In [1]: deployment_types = ["template", "onprem", "linux-cluster"]

# Without any specifics, how do we differentiate onprem vs linux-cluster? 
# Is Linux-cluster a cloud-cluster that is not template? 
# Is onprem is what mesosphere defines a group of resources?

In [2]: provider = ["aws", "azure", "existing-cluster"]

# I am assuming existing-cluster can be anywhere, including aws, azure.

In [4]: for p in provider:
   ...:     for d in deployment_types:
   ...:         print "%s-%s" % (p, d)
   ...:
aws-template
aws-onprem
aws-linux-cluster
azure-template
azure-onprem
azure-linux-cluster
existing-cluster-template
existing-cluster-onprem
existing-cluster-linux-cluster
```

How do we differentiate between aws-onprem vs aws-linux-cluster *in abstract* ? (Same with azure-onprem vs azure-linux-cluster) 

*  Those feel same to me.
* Does that mean onprem can be reduced to a sub-type of "linux-cluster" ?

Also, which of the configurations above do not make sense?

* `existing-cluster-template`?

test_util/launch_configs/aws-linux-cluster.yaml
mellenburg: I think you might be mis-interpreting what this is specifying. This is going to *create* those instances in that count. There is no picking involved.

As I just commented, the deployment name is wrong in this example. This example is for creating a plain linux cluster of homogeneous instances

the field os_name is dcos/dcos/test_util/aws specific. We have a mapping from strings to region to AMI. For each of those OS-strings, we also know the default user and home dir. There is no reason for the user to provide information for hard-coded AMIs. No passwords are ever used anywhere (key only) so that is irrelevant.

mellenburg:
@orsenthil:

> Is onprem is what mesosphere defines a group of resources?

onprem == on-prem == on premise == arbitrary hardware that does not leverage a cloud provisioning API for bulk host operations

Thus, onprem refers to provisioning (i.e. deploying) DC/OS by installing via bash/docker (i.e. the installer) and nothing else.

linux_cluster (to be renamed cloud_cluster) is a baked in template for deploying a bunch of cloud hosts to a given provider. We currently only have this for AWS. The purpose of this is to allow the user to say '6 CentOS-7 hosts in AWS' and nothing else. If someone really wants to deploy their own customized linux cluster through a template, they can easily use this tool with a custom template with aws-template

Re: Your permutations
```
# Currently supportable combinations:
aws-template
aws-linux-cluster
aws-onprem # this is two steps, 1. aws-linux-cluster 2. existing-cluster-onprem
azure-template
existing-cluster-onprem
# Possible but unsupported
azure-linux-cluster  # i do not have offhand tools for creating and interacting with a homogeneous group of azure instances
azure-onprem  # I need azure-linux-cluster to do this
# Do not make any sense:
existing-cluster-template   # you cannot make a template deploy from a cluster. That is nonsensical
existing-cluster-linux-cluster # this would be making a cluster from ... a cluster? again, nonsensical
```

BenWhitehead:
:+1: to renaming `linux_cluster` to `cloud_cluster`

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+  * `onprem` refers to using the DC/OS installer (`dcos_generate_config.sh`) to deploy DC/OS onto a set of provided hosts. Those hosts can either be provisioned with keywords or by having previously created a `cloud_cluster` with dcos-launch. `installer_url` must be set as must `dcos_config` (see below)
+    * If `provider` is set to `cloud_cluster`, then the only other required field is `provider_info_path`, which points towards the cluster info JSON created by dcos-launch for which you want to install DC/OS
+    * If `provider` is not set to `cloud_cluster`, dcos-launch will use that provider to create the linux cluster and then immediately start deploying DC/OS to it (currently only supports `aws` provider). The linux cluster can be configured in the same manner as deploying a stand alone linux cluster (see `cloud_cluster` below), however the `instance_count` is omitted as it will be implied from `num_masters`, `num_public_agents`, and `num_private_agents`
```
lingmann: Can the two `provider` sections above be reworked or moved to underneath the provider section below? It was pretty confusing to me to show up here after the onprem provider details.

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+* `deployment_name`: The high-level name assigned in the provider for this resource allocation. In AWS, this is equivalent to stack name. In Azure, this is the resource group name, etc
+* `provider`: Can be `aws`, `cloud_cluster`, or `azure`
+  * If `aws` is set, then `provider_info` must be supplied with subfields for `region`, `access_key_id`, and `secret_access_key`
+  * If `azure` is set, then `provider_info` must be supplied with subfields for `location`, `tenant_id`, `subscription_id`, `client_id`, and `client_secret`
+  * If `cloud_cluster` is set, then that means you wish to use a pre-existing linux-cluster deployment that you previously launched. In this case, you do not specific a `provider_info` and instead use a `provider_info_path` that points to the cluster info JSON supplied by DC/OS Launch to provision the linux cluster
+
+### Potentially Optional Fields
```
lingmann: s/Potentially Optional/Optional/?

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+* `ssh`: In deployment types `template` and `cloud_cluster` this section can be entirely ommitted and the deployment will still occur. However, use of the pytest command will not be possible (this is irrelevant for the case of `cloud_cluster` where there is no purpose for pytest). In `onprem` deployment type, ssh capability is required to use the installer.
+  * `user`: User that will be able to SSH into the cluster and run test. Cannot be set with `deployment_type: cloud_cluster` as it will be inferred from the baked-in linux cluster template.
+  * `private_key_path`: path to the private key that will allow the above user to SSH. Cannot be set if `key_helper` is true (see below). Must be set with `deployment_type: onprem`
```
lingmann: How about flattening this to `ssh_user` and `ssh_private_key_path` at the top level? I think keeping the config layers shallow makes for an easier to grok interface.

test_util/launch_configs/onprem-from-info.yaml
```diff
@@ -0,0 +1,15 @@
+---
+# no deployment_name provided because it was set in the cloud_cluster
+deployment_type: onprem
+installer_url: https://downloads.dcos.io/dcos/testing/master/dcos_generate_config.sh
+provider: cloud_cluster
+provider_info_json_path: /foo/bar/important/path
```
lingmann: This is currently referenced as `provider_info_path` in the docs. Update either here or there to be consistent.

test_util/launch_configs/onprem-from-info.yaml
```diff
@@ -0,0 +1,15 @@
...
+dcos_config:
+    num_private_agents: 2
+    num_public_agents: 1
+    num_masters: 3
+    cluster_name: My Awesome DC/OS
+    resolvers:
+        - 8.8.4.4
+        - 8.8.8.8
+    security: strict
```
lingmann: Instead of adding new options like `num_masters` mixed in with standard config.yaml, is there any way we could have this section map 1 to 1 with our existing installer config format and just make some incompatible options disallowed and infer the rest from the provider info or via configuration at an earlier provision stage? I'm concerned that this is going to get really confusing, since the existing config.yaml format is already very complicated.

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+- [Onprem Install on AWS Linux Cluster](aws-onprem.yaml)
+- [Onprem Install on Previously Provisioned Linux Cluster](onprem-from-info.yaml)
+- [Azure Template Deployment](azure.yaml)
+
+## Keywords and Definitions
+
+### Required for Deployment
+* `deployment_type`: Can be one of `template`, `onprem`, or `cloud_cluster`.
+  * `template` refers to either an AWS cloudformation template or an Azure Resource Manager deployment template. Use of this options will require setting `deployment_name`, `template_url`, `provider`, and `provider_info`
+  * `cloud_cluster` will used hard-coded templates to deploy a cluster of machines in a shared network on the provider of choice (currently only supports AWS). `deployment_name` and `provider` must be set. See `cloud_cluster` below for options. Only `aws` provider is currently supported for this deploy option
```
branden: I'll toss another suggestion onto the pile: `bare_cluster`?

mellenburg: I feel any term we use should probably include the term `cluster` If I see `infra_pre_dcos`, `infra_only`, or `infra_without_dcos` I don't think I would naturally make the leap to seeing those as cluster provisioned in the cloud. I second `bare_cluster`

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Design Intention
+The intention of this configuration file is to provide an interface by which
+all deployments of DC/OS, regardless of provider have a similar format, thus
+complementing the goal of dcos-launch to provide a single tool for launching
+across a variety of provider APIs
+
+## Supported Deployments and Examples
+- [Simple AWS Cloudformation](aws-cf.yaml)
+- [Zen AWS Cloudformation](aws-zen-cf.yaml)
+- [Non-DC/OS AWS Linux Cluster](aws-linux-cluster.yaml)
```
mellenburg: This is what is currently called a "VPC" by CCM. I.E. a bunch of hosts with some standard distro loaded that has been provided by AWS

test_util/launch_configs/onprem-from-info.yaml
mellenburg: So, this gets a little tricky. You say:

> Instead of adding new options like `num_masters`

This is the opposite of the situation. In the *real* config.yaml, num_masters is a real term. Rather, its the only the *installer* config that requires a `master_list` which is then converted into `num_masters` in a "real" config.yaml

On the other hand, the `num_agents` and `num_public_agents` come from `agent_list` and `public_agent_list` which are *installer only* names. The "real" config.yaml doesnt care about those counts and they are just used to deploy

Finally, you have the other odd connectors:
- ip-detect. Passed to the installer via magic file path (genconf/ip-detect). Installer injects this magic file path (because it is closer in line to the legacy '--genconf' instructions). Thus, using 'detect_ip_filename' is something we can directly pass to config.yaml with impunity
- ssh_key. Passed via another magic path (genconf/ssh_key). This is an installer-only term and is already covered else where in dcos-launch config doc
- rexray: Passed via yet another magic path that is injected into config.yaml. Thus, we can just use native calc.py's 'rexray_config' which reads in a file path and coverts to contents
- detect-ip-public: The only hook we have for this is to pass the direct contents of the file. The installer does this by, guess what, another magic path. I would suggest we add to gen a detect_ip_public_filename as to maintain the same model as the regular detect-ip

So, what I am getting at here is that this would be a config file format that is less magic than the installer (which again, you need to thoroughly read through the docs to find implicit settings, rather than explicitly declaring in a config file). Only for a the agent lists does this breakdown because agent lists are not something the config cares about.

Thus, I think the question should be, at which level do we declare the agent list counts without confusing the user

test_util/launch_configs/aws-cf.yaml
```diff
@@ -0,0 +1,17 @@
...
+deployment_name: This-is-a-test-stack
+deployment_type: template
+template_url: https://s3.amazonaws.com/downloads.dcos.io/dcos/testing/master/cloudformation/single-master.cloudformation.json
+provider: aws
+provider_info:
+    region: us-west-2
+    access_key_id: ALSAFLAJ4L20922KF20TL294
+    secret_access_key: ADWOFIAOFWJTPW3924JR
+    key_helper: true
+dcos_parameters:
```
mellenburg: template_parameters

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,46 @@
+# DC/OS Launch Configuration YAML
+## Design Intention
```
mellenburg: There will be a required versioning field to explicitly link to the configuration schema required in the given version of launch

test_util/launch_configs/README.md
```diff
@@ -0,0 +1,46 @@
...
+* `bare_cluster`: Used with `deployment_type: bare_cluster` or `onprem` while provider is not `bare_cluster`
+  * `os_name`: see [AWS AMI-OS map](../aws.py)
+  * `instance_type`: provider-specific string describing what type of hardware to request
+  * `instance_count`: number of instances to provision. This field is only used if `deployment_type` is not `onprem`. In that case, there are `num_masters`, `num_private_agents`, and `num_public_agents` fields which will implicitly determine `instance_count`
+* `dcos_parameters`: Used with `deployment_type: template` . Set as a nested key-value mapping describing which parameters to pass to the template pointed to with `template_url`. Not necessarily required as some templates have enough default settings to be launched without parameters
+* `dcos_config`: Used with `deployment_type: onprem`. This field will be the config to be passed to the onprem installer. There are a few difference between specifying a config here and one for the installer
```
mellenburg: dcos_config should only include gen-native terms. use 'dcos_installer_config' for the installer specific bits

flake8_dcos_lint/setup.py
```diff
@@ -18,7 +18,7 @@
     install_requires=[
         'pycodestyle==2.2.0',
         'flake8',
-        'flake8-import-order==0.9.2',
```
mellenburg: I did this while trying to fix some other naming errors. Any objections to leaving it?

launch/config.py
```diff
@@ -0,0 +1,385 @@
...
+    user_source = get_user_argument_source(user_config)
+    t = get_target()
+    sources = [get_source(), user_source]
+    resolver = resolve_configuration(sources, [t])
+    status = resolver.status_dict
+
+    if status['status'] == 'errors':
+        raise launch.util.LauncherError('ValidationError', pretty_print_validate_error(status))
+
+    late_variables = dict()
```
mellenburg: @branden I copied this for completeness of validation, but I think I should just ban all late usage for this module. What do you think?

launch/config.py
```diff
@@ -0,0 +1,385 @@
...
+
+
+def get_validated_config(config_path: str):
+    """ Takes a dcos-launch configuration file and:
+    - string-ifies nested dicts
+    - expands paths into absolute paths, supporting tilde expansion
+    - converts provider_info_filename into provider_info
+    """
+    config_dir = os.path.dirname(os.path.abspath(config_path))
+    # Try and catch YamlParseError
```
mellenburg: Will add for the express purpose of providing a cleaner message when a garbled config is provided

launch/config.py
```diff
@@ -0,0 +1,385 @@
...
+
+def expand_path(path: str, config_dir: str):
+    """ Returns absolute paths by performing ~ expansion or prepending the
+    config_dir to make an absolute path
+    """
+    if path.startswith('/'):
+        # nothing to do, return the path
+        return path
+    if path.startswith('~'):
+        # Try this and raise LauncherError
```
mellenburg: Note: add this catch

mellenburg:
Most of the changes in this PR are moving code and adding sample/documentation content. For reviewing, please focus your attention on the following changes:
- adding a new, minor input to gen: https://github.com/dcos/dcos/pull/1214/files#diff-3cd4ddcecc89d0887dd9cb3d5fabe995
- adding the config validation: https://github.com/dcos/dcos/pull/1214/files#diff-f9a1c04475b0653bde7eec04c87148dc
- the new tests for the config (best way to start playing around with this code) https://github.com/dcos/dcos/pull/1214/files#diff-a784ca0e06a4ba7a8dea729e19323e25

launch/config.py
branden: We should factor out the chunk of `gen.generate()` that's copied here and call it rather than copy it.

launch/config.py
```diff
@@ -0,0 +1,385 @@
...
+                # Validate a late variable should only have one source.
+                assert setter.name not in late_variables
+
+                late_variables[setter.name] = setter.late_expression
+
+    final_args = {k: v.value for k, v in resolver.arguments.items() if v.is_finalized}
+    unrecognized_args = set(user_config.keys()) - set(final_args.keys())
+    if len(unrecognized_args) > 0:
+        raise launch.util.LauncherError(
+            'ValidationError', 'Unrecognized/incompatible arguments: {}'.format(unrecognized_args))
```
branden: That gen silently accepts unrecognized parameters is IMO a bug, and should be fixed there. It's caused some user confusion in the past, though I can't find a ticket for it. IIRC gen used to error when it saw an unrecognized parameter, but at some point that was broken and never got fixed.

mellenburg: Yeah, I was also very surprised to see that this wasn't caught, but then I saw this:
https://github.com/dcos/dcos/blob/master/gen/internals.py#L773

So JIRA that or fix it now? I lean towards JIRA when this PR is so huge and that is a gen change

branden: Sounds like that'll take some work, bummer. Yeah if you could file a ticket and link it here in a comment, that'll help surface the problem.

flake8_dcos_lint/setup.py
amitaekbote: as long as it doesnt break things in new fun ways i am ok with this being here

gen/calc.py
```diff
@@ -172,7 +172,11 @@ def calculate_ip_detect_contents(ip_detect_filename):
     return yaml.dump(open(ip_detect_filename, encoding='utf-8').read())
 
 
-def calculate_ip_detect_public_contents(ip_detect_contents):
+def calculate_ip_detect_public_contents(ip_detect_contents, ip_detect_public_filename):
+    if ip_detect_public_filename != '':
```
amitaekbote: The function on line 170 does the exact same thing, use that?

launch/__init__.py
```diff
@@ -0,0 +1,14 @@
+import launch.aws
+import test_util.aws
+
+
+def get_launcher(config):
+    """Returns a launcher given a launcher type (string) and a dictionary of
+    appropriate provier_info
```
amitaekbote: typo, s/provier/provider

gen/calc.py
mellenburg: The function on line 170 does *not* do the same thing. These are calc functions and so their arguments become gen.internals.Target's . Thus, this allows you to provide 'ip_detect_public_filename' which was not the case before

mellenburg:
@branden @amitaekbote Comments addressed PTAL

gen/__init__.py
```diff
@@ -524,20 +524,9 @@ def build_late_package(late_files, config_id, provider):
...
-        extra_sources=list(),
-        extra_targets=list()):
-    # To maintain the old API where we passed arguments rather than the new name.
-    user_arguments = arguments
-    arguments = None
-
-    sources, targets, templates = get_dcosconfig_source_target_and_templates(
-        user_arguments, extra_templates, extra_sources)
-
+def validate_and_get_arguments(sources, targets):
```
branden: This function is doing a few things (resolving/validating config, collecting argument dict, collecting late variables) that seem like they could be split up. Collecting late variables is a clear candidate for a separate function -- that this function returns two unrelated values is awkward.

launch/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+def get_launcher(config):
+    """Returns the correct class of launcher from a validated launch config dict
+    """
+    platform = config['platform']
+    if platform == 'aws':
+        boto_wrapper = test_util.aws.BotoWrapper(
+            config['aws_region'], config['aws_access_key_id'], config['aws_secret_access_key'])
+        return launch.aws.AwsCloudformationLauncher(boto_wrapper)
+    raise launch.util.LauncherError('UnsupportedAction', 'Launch platform not supported: {}'.format(platform))
```
mellenburg: This was just for keeping life simple while implementing the large swathes of code. Now that a lot more of this has been implemented, in the next iteration we should definitely add a launch.exceptions with 'UnsupportedActionError' and 'ValidationError' etc

launch/config.py
```diff
@@ -0,0 +1,393 @@
...
+    if path.startswith('~'):
+        try:
+            home_dir = os.environ['HOME']
+        except KeyError as ex:
+            raise launch.util.LauncherError(
+                'MissingInput', 'HOME must be set in the current environment '
+                'in order to use "~" expansion in config files')
+        return path.join(home_dir, path[1:])
+    # path is relative to config_dir
+    return os.path.join(config_dir, path)
```
mellenburg: Thanks! I did not know about expanduser.

launch/config.py
```diff
@@ -0,0 +1,393 @@
...
+    config['ssh_user'] = ssh_user
+    return yaml.dump(config)
+
+
+def validate_provider_platform(provider, platform):
+    """ TODO: Remove conditional logic here by placing this function in a
+    conditional entry. At present, all validates are global. See: DCOS-14033
+    """
+    if platform in ('aws', 'azure'):
+        assert platform != 'bare_cluster', 'bare_cluster only supports "provider: onprem"'
```
mellenburg: onprem is never a platform. bare_cluster is a platform that supports onprem, nonetheless, I see now that this validate is a bit of non-sense and will fix

gen/__init__.py
mellenburg: :+1: will refactor

launch/config.py
mellenburg: Also shows this needs a specific test case. Adding that now as well...

mellenburg:
@branden Thanks for the first pass. Comments have been addressed

mellenburg:
code-quality failure is broken S3

gen/calc.py
amitaekbote: Branden explained it better :)

gen/__init__.py
```diff
@@ -544,7 +537,28 @@ def generate(
...
+
+
+def generate(
+        arguments,
+        extra_templates=list(),
+        extra_sources=list(),
+        extra_targets=list()):
+    # To maintain the old API where we passed arguments rather than the new name.
+    user_arguments = arguments
+    arguments = None
```
mellenburg: This is copy/paste code and I do not want to be distracted by inherited issues. If you feel this is worth addressing, we can open a PR and/or JIRA

That being said, the point of this seems to be so that you can call gen.generate(arguments={}) as opposed to the more correct gen.generate(user_arguments={}). The distinction being that the name arguments implies the general complete set, whereas user arguments are a non-optional source.

launch/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Options
+* `key_helper`: generate private SSH keys for the underlying hosts if `true`. In `platform: aws`, this means the user does not have to supply `KeyName` in the template parameters and dcos-launch will fill it in. Similarly, in `platform: azure`, `sshRSAPublicKey` is populated automatically. In the aws case, this key will be deleted from EC2 when the deployment is deleted with dcos-launch
+* `zen_helper`: only to be used with `provider: aws` and zen templates. If `true` the network prerequisites for launching a zen cluster will be provided if missing. As with `key_helper`, these resources will be deleted if dcos-launch is used for destroying the deployment
+* `deploy_bare_cluster_only`: only to be used with `provider: onprem`. If `true`, the onprem deployment will stop after the bare cluster hosts have been provisioned. Like any dcos-launch deployment, this will generate a cluster info JSON. This JSON can then be passed back into another dcos-launch deployment for `provider: onprem` with `platform: bare_cluster`, by providing the field `platform_info_filename`
```
mellenburg: Eh, it's not a variable name and it's not proper English, so I will compromise with `cluster-info`

mellenburg:
@orsenthil Comments addressed PTAL

orsenthil:
@mellenburg - please squash them into a single change. I will include this in the train. Thank you!

PR #1223, opened 2017-02-08
Convert acs deploy to generic utilities

mellenburg:
The current tooling only allows deploying a specific template and does a lot of custom logic. This PR breaks that logical into sensible chunks that are independent of DC/OS environment

Included in this PR is a helper function that will pull down a Azure template, parse the parameters, and then type cast the user parameters as indicated. The motivation for this is that Azure parameters are naturally JSON, so a user should be able to pass a JSON or YAML parameter specification to the azure wrapper and launch. The YAML or JSON parsers have to way to tell if a string should be a string or an integer or a boolean, and as such the user would have to explicitly detail the types for no benefit what so ever.

Edit:
PR also includes a small shuffling of unit tests relevant to the code being modified

## Related Issues
https://mesosphere.atlassian.net/browse/DCOS-13300

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: N/A Refactor work

### DIFF
--- a/test_util/azure.py
+++ b/test_util/azure.py
@@ -1,26 +1,29 @@
+""" This module is intended to allow deploying of arbitrary Azure Resource
+Manager templates. For more information on how to configure and interpret these
+templates, see:
+
+https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates
+"""
 import contextlib
+import copy
 import logging
 import re
-from typing import Union
 
-import azure.common.credentials
+import requests
+import retrying
+from azure.common.credentials import ServicePrincipalCredentials
 from azure.common.exceptions import CloudError
 from azure.mgmt.network import NetworkManagementClient
 from azure.mgmt.resource.resources import ResourceManagementClient
 from azure.mgmt.resource.resources.models import (DeploymentMode,
                                                   DeploymentProperties,
-                                                  ResourceGroup, TemplateLink)
-from retrying import retry
+                                                  ResourceGroup)
 
 from test_util.helpers import lazy_property
 
-
-# Very noisy loggers that do not help much in debug mode
-logging.getLogger("msrest").setLevel(logging.INFO)
-logging.getLogger("requests_oauthlib").setLevel(logging.INFO)
 log = logging.getLogger(__name__)
 
-# This interface is designed to only use a single deployment
+# This interface is designed to only use a single deployment.
 # Being as the azure interface is based around resource groups, deriving
 # deployment name from group names makes it easier to attach to creating deploys
 DEPLOYMENT_NAME = '{}-Deployment'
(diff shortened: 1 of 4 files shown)


mellenburg:
@orsenthil @amitaekbote @lingmann PTAL This is a split out prerequisite for dcos-launch azure support

test_util/azure.py
```diff
@@ -35,77 +35,123 @@ def validate_hostname_prefix(prefix):
...
                 group_name, deployment_name, deployment_properties)
             stack.pop_all()
+        log.info('Successfully started template deployment')
+
+    def create_deployment_properties(self, template_url, parameters):
+        """ Pulls the targeted template, checks parameter specs and casts
+        user provided parameters to the appropriate type. Assertion is raised
+        if there are unused parameters
+        """
+        user_parameters = copy.copy(parameters)
```
orsenthil: If there are going to be list of lists or dictionary with list, it should be `copy.deepcopy`

1. My rule of thumb. use deepcopy unless you worry about space explosion.

test_util/azure.py
```diff
@@ -35,77 +35,123 @@ def validate_hostname_prefix(prefix):
...
+
+
+class DcosAzureResourceGroup:
+    def __init__(self, group_name, azure_wrapper):
+        self.group_name = group_name
+        self.azure_wrapper = azure_wrapper
+
+    @classmethod
+    def deploy_acs_template(
+            cls, azure_wrapper, template_url, group_name,
```
orsenthil: Using `mypy` all these can be checked for type at the parameter level.  We dont have a write a method to type checkit.

test_util/azure.py
```diff
@@ -35,77 +35,123 @@ def validate_hostname_prefix(prefix):
     assert re.match('^[a-z][a-z0-9-]{1,61}[a-z0-9]$', prefix), 'Invalid DNS prefix: {}'.format(prefix)
 
 
+def check_json_object(obj):
```
orsenthil: -ve on this.

If we use type-hints and mypy, then we dont need this. 

And in this code base we already have started using type hints.

test_util/azure.py
mellenburg: I can type this out, but this whole function should die. Mapping parameters 1:1 is a waste of maintenance time and energy. Once the dcos-launch hits the CI we will be doing it like so: https://github.com/mellenburg/dcos/blob/8d3f5e60a48c68ac453d01da4c8dc830db937d64/test_util/launch_configs/azure.yaml

In which case, the python code doesn't care or even need to know what the type on the parameters are. The template is what determines that

test_util/azure.py
mellenburg: I don't think you are seeing the intention of this code. This is to fill in the cast map. It doesn't make any sense to try to cast something to an object or array when we are already using JSON/YAML.

mellenburg:
@orsenthil I updated the description of this PR to reflect this, but I think you are missing the intention of the validation code here. The comments you have lean towards how to validate python calling this function. That is not a use case that will be exercised. Rather, what we want to do is feed JSON/YAML directly in as parameters without having to do any intrusive logic. We cannot do that currently, because its up the YAML/JSON parser which will often cast numbers as ints, but the template strictly expects one or the other. Thus, either the user would have to declare a type with every field (yikes) or we do a 1:1 mapping of arguments for user (wayyy too much overhead)

mellenburg:
@orsenthil Thanks for calling me out on the force-push. Bad habits die hard :)

test_util/azure.py
gpaul: Yeah - this code is intentionally dynamic - in go you'd be using reflection to populate templates.

test_util/azure.py
```diff
@@ -35,77 +38,126 @@ def validate_hostname_prefix(prefix):
...
+
+
+class DcosAzureResourceGroup:
+    def __init__(self, group_name, azure_wrapper):
+        self.group_name = group_name
+        self.azure_wrapper = azure_wrapper
+
+    @classmethod
+    def deploy_acs_template(
+            cls, azure_wrapper, template_url, group_name,
```
gpaul: Would you mind adding a docstring that explains what the method parameters are supposed to be, along with a description of what the function returns?

```
return cls(group_name, azure_wrapper)
```
Is tricky to reason about given that `cls` isn't strictly defined.

test_util/azure.py
```diff
@@ -35,77 +38,126 @@ def validate_hostname_prefix(prefix):
...
-
+    def deploy_template_to_new_resource_group(self, template_url, group_name, parameters):
         deployment_name = DEPLOYMENT_NAME.format(group_name)
-
         # Resource group must be created before validation can occur
-        if azure_wrapper.rmc.resource_groups.check_existence(group_name):
+        if self.rmc.resource_groups.check_existence(group_name):
             raise Exception("Group name already exists / taken: {}".format(group_name))
         log.info('Starting resource group_creation')
         with contextlib.ExitStack() as stack:
```
gpaul: any reason you don't just use try/finally?

test_util/azure.py
branden: The `cls` param is a common Python idiom that shouldn't need much documentation. This is a `@classmethod`, so the first argument is going to be the class to which this method belongs. https://docs.python.org/3/library/functions.html#classmethod.

Not opposed to a docstring and/or type annotations for the other args.

test_util/azure.py
mellenburg: Yes, finally executes always. We do not want to delete the resource group if everything for deployment is looking green, so then we pop it from the stack before the context exits

test_util/azure.py
mellenburg: Sure thing, however I will not spend too much time on describing the parameters, as we do not want to be hard-coding those anywhere in the long run. Once dcos-launch is ready, methods like this will be purged in favor of having config files in CI where we can tweak the parameters whenever

mellenburg:
@gpaul Comments addressed, PTAL

mellenburg:
@lingmann @spahl @alberts This is looking ready to :ship:

gpaul:
LGTM

PR #1224, opened 2017-02-08
anatoly/dcos 13619 mesos ucr minuteman tests

aeyakovenko:
## High Level Description

Test for vips from UCR containers.

networking-team reviewers:
* @asridharan @GoelDeepak 

## Related Issues

  - [DCOS-13619](https://mesosphere.atlassian.net/browse/DCOS-13619) ucr tests for minuteman

## Checklist for all PR's

this is the test

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review):
this is all the changes
this is the test
na

### DIFF
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -1,8 +1,10 @@
+import concurrent.futures
 import contextlib
 import json
 import logging
 import random
 import threading
+import time
 from collections import deque
 from subprocess import check_output
 
@@ -11,12 +13,13 @@ import requests
 import retrying
 
 from pkgpanda.build import load_json
-from test_util.marathon import get_test_app, get_test_app_in_docker
+from test_util.marathon import get_test_app, get_test_app_in_docker, get_test_app_in_ucr
 
 
 log = logging.getLogger(__name__)
 timeout = 500
-maxthreads = 8
+maxthreads = 16
+backend_port_st = 8000
 
 
 def lb_enabled():
@@ -28,8 +31,8 @@ def lb_enabled():
                 stop_max_delay=timeout * 1000,
                 retry_on_result=lambda ret: ret is False,
                 retry_on_exception=lambda x: True)
-def ensure_routable(cmd, service_points):
-    proxy_uri = 'http://{}:{}/run_cmd'.format(service_points[0].host, service_points[0].port)
+def ensure_routable(cmd, host, port):
+    proxy_uri = 'http://{}:{}/run_cmd'.format(host, port)
     log.info('Sending {} data: {}'.format(proxy_uri, cmd))
     r = requests.post(proxy_uri, data=cmd)
     log.info('Requests Response: %s', repr(r.json()))
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
 
 
 class VipTest:
-    def __init__(self, num, vip, vipaddr, samehost, vipnet, proxynet):
+    def __init__(self, num, container, vip, vipaddr, samehost, vipnet, proxynet):
+        self.num = num
         self.vip = vip.format(num, 7000 + num)
         self.vipaddr = vipaddr.format(num, 7000 + num)
+        self.container = container
         self.samehost = samehost
         self.vipnet = vipnet
         self.proxynet = proxynet
+        self.notes = ""
 
     def __str__(self):
-        return ("VipTest(vip={},vipaddr={},samehost={},"
-                "vipnet={},proxynet={})").format(self.vip, self.vipaddr, self.samehost, self.vipnet, self.proxynet)
+        return ('VipTest(container={}, vip={},vipaddr={},samehost={},'
+                'vipnet={},proxynet={}, notes={})').format(self.container, self.vip, self.vipaddr, self.samehost,
+                                                           self.vipnet, self.proxynet, self.notes)
 
     def log(self, s, lvl=logging.DEBUG):
         m = 'VIP_TEST {} {}'.format(s, self)
         log.log(lvl, m)
 
 
+def docker_vip_app(network, host, vip):
+    # docker app definition defines its own healthchecks
+    app, uuid = get_test_app_in_docker()
+    app['id'] = '/viptest/' + app['id']
+    app['container']['docker']['network'] = network
+    app['mem'] = 16
+    app['cpu'] = 0.01
+    if network == 'HOST':
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        del app['container']['docker']['portMappings']
+        if vip is not None:
+            app['portDefinitions'] = [{'labels': {'VIP_0': vip}}]
+    else:
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py 9080'
+        app['container']['docker']['portMappings'] = [{
+            'hostPort': 0,
+            'containerPort': 9080,
+            'protocol': 'tcp',
+            'name': 'test',
+            'labels': {}
+        }]
+        if vip is not None:
+            app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': vip}
+        if network == 'USER':
+            app['ipAddress'] = {'networkName': 'dcos'}
+    app['constraints'] = [['hostname', 'CLUSTER', host]]
+    return app, uuid
+
+
+def mesos_vip_app(num, network, host, vip, ucr=False):
+    app = None
+    uuid = None
+    port = backend_port_st + num
+    if ucr is False:
+        app, uuid = get_test_app()
+    else:
+        app, uuid = get_test_app_in_ucr()
+    app['id'] = '/viptest/' + app['id']
+    app['mem'] = 16
+    app['cpu'] = 0.01
+    # define a health check that works with all the network options
+    app['healthChecks'] = [{
+        'protocol': 'MESOS_HTTP',
+        'path': '/ping',
+        'gracePeriodSeconds': 5,
+        'intervalSeconds': 10,
+        'timeoutSeconds': 10,
+        'maxConsecutiveFailures': 3,
+    }]
+    assert network != 'BRIDGE'
+    if network == 'HOST':
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        app['portDefinitions'] = [{
+            'protocol': 'tcp',
+            'port': 0
+        }]
+        if vip is not None:
+            app['portDefinitions'][0]['labels'] = {'VIP_0': vip}
+        app['healthChecks'][0]['portIndex'] = 0
+    if network == 'USER':
+        app['ipAddress'] = {
+            'discovery': {
+                'ports': [{
+                    'protocol': 'tcp',
+                    'name': 'test',
+                    'number': port,
+                }]
+            }
+        }
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py {}'.format(port)
+        app['ipAddress']['networkName'] = 'dcos'
+        if vip is not None:
+            app['ipAddress']['discovery']['ports'][0]['labels'] = {'VIP_0': vip}
+        app['healthChecks'][0]['port'] = port
+        app['portDefinitions'] = []
+    app['constraints'] = [['hostname', 'CLUSTER', host]]
+    log.info('app: {}'.format(json.dumps(app)))
+    return app, uuid
+
+
+def vip_app(num, container, network, host, vip):
+    if container == 'UCR':
+        return mesos_vip_app(num, network, host, vip, ucr=True)
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    assert container == 'NONE', 'unkown container option {}'.format(container)
+    return mesos_vip_app(num, network, host, vip, ucr=False)
+
+
 def vip_test(dcos_api_session, r):
     r.log('START')
-    agents = list(dcos_api_session.slaves)
+    agents = list(dcos_api_session.all_slaves)
     # make sure we can reproduce
     random.seed(r.vip)
     random.shuffle(agents)
@@ -66,43 +166,28 @@ def vip_test(dcos_api_session, r):
         host2 = agents[1]
     log.debug('host1 is is: {}'.format(host1))
     log.debug('host2 is is: {}'.format(host2))
-    origin_app, app_uuid = get_test_app_in_docker()
-    origin_app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': r.vip}
-    origin_app['container']['docker']['network'] = r.vipnet
-    origin_app['mem'] = 0
-    origin_app['cpu'] = 0
-    if r.vipnet == 'USER':
-        origin_app['ipAddress'] = {'networkName': 'dcos'}
-    if r.vipnet == 'HOST':
-        origin_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
-                            '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
-        origin_app['container']['docker']['portMappings'][0]['hostPort'] = 0
-        origin_app['container']['docker']['portMappings'][0]['containerPort'] = 0
-        origin_app['container']['docker']['portMappings'][0]['labels'] = {}
-        origin_app['portDefinitions'] = [{'labels': {'VIP_0': r.vip}}]
 
-    origin_app['constraints'] = [['hostname', 'CLUSTER', host1]]
-
-    proxy_app = get_test_app_in_docker()[0]
-    proxy_app['container']['docker']['network'] = r.proxynet
-    proxy_app['constraints'] = [['hostname', 'CLUSTER', host2]]
-    proxy_app['mem'] = 0
-    proxy_app['cpu'] = 0
-    if r.proxynet == 'USER':
-        proxy_app['ipAddress'] = {'networkName': 'dcos'}
-    if r.proxynet == 'HOST':
-        proxy_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
-                           '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
-        proxy_app['container']['docker']['portMappings'][0]['hostPort'] = 0
-        proxy_app['container']['docker']['portMappings'][0]['containerPort'] = 0
+    origin_app, app_uuid = vip_app(r.num, r.container, r.vipnet, host1, r.vip)
+    # allow to run on both public and private agents
+    origin_app['acceptedResourceRoles'] = ["*", "slave_public"]
+    proxy_app, _ = vip_app(r.num, r.container, r.proxynet, host2, None)
+    # allow to run on both public and private agents
+    proxy_app['acceptedResourceRoles'] = ["*", "slave_public"]
 
     returned_uuid = None
-
     with contextlib.ExitStack() as stack:
+        # try to avoid thundering herd
+        time.sleep(random.randint(0, 30))
         stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app, timeout=timeout))
         sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app, timeout=timeout))
+        proxy_host = sp[0].host
+        proxy_port = sp[0].port
+        if proxy_port == 0 and sp[0].ip is not None:
+            proxy_port = backend_port_st + r.num
+            proxy_host = sp[0].ip
+        log.info("proxy endpoints are {}".format(sp))
         cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(r.vipaddr)
-        returned_uuid = ensure_routable(cmd, sp)
+        returned_uuid = ensure_routable(cmd, proxy_host, proxy_port)
         log.debug('returned_uuid is: {}'.format(returned_uuid))
     assert returned_uuid is not None
     assert returned_uuid['test_uuid'] == app_uuid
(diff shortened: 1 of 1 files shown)


asridharan:
@aeyakovenko this has pull/1169 as well? Probably keep it in WIP for the time being till 1169 gets merged, and then you can rebase and remove WIP?

orsenthil:
@mesosphere-mergebot bump-ee

aeyakovenko:
@adam-mesos ready for review.  can you add @asridharan and @GoelDeepak from networking

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -120,42 +195,48 @@ def reduce_logging():
...
+    tests = [VipTest(i, c, vi, va, sh, vn, pn) for i, [c, vi, va, sh, vn, pn] in enumerate(permutations)]
+    executor = concurrent.futures.ThreadPoolExecutor(max_workers=maxthreads)
     # deque is thread safe
     failed_tests = deque(tests)
     passed_tests = deque()
     skipped_tests = deque()
+    # skip certain tests
     for r in tests:
+        if r.container is 'UCR' or r.container is 'NONE':
+            if r.vipnet is 'BRIDGE' or r.proxynet is 'BRIDGE':
```
jgehrcke: Never use the identity operator for value comparison:

```
>>> a = 's'
>>> b = 's'
>>> a is b
True
>>> c = 1000 * 's'
>>> d = 1000 * 's'
>>> c is d
False
```

You're hitting CPython-specific implementation detail here, where short strings with equal values -- for optimization -- map to the same object. But that's not a language guarantee.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,22 +40,114 @@ def ensure_routable(cmd, service_points):
...
+    app['healthChecks'] = [{
+        'protocol': 'MESOS_HTTP',
+        'path': '/ping',
+        'gracePeriodSeconds': 5,
+        'intervalSeconds': 10,
+        'timeoutSeconds': 10,
+        'maxConsecutiveFailures': 3,
+    }]
+    assert network is not 'BRIDGE'
+    if network is 'USER':
```
jgehrcke: Same here (see below).

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,22 +40,114 @@ def ensure_routable(cmd, service_points):
...
+    app['cpu'] = 0.01
+    if network == 'USER':
+        app['ipAddress'] = {'networkName': 'dcos'}
+    if network == 'HOST':
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        app['container']['docker']['portMappings'][0]['hostPort'] = 0
+        app['container']['docker']['portMappings'][0]['containerPort'] = 0
+        app['container']['docker']['portMappings'][0]['labels'] = {}
+        if vip is not None:
```
jgehrcke: Here, the identity operator `is` is fine, because `None` is a singleton.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,22 +40,114 @@ def ensure_routable(cmd, service_points):
...
+
+    app['constraints'] = [['hostname', 'CLUSTER', host]]
+    return app, uuid
+
+
+def mesos_vip_app(num, network, host, vip, ucr=False):
+    app = None
+    uuid = None
+    port = backend_port_st + num
+    if ucr is False:
```
jgehrcke: Here, it's also fine, because `False` is a singleton, too.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,22 +40,114 @@ def ensure_routable(cmd, service_points):
...
+        }
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py {}'.format(port)
+        app['ipAddress']['networkName'] = 'dcos'
+        if vip is not None:
+            app['ipAddress']['discovery']['ports'][0]['labels'] = {'VIP_0': vip}
+        app['healthChecks'][0]['port'] = port
+        app['portDefinitions'] = []
+
+    if network is 'HOST':
```
jgehrcke: Use `==` (see below).

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -66,43 +160,24 @@ def vip_test(dcos_api_session, r):
...
     with contextlib.ExitStack() as stack:
-        stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app, timeout=timeout))
-        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app, timeout=timeout))
+        stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(origin_app, timeout=timeout,
+                                                                         check_health=False))
+        sp = stack.enter_context(dcos_api_session.marathon.deploy_and_cleanup(proxy_app, timeout=timeout,
+                                                                              check_health=False))
+        proxy_host = sp[0].host
+        proxy_port = sp[0].port
+        if proxy_port is 0 and sp[0].ip is not None:
```
jgehrcke: `is 0` is deadly -- see below. Use `==` instead.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: what a horrible language

jgehrcke: Well, you have misused `is`. :D The operators are well-defined, and the purpose of `is` is limited to a _single_ thing.

aeyakovenko:
@jgehrcke fixed the `is` usage.  any other comments?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +40,126 @@ def ensure_routable(cmd, service_points):
...
+    log.info('app: {}'.format(json.dumps(app)))
+    return app, uuid
+
+
+def vip_app(num, container, network, host, vip):
+    if container == 'UCR':
+        return mesos_vip_app(num, network, host, vip, ucr=True)
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    if container == 'NONE':
```
asridharan: I think this should be 
assert container == NONE, 'unkown container option {}'.format(container)
return mesos_vip_app(num, network, host, vip, ucr=False)

assert False doesn't seem idiomatic?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -66,43 +169,26 @@ def vip_test(dcos_api_session, r):
...
-        proxy_app['ipAddress'] = {'networkName': 'dcos'}
-    if r.proxynet == 'HOST':
-        proxy_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
-                           '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
-        proxy_app['container']['docker']['portMappings'][0]['hostPort'] = 0
-        proxy_app['container']['docker']['portMappings'][0]['containerPort'] = 0
 
-    returned_uuid = None
+    origin_app, app_uuid = vip_app(r.num, r.container, r.vipnet, host1, r.vip)
+    origin_app['acceptedResourceRoles'] = ["*", "slave_public"]
```
asridharan: Why is the `origin_app` always on `slave_public`?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -120,46 +206,52 @@ def reduce_logging():
...
+        vip_test(dcos_api_session, test)
+        failed_tests.remove(test)
+        passed_tests.append(test)
+
+    tasks = [executor.submit(run, t) for t in failed_tests]
+    for t in concurrent.futures.as_completed(tasks):
+        try:
+            t.result()
+        except Exception as exc:
+            log.info('vip_test generated an exception: {}'.format(exc))
```
asridharan: Shouldn't this be a `raise AssertionError` ? Why aren't we asserting on the first exception? Because we want all tests to be complete an report the failed tests?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +40,126 @@ def ensure_routable(cmd, service_points):
...
+            'containerPort': 9080,
+            'protocol': 'tcp',
+            'name': 'test',
+            'labels': {}
+        }]
+        if vip is not None:
+            app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': vip}
+        if network == 'USER':
+            app['ipAddress'] = {'networkName': 'dcos'}
+
```
asridharan: We are setting up health check for `Mesos` but not for `Docker`?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +40,126 @@ def ensure_routable(cmd, service_points):
...
+        }
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py {}'.format(port)
+        app['ipAddress']['networkName'] = 'dcos'
+        if vip is not None:
+            app['ipAddress']['discovery']['ports'][0]['labels'] = {'VIP_0': vip}
+        app['healthChecks'][0]['port'] = port
+        app['portDefinitions'] = []
+
+    if network == 'HOST':
```
asridharan: Would be nice to have the same ordering as docker. Move `HOST` above `USER`.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -66,43 +169,26 @@ def vip_test(dcos_api_session, r):
...
-        proxy_app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
-                           '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
-        proxy_app['container']['docker']['portMappings'][0]['hostPort'] = 0
-        proxy_app['container']['docker']['portMappings'][0]['containerPort'] = 0
 
-    returned_uuid = None
+    origin_app, app_uuid = vip_app(r.num, r.container, r.vipnet, host1, r.vip)
+    origin_app['acceptedResourceRoles'] = ["*", "slave_public"]
+    proxy_app, _ = vip_app(r.num, r.container, r.proxynet, host2, None)
+    proxy_app['acceptedResourceRoles'] = ["*", "slave_public"]
```
asridharan: This is confusing, even proxy_app is on `slave_public`? Maybe I am not following the topology of the test. Could you add a comment of how the test topology looks and what exactly is the state machine of the test.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -120,46 +206,52 @@ def reduce_logging():
 
 @pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
 def test_vip(dcos_api_session, reduce_logging):
-    """Test every permutation of VIP
-    """
+    '''Test every permutation of VIP
```
asridharan: Would be nice to have a comment here explaining the different options for docker and mesos containers, and the permutations we are aiming to test. For someone not familiar with our networking semantics this would be pretty incomprehensible.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +40,126 @@ def ensure_routable(cmd, service_points):
...
+
+
+def vip_app(num, container, network, host, vip):
+    if container == 'UCR':
+        return mesos_vip_app(num, network, host, vip, ucr=True)
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    if container == 'NONE':
+        return mesos_vip_app(num, network, host, vip, ucr=False)
+    assert False, 'unkown container option {}'.format(container)
```
jgehrcke: yeah, `raise AssertionError('...')` is the more explicit way to raise an AssertionError unconditionally.

Here, I think we don't need to have an AssertionError -- py.test will record this test as failed for any unhandled exception. So, `raise Exception('...')` would be sufficient.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: '*' means any

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: '*' means any, so it can run on both

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: Yep.  So you can look at the set of failed tests and see if there is a common thread, like all the "bridge" tests failed.

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: the marathon applications define the health checks for docker

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: i think `assert container == 'NONE'` makes more sense

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -206,7 +204,12 @@ def reduce_logging():
 
 @pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
 def test_vip(dcos_api_session, reduce_logging):
-    '''Test every permutation of VIP
+    '''Test VIPs between the follow source and destination configurations:
```
asridharan: s/follow/following

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    assert container == 'NONE', 'unkown container option {}'.format(container)
+    return mesos_vip_app(num, network, host, vip, ucr=False)
+
+
 def vip_test(dcos_api_session, r):
     r.log('START')
-    agents = list(dcos_api_session.slaves)
+    agents = list(dcos_api_session.all_slaves)
```
mellenburg: Nit: the intention of this line is to make a new list that is a copy of dcos_api_session.all_slaves so that it can be mutated; in such a case, it is helpful to do something like ```agents = copy.copy(dcos_api_session.all_slaves)``` to explicitly show that this is a pure copy and not a list type-cast

mellenburg: Nit: the intention of this line is to make a new list that is a copy of dcos_api_session.all_slaves so that it can be mutated; in such a case, it is helpful to do something like ```agents = copy.copy(dcos_api_session.all_slaves)``` to explicitly show that this is a pure copy and not a list type-cast

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -66,43 +166,28 @@ def vip_test(dcos_api_session, r):
...
+    # allow to run on both public and private agents
+    origin_app['acceptedResourceRoles'] = ["*", "slave_public"]
+    proxy_app, _ = vip_app(r.num, r.container, r.proxynet, host2, None)
+    # allow to run on both public and private agents
+    proxy_app['acceptedResourceRoles'] = ["*", "slave_public"]
 
+    returned_uuid = None
     with contextlib.ExitStack() as stack:
+        # try to avoid thundering herd
+        time.sleep(random.randint(0, 30))
```
mellenburg: Is this wise? Why are we handicapping our tests? Can the marathon/mesos really not handle so many app POST's?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+def vip_app(num, container, network, host, vip):
+    if container == 'UCR':
+        return mesos_vip_app(num, network, host, vip, ucr=True)
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    assert container == 'NONE', 'unkown container option {}'.format(container)
+    return mesos_vip_app(num, network, host, vip, ucr=False)
+
+
 def vip_test(dcos_api_session, r):
```
mellenburg: Not from this PR, but this paradigm gets used more in this PR, so I would prefer that we fix it now as it will greatly improve readability/reviewability.

Please do not name variables something like 'r' unless there is a clear and *limited* context. E.G. this is OK:
```
for r in response_list:
    r.raise_for_status()
```
It is clear in this case that r represents a response from that list and will only be used within the context of that iterator loop. In this PR here is a great example where this is totally fine: https://github.com/dcos/dcos/pull/1224/files#diff-89643056102d7647724e7dc94be80843R219

However, in this case, r seems to be used for no logical reason. Specifically, r is of type ```VipTest``` which has no relation to the letter ```r``` I can understand.

Later on, in totally unrelated code, this happens again with ```for r in tests:``` How was r an appropriate variable name here? https://github.com/dcos/dcos/pull/1224/files#diff-89643056102d7647724e7dc94be80843R234

So, I would recommend that we call this function like so: ```def run_vip_test(dcos_api_session, vip_test: VipTest)```
and then later on in the for-loop do: ```for vip_test in tests:```

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+    app['cpu'] = 0.01
+    if network == 'HOST':
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
+        del app['container']['docker']['portMappings']
+        if vip is not None:
+            app['portDefinitions'] = [{'labels': {'VIP_0': vip}}]
+    else:
+        app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
+                     '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py 9080'
```
orsenthil: Please extract out this port definition in line 79 and line 82.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+        if vip is not None:
+            app['container']['docker']['portMappings'][0]['labels'] = {'VIP_0': vip}
+        if network == 'USER':
+            app['ipAddress'] = {'networkName': 'dcos'}
+    app['constraints'] = [['hostname', 'CLUSTER', host]]
+    return app, uuid
+
+
+def mesos_vip_app(num, network, host, vip, ucr=False):
+    app = None
```
orsenthil: Kill `app` and `uuid` declarations. unnecessary.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -11,12 +13,13 @@
...
 from pkgpanda.build import load_json
-from test_util.marathon import get_test_app, get_test_app_in_docker
+from test_util.marathon import get_test_app, get_test_app_in_docker, get_test_app_in_ucr
 
 
 log = logging.getLogger(__name__)
 timeout = 500
-maxthreads = 8
+maxthreads = 16
+backend_port_st = 8000
```
orsenthil: Please comment the meaning of this. I guess it is `start_number for the ports used in networking test`?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+    # define a health check that works with all the network options
+    app['healthChecks'] = [{
+        'protocol': 'MESOS_HTTP',
+        'path': '/ping',
+        'gracePeriodSeconds': 5,
+        'intervalSeconds': 10,
+        'timeoutSeconds': 10,
+        'maxConsecutiveFailures': 3,
+    }]
+    assert network != 'BRIDGE'
```
orsenthil: newline after hanging indent.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+        return ('VipTest(container={}, vip={},vipaddr={},samehost={},'
+                'vipnet={},proxynet={}, notes={})').format(self.container, self.vip, self.vipaddr, self.samehost,
+                                                           self.vipnet, self.proxynet, self.notes)
 
     def log(self, s, lvl=logging.DEBUG):
         m = 'VIP_TEST {} {}'.format(s, self)
         log.log(lvl, m)
 
 
+def docker_vip_app(network, host, vip):
```
orsenthil: In this method, please use newlines generously to enhance readability.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+    log.info('app: {}'.format(json.dumps(app)))
+    return app, uuid
+
+
+def vip_app(num, container, network, host, vip):
+    if container == 'UCR':
+        return mesos_vip_app(num, network, host, vip, ucr=True)
+    if container == 'DOCKER':
+        return docker_vip_app(network, host, vip)
+    assert container == 'NONE', 'unkown container option {}'.format(container)
```
orsenthil: `unknown`

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -38,25 +41,122 @@ def ensure_routable(cmd, service_points):
...
+    return mesos_vip_app(num, network, host, vip, ucr=False)
+
+
 def vip_test(dcos_api_session, r):
     r.log('START')
-    agents = list(dcos_api_session.slaves)
+    agents = list(dcos_api_session.all_slaves)
     # make sure we can reproduce
```
orsenthil: Mark a TODO here.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -120,46 +205,58 @@ def reduce_logging():
...
+        vip_test(dcos_api_session, test)
+        failed_tests.remove(test)
+        passed_tests.append(test)
+
+    tasks = [executor.submit(run, t) for t in failed_tests]
+    for t in concurrent.futures.as_completed(tasks):
+        try:
+            t.result()
+        except Exception as exc:
+            # just log the exception, each failed test is recored in the `failed_tests` array
```
orsenthil: `recorded`

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -120,46 +205,58 @@ def reduce_logging():
...
+            log.info('vip_test generated an exception: {}'.format(exc))
     [r.log('PASSED', lvl=logging.INFO) for r in passed_tests]
```
orsenthil: Why are we doing all these logging? Once this test is stabilized, you could kill these logging and leave it at the assert statement only?

mellenburg:
@skumaran  Your comments seem like mine in that they relate mostly to code clarity. Could you consider migrating them to https://github.com/dcos/dcos/pull/1289  ?

orsenthil:
@mellenburg  sure.

orsenthil:
@aeyakovenko  - before making this ready for merge, please rebase and squash all your commits to a single one. (By this, when it lands in master, it is much easier to track the problems in TC or with History).

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: python is a weird language

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: can you elaborate?

packages/dcos-integration-test/extra/test_networking.py
aeyakovenko: what do you mean?

packages/dcos-integration-test/extra/test_networking.py
orsenthil: I meant, adding a newline at some places in the code, and presenting the logical chunks together improved the readability.

this diff:


```
diff --git a/packages/dcos-integration-test/extra/test_networking.py b/packages/dcos-integration-test/extra/test_networking.py
index a146a6c9..2fed4816 100644
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -68,6 +68,7 @@ def docker_vip_app(network, host, vip):
     app['container']['docker']['network'] = network
     app['mem'] = 16
     app['cpu'] = 0.01
+
     if network == 'HOST':
         app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
                      '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
@@ -112,7 +113,9 @@ def mesos_vip_app(num, network, host, vip, ucr=False):
         'timeoutSeconds': 10,
         'maxConsecutiveFailures': 3,
     }]
+
     assert network != 'BRIDGE'
+
     if network == 'HOST':
         app['cmd'] = '/opt/mesosphere/bin/dcos-shell python '\
                      '/opt/mesosphere/active/dcos-integration-test/util/python_test_server.py $PORT0'
@@ -123,6 +126,7 @@ def mesos_vip_app(num, network, host, vip, ucr=False):
         if vip is not None:
             app['portDefinitions'][0]['labels'] = {'VIP_0': vip}
         app['healthChecks'][0]['portIndex'] = 0
+
     if network == 'USER':
         app['ipAddress'] = {
             'discovery': {
@@ -140,8 +144,11 @@ def mesos_vip_app(num, network, host, vip, ucr=False):
             app['ipAddress']['discovery']['ports'][0]['labels'] = {'VIP_0': vip}
         app['healthChecks'][0]['port'] = port
         app['portDefinitions'] = []
+
     app['constraints'] = [['hostname', 'CLUSTER', host]]
+
     log.info('app: {}'.format(json.dumps(app)))
+
     return app, uuid


@@ -189,6 +196,7 @@ def vip_test(dcos_api_session, r):
         cmd = '/opt/mesosphere/bin/curl -s -f -m 5 http://{}/test_uuid'.format(r.vipaddr)
         returned_uuid = ensure_routable(cmd, proxy_host, proxy_port)
         log.debug('returned_uuid is: {}'.format(returned_uuid))
+
     assert returned_uuid is not None
     assert returned_uuid['test_uuid'] == app_uuid
     r.log('PASSED')
```

PR #766, opened 2016-09-22
Add enable_gpu_isolation flag

klueska:
This patch creates a new config.yaml parameter called `enable_gpu_isolation` that enables (cluster-wide) GPU isolation from mesos. This means that *all* machines in your cluster must have GPUs installed on them in order to enable this feature. Otherwise starting the mesos-agent on non-GPU capable machines will fail.

For now, this support is considered experimental, so requiring all machines in a cluster to have GPUs is acceptable for testing out the feature. In future releases we will proide support for having a mixed cluster of GPU and non GPU machines, where only the GPU-capable machines have this feature enabled.

PR #1227, opened 2017-02-10
Bump Mesos, Turn on GPU support by default.

adam-mesos:
## High Level Description

This PR enables GPU support in DC/OS by default.

This means that if your agents have Nvidia GPUs installed on them as well as the proper Nvidia NVML library (> 340.29) installed on them, GPUs will be offered to your frameworks as a resource by those agents.

Previously, GPU support was an install-time parameter that required *all* agents in your cluster to have the proper NVML libraries installed on them. This PR removes this restriction and turns on GPU support by default. Operators should not have to perform any extra configuration to take advantage of this feature.

One caveat to be aware of, however, is that mesos will only offer GPU resources to frameworks that *opt-in* to receive these resources. If your framework has not opted-in, it will receive *no* offers from agents that have GPUs installed on them (these agents will essentially appear not to exist in your cluster from the perspective of those frameworks). The choice to make frameworks explicitly opt-in to receive offers that contain GPUs was to keep legacy frameworks from accidentally consuming non-GPU resources on GPU-capable machines (and thus blocking GPU jobs from running).

This is not an issue for frameworks that ship with DC/OS itself, as they have all been updated to opt-in to receive GPU resources. Details about how to update your framework so that it properly opts-in for GPU resources can be found here:
https://github.com/apache/mesos/blob/master/docs/gpu-support.md#framework-capabilities

## Related Issues

https://jira.mesosphere.com/browse/DCOS-13917

## Checklist for all PR's

There is no test included, because adding an automatically triggered integration test that tests GPU functionality is not yet supported (we would need to run the integration tests on nodes equipped with GPUs). However, because this PR actually turns GPU support on by default, the real test is whether we still pass all of the already-included tests with this feature enabled (since they are run on machines without GPUs). Manual tests for making sure GPU support worked in general were performed in https://github.com/dcos/dcos/pull/766.

  - [X] Change log from the last mesosphere/mesos version integrated: https://github.com/mesosphere/mesos/compare/abe6e6390612f523a3106192d719982c70c31f3e...dc01a78680410d7fe49e096508b64e94d991471c

  - [X] Test Results: Verified manually
  - [ ] Code Coverage (if available): *None*

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -609,7 +609,7 @@ entry = {
                 }
             }
         }),
-        'enable_gpu_isolation': 'false',
+        'enable_gpu_isolation': 'true',
         'cluster_docker_registry_url': '',
         'cluster_docker_credentials_dcos_owned': calculate_docker_credentials_dcos_owned,
         'cluster_docker_credentials_write_to_etc': 'false',
--- a/packages/mesos/README.md
+++ b/packages/mesos/README.md
@@ -1,7 +1,8 @@
 <H2>Patches to cherry-pick on top of open source Apache Mesos before building in DC/OS</h2>
 These commits can be found in the repository at <a href="https://github.com/mesosphere/mesos/">https://github.com/mesosphere/mesos/</a>:
-<li>[c81bea7092bea8a2912237abdcab027e67c1dd91] Set LIBPROCESS_IP into docker container.
-<li>[5720037ac5921b675d59b670ecf6ea24cfbdf317] Changed agent_host to expect a relative path.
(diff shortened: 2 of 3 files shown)


klueska:
@adam-mesos Looks like everything is green now except the non-rquired one.

adam-mesos:
Great. Those azure failures all look "flaky" (note the :game_die: )
However, we're currently investigating a regression somehow caused by the previous Mesos bump, with a fix we're cherry-picking in via https://github.com/dcos/dcos/pull/1242 until https://reviews.apache.org/r/56711/ can land in Mesos master (and be backported to 1.2.0)
So, we'll put this PR on hold for a bit until we can land the fix from pull/1242, then update this PR to include that cherry-pick as well.

klueska:
Rebased on merge of #1242 into master.

klueska:
All green again.

PR #1230, opened 2017-02-10
Support for Oracle Enterprise Linux

jibsonline:
## High Level Description

DCOS can be installed on Oracle Enterprise Linux 7.2 and 7.3(Fork of RHEL 7.2/7.3) if booted from kernel-3.10.0-514.el7 which ships along with the default kernel kernel-uek-4.1.12-61.1.18.el7uek

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -360,8 +360,15 @@ if ([ x$dist == 'xcoreos' ]); then
   exit 0
 fi
 
-if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
-  echo "$dist is not supported. Only RHEL and CentOS are supported" >&2
+if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ] && [ x$dist != 'xol' ]); then
+  echo "$dist is not supported. Only RHEL, CentOS and Oracle Enterprise Linux are supported" >&2
+  exit 0
+fi
+
+kernel_version=`uname -r`
+
+if ([ x$dist != 'xol' ] && [ kernel_version == *"uek"* ]); then
+  echo "Unbreakable Enterprise Kernel is not supported on Oracle Enterprise Linux. Switch to the other Kernel" >&2
   exit 0
 fi
 


adam-mesos:
@lingmann Can you get somebody to review this?

orsenthil:
For the support of Oracle Enterprise Linux to be added "officially", we might need more than just this change.

1. Community Demand
2. Than extenstive testing of DCOS with the applications
3. Handle corner cases with Oracle Linux. One such is handled by this patch, but is not extensive.
4. We will also need our test-suite be run completely on Oracle Linux. We don't have that in our infrastructure.

In general, I would say that if folks want to run open DCOS on a particular flavor of linux, they can run by using the open source code. 

Merging the config back upstream is not applicable due the above 4 points.

I think, we sould close this issue as and revisit only if we decide to add it, and have instructure to test it.

jibsonline:
@orsenthil and @adam-mesos , OEL 7 is identical to RHEL 7/CentOS 7 in every aspect apart from having the extra UEK kernel. That is why the extra condition is added to make sure it is not booted on the UEK kernel. This change should make DC/OS comfortable to be compatible with Oracle Public Cloud and Oracle Bare Metal Cloud.

Reference : https://en.wikipedia.org/wiki/Oracle_Linux#RHEL_compatibility

dcos_installer/action_lib.py
```diff
@@ -360,8 +360,15 @@ def _add_prereqs_script(chain):
...
-if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
-  echo "$dist is not supported. Only RHEL and CentOS are supported" >&2
+if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ] && [ x$dist != 'xol' ]); then
+  echo "$dist is not supported. Only RHEL, CentOS and Oracle Enterprise Linux are supported" >&2
+  exit 0
+fi
+
+kernel_version=`uname -r`
+
+if ([ x$dist != 'xol' ] && [ kernel_version == *"uek"* ]); then
```
lingmann: Should this be `x$dist == 'xol'`?

dcos_installer/action_lib.py
```diff
@@ -360,8 +360,15 @@ def _add_prereqs_script(chain):
   exit 0
 fi
 
-if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
-  echo "$dist is not supported. Only RHEL and CentOS are supported" >&2
+if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ] && [ x$dist != 'xol' ]); then
+  echo "$dist is not supported. Only RHEL, CentOS and Oracle Enterprise Linux are supported" >&2
```
lingmann: Even if users are running Oracle Enterprise Linux with the Red Hat Compatible Kernel (RHCK) enabled, I'm not comfortable using the word supported here unless we actually have some basic automated test infrastructure to verify that it will work. @adam-mesos or @karlkfi is there some terminology that the PMC would approve for 'unsupported' DC/OS installation environments?

adam-mesos: "Supported" implies a lot of things, including testing, bug-fixing, and answering questions. Maybe we could call it "community-supported" instead of "supported" or "Mesosphere-supported", but I'd still expect somebody to be responsible for the above. A softer word would be to say "recommended".
Is @jibsonline running DC/OS on Oracle Enterprise Linux? We could perhaps say "community-tested"?

karlkfi: I usually use "reported to work" vs "tested regularly". But I'm not sure either are great in the context of validation. 

That said, I'm not even sure we should have hard validation like this in the open source project. 

Tho maybe we can hedge and use the product maturity labels. That would make OEL compatibility experimental, while coreos, centos, rhel compatibility is stable.

The key is not using "support" in the name of the feature. It's "OEL compatibility" not "OEL support". 

Customers are supported. 
Features are experimental, preview, or stable.

janisz:
Can we close this PR?

PR #1232, opened 2017-02-11
Update JIRA links in PR template.

adam-mesos:
We recently moved to new on-prem servers, so the old URLs:
https://dcosjira.atlassian.net/browse/DCOS
https://mesosphere.atlassian.net/browse/DCOS
won't work anymore, and instead we'll need to use
https://jira.dcos.io/browse/DCOS_OSS
https://jira.mesosphere.com/browse/DCOS

More recently, the public dcosjira projects were imported into the same JIRA instance, which means all the tickets from https://jira.dcos.io/browse/DCOS were renamed to https://jira.dcos.io/browse/DCOS_OSS (also accessible via https://jira.mesosphere.com/browse/DCOS_OSS )

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -4,8 +4,8 @@ What features does this change enable? What bugs does this change fix? High-Leve
 
 ## Related Issues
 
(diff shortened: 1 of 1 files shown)


orsenthil:
Please make a corresponding enterprise PR template change as well.

adam-mesos:
See also https://github.com/mesosphere/dcos-enterprise/pull/672

PR #1238, opened 2017-02-14
Use generated SSH keys

mellenburg:
Emergency CI Fix

- removes usage of default key in favor of generating a new key each time
- extra keys will be garbage colected by cloud cleaner
- user can no longer specify pre-existing clusters (to aid simplicity)
- standardizes tests to start with the string 'dcos-ci-test' to allow easier resource cleanup

PR #1240, opened 2017-02-14
Provide an Ability to give AWS Stack name prefix.

orsenthil:
## High Level Description

Since Upgrade Integration Tests are inherited and used in Enterprise tests too, it will be good to know the stacks that were launched for enterprise vs the cloudformation stacks that were launched for open.

## Related Issues

- https://jira.mesosphere.com/browse/DCOS-13979 

## Checklist for all PR's

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -277,6 +277,7 @@ class DcosCfStack(CleanupS3BucketMixin, CfStack):
     @classmethod
     def create(cls, stack_name: str, template_url: str, public_agents: int, private_agents: int,
                admin_location: str, key_pair_name: str, boto_wrapper: BotoWrapper):
+
         parameters = {
             'KeyName': key_pair_name,
             'AdminLocation': admin_location,
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -206,6 +206,7 @@ class VpcClusterUpgradeTest:
                  config_yaml_override_install: str, config_yaml_override_upgrade: str,
                  dcos_api_session_factory_install: VpcClusterUpgradeTestDcosApiSessionFactory,
                  dcos_api_session_factory_upgrade: VpcClusterUpgradeTestDcosApiSessionFactory):
+
         self.dcos_api_session_factory_install = dcos_api_session_factory_install
         self.dcos_api_session_factory_upgrade = dcos_api_session_factory_upgrade
         self.num_masters = num_masters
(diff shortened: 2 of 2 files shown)


test_util/test_upgrade_vpc.py
```diff
@@ -271,7 +272,8 @@ def test_app_dns_survive_upgrade():
             self.log_test("test_upgrade_vpc.test_app_dns_survive_upgrade", test_app_dns_survive_upgrade)
 
     def run_test(self) -> int:
-        stack_name = 'upgrade-test-' + random_id(10)
+        stack_prefix = os.getenv("CF_STACK_PREFIX", "dcos-upgrade-test")
```
branden: `CF_STACK_NAME_PREFIX`?

test_util/aws.py
```diff
@@ -424,6 +424,7 @@ class VpcCfStack(CfStack):
     @classmethod
     def create(cls, stack_name, instance_type, instance_os, instance_count,
                admin_location, key_pair_name, boto_wrapper):
+
```
branden: It'd be helpful to separate style changes from behavior changes in separate commits.

orsenthil: The problem with this is, it is difficult to keep looking for style changes only and that same time wait for future to do only the style changes. I will make this in two PR.  We can find it uncomfortable to review and approve a new line PR. :)

test_util/test_upgrade_vpc.py
orsenthil: updated.

test_util/aws.py
orsenthil: Also, not sure if I understood your comment properly.  If requested for a different PR: https://github.com/dcos/dcos/pull/1249 

And my intention is to squash all the commits into a single One before merging.

test_util/test_upgrade_vpc.py
```diff
@@ -271,7 +271,8 @@ def test_app_dns_survive_upgrade():
             self.log_test("test_upgrade_vpc.test_app_dns_survive_upgrade", test_app_dns_survive_upgrade)
 
     def run_test(self) -> int:
-        stack_name = 'upgrade-test-' + random_id(10)
+        stack_prefix = os.getenv("CF_STACK_NAME_PREFIX", "dcos-upgrade-test")
```
mellenburg: This will allow you to deploy things that will not get cleaned by cloud cleaner. We are using generated keys now and want to use the right magic strings everywhere (generated keys are named with the stack name). https://github.com/dcos/dcos/pull/1238/files#diff-be106601b4bf185940d95c6bce4a625bR270

So, instead, just make this so that dcos-ci-test is still used, e.g.
```stack_prefix = os.getenv('CF_STACK_NAME_PREFIX', '')
stack_name = f"{stack_prefix}-dcos-ci-test"```

orsenthil: Addressed. I turned it into a SUFFIX. My need was to have a differentiator that can be set from CI. Having it as a suffix was better, because if we have it a prefix and it is set to blank, the string could start with `-`.

test_util/aws.py
branden: I'd be fine with keeping the style commit on this PR, that way I can still review the behavior changes in isolation. No big deal either way.

orsenthil:
@lingmann - added the PR template. ( I initially had the High Level Description without title). Created a JIRA and added a link.

orsenthil:
@lingmann , review comments have been addressed. PTAL again.

PR #1278, opened 2017-02-22
Fix the sort -V capability check

surdy:
## High Level Description

I think its a minor bug. We would never run into this in currently supported distros. But I happened to stumble upon the logic which seems incorrect.

I believe there is a bug in the code block : 
```
     $( echo '1' | sort -V >/dev/null 2>&1 || exit 1 )
     RC=$?
     if [[ "$RC" -eq "2" ]]; then
```

Bash is confusing in that 0 is true(success) non-zero is false(failure) 

Current logic
If `sort -V` is valid, its retncode will be 0, `||` will short circuit and `$?`=0
If  `sort -V` is invalid its retncode will be 2, `||` will not short circuit and exit 1 will be executed  and `$?`=1

I think && is what was intended to be used.

New logic 

If `sort -V` is valid, its retncode will be 0, `&&` will not short circuit and exit 1 will be executed  and `$?`=1
If  `sort -V` is invalid its retncode will be 2, `&&` will short circuit and `$?`= 2

## Checklist for all PR's

This succeeded so far because we only test on images with sort -V functionality so we never hit a path where we have to disable version check. I ran the modified script in an image without sort -V functionality and got version check disabled using new logic but not using old logic

  - [X ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [X ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

___

### DIFF
--- a/gen/build_deploy/bash.py
+++ b/gen/build_deploy/bash.py
@@ -136,9 +136,11 @@ declare -i DISABLE_VERSION_CHECK=0
 
 # check if sort -V works
 function check_sort_capability() {
-    $( echo '1' | sort -V >/dev/null 2>&1 || exit 1 )
-    RC=$?
-    if [[ "$RC" -eq "2" ]]; then
+    $( command -v sort >/dev/null 2>&1 || exit 1 )
+    RC1=$?
+    $( echo '1' | sort -V >/dev/null 2>&1 )
+    RC2=$?
+    if [[ "$RC1" -eq "1" || "$RC2" -eq "2" ]]; then
         echo -e "${RED}Disabling version checking as sort -V is not available${NORMAL}"
         DISABLE_VERSION_CHECK=1
     fi


surdy:
@mellenburg @branden Please review

branden:
Good catch. It looks like we can get rid of the `exit 1` too, it doesn't appear to be helping any.

surdy:
Makes sense. How about 

```
 $( echo '1' | sort -V >/dev/null 2>&1 )
     RC=$?
     if [[ "$RC" -ne "0" ]]; then
```

branden:
LGTM 👍

orsenthil:
Yeah, avoid the `&&` logic, it can be more confusing.

My suspicion is `||` was intentional (and common like:http://stackoverflow.com/a/677212/18852)  but incorrectly used.

surdy:
Updated with a simplified test

gen/build_deploy/bash.py
```diff
@@ -136,9 +136,9 @@
 
 # check if sort -V works
 function check_sort_capability() {
-    $( echo '1' | sort -V >/dev/null 2>&1 || exit 1 )
+    $( echo '1' | sort -V >/dev/null 2>&1 )
     RC=$?
-    if [[ "$RC" -eq "2" ]]; then
+    if [[ "$RC" -eq "0" ]]; then
```
branden: `-ne`

branden: Actually I just took a look at the docs for GNU sort: https://www.gnu.org/software/coreutils/manual/html_node/sort-invocation.html
> Exit status:
>    0 if no error occurred
>    1 if invoked with -c or -C and the input is not sorted
>    2 if an error occurred

An exit status of 1 doesn't indicate an error, so let's stick with the `-eq "2"`.

orsenthil: It gets a little tricky now, because we are hiding this 'gnu sort specific exit code' check with-in this statement.

I think, the fix could be simpler and that's the reason the `||` logic is used in the first place.

```
$( echo 'a' | sort -V >/dev/null 2>&1 || exit 1 )
RC=$?
if [[ "$RC" -eq "1" ]]; then
```

* We are checking against the implanted value of 1, that is reached only when the command fails.

branden: We don't want to collapse all nonzero exit statuses into 1, because not all of them indicate an error. An exit status of 1 doesn't mean the command failed, it means input wasn't sorted.

orsenthil: We could use something else instead of 1 too. (like 3, 4, 5 ...). It's not about collapsing. The idea is about a known value to exit with, when it reaches the `||` condition.

> An exit status of 1 doesn't mean the command failed, it means input wasn't sorted.

That is gnu sort oddity and as you shared, it will be applicable "if invoked with -c or -C".

branden: `||` is going to collapse all nonzero `sort` statuses into whatever status gets passed to the following `exit`.

This check is already assuming we have GNU sort. It's just making sure it's a version that supports `-V`. Distinguishing GNU sort from BSD support would probably be a separate check, and one that's not worth doing since we don't support any BSD-based OS'es yet.

surdy: brainfart using `-eq` instead of `-ne`
`-eq "2"` seems fine until you consider the case of sort not being available all together on the machine. which would give 127. Oh the pleasures of coding in bash. 

Maybe is should do a `check_command_exists sort` along with `sort -V >/dev/null 2>&1 `check

orsenthil: I was suggesting that:

> command || return

is fairly "common pattern" in bash and we can find many references to this pattern it in documents. 

My intention was not to go into specifics with exit code and or discussing with gnu vs bsd sort.

Any proper fix is fine with me.

surdy: oops forgot to publish my comment

branden: @surdy I'm 👍 putting a `which sort` at the top of this function and exiting with `1` if its status isn't `0`.

orsenthil: @branden - I think, you missed the stackoverflow link that I posted early in this. 
They have covered all these well.

@surdy  - did you happen to check that? Did it help with forming any opinions?

surdy: Sorry for delayed response. I was tied up with customer requests. 
I have updatedt he test based on feed back from you guys.
@branden I have used `command` instead of `which`, seems more robust (Also @orsenthil 's SO link re-inforces that)

@orsenthil Yes, I did the review the SO link. Thanks for sharing.

orsenthil:
Thanks for addressing the comments.

orsenthil:
@surdy , could you please squash them into a single commit.

PR #1284, opened 2017-02-23
Add acs token to test upgrades

amitaekbote:
## High Level Description

Refactor the upgrade tests to clean up cluster.py. Add option for acs token

## Related Issues
DCOS-14031
https://jira.mesosphere.com/browse/DCOS-14031

## Checklist for all PR's

___

Reviewers should be:

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -35,20 +35,25 @@ CONFIG_YAML_OVERRIDE_UPGRADE: file path(default=None)
     This value will be used when upgrading the cluster.
 
 """
+import json
 import logging
 import os
 import pprint
+import random
 import sys
 import traceback
 import uuid
+from subprocess import CalledProcessError
 from typing import Callable, List
 
 import retrying
+from retrying import retry, RetryError
 from teamcity.messages import TeamcityServiceMessages
 
 import test_util.aws
 import test_util.cluster
 from pkgpanda.util import logger, write_string
+from test_util.cluster import run_bootstrap_nginx
 from test_util.dcos_api_session import DcosApiSession, DcosUser
 from test_util.helpers import CI_CREDENTIALS, marathon_app_id_to_mesos_dns_subdomain, random_id
 
@@ -57,6 +62,19 @@ logging.basicConfig(format='[%(asctime)s|%(name)s|%(levelname)s]: %(message)s', 
 
 
 TEST_APP_NAME_FMT = 'upgrade-{}'
+
+curl_cmd = [
+    'curl',
+    '--silent',
+    '--verbose',
+    '--show-error',
+    '--fail',
+    '--location',
+    '--keepalive-time', '2',
+    '--retry', '20',
+    '--speed-limit', '100000',
+    '--speed-time', '60',
+]
 
 
 def create_marathon_viplisten_app():
@@ -219,6 +237,7 @@ class VpcClusterUpgradeTest:
         self.default_os_user = default_os_user
         self.config_yaml_override_install = config_yaml_override_install
         self.config_yaml_override_upgrade = config_yaml_override_upgrade
+        self.acs_token = None
 
         self.teamcity_msg = TeamcityServiceMessages()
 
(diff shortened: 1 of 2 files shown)


orsenthil:
@amitaekbote , could you please update the PR template of this PR.

amitaekbote:
Its WIP , I will update it once its ready

orsenthil:
Actually, the template could updated to reflect the state and then then, it could be updated with the findings once ready for review.

---- 

**High Level Description**

Testing sending of authentication token during retrieval of information from Mesos

Work in progress.

----

amitaekbote:
@mellenburg @branden ready for review

test_util/test_upgrade_vpc.py
```diff
@@ -360,6 +379,104 @@ def test_app_dns_survive_upgrade():
...
+        stop_max_delay=(1000 * 60 * 5),
+        # Retry on SSH command error or metric not equal to expected value.
+        retry_on_exception=(lambda exc: isinstance(exc, CalledProcessError)),
+        retry_on_result=(lambda result: not result),
+    )
+    def wait_for_mesos_metric(self, cluster, host, key, value):
+        """Return True when host's Mesos metric key is equal to value."""
+        return self.mesos_metrics_snapshot(cluster, host).get(key) == value
+
+    def upgrade_dcos(self, cluster, installer_url, version, add_config_path=None):
```
branden: This function may be useful outside the context of an upgrade test. It should stay in the `cluster` lib.

test_util/test_upgrade_vpc.py
```diff
@@ -473,6 +590,21 @@ def apply(self, dcos_url: str, masters: List[str], public_masters: List[str], sl
                               default_os_user, DcosUser(CI_CREDENTIALS))
 
 
+class VpcClusterUpgradeTest(VpcClusterUpgradeTestFactory):
+    def mesos_metrics_snapshot(self, cluster, host):
```
branden: This should stay a part of the `Cluster` class. We'll likely find other reasons to get a Mesos metrics snapshot aside from the upgrade test. Also the knowledge of which port Mesos masters listen on vs. agents and the URL for the metrics endpoint fit better in the `Cluster` class.

test_util/test_upgrade_vpc.py
mellenburg: @branden how do you see this being useful elsewhere? As it stands, the entire cluster.py file will only be used by the upgrade test after dcos-launch is used to drive the CI, and even then, its only purpose for those tests now is having the shared constructor for running tests. What I am thinking for the long term is that cluster will lose all the provisioning and cloudformation bits, and become an abstraction whose main purpose is running SSH commands against roles. E.G. I don't need to say "do this to this host" but rather "do this to all agents in parallel or serial"

The goal being that many people have wanted to write tests to perform remote commands and so this would be one interface. There might be an "upgrader" module after that which combines the api_session objects with the ssh controller.

What do you think?

test_util/test_upgrade_vpc.py
mellenburg: This is running in Ssher now, so its a pain to override this in a child of cluster because then you need to remake the cluster constructor. However, I don't think this should be made into a new class. The old class is fine and the downstream can just override...

test_util/test_upgrade_vpc.py
```diff
@@ -195,7 +213,7 @@ def apply(self, dcos_url: str, masters: List[str], public_masters: List[str], sl
         pass
 
 
-class VpcClusterUpgradeTest:
+class VpcClusterUpgradeTestFactory:
```
mellenburg: I dont think this needs to change to a factory so mesos_metrics_snapshot can be overriden. The downstream can just inherit this class and override. Am I missing something?

amitaekbote:
@mellenburg @branden ready for another pass

test_util/test_upgrade_vpc.py
branden: I was thinking that this code will be useful come time we implement orchestrated upgrades, but if the `cluster` lib is going away then it makes sense to pull this out.

test_util/test_upgrade_vpc.py
```diff
@@ -463,6 +590,9 @@ def run_test(self) -> int:
             self.log.info("Test failed! VPC cluster will remain available for "
                           "debugging for 2 hour after instantiation.")
 
+        if os.getenv('CI_FLAGS'):
+            result = 0
+
```
branden: This bit shouldn't be here, looks like it came from https://github.com/dcos/dcos/pull/1373.

amitaekbote:
@lingmann @adam-mesos can we add a ship-it label on this?

amitaekbote:
@mesosphere-mergebot bump-ee

PR #1291, opened 2017-02-23
Merge Adminrouter repo into DC/OS

vespian:
## High Level Description

* merge AR into DC/OS repository
* add Jenkins integration for AR CI: https://jenkins.mesosphere.com/service/jenkins/job/security/job/dcos-open-adminrouter/view/change-requests/job/PR-1291/

## Related Issues

  - https://jira.mesosphere.com/browse/DCOS-11919?filter=-1

## Related PRs

   - https://github.com/dcos/adminrouter/pull/53
   - https://github.com/mesosphere/dcos-enterprise/pull/637

## Other

   - this PR is based on https://github.com/dcos/adminrouter/commit/a13d8c2223ea866746d1323c56431a03fd37d1ec commit of Open AR repo. Please feel free to compare the AR code :)
   - we are running into resource limits for Azure, @mellenburg is working on that, for now it's safe to ignore it.

## Quick howto on reviewing this monster:
This PR in 95% is just AR repo contents that can be easily verified by:
* create a temporary directory, `cd` into it
* git checkout `dcos/dcos` repo
* git checkout `dcos/adminrouter` repo
* cd into `dcos/packages/adminrouter/` directory
* `diff -Nwrup ../../../adminrouter/docker ./docker`
* ` diff -Nwrup ../../../adminrouter/ ./extra/src/ --exclude '*docker*' --exclude '*git*'`

Remaining stuff are the changes/files that need to be reviewed:
* addition of `Jenkinsfile*` files in root repo dir
* `dcos/packages/adminrouter/build*` files changes
* move of the systemd unit files 
* move of the OpenResty download into the docker container build stage (more in https://github.com/dcos/adminrouter/pull/53)
* general `packages/adminrouter` directory structure

Some extra context/information can also be found in my comment (https://github.com/dcos/dcos/pull/1291#issuecomment-282869996) further down below.


spahl:
It's happening!

vespian:
@jgehrcke This is a longer story and a bit of a compromise.

There are three options to build AR now:
a) developer launches AR on his laptop during development, he/she just wants to launch tests
b) Jenkins builds AR in order to tests AR
c) TC builds AR to integrate it in DC/OS.

On top of that we want:
x) have the build configuration in just one place
y) be sure that if a) succeeds, b) and c) also will
z) keep everything easy to use and avoid manual steps

If we put `build-resty.sh` in e.g. `extra/` directory, we will still have to supply this file (or inline it) in `docker/` directory so that a) and b) can build AR (they do not have access to packagepanda stuff. This will violate x) and make y) much more probable due to copy-paste errors.

The same applies for Openresty tarball download - we can put it inside `buildinfo.json` and this will make stuff more usable. But this will also make a)'s and b)'s life more difficult as they will have to download the tarball during some extra step and we will have openresty shasums/download URL/version in two places, which violates x) and makes y) more likely again.

The current solution is still not ideal, as we are using different libraries (openssl, possibly others, would have to `ldd` a bit) for a) and b) vs. c). But the risk of this violating y) is small I think and fixing this would make build scripts much more obscure. On top of that in order to satisfy c), a) and b) prepare devkit container in two stages now: https://github.com/dcos/dcos/pull/1291/files#diff-d1e7c76e752b20cc50635a6ca3a47036R50

So we end up with a docker container that:
* includes OpenResty tarball, verified using SHAsum
* ships the `build-resty.sh` script *inside the container* so that a), b) and c) can use the same compilation cmdline

Does it make more sense now?

vespian:
EDIT - added some quick how-to on reviewing this monster to the PR description.

packages/adminrouter/extra/src/.flake8
```diff
@@ -0,0 +1,13 @@
+[flake8]
+
+# Ignore the following errors:
```
mellenburg: Nit: I share your opinion on style here, but we should probably try to make this identical to the style used on the rest of the python in this repo if its living here forever

vespian: I tried to be consistent with other repos, i.e. https://github.com/mesosphere/bouncer/blob/master/.flake8

From what I can see the difference is only in the max line length. 95 < 120, so AR code will definitely pass general DC/OS ones. On top of that - I ofter use my laptop and it has very narrow screen ;)

Would it be OK for you to have two sets of rules as long as the "local" ones are stricter than the DC/OS ones and not the other way round?

vespian:
> It would be nice if these unit tests were trigger-able inline with the rest of the DC/OS unit tests, but given the breadth of Adminrouter, we would be shooting ourselves in the foot if we enforced that here. Rather, this containerized unit test suite and the built-out Jenkins CI have some pretty tremendous advantages for developing, so I am totally fine with introducing a CI testing flow that is a somewhat different than what is used for the rest of DC/OS

I am very happy that you are ok with it :) Let me just give you some more arguments so that you have better context.

I believe that teams should be asynchronous as much as possible/if possible. If we join DC/OS unitests with AR tests, some of this flexibility may be lost - changes in one part of DC/OS may result in changes to AR buildchain/AR itself. On top of that - having separate CI just for AR allows for shorter RTT. Developer just needs to check AR CI and DC/OS integration tests, there is no need to try to isolate AR tests from all the DC/OS unittests if all he/she wants to do is just add extra 'Cache-Control' header.

Thanks for review!

jgehrcke:
Above, I wrote

>  it looks like this file, despite its full file path, actually (also) determines the pkgpanda build for DC/OS. I therefore think that it should not be in the /docker directory. That can be misleading.

I missed the fact that this PR also switches the adminrouter package build from the classical/default container to the custom container build, which is defined by the `docker` directory contents.

spahl:
It is not risky. We put a lot of work into making sure the risk is gone and the actual merge is not adding risk, only moving code without changing it and that only after having added a fully new test harness. I will comment on the motivations behind this in the ticket.

vespian:
Some more context: https://jira.mesosphere.com/browse/DCOS-11919?focusedCommentId=76505&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-76505

vespian:
@adam-mesos Spelling mistakes fixed.

adam-mesos:
Hmm... looks like the pre-spellcheck version got merged into train 134. You might have to address the spelling errors in a follow-up PR.

adam-mesos:
Merged (except for spelling changes) in #1303 (train 134)

packages/adminrouter/extra/src/.flake8
gpaul: I see our tox.ini ignores packages' src code anyway.

vespian: Discussed in person:
I will talk with @mellenburg  about unifying it. The idea is to have DC/OS create general set of rules to which all the packages need to adhere to. Packages themselves may be stricter if they want to, but never move lenient. The flake8 run will take place both in code-quality and Jenkins tests.

mellenburg: To get here I think we need to be at a point where we run unittests and code quality with the packages we build rather than the packages that setuptools/on-host tox decides to build in. So, I am very OK with other projects having their own settings, but they should still use the same special linter. See: https://jira.mesosphere.com/browse/DCOS-14305

PR #1294, opened 2017-02-24
dcos-integration-test: use dcos_api_session in test_mesos.py

gpaul:
## High Level Description

In Enterprise `security:strict` mode the `mesos-master` and `mesos-agent` processes check the `Authorization` header and forbids unauthorized requests. They also supports only `https`. 

This PR fixes a test by replacing direct use of the `requests` library with the recently added `dcos_api_session` fixture that provides a similar API. Requests made using this fixture have a valid `Authorization` header injected. It also chooses the scheme (ie., `http` / `https`) appropriate to the cluster's security mode.

The initial `Connection Refused` failure was due to use of `http://` instead of `https://`. Fixing that revealed what we already knew: the `mesos-master` and `mesos-agent` processes would forbid unauthorized requests.

## Related Issues

  - [DCOS-14161](https://jira.mesosphere.com/browse/DCOS-14161) Fix `test_mesos.test_if_marathon_app_can_be_debugged` test

## Checklist for all PR's

___

### DIFF
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -38,11 +38,8 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
     app_id = 'integration-test-{}'.format(test_uuid)
     with dcos_api_session.marathon.deploy_and_cleanup(app):
         # Fetch the mesos master state once the task is running
-        scheme = dcos_api_session.default_url.scheme
         master_ip = dcos_api_session.masters[0]
-        master_state_url = '{}://{}:{}/state'.format(scheme, master_ip, 5050)
         r = dcos_api_session.get('/state', host=master_ip, port=5050)
-        logging.debug('Got %s with request for %s. Response: \n%s', r.status_code, master_state_url, r.text)
         assert r.status_code == 200
         state = r.json()
 
@@ -65,7 +62,6 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
         assert agent_hostname is not None, 'Agent hostname not found for agent_id {}'.format(agent_id)
         logging.debug('Located %s with containerID %s on agent %s', app_id, container_id, agent_hostname)
 
-        # Wrapper to post, log, and validate return status
         def _post_agent(url, headers, json=None, data=None, stream=False):
             r = dcos_api_session.post(
                 url,
@@ -75,13 +71,6 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
                 json=json,
                 data=data,
                 stream=stream)
-            agent_url = '{}://{}:{}/state'.format(scheme, agent_hostname, 5051)
-            logging.info(
-                'Got %s with POST request to %s with headers %s and json data %s.',
-                r.status_code,
-                agent_url,
-                headers,
-                json)
             assert r.status_code == 200
             return r
 
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -2,24 +2,9 @@ import json
 import logging
 import uuid
 
-import requests
 
 from test_util.marathon import get_test_app
 from test_util.recordio import Decoder, Encoder
-
-
-# Wrapper to post, log, and validate return status
-def post(url, headers, json=None, data=None, stream=False):
-    r = requests.post(url, headers=headers, json=json, data=data, stream=stream)
-    logging.info(
-        'Got %s with POST request to %s with headers %s and json data %s.',
-        r.status_code,
-        url,
-        headers,
-        json
-    )
-    assert r.status_code == 200
-    return r
 
 
 # Creates and yields the initial ATTACH_CONTAINER_INPUT message, then a data message,
@@ -53,8 +38,10 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
     app_id = 'integration-test-{}'.format(test_uuid)
     with dcos_api_session.marathon.deploy_and_cleanup(app):
         # Fetch the mesos master state once the task is running
-        master_state_url = 'http://{}:{}/state'.format(dcos_api_session.masters[0], 5050)
-        r = requests.get(master_state_url)
+        scheme = dcos_api_session.default_url.scheme
+        master = dcos_api_session.masters[0]
+        master_state_url = '{}://{}:{}/state'.format(scheme, master, 5050)
+        r = dcos_api_session.get('/state', host=dcos_api_session.masters[0], port=5050)
         logging.debug('Got %s with request for %s. Response: \n%s', r.status_code, master_state_url, r.text)
         assert r.status_code == 200
         state = r.json()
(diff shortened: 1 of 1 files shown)


gpaul:
@mellenburg @vespian @malnick @orsenthil PTAL

gpaul:
@mesosphere-mergebot bump-ee

gpaul:
Sure thing, it was blocking my exhibitor PR. I'm pretty confident that the idea of the patch is sound, but please review for coding style, etc. Thanks!

jgehrcke:
> This fixes a test failure where the Authorization header wasn't being provided to the mesos-master and mesos-agent processes.

Mhm. The reported symptom (a 'Connection reset by peer') cannot be addressed with adding an Authorization header. Also, from the diff it is not obvious to me what actually changed with respect to the request authentication. Is it the dynamic URL scheme handling of `dcos_api_session` which actually fixes the problem (before, we hard-coded `http` -- now it's dynamic and can be `https`)?

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -53,8 +38,10 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
     app_id = 'integration-test-{}'.format(test_uuid)
     with dcos_api_session.marathon.deploy_and_cleanup(app):
         # Fetch the mesos master state once the task is running
-        master_state_url = 'http://{}:{}/state'.format(dcos_api_session.masters[0], 5050)
-        r = requests.get(master_state_url)
+        scheme = dcos_api_session.default_url.scheme
+        master = dcos_api_session.masters[0]
```
orsenthil: `master_ip` to be explicit? 

Because we use the `agent_host` when it is a hostname.

gpaul:
@jgehrcke I've updated the Jira with the exact failure details. I've updated the PR description with a summary thereof. Indeed the initial `Connection Refused` error was due to our using `http` instead of `https`. Fixing that showed that we still need to send the `Authorization` header as we are then faced with 401 responses. The `dcos_api_session` fixture does exactly both these things.

packages/dcos-integration-test/extra/test_mesos.py
gpaul: Good catch. I've renamed it to `master_ip` and use it in the `dcos_api_session.get(...)` on the next two lines instead of calling `dcos_api_session.masters[0]` again.

jgehrcke:
Thanks for the update, Gustav.

> Fixing that showed that we still need to send the Authorization header as we are then faced with 401 responses.

Which authenticator changed behavior, and (why) is that expected? CC @adam-mesos

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -76,9 +63,29 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
...
+        def _post_agent(url, headers, json=None, data=None, stream=False):
+            r = dcos_api_session.post(
+                url,
+                host=agent_hostname,
+                port=5051,
+                headers=headers,
+                json=json,
+                data=data,
+                stream=stream)
+            agent_url = '{}://{}:{}/state'.format(scheme, agent_hostname, 5051)
```
adam-mesos: Why do you need `agent_url` here? Shouldn't you be logging the actual `url` that you're posting to?

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -76,9 +63,29 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
...
+            logging.info(
+                'Got %s with POST request to %s with headers %s and json data %s.',
+                r.status_code,
+                agent_url,
+                headers,
+                json
+            )
```
adam-mesos: Nit: Is the proper style to wrap the closing parens like this? I see it unwrapped elsewhere.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -53,8 +38,10 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
     app_id = 'integration-test-{}'.format(test_uuid)
     with dcos_api_session.marathon.deploy_and_cleanup(app):
         # Fetch the mesos master state once the task is running
-        master_state_url = 'http://{}:{}/state'.format(dcos_api_session.masters[0], 5050)
-        r = requests.get(master_state_url)
+        scheme = dcos_api_session.default_url.scheme
+        master_ip = dcos_api_session.masters[0]
+        master_state_url = '{}://{}:{}/state'.format(scheme, master_ip, 5050)
```
adam-mesos: I don't think you need `master_state_url` (or `scheme`) either, since the line below just uses dcos_api_session's defaults with a hardcoded `/state`.

jgehrcke:
>  We used to mute this OSS test in strict mode since we couldn't get it to pass without authz

Ah, I see. So it all makes sense. :ship:

packages/dcos-integration-test/extra/test_mesos.py
gpaul: If you look at where I use `master_state_url`, you'll notice this is purely for logging. The original code logged the full URL, I decided to keep that behaviour to make this PR as non-invasive as possible.

packages/dcos-integration-test/extra/test_mesos.py
gpaul: Same reasoning as for `master_state_url`. The `url` parameter is just the request path, eg. "/state".

packages/dcos-integration-test/extra/test_mesos.py
gpaul: It's not, thanks I'll fix it.

gpaul: Done.

gpaul:
@adam-mesos PTAL

packages/dcos-integration-test/extra/test_mesos.py
adam-mesos: But in the case of the current callers of this function, it's `/api/v1` not `/state`, and there's no way to guarantee that's always the case since `url` is a parameter to `_post_agent` and this function can and likely will be reused elsewhere, with different urls. Please log the actual URL being 'post'ed. That was my original intention in these log lines.

jgehrcke: > Please log the actual URL being 'post'ed 

Then we probably don't need to do anything here, because I think every HTTP interaction through the API object(s) is already logged accordingly.

adam-mesos: Great, then we can remove these log lines. I originally put them in there for purposes of debugging test failures, but if the api_session already logs requests and responses, then it's redundant.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -76,9 +60,21 @@ def test_if_marathon_app_can_be_debugged(dcos_api_session):
             if agent['id'] == agent_id:
                 agent_hostname = agent['hostname']
         assert agent_hostname is not None, 'Agent hostname not found for agent_id {}'.format(agent_id)
-        agent_v1_url = 'http://{}:{}/api/v1'.format(agent_hostname, 5051)
         logging.debug('Located %s with containerID %s on agent %s', app_id, container_id, agent_hostname)
 
+        # Wrapper to post, log, and validate return status
```
adam-mesos: Nit: s/post, log,/post to agent/

gpaul: I dropped the comment - it's a two-line function that is clearer than the comment itself.

PR #1304, opened 2017-03-01
[dcos-launch] azure support

mellenburg:
Refactors azure lib to be cleaner and further inline with the other libraries
Builds on dcos-launch packaging from https://github.com/dcos/dcos/pull/1353
## Related Issues
https://jira.mesosphere.com/browse/DCOS-13300?filter=-1
## Checklist for all PR's


gen/build_deploy/bash.py
```diff
@@ -669,24 +668,31 @@ def fill_template(base_name, format_args):
...
-
-    return "dcos-launch"
+        subprocess.check_call(['xz', '-d', util_bootstrap_filename])
+        subprocess.check_call(['tar', 'xf', util_bootstrap_filename.rstrip('.xz')])
+    prefix = 'packages/cache/'
+    if variant is not None:
+        launch_path = prefix + 'dcos-launch-{}'.format(pkgpanda.util.variant_name(variant))
+    else:
+        launch_path = prefix + 'dcos-launch'
+    subprocess.check_call([
```
amitaekbote: catch exception for all subprocess calls?

mellenburg: This is actually a change from https://github.com/dcos/dcos/pull/1353 and unrelated to the azure changes happening here. In any case, for what purpose would I catch this exception? I see no reason to here...

packages/dcos-launch/build
```diff
@@ -8,6 +8,7 @@ cp -r /pkg/extra/* /build
 cd /build
 pip3 install --no-deps --install-option="--prefix=$PKG_PATH" /pkg/extra/
 # verify the build
+ls /opt/mesosphere/lib/python3.5/site-packages/azure
```
orsenthil: Is `ls` and not checking the exit code fine for the test?

packages/dcos-launch/extra/launch/config.py
```diff
@@ -1,15 +1,37 @@
...
 from gen.internals import resolve_configuration, Scope, Source, Target, validate_one_of
-from pkgpanda.util import load_string, load_yaml, YamlParseError
+from pkgpanda.util import load_string, load_yaml, YamlParseError  # n
+
+# dcos_installer.config will directly import gen.build_deploy.util
+# which expects to find the image commit in the environment or in
+# a directory-local git tree **at import time**. Therefore, the
+# environment variable DCOS_IMAGE_COMMIT must be set here.
+if 'DCOS_IMAGE_COMMIT' not in os.environ:
+    os.environ['DCOS_IMAGE_COMMIT'] = 'deadbeef'
```
orsenthil: I guess it's the same commit. But prefer '' string here for value.

gen/build_deploy/bash.py
```diff
@@ -701,24 +707,26 @@ def do_create(tag, build_name, reproducible_artifact_path, commit, variant_argum
...
-            installer_filename = make_installer_docker(variant, all_completes[variant],
-                                                       all_completes[bootstrap_installer_name])
-
-            yield {
-                'channel_path': 'dcos_generate_config.{}sh'.format(pkgpanda.util.variant_prefix(variant)),
-                'local_path': installer_filename
-            }
+    for variant in sorted(variant_arguments.keys(), key=lambda k: pkgpanda.util.variant_str(k)):
+        variant_name = pkgpanda.util.variant_name(variant)
+        bootstrap_installer_name = '{}installer'.format(pkgpanda.util.variant_prefix(variant))
```
orsenthil: `TODO`:

Everywhere  we are doing a string interpolation for the path, I think, defining template upfront will be a  preferable pattern. 

I am thinking something like this

```
dcos_launch_path= "{dcos_path_prefix}/dcos-launch"
dcos_launch_variant_path= "{dcos_path_prefix}/dcos-launch-{variant_info}"
bootstrap_name_template = "{pkgpanda_variant_path}/installer"
bootstrap_util_name_template = "{pkgpanda_variant_path}/util"
```

I'd define them all upfront, so that these are paths are like "routes are templatized" (as in webapps) and the code just substitutes the values at appropriate time.  Currently, everything is created when needed.

launch/azure.py
```diff
@@ -0,0 +1,71 @@
...
+            'private_agents': launch.util.convert_host_list(rg.get_private_agent_ips()),
+            'public_agents': launch.util.convert_host_list(rg.get_public_agent_ips()),
+            'master_fqdn': rg.public_master_lb_fqdn,
+            'public_agent_fqdn': rg.public_agent_lb_fqdn}
+
+    def delete(self, info):
+        self.get_resource_group(info).delete()
+
+    def key_helper(self, config):
+        """ Adds private key to the config and injects the public key into
```
orsenthil: Kill space before `Adds`

packages/dcos-launch/extra/launch/util.py
```diff
@@ -66,27 +74,41 @@ def delete(self, info):
...
-                tunnel=test_tunnel,
-                dcos_dns=test_host,
-                master_list=[m['private_ip'] for m in details['masters']],
-                agent_list=[a['private_ip'] for a in details['private_agents']],
-                public_agent_list=[a['private_ip'] for a in details['public_agents']],
-                aws_access_key_id=info['aws_access_key_id'],
-                aws_secret_access_key=info['aws_secret_access_key'],
-                region=info['aws_region'],
-                test_cmd=test_cmd)
+        raise NotImplementedError()
```
orsenthil: +1 :) Abstracts stay Abstract.

packages/dcos-launch/extra/launch/config.py
mellenburg: This is not apart of this PR. This is from https://github.com/dcos/dcos/pull/1353/files#diff-870f931cca2a43e3f3b4881046e29c57R19

And will be removed once this is rebased.

gen/build_deploy/bash.py
mellenburg: This work is not relevant to this PR. This comes from https://github.com/dcos/dcos/pull/1353

That being said, I do not see how templates help here. These strings just need to be defined once. In the create docker installer method this comment is much more appropriate, but out of scope

orsenthil: templates help in readability. For e.g. in a different project these are my templatized URLS: https://github.com/mesosphere/mergebot/blob/master/mergeservice/lambda/mergeconfig.json

mellenburg: That doesn't convince me of anything. That is a good time to use a template because it is a template. I.E. a data structure parameterized for a bunch of variables. Your source is separate from your structure; you need to be explicit with your templating. You also have more than one varaible

Now compare that to here. I am constructing a string with ONE parameter exactly ONCE to use it ONCE. As I see it, you are just asking me to take one line and split it into 2 so that I am defining what my string *could* look like on one line and then how it actually looks on another. That doesnt seem much more readable to me.

In any case, as I already stated, that is not in this PR and if you want to see that changed, take this over to https://github.com/dcos/dcos/pull/1353. This is a PR for the Azure support in dcos-launch

orsenthil: The comment was a `TODO` and no action is required. The magical creation of the string at every place because it was being done only once was bit bothering to me and that's the reason the suggestion.   We can try in a different PR and vet the suggestion. As you say, probably, it's not worth it too.

(Not sure why I looked at the whole thing and commented on other commits. We can close this discussion here (and now :)) and instead of taking it to #1353)

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
In order for this PR to build EE, downstream packages of python-cryptography and libffi need to be moved removed downstream

PR #1320, opened 2017-03-03
Make AdminRouter `make test` work on Mac (docker-machine)

karlkfi:
## High Level Description

Make AdminRouter `make test` work on Mac (docker-machine).

The `ip` command isn't present on Mac and even if it were, when using docker-machine the bridge interface is not present on the host machine, only inside the docker-machine VM. So to get that IP, you can run the IP detection inside a docker container (using alpine). This should work on both Mac and any variety of Linux.

Since `make test` already runs the tests in a docker container, it's no bid deal to use docker to look up the IP to use for docker dns.

### DIFF
--- a/packages/adminrouter/extra/src/Makefile
+++ b/packages/adminrouter/extra/src/Makefile
@@ -7,7 +7,7 @@ DCOSAR_AR_LOCAL_PATH := $(CURDIR)
 DCOSAR_AR_CTR_MOUNT := /usr/local/adminrouter/nginx/conf/
 
 BRIDGE_DEVNAME := $(shell docker network inspect -f '{{ index .Options "com.docker.network.bridge.name" }}' bridge | awk 'NF')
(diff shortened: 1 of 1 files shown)


karlkfi:
@vespian: please nominate another reviewer familiar with AR.

karlkfi:
Looks like the docker version used by jenkins doesn't know about `--network`...

```
flag provided but not defined: --network
See 'docker run --help'.
```

karlkfi:
Alright, I changed it to use `--net=host` to support docker 1.11

karlkfi:
Oops
```
cannot enable tty mode on non tty input
```

karlkfi:
Removed unneeded `-it`

karlkfi:
Looks like the tests passed this time.

karlkfi:
Comments added.

vespian:
@karlkfi Could you also prepare the same PR for EE?

karlkfi:
@vespian: Do I actually need to? The Makefile.common is used by both...

vespian:
Between then and now I have deduplicated Makefile so no - it's no longer
needed.

PR #1325, opened 2017-03-03
Add valid ip-detect-public for onprem

mellenburg:
The public IP shown in the configuration page of the DC/OS is often incorrect in onprem deploys because the user is not required to provide an ip-detect-public script, and if one is not provided, then the ip-detect (private) is used for both. This PR will simply add a more sensible default value that can be easily overridden.

## Known Issues

https://jira.mesosphere.com/browse/DCOS-11602

## Checklist for all PR's

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -520,6 +520,12 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
 __dcos_overlay_network_default_name = 'dcos'
 
 
+ip_detect_public = """#!/bin/sh
+set -o nounset -o errexit
+curl -fsSL https://ipinfo.io/ip
+"""
+
+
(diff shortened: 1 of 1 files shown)


tamarrow:
👍

alberts:
Not really sure this is a good idea. Do we really want everybody's DC/OS clusters hitting ipinfo.ip?

tamarrow:
can we have this as a on-prem default if no `ip-detect-public ` script is provided?

mellenburg:
@tamarrow That is what this PR does. @alberts This is override-able here

mellenburg:
This issue is already tracked in other JIRAs so no point in leaving this PR up if no one thinks this is a valid way to go. Opened more as a proof-of-concept anyway

PR #1330, opened 2017-03-07
AdminRouter: Test library improvements

mhrabovcin:
## High Level Description

* Adds new method to send arbitrary encoded response from Recording mock endpoint
* Improves startup and teardown times for mock servers

## Related Issues

  - [DCOS-14067](https://jira.mesosphere.com/browse/DCOS-14067) add tests for EE auth AR

## Checklist for all PR's

### DIFF
--- a/packages/adminrouter/extra/src/test-harness/modules/mocker/common.py
+++ b/packages/adminrouter/extra/src/test-harness/modules/mocker/common.py
@@ -4,6 +4,7 @@
 Shared management code for DC/OS mocks used by AR instances, both EE and Open.
 """
 
+import concurrent.futures
 import logging
 
 from mocker.endpoints.reflectors import (
@@ -98,16 +99,18 @@ class MockerBase():
 
     def start(self):
         """Start all endpoints registered with this Mocker instance"""
-        for endpoint in self._endpoints.values():
-            endpoint.start()
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            for endpoint in self._endpoints.values():
+                executor.submit(endpoint.start)
 
     def stop(self):
         """Stop all endpoints registered with this Mocker instance.
 
         Usually called right before object destruction
         """
-        for endpoint in self._endpoints.values():
-            endpoint.stop()
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            for endpoint in self._endpoints.values():
+                executor.submit(endpoint.stop)
 
     def reset(self):
         """Reset all the endpoints to their initial state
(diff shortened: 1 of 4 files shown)


packages/adminrouter/extra/src/test-harness/modules/mocker/common.py
```diff
@@ -98,16 +99,30 @@ def __init__(self, extra_endpoints=None):
 
     def start(self):
         """Start all endpoints registered with this Mocker instance"""
+        threads = []
         for endpoint in self._endpoints.values():
-            endpoint.start()
+            threads.append(threading.Thread(target=endpoint.start))
+
+        for thread in threads:
```
vespian: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor
```
with concurrent.futures.ThreadPoolExecutor() as e:
    for endpoint in self._endpoints.values():
         e.submit(endpoint.start)
```

jgehrcke:
Looks good.

Remarks:

* In general, if "starting and endpoint" lasts ~ 1 s, then this is a strong indicator that there is some unnecessary waiting involved. Might be worth addressing in the future.

* Before we grow the number of threads spawned simultaneously here (by adding more endpoints in the future), we should use a thread pool concept. Can use either the [concurrent.futures.ThreadPoolExecutor](https://docs.python.org/3.5/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) or the undocumented [multiprocessing.pool.ThreadPool](http://stackoverflow.com/a/3386632/145400).

packages/adminrouter/extra/src/test-harness/modules/mocker/endpoints/recording.py
```diff
@@ -77,6 +77,12 @@ def _process_commands(self, blob):
             self._record_request()
             # Recording does not end the request processing, so we do not
             # return anything here
+        if ctx.data["encoded_response"]:
+            msg_fmt = "Endpoint `%s` sending encoded response `%s` as requested"
+            log.debug(
+                msg_fmt, ctx.data['endpoint_id'], ctx.data["encoded_response"])
+            self._finalize_request(
+                200, 'text/plain; charset=utf-8', ctx.data["encoded_response"])
```
vespian: We should be returning True/False as per https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/test-harness/modules/mocker/endpoints/basehandler.py#L81-L84 Otherwise request processing will continue while the response has already been sent.

vespian:
BTW, flake8 tests are red: https://jenkins.mesosphere.com/service/jenkins/job/security/job/dcos-open-adminrouter/view/change-requests/job/PR-1330/

packages/adminrouter/extra/src/test-harness/modules/mocker/common.py
mhrabovcin: Changed.

PR #1341, opened 2017-03-09
Add script that fixes /etc/hosts in CentOS 7 on EC2 upon boot

lloesche:
## High Level Description

Several Frameworks rely on the local FQDN being present in /etc/hosts. This is currently not the case in CentOS 7 on EC2. This PR adds a script that, if necessary, updates /etc/hosts with metadata retrieved from the AWS API upon boot.

We generally recommend for frameworks not to rely on working DNS resolution. However according to Julien Eid Zeppelin, Elasticsearch, Kafka and Chronos do.

## Related Issues

  - [DCOS-14302](https://jira.mesosphere.com/browse/DCOS-14302) Chronos requires the hostname inside /etc/hosts.
  - [OPS-578](https://jira.mesosphere.com/browse/OPS-578) DC/OS CentOS 7 AMI has broken hostname config.

### DIFF
--- a/cloud_images/centos7/install_prereqs.sh
+++ b/cloud_images/centos7/install_prereqs.sh
@@ -99,6 +99,41 @@ echo ">>> Remove history"
 unset HISTFILE
 rm -rf /home/*/.*history /root/.*history
 
+echo ">>> Update /etc/hosts on boot"
+update_hosts_script=/usr/local/sbin/dcos-update-etc-hosts
+update_hosts_unit=/etc/systemd/system/dcos-update-etc-hosts.service
+
+mkdir -p "$(dirname $update_hosts_script)"
+
+cat << 'EOF' > "$update_hosts_script"
+#!/bin/bash
+export PATH=/opt/mesosphere/bin:/sbin:/bin:/usr/sbin:/usr/bin
+curl="curl -s -f -m 30 --retry 3"
+fqdn=$($curl http://169.254.169.254/latest/meta-data/local-hostname)
+ip=$($curl http://169.254.169.254/latest/meta-data/local-ipv4)
+echo "Adding $fqdn if $ip is not in /etc/hosts"
+grep ^$ip /etc/hosts > /dev/null || echo -e "$ip\t$fqdn ${fqdn%%.*}" >> /etc/hosts
+EOF
+
+chmod +x "$update_hosts_script"
+
+cat << EOF > "$update_hosts_unit"
+[Unit]
+Description=Update /etc/hosts with local FQDN if necessary
+After=network.target
+
+[Service]
+Restart=no
+Type=oneshot
+ExecStart=$update_hosts_script
+
+[Install]
+WantedBy=multi-user.target
+EOF
+systemctl daemon-reload
+systemctl enable $(basename "$update_hosts_unit")
+
+
 # Make sure we wait until all the data is written to disk, otherwise
 # Packer might quite too early before the large files are deleted
 sync


cloud_images/centos7/install_prereqs.sh
```diff
@@ -99,6 +99,41 @@ echo ">>> Remove history"
...
+update_hosts_unit=/etc/systemd/system/dcos-update-etc-hosts.service
+
+mkdir -p "$(dirname $update_hosts_script)"
+
+cat << 'EOF' > "$update_hosts_script"
+#!/bin/bash
+export PATH=/opt/mesosphere/bin:/sbin:/bin:/usr/sbin:/usr/bin
+curl="curl -s -f -m 30 --retry 3"
+fqdn=$($curl http://169.254.169.254/latest/meta-data/local-hostname)
+ip=$($curl http://169.254.169.254/latest/meta-data/local-ipv4)
```
lingmann: Error checking on `fqdn` and `ip` is effectively managed by the `curl` command.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -99,6 +99,41 @@ echo ">>> Remove history"
...
+echo "Adding $fqdn if $ip is not in /etc/hosts"
+grep ^$ip /etc/hosts > /dev/null || echo -e "$ip\t$fqdn ${fqdn%%.*}" >> /etc/hosts
```
lingmann: There is a one to one mapping between hostnames and IP addresses on AWS, so even if the IP/Hostname changes this will not result in bad data in /etc/hosts.

lingmann:
@lloesche in addition to this PR, we'll need a secondary PR which updates all DC/OS AMI references to use the AMI's built with this updated prereqs script. Ideally, we can merge them both at the same time and this can land with 1.9.1. Sound good?

lloesche:
@lingmann I did some other EL7 AMI changes that fix journald permissions in parallel. I'd like to see this merged in master and then create a separate PR that updated AMIs based on current master with all the changes combined, instead of building AMIs from individual PRs that might be missing features or would have to cherry-pick them into this PR.

lingmann:
@lloesche the thing I'm concerned about is that if we change the AMI builder but don't also update the AMI's our deployments use at the same time (ideally in the same PR)... we won't have any test coverage to ensure these AMI builder changes are valid / don't break other stuff. Does that make sense?

lloesche:
@lingmann it does, but our current CI pipeline doesn't really support that. I have to manually generate AMIs from PRs which based on the TeamCity history has never been done before. Then when I generate new AMIs from a PR instead of master the PR will be missing features from other open PRs that might change the AMI build scripts in parallel (as is the case with this PR).

So for instance if I build AMIs from this PR they will be missing the journald permissions fix that I submitted around the same time. For this single instance I could fix that manually by rebasing, but only because I personally have the knowledge of this other PR.

If somebody else is working on a PR that changes the AMIs his change might override my changes and vice versa. So the only safe way to avoid this that I see is to always build AMIs from master. Essentially we'd need these kinds of PR's that make changes to the AMIs to trigger a "AMIs require rebuild from master after changes have been merged" flag.

lingmann:
@lloesche understood, the case we discussed with two independent PR's that are changing AMI builder code + AMI's deployed at the same time is a good example of where this approach could be a problem and both PR's are missing changes from the other. Let's proceed with always having separate PR's for AMI builder changes and AMI's deployed changes. That way the AMI's deployed are always built "against master". Thanks!

PR #1345, opened 2017-03-09
Fix intended exception

mellenburg:
This code is not funtional


orsenthil:
Not sure if we have test for this. (If yes., these are the right candidates for the test).

PR #1353, opened 2017-03-11
Launch package

mellenburg:
In order to unblock https://github.com/dcos/dcos/pull/1304 I needed to be able to reliably pull in python-cryptography to build dcos-launch. That library has some system requirements that make it difficult to simply pull in with setuptools or pip as we do now. Rather, I want to build cryptography as a package and make dcos-launch a dependent package.

Furthermore, as I went through this implementation, I found another reason to make this change: PyInstaller has many releases that will not work with our build. Specifically:
- the current 3.2.1 release has a broken hook for botocore that pulls in an import (from within pyinstaller) that does not exist. (pyinstaller has terrible test coverage apparently)
- the releases between 3.2.0-3.1.0 have a different bug in their python-six hook, so that is another blocker
- the 3.0 would work, but then it doesnt run inside of a pkgpanda build environment because 3.0 strictly forbids running as root.

Thus, the best solution at this time was to grab one specific dev commit off github that worked and pinned that. No other python package manager would be able to cover this specific scenario. The only way this could have worked before is if the version of 3.2 that pip/tox pulls in is different than the version on github.

In any case, I think taking this approach and making docker the main dependency of dcos-launch as it already is for pkgpanda will be more reliable than hoping pip grabs a "good" pyinstaller or compatible cryptography build.

Main changes:
- dcos-launch is now a package and is no longer a part of the dcos-image library
- dcos-launch is built by making a bootstrap tree called 'util'. the resulting bootstrap tarball will be expanded to pull out the py-installer built dcos-launch binary that will then be uploaded. This is necessary so dcos-launch can come in variant flavors (like the installer) as well as because dcos-launch has no purpose on hosts running DC/OS and therefore is not shipped with the regular bootstrap  tarball
- the launch tests were not being run in CI and were relying on local settings to pass. This PR enabled the tests as a part of the build steps and fixes the tests so they pass.

## Related Issues
https://jira.mesosphere.com/browse/DCOS-14305?filter=-1


gen/build_deploy/bash.py
```diff
@@ -670,23 +668,8 @@ def fill_template(base_name, format_args):
...
-    # put the launch spec into the root of the tree before running pyinstaller
-    work_dir.join('dcos-launch.spec').write(pkg_resources.resource_string('launch', 'dcos-launch.spec'))
-    with work_dir.as_cwd():
-        subprocess.check_call(['pyinstaller', '--log-level=DEBUG', 'dcos-launch.spec'])
-    subprocess.check_call(['mv', str(work_dir.join('dist').join('dcos-launch')), 'dcos-launch'])
-    work_dir.remove()
-
-    return "dcos-launch"
+    # Make this its own function for easy mocking
+    subprocess.check_call(['./build_dcos_launch.sh'])
```
mellenburg: This is a terrible approach. Reworking to invoke pkgpanda artifacts without re-starting a build (even though the build would be cached and therefore trivial)

packages/dcos-launch/extra/conftest.py
```diff
@@ -1,13 +1,23 @@
...
 import pytest
 
+import dcos_installer
 import launch
 import ssh
 import test_util
 from launch.util import get_temp_config_path, stub
 from test_util.helpers import Host
 
+if 'DCOS_IMAGE_COMMIT' not in os.environ:
```
mellenburg: This is an abomination. Considering a PR to land before this that cleans it up

packages/dcos-launch/extra/launch/config.py
```diff
@@ -1,15 +1,24 @@
...
-import dcos_installer.config
-import gen
 import launch.util
 import ssh.validate
 import test_util.aws
-from gen.internals import resolve_configuration, Scope, Source, Target, validate_one_of
-from pkgpanda.util import load_string, load_yaml, YamlParseError
+
+# This is a disgusting hack
+if 'DCOS_IMAGE_COMMIT' not in os.environ:
```
mellenburg: Same as above, mulling over a PR to fix this before this PR

mellenburg:
@branden @orsenthil PTAL

packages/dcos-launch/extra/launch/config.py
```diff
@@ -1,15 +1,37 @@
...
 from gen.internals import resolve_configuration, Scope, Source, Target, validate_one_of
-from pkgpanda.util import load_string, load_yaml, YamlParseError
+from pkgpanda.util import load_string, load_yaml, YamlParseError  # n
+
+# dcos_installer.config will directly import gen.build_deploy.util
+# which expects to find the image commit in the environment or in
+# a directory-local git tree **at import time**. Therefore, the
+# environment variable DCOS_IMAGE_COMMIT must be set here.
+if 'DCOS_IMAGE_COMMIT' not in os.environ:
+    os.environ['DCOS_IMAGE_COMMIT'] = 'deadbeef'
```
orsenthil: prefer setting the value to '' as only key is required.

packages/dcos-launch/extra/launch/config.py
```diff
@@ -1,15 +1,37 @@
...
 import yaml
 
-import dcos_installer.config
 import gen
 import launch.util
 import ssh.validate
 import test_util.aws
 from gen.internals import resolve_configuration, Scope, Source, Target, validate_one_of
-from pkgpanda.util import load_string, load_yaml, YamlParseError
+from pkgpanda.util import load_string, load_yaml, YamlParseError  # n
```
mellenburg: it used to be a `# noqa`. Good catch!

packages/dcos-launch/extra/launch/config.py
mellenburg: Sure. I didn't feel too strongly, but I like deadbeef because its obviously wrong on visual inspection. However deadbeef could fool a small sanity test, so its probably safer to use `''`

mellenburg:
@orsenthil RE: strings that are templated. If this https://github.com/dcos/dcos/blob/master/gen/build_deploy/bash.py#L590 is alright then I can totally agree to that, it was really just the adding the named variabled in the format string that seemed over the top

packages/dcos-launch/extra/launch/config.py
```diff
@@ -1,16 +1,38 @@
...
 
+# dcos_installer.config will directly import gen.build_deploy.util
+# which expects to find the image commit in the environment or in
+# a directory-local git tree **at import time**. Therefore, the
+# environment variable DCOS_IMAGE_COMMIT must be set here.
+if 'DCOS_IMAGE_COMMIT' not in os.environ:
+    os.environ['DCOS_IMAGE_COMMIT'] = ''
+    atexit.register(os.unsetenv, 'DCOS_IMAGE_COMMIT')
+
+import dcos_installer.config  # noqa
```
branden: 😬 Have you tried putting that env lookup inside a function that get called at runtime, instead of import time?

build_dcos_launch.sh
```diff
@@ -1,32 +1,19 @@
...
-fi
-rm -rf /tmp/dcos-installer-build
-git clone -q "file://$PWD" /tmp/dcos-installer-build/
-pushd /tmp/dcos-installer-build
-# Install the DC/OS tools
-pip install -e /tmp/dcos-installer-build
-cp launch/dcos-launch.spec ./
-pyinstaller --log-level=DEBUG dcos-launch.spec
+pushd packages/dcos-launch
+new_package_name=`mkpanda | tail -n 1 | cut -d':' -f2`
```
mellenburg: To be honest, I feel this entire file should be removed. It was only updated because a helper script like this already exists. In truth, this will not be able to build at all on the first build because mkpanda will not evaluate any other part of the tree. Alternatively, you do not want to check every part of the tree when you are just developing dcos-launch. Rather, this should be a Makefile so you can do make/make deps/make launch-deps/make installer/make clean

gen/build_deploy/bash.py
```diff
@@ -669,24 +668,31 @@ def fill_template(base_name, format_args):
...
+        'cp', os.getcwd() + '/packages/cache/bootstrap/' + util_bootstrap_filename,
+        str(work_dir.join(util_bootstrap_filename))])
     with work_dir.as_cwd():
-        subprocess.check_call(['pyinstaller', '--log-level=DEBUG', 'dcos-launch.spec'])
-    subprocess.check_call(['mv', str(work_dir.join('dist').join('dcos-launch')), 'dcos-launch'])
-    work_dir.remove()
-
-    return "dcos-launch"
+        subprocess.check_call(['xz', '-d', util_bootstrap_filename])
+        subprocess.check_call(['tar', 'xf', util_bootstrap_filename.rstrip('.xz')])
```
mellenburg: It is not at all clear where I can find the uncompressed packages and their artifacts from builds. The packages were the only thing being passed around here. If you want to instruct me on how exactly I can go from `variant_arguments` and `all_completes` to having uncompressed packages that were guaranteed to be from this same build, then I would be glad to make the change. Otherwise, this is the best I can do from guessing the magic state of the system my looking at these few lines of code https://github.com/dcos/dcos/blob/master/gen/build_deploy/bash.py#L635 (which don't tell me anything about uncompressed packages).

gen/build_deploy/bash.py
```diff
@@ -701,24 +707,26 @@ def do_create(tag, build_name, reproducible_artifact_path, commit, variant_argum
...
+                    'local_path': make_installer_docker(variant, all_completes[variant],
+                                                        all_completes[bootstrap_installer_name])
+                }
+
+        if bootstrap_util_name not in all_completes:
+            print('WARNING: No util tree for variant: {}'.format(variant_name))
+        else:
+            with logger.scope("building dcos-launch for variant: {}".format(variant_name)):
+                yield {
+                    'channel_path': 'dcos-launch-' + variant_name if variant is not None else 'dcos-launch',
```
mellenburg: I see now that this is confusing/misleading syntax but it works as I intended:
`('dcos-launch-' + variant_name) if variant is not None else 'dcos-launch'`

packages/dcos-launch/extra/launch/config.py
mellenburg: Too much work for this PR. That function can also parse the current git head instead of reading from the environment, so that means that it relies on current working directory which means anytime pkpanda uses it will be prone to breaking if called at runtime. I strongly feel trying to tackle that in this PR is out of scope.

gen/build_deploy/bash.py
```diff
@@ -669,24 +668,31 @@ def fill_template(base_name, format_args):
...
+def make_dcos_launch(variant, launch_info):
+    util_bootstrap_id = launch_info['bootstrap']
+    launch_package_id = None
+    for package_id in launch_info['packages']:
+        if pkgpanda.PackageId(package_id).name == 'dcos-launch':
+            launch_package_id = package_id
+            break
+    assert launch_package_id, 'dcos-launch package cannot be found!'
+    util_bootstrap_filename = util_bootstrap_id + '.bootstrap.tar.xz'
+    # Make this its own function for easy mocking
```
mellenburg: Just forgot to delete. Good eye

gen/build_deploy/bash.py
```diff
@@ -701,24 +707,26 @@ def do_create(tag, build_name, reproducible_artifact_path, commit, variant_argum
...
-            yield {
-                'channel_path': 'dcos_generate_config.{}sh'.format(pkgpanda.util.variant_prefix(variant)),
-                'local_path': installer_filename
-            }
+    for variant in sorted(variant_arguments.keys(), key=lambda k: pkgpanda.util.variant_str(k)):
+        variant_name = pkgpanda.util.variant_name(variant)
+        bootstrap_installer_name = '{}installer'.format(pkgpanda.util.variant_prefix(variant))
+        bootstrap_util_name = '{}util'.format(pkgpanda.util.variant_prefix(variant))
+        if bootstrap_installer_name not in all_completes:
+            print('WARNING: No installer tree for variant: {}'.format(variant_name))
```
mellenburg: The code is currently written so that if you do not include an installer tree with your variant, then it will fail hard:
https://github.com/dcos/dcos/blob/master/gen/build_deploy/bash.py#L707

gen/build_deploy/bash.py
branden: Oh yup. Please add the parens, they clarify this line for anyone who doesn't know Python's grammar well enough to know the precedence order between addition and the ternary operator (AKA me).

build_dcos_launch.sh
```diff
@@ -1,32 +1,19 @@
...
 popd
-cp /tmp/dcos-installer-build/dist/dcos-launch ./
+mkdir -p tmp/dcos_build
+cp $new_package_name tmp/dcos_build/dcos-launch.tar.xz
+pushd tmp/dcos_build
+xz -d dcos-launch.tar.xz
+tar xf dcos-launch.tar
+popd
+cp tmp/dcos_build/dcos-launch .
+rm -rf tmp/dcos_build/
```
mellenburg: These are not gzipped tarballs: these are xz compressed tarballs, so that command is invalid

branden: My mistake, recent versions of `tar` auto-detect compression format these days. You can use `-J` or `--xz` to explicitly decompress with `xz` (added to `tar` in 2009, doubt we'll run into a version that doesn't support it). https://www.gnu.org/software/tar/manual/tar.html#SEC136

packages/dcos-launch/extra/launch/config.py
branden: SGTM. Mind adding a tech debt ticket and linking it here?

gen/build_deploy/bash.py
branden: This code is being executed in a build context, so any package should be available at `packages/cache/packages/<package-name>/<package-id>.tar.xz`.

branden: https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L341 ->
https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L344 ->
https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L255 ->
https://github.com/dcos/dcos/blob/master/pkgpanda/build/cli.py#L29

build_dcos_launch.sh
mellenburg: Sweet; I didnt know about that option, thanks

packages/dcos-launch/extra/launch/config.py
mellenburg: Sure thing. I figured that I will one day soon break the circular dependency between dcos_installer and gen which breaks in all kinds of weird ways, so this would be one of the main tasks for that

gen/build_deploy/bash.py
mellenburg: I just did the most pythonic thing and tossed it into a method for maximum clarity :D

gen/build_deploy/bash.py
```diff
@@ -668,33 +668,28 @@ def fill_template(base_name, format_args):
...
-    By checking the specific package id, we avoid grabbing something unintentionally
-    """
+def fetch_dcos_launch_bin(launch_name, launch_package_id):
     work_dir = py.path.local.mkdtemp()
-    unpacked_bin_path = str(work_dir.join('packages/{}/dcos-launch'.format(launch_package_id)))
-    cached_util_bootstrap = os.getcwd() + '/packages/cache/bootstrap/' + util_bootstrap_filename
-    temp_util_bootstrap = str(work_dir.join(util_bootstrap_filename))
-    subprocess.check_call(['cp', cached_util_bootstrap, temp_util_bootstrap])
+    # inspect the local packages cache to grab the desired artifact
+    cached_package_path = 'packages/cache/packages/' + launch_package_id + '.tar.xz'
```
branden: The package is at `packages/cache/packages/<package_name>/<package_id>.tar.xz`. https://github.com/dcos/dcos/blob/master/pkgpanda/build/__init__.py#L341

branden: It'd be nice to reuse `PackageSet.get_package_path()` here, but its `__init__()` does a bunch of unrelated work :\

mellenburg: Yeah, really, it would be ideal if the make_installer code did that as it does a bunch more

mellenburg:
@branden comments addressed. Is this good to go?

mellenburg:
@mesosphere-mergebot bump-ee

gen/build_deploy/bash.py
```diff
@@ -669,24 +668,36 @@ def fill_template(base_name, format_args):
...
+    subprocess.check_call(['cp', str(work_dir.join('dcos-launch')), launch_path])
+    return launch_path
+
 
-    return "dcos-launch"
+def get_dcos_launch_name(variant):
+    if variant is None:
+        return 'dcos-launch'
+    else:
+        return 'dcos-launch-' + pkgpanda.util.variant_name(variant)
```
branden: Could we replace this with something in `pkgpanda.util`? It looks similar to what `variant_prefix()` is doing. Maybe add a `def variant_suffix(variant, delim='.')` and use `launch_name = 'dcos-launch' + variant_suffix(variant, delim='-')`.

build_dcos_launch.sh
```diff
@@ -11,7 +11,8 @@ export PYTHONUNBUFFERED="notemtpy"
 pushd packages/dcos-launch
 # mkpanda will pump out the package path in the last line;
 # this is required to grab only the binary built here
-new_package_name=`mkpanda | tail -n 1 | cut -d':' -f2`
+user_variant=$1
+new_package_name=`mkpanda ${user_varaint:=default} | tail -n 1 | cut -d':' -f2`
```
branden: `variant`

pkgpanda/build/cli.py
```diff
@@ -23,11 +23,17 @@ def main():
...
+        variant_arg = arguments['<variant>']
+        # map the keyword 'default' to None to build default, but use the null arguments
+        # to trigger all variants
+        target_variant = variant_arg if variant_arg != 'default' else None,
         # Make a local repository for build dependencies
         if arguments['tree']:
             package_store = pkgpanda.build.PackageStore(getcwd(), arguments['--repository-url'])
-            pkgpanda.build.build_tree(package_store, arguments['--mkbootstrap'], arguments['<variant>'])
+            if variant_arg is None:
+                pkgpanda.build.build_tree_varaints(package_store, arguments['--mkbootstrap'])
```
branden: `variants`

pkgpanda/build/cli.py
```diff
@@ -23,11 +23,18 @@ def main():
...
+        # map the keyword 'default' to None to build default, but use the null arguments
+        # to trigger all variants
+        target_variant = variant_arg if variant_arg != 'default' else None
         # Make a local repository for build dependencies
         if arguments['tree']:
             package_store = pkgpanda.build.PackageStore(getcwd(), arguments['--repository-url'])
-            pkgpanda.build.build_tree(package_store, arguments['--mkbootstrap'], arguments['<variant>'])
+            if variant_arg is None:
+                for variant in pkgpanda.build.get_variants_from_filesystem(getcwd(), 'treeinfo.json'):
+                    pkgpanda.build.build_tree(package_store, arguments['--mkbootstrap'], variant)
```
branden: The `get_variants_from_filesystem()` call feels overly detailed for a CLI script's main function. Let's stuff this loop into a `build_tree_variants()` helper, similar to `build_package_variants()`.

gen/build_deploy/bash.py
```diff
@@ -669,24 +668,35 @@ def fill_template(base_name, format_args):
     return installer_filename
 
 
-def make_dcos_launch():
-    # NOTE: this needs to be kept in sync with build_dcos_launch.sh
+def get_launch_package_id(package_list):
+    """scans througha the package list of the
```
branden: Incomplete docstring.

build_dcos_launch.sh
```diff
@@ -1,32 +1,21 @@
...
-pushd /tmp/dcos-installer-build
-# Install the DC/OS tools
-pip install -e /tmp/dcos-installer-build
-cp launch/dcos-launch.spec ./
-pyinstaller --log-level=DEBUG dcos-launch.spec
+pushd packages/dcos-launch
+# mkpanda will pump out the package path in the last line;
+# this is required to grab only the binary built here
+user_variant=$1
+new_package_name=`mkpanda --variant ${user_varaint:=default} | tail -n 1 | cut -d':' -f2`
```
branden: `variant`

gen/build_deploy/bash.py
```diff
@@ -669,24 +668,35 @@ def fill_template(base_name, format_args):
...
     work_dir.remove()
+    return launch_path
+
 
-    return "dcos-launch"
+def get_dcos_launch_name(variant):
+    if variant is None:
+        return 'dcos-launch'
+    else:
+        return 'dcos-launch' + pkgpanda.util.variant_suffix(variant, delim='-')
```
branden: No need for this function now that we have `variant_suffix()`. That'll do the work of returning an empty suffix `if variant is None`.

pkgpanda/build/cli.py
```diff
@@ -5,8 +5,8 @@
 the necessary dependencies.
 
 Usage:
-  mkpanda [--repository-url=<repository_url>] [--dont-clean-after-build] [--recursive]
-  mkpanda tree [--mkbootstrap] [--repository-url=<repository_url>] [<variant>]
+  mkpanda [--repository-url=<repository_url>] [--dont-clean-after-build] [--recursive] [--variant=<variant>]
+  mkpanda tree [--mkbootstrap] [--repository-url=<repository_url>] [--variant=<variant>]
```
branden: Why make this a flag instead of leaving it as an arg? Seems unnecessary.

mellenburg: Because then mkpanda tree is borked forever as when you do:
`mkpanda tree` it thinks you are doing mkpanda with variant=tree. Its either do it this way or change the form of the command to be something like `mkpanda package` and `mkpanda tree` so that way a command could never be parsed as an argument by not having a command `mkpanda` that takes a positional arg

branden: Ah good point.

spahl:
Holding for 1.9 until https://github.com/dcos/dcos/pull/1304 and this are in sync

PR #1354, opened 2017-03-11
Fix erlang ref

mellenburg:
I tried building locally and failed because this ref is not in the repo.

Looks like this was pinned against a merge commit that is no longer in the branch?


nlsun:
@GoelDeepak what happened here?

GoelDeepak:
Community asked me to remove the merge commit and I didn't realized that dcos is pointing to that.
@mellenburg thanks for taking care of it

PR #1091, opened 2016-12-23
Integration test for dockercfg remover hook

adam-mesos:
Enables the docker credential cleanup hook by default, and adds an integration test.

# Issues

[DCOS-5255](https://jira.mesosphere.com/browse/DCOS-5255) Docker credential cleanup hook
[DCOS-8955](https://jira.mesosphere.com/browse/DCOS-8955) Docker credential cleanup DC/OS integration test

# Checklist

 - [X] Included a test which will fail if code is reverted but test is not: This _is_ the test for [DCOS-5255](https://jira.mesosphere.com/browse/DCOS-5255)
 - [X] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #1359, opened 2017-03-14
Move dcos-log configs out of the middle of mesos-slave-common.

adam-mesos:
## High Level Description

While investigating failures in https://github.com/dcos/dcos/pull/1091 we discovered that sometime early in 1.9, we injected dcos-log config files in the middle of the `/etc/mesos-slave-common` definition, which renders the (conditional) setting of MESOS_DOCKER_REGISTRY, MESOS_DOCKER_CONFIG, and MESOS_HOOKS futile, since they are now set in `/etc_slave_public/dcos-log-config.json` instead of becoming Mesos flags. This patch moves the dcos-log configs out, and reunites the conditional parts of mesos-slave-common with the rest.

## Related Issues

  - [DCOS-14473](https://jira.mesosphere.com/browse/DCOS-14473) MESOS_DOCKER_REGISTRY, MESOS_DOCKER_CONFIG, and MESOS_HOOKS ignored in OSS dcos

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: It's really hard to test conditional code that is disabled-by-default.

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -406,24 +406,6 @@ package:
       MESOS_GC_DELAY={{ gc_delay }}
       MESOS_HOSTNAME_LOOKUP=false
       GLOG_drop_log_memory=false
-  - path: /etc/dcos-log.env
-    content: |
-      DCOS_LOG_CONFIG_PATH=/opt/mesosphere/etc/dcos-log-config.json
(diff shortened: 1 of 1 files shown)


spahl:
:ship:

PR #1361, opened 2017-03-15
Fix logrotate URL and sha1

mellenburg:
fedorahosted.org was retired on March 1st, 2017.

### DIFF
--- a/packages/logrotate/buildinfo.json
+++ b/packages/logrotate/buildinfo.json
@@ -1,7 +1,7 @@
 {
   "single_source": {
     "kind": "url_extract",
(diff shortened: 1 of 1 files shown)


BenWhitehead:
@mesosphere-mergebot bump-ee

mellenburg:
Yikes, it looks like that 3.9.1 is not identical as stuff involving 3dt and logs just failed. @darkonie @malnick can you take a look and find a version that works? Right now we are relying on our caches for logrotate...

darkonie:
@mellenburg sorry can you elaborate? i don't see 3dt or dcos-log failures...

mellenburg:
@darkonie oops, I'm an idiot and was looking at the wrong logs. I think these are some insane AWS flake-outs here. Sorry to have bothered you. :grimacing:

PR #1374, opened 2017-03-16
Fix logging statements

mellenburg:
went unnoticed because logging exceptions wont interrupt
the main thread so you actually have to view the logs to notice

### DIFF
--- a/packages/dcos-integration-test/extra/test_composition.py
+++ b/packages/dcos-integration-test/extra/test_composition.py
@@ -87,9 +87,9 @@ def test_signal_service(dcos_api_session):
     if enabled == 'false':
         pytest.skip('Telemetry disabled in /opt/mesosphere/etc/dcos-signal-config.json... skipping test')
 
-    logging.info("Version: ", dcos_version)
-    logging.info("Customer Key: ", customer_key)
-    logging.info("Cluster ID: ", cluster_id)
+    logging.info("Version: " + dcos_version)
+    logging.info("Customer Key: " + customer_key)
+    logging.info("Cluster ID: " + cluster_id)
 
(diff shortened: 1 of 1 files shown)


packages/dcos-integration-test/extra/test_composition.py
```diff
@@ -87,9 +87,9 @@ def test_signal_service(dcos_api_session):
     if enabled == 'false':
         pytest.skip('Telemetry disabled in /opt/mesosphere/etc/dcos-signal-config.json... skipping test')
 
-    logging.info("Version: ", dcos_version)
```
orsenthil: Just an FYI.  Here is an alternate way to fix this.

```
logging.info("Version: %s", dcos_version)
```

mellenburg: I personally hate %-based format strings. You can make the argument that .format and f-strings are just as bad, but no one will ever have to lookup %i vs %d vs %s

orsenthil: Glad you didn't program in C, coz' that's the only way to use in `printf`.

The other style that I prefer (both in Python and Java) is

1) Create a logger with formatter (one time in common util class).


```
def get_logger():
    logger = logging.getLogger(name='mergebot')
    handler = logging.StreamHandler()
    formatter = logging.Formatter('%(asctime)s %(name)-12s %(levelname)-8s %(message)s')
    handler.setFormatter(formatter)
    logger.addHandler(handler)
    logger.setLevel(logging.DEBUG)
    return logger


log = get_logger()
```

2) Use .formatted string.

```
log.debug("Failed parsing r{}".format(config_file_contents))
```

PR #1372, opened 2017-03-16
make test ENTER context instead of generating it

mellenburg:
This test has been disabled for an ungodly amount of time. My bad. https://github.com/dcos/dcos/commit/0e12f32912679bf50339e86e012b129540d47c59

PR #1376, opened 2017-03-17
Fix octarine, tests, and increase coverage

nlsun:
## High Level Description

@mellenburg discovered that the octarine tests were muted. Upon fixing the test code itself and turning them back on, we discovered that they are now failing due to a bug in octarine. This fixes octarine, the existing octarine tests, and expands the test coverage.

## Related Issues

  - [DCOS_OSS-780](https://jira.mesosphere.com/browse/DCOS_OSS-780) octarine's history of why it was broken.

## Checklist for all PR's

___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -140,7 +140,8 @@ def test_octarine_http(dcos_api_session, timeout=30):
         }]
     }
 
-    dcos_api_session.marathon.deploy_and_cleanup(app_definition)
+    with dcos_api_session.marathon.deploy_and_cleanup(app_definition):
+        pass
 
 
 def test_octarine_srv(dcos_api_session, timeout=30):
@@ -187,7 +188,8 @@ def test_octarine_srv(dcos_api_session, timeout=30):
         }]
     }
 
-    dcos_api_session.marathon.deploy_and_cleanup(app_definition)
+    with dcos_api_session.marathon.deploy_and_cleanup(app_definition):
+        pass
 
 
 def test_pkgpanda_api(dcos_api_session):
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -1,8 +1,11 @@
+import logging
 import uuid
 
 import pytest
 
 from test_util.marathon import get_test_app, get_test_app_in_docker, get_test_app_in_ucr
+
+log = logging.getLogger(__name__)
 
 
 def test_if_marathon_app_can_be_deployed(dcos_api_session):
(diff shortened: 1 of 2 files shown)


nlsun:
Supercedes https://github.com/dcos/dcos/pull/1372

nlsun:
@mellenburg @orsenthil @spahl @alberts please review.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+def octarine_runner(octarine, mode, uuid, uri, bind_port=None):
+    log.info("Running octarine(mode={}, uuid={}, uri={}".format(mode, uuid, uri))
+
+    bind_port_str = ""
+    if bind_port is not None:
+        bind_port_str = "-bindPort {}".format(bind_port)
+
+    server_cmd = "{} -mode {} {} {}".format(octarine, mode, bind_port_str, uuid)
+    log.info("Server: {}".format(server_cmd))
+    server = threading.Thread(target=subprocess.call, args=(shlex.split(server_cmd),), daemon=True)
```
mellenburg: Why are we running this directly on the host instead of a mesos container? And why are we using a thread to control a subprocess, that seems redundant any python threads tend to introduce more bugs. E.G. you use daemon mode so if this thread hangs it will just be cleaned up with the the entire interpreter, but because that thread is starting a subprocess, I think its possible the subprocess just lives on forever

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
-    dcos_api_session.marathon.deploy_and_cleanup(app_definition)
+def test_octarine(dcos_api_session, timeout=30):
+    # This app binds to port 80. This is only required by the http (not srv)
+    # transparent mode test. In transparent mode, we use ".mydcos.directory"
+    # to go to localhost, the port attached there is only used to
+    # determine which port to send traffic to on localhost. When it
+    # reaches the proxy, the port is not used, and a request is made
+    # to port 80.
+
+    octarine = "/opt/mesosphere/bin/octarine"
```
mellenburg: Why is this a variable at all? Its never mutated and could just be hard coded into the octarine_runner method.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+    server = threading.Thread(target=subprocess.call, args=(shlex.split(server_cmd),), daemon=True)
+    server.start()
+
+    proxy = ('"http://127.0.0.1:$({} --client --port {})"'.format(octarine, uuid))
+    query_cmd = "sh -c 'curl --fail --proxy {} {}'".format(proxy, uri)
+    log.info("Query: {}".format(query_cmd))
+    res = shell_retry(query_cmd)
+    assert json.loads(res)["pong"] is True
+
+    kill_cmd = "pkill octarine"
```
mellenburg: This is the kind of stuff that needs to be done in a fixture in general. See above for greater complications when you call a subprocess from a thread.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+    log.info("Kill: {}".format(kill_cmd))
+    subprocess.call(shlex.split(kill_cmd))
+    log.info("Wait for server to terminate")
+    server.join()
+
+
+@retrying.retry(wait_fixed=1000, stop_max_delay=30 * 1000,
+                retry_on_exception=lambda x: True)
```
mellenburg: This is superfluous. Every retry decorator has a default of `retry_on_exception=lambda x: True`

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+
+    server_cmd = "{} -mode {} {} {}".format(octarine, mode, bind_port_str, uuid)
+    log.info("Server: {}".format(server_cmd))
+    server = threading.Thread(target=subprocess.call, args=(shlex.split(server_cmd),), daemon=True)
+    server.start()
+
+    proxy = ('"http://127.0.0.1:$({} --client --port {})"'.format(octarine, uuid))
+    query_cmd = "sh -c 'curl --fail --proxy {} {}'".format(proxy, uri)
+    log.info("Query: {}".format(query_cmd))
+    res = shell_retry(query_cmd)
```
mellenburg: Get rid of the shell_retry function completely; it seems clear its intention is to do retrying logic for getting an assert case, so then make a retry loop specific for that case.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +117,81 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+    if bind_port is not None:
+        bind_port_str = "-bindPort {}".format(bind_port)
+
+    server_cmd = "{} -mode {} {} {}".format(octarine, mode, bind_port_str, uuid)
+    log.info("Server: {}".format(server_cmd))
+    server = threading.Thread(target=subprocess.call, args=(shlex.split(server_cmd),), daemon=True)
+    server.start()
+
+    proxy = ('"http://127.0.0.1:$({} --client --port {})"'.format(octarine, uuid))
+    query_cmd = "sh -c 'curl --fail --proxy {} {}'".format(proxy, uri)
```
mellenburg: do you really need to be shelling out to curl here? Can't you just use python requests lib for this?

packages/dcos-integration-test/extra/test_applications.py
nlsun: The first reason is that the server and query must run on the same host (for transparent mode test) at least. Implementing the query logic and success condition in a mesos container is nontrivial.

I can change to using the subprocess popen instead of threading+call. Although that has the same issue with processes living forever.

packages/dcos-integration-test/extra/test_applications.py
nlsun: I could, this was just the first thing that came to mind.

packages/dcos-integration-test/extra/test_applications.py
nlsun: Ah, this is an artifact of when I was using it in more places.

packages/dcos-integration-test/extra/test_applications.py
nlsun: Is it good enough to just switch to subprocess for now?

packages/dcos-integration-test/extra/test_applications.py
mellenburg: How about this: your server is run in a container and the query is the health check for that container? That way you get marathon to do the retrying on a HTTP request logic for you

packages/dcos-integration-test/extra/test_applications.py
mellenburg: If we can't switch to container + health check to do server + retrying query, then let's just make some fixtures out of this (reach out to me and I can help here). Let's consider the container first as any test that can leave state outside of DC/OS is non trivial technical debt

packages/dcos-integration-test/extra/test_applications.py
nlsun: Yeah, I think I'll try a mesos container with command health check since i have it all as shell commands anyways

nlsun:
@mellenburg Changing to a mesos task, now it no longer uses subprocess or threading.

nlsun:
@mellenburg octarine tests are passing again

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -109,85 +112,83 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
...
+    proxy = ('http://127.0.0.1:$({} --client --port {})'.format(octarine, uuid))
+    curl_cmd = '''"$(curl --fail --proxy {} {})"'''.format(proxy, uri)
+    expected_output = '''"$(printf "{\\n    \\"pong\\": true\\n}")"'''
+    check_cmd = """sh -c '[ {} = {} ]'""".format(curl_cmd, expected_output)
+    log.info("Check: {}".format(check_cmd))
+
+    app, uuid = get_test_app()
+    app['requirePorts'] = True
+    app['cmd'] = server_cmd
+    app['healthChecks'] = [{
```
mellenburg: :+1:  way easier to comprehend

PR #1378, opened 2017-03-18
Clear out old irrelevent pkgpanda docs

mellenburg:
These were never relevant or are now incorrect


mellenburg:
@orsenthil @lingmann can I get a second :ship: on this?

PR #1380, opened 2017-03-19
Enable PID namespace isolator in mesos for the UCR.

klueska:
## High Level Description

This change enables the pid isolator for all containers launched by the UCR.

## Related Issues

  - [DCOS_OSS-790](https://jira.mesosphere.com/browse/DCOS_OSS-790) Always enable the pid isolator for the UCR in Mesos.

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -387,7 +387,7 @@ def calculate_config_yaml(user_arguments):
 def calculate_mesos_isolation(enable_gpu_isolation):
     isolators = ('cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,'
                  'docker/runtime,docker/volume,volume/sandbox_path,posix/rlimits,'
-                 'com_mesosphere_MetricsIsolatorModule')
+                 'namespaces/pid,com_mesosphere_MetricsIsolatorModule')
     if enable_gpu_isolation == 'true':
         isolators += ',cgroups/devices,gpu/nvidia'
     return isolators
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -3,7 +3,7 @@ import logging
 import uuid
 
 
-from test_util.marathon import get_test_app
+from test_util.marathon import get_test_app, get_test_app_in_ucr
 from test_util.recordio import Decoder, Encoder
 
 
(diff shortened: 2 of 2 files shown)


adam-mesos:
@mesosphere-mergebot bump-ee

klueska:
Not very likely -- it's only relevant for containers launched via the UCR
-- it also is likely the "expected" behavior and anything that relied on it
not being set should be considered a bug.

klueska:
Can I get a milestone added to this? @adam-mesos @spahl @lingmann

adam-mesos:
TBD which milestone/release. Let's discuss.

lingmann:
1.10 would be my vote for milestone. people have lived with the existing behavior for quite some time now so enabling it with the next release seems the safest option.

adam-mesos:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -141,3 +141,21 @@ def test_files_api(dcos_api_session):
...
             assert content, 'File {} should not be empty'.format(required_sandbox_file)
+
+
+def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
+    app, test_uuid = get_test_app_in_ucr()
+
+    ps_output_file = 'ps_output'
+    ps_cmd = 'ps ax -o pid= > {};'.format(ps_output_file)
+
+    app['cmd'] = ps_cmd + app['cmd']
```
adam-mesos: Why the `+ app['cmd']`? Is there some reason you want the python_test_server running after your ps command?
Seems like this could be a metronome job.

gen/calc.py
```diff
@@ -387,7 +387,7 @@ def calculate_config_yaml(user_arguments):
 def calculate_mesos_isolation(enable_gpu_isolation):
     isolators = ('cgroups/cpu,cgroups/mem,disk/du,network/cni,filesystem/linux,'
                  'docker/runtime,docker/volume,volume/sandbox_path,posix/rlimits,'
-                 'com_mesosphere_MetricsIsolatorModule')
+                 'namespaces/pid,com_mesosphere_MetricsIsolatorModule')
```
adam-mesos: Does the ordering here matter? If not, shouldn't we strive for alphabetical? Probably good to separate the `com_mesosphere` ones though.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -141,3 +141,21 @@ def test_files_api(dcos_api_session):
...
+
+    with dcos_api_session.marathon.deploy_and_cleanup(app):
+        marathon_framework_id = dcos_api_session.marathon.get('/v2/info').json()['frameworkId']
+        app_task = dcos_api_session.marathon.get('/v2/apps/{}/tasks'.format(app['id'])).json()['tasks'][0]
+
+        content = dcos_api_session.mesos_sandbox_file(
+            app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
+
+        assert len(content.split()) < 5, 'UCR app has more than than 5 processes running in its pid namespace'
```
adam-mesos: Technically, you're checking `< 5`, so an app with 5 processes would fail, which isn't what the error message says.
Either do `<= 5` or update the message to `5 or more processes`.
Also, s/than than/than/

gen/calc.py
klueska: Sometimes the order matters (but it doesn't for this one). These are already not in alphabetical order (look at`volume/sandbox_path` and networks/cni`, and `filesystem/linux` for example). I know that `filesystem/linux` has to come before the docker ones (and the gpu one which gets added later based on the config). I'm fine with moving it if you want, just not sure where it best fits.

adam-mesos: I believe we actually make a special exception in the code to reorder `filesystem/linux` to always be first.

PR #1395, opened 2017-03-23
Add HTTP routing to master for Azure

mellenburg:
Currently, the acs template does not provide the user with a route to the master for DC/OS UI usage. This was due to security requirements and previous issues with oauth. Now that we can oauth in the acs templates without issue, this patch will add routes for http/https when oauth is *available* and then will only add extra network security rules when oauthEnabled=true at deploy time.

Thus, with this PR, when someone deploys a oauthEnabled=false cluster, the traffic will not be routed to the master, but it can easily be turned on by opening the Network Security Group rule. Additionally, if a downstream DC/OS variant is built and if it uses `custom_auth`  the routes will be enabled.

## Checklist for all PR's

### DIFF
--- a/gen/azure/templates/acs.json
+++ b/gen/azure/templates/acs.json
@@ -764,7 +764,7 @@
         "frontendIPConfiguration": {
           "id": "[variables('masterLbIPConfigID')]"
         },
-        "frontendPort": "22",
+        "frontendPort": 22,
         "protocol": "tcp"
       }
     },
@@ -781,7 +781,7 @@
               "description": "Allow SSH",
               "protocol": "Tcp",
               "sourcePortRange": "*",
-              "destinationPortRange": "22",
+              "destinationPortRange": 22,
               "sourceAddressPrefix": "*",
               "destinationAddressPrefix": "*",
               "access": "Allow",
@@ -795,7 +795,7 @@
               "description": "Allow SSH",
               "protocol": "Tcp",
               "sourcePortRange": "*",
-              "destinationPortRange": "2222",
+              "destinationPortRange": 2222,
               "sourceAddressPrefix": "*",
               "destinationAddressPrefix": "*",
               "access": "Allow",
@@ -809,7 +809,7 @@
               "description": "HTTP Port 80",
               "protocol": "Tcp",
               "sourcePortRange": "*",
-              "destinationPortRange": "80",
+              "destinationPortRange": 80,
               "sourceAddressPrefix": "*",
               "destinationAddressPrefix": "*",
               {% switch oauth_available %}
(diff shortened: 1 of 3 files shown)


mellenburg:
@lingmann PTAL This is a little uglier than I wanted, but then again, templates like this weren't designed for conditionals in mind and only really support it with nested templates. I get around that by being a little tricky with maps, but it allows us to deploy different resources dynamically

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot  bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
Template artifacts produced in https://github.com/mesosphere/dcos-enterprise/pull/748 were verified to work so this PR is safe for downstream consumption

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

gen/build_deploy/azure.py
```diff
@@ -43,13 +43,21 @@ def validate_provider(provider):
     assert provider == 'azure'
 
 
+def calculate_public_route_to_master_available(oauth_available, custom_auth):
+    if oauth_available == 'true' or custom_auth == 'true':
+        return 'true'
+    else:
+        return 'false'
```
mellenburg: Its more that if you decide to roll your own, you can just set your config preferences to be oauth_availble = "false" in which case the generated template will have no auth in DC/OS and an HTTP/S route to master. There is no guarantee that "custom_auth" is true if oauth is unavailable

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@lingmann Can we move this forward? I have manually verified downstream. Still looking at a 100% success rate for launching with this template, so I am not worried about flakes

gen/azure/templates/acs.json
```diff
@@ -789,12 +802,96 @@
...
+      "dependsOn": [
+        "[variables('masterLbID')]"
+      ],
+      "properties": {
+        "backendPort": 443,
+        "enableFloatingIP": false,
+        "frontendIPConfiguration": {
+          "id": "[variables('masterLbIPConfigID')]"
+        },
+        "frontendPort": "443",
```
orsenthil: Let's have `int` type here. Otherwise it reads inconsistent while reviewing.

PR #1404, opened 2017-03-27
Admin Router: Deduplication of auth-related code

vespian:
## High Level Description

This PR:
* Updates following libraries:
  * lua-resty-http: 0.09 -> 0.10
  * lua-resty-jwt: 0.1.2 -> 0.1.9
  * lua-resty-hmac: ? -> commit 6a5b67
* adds `jwt-validators.lua` library in order to unify how Open and EE code handles JWT validation
* improve test-harness:
  * enable tests with forged JWTs
  * enable tests with missing claims
* refactor `auth.lua` into two parts:
  * `auth/common.lua` which contains code shared between both flavours
  * `auth/open.lua` which contains code used only in "Open" flavour
* implement shim-like approach for auth LUA code so that endpoints can share the same config at the `nginx.(master|agent).conf` level, but different behavriour for different flavours at the LUA code level
* in case of Open autzh, anything else than 200 was treated as "deny". This PR changes this behaviour so that 200 is "Permitted", 404 is "Deny" and anything else results in 500 sent to the client

## Related Issues

  - [DCOS-14583](https://jira.mesosphere.com/browse/DCOS-14583) Admin Router: Dedup of auth.lua

## Dependencies

EE counterpart: https://github.com/mesosphere/dcos-enterprise/pull/759


packages/adminrouter/extra/src/common/auth/common.lua
```diff
@@ -32,53 +28,36 @@ else
...
 
-    if SECRET_KEY == nil then
+    -- Refs:
+    -- https://github.com/openresty/lua-nginx-module#access_by_lua
+    -- https://github.com/SkyLothar/lua-resty-jwt
+
+    if secret_key == nil then
         ngx.log(ngx.ERR, "Secret key not set. Cannot validate request.")
-        return exit_401()
+        return nil, 401
```
mhrabovcin: Why is this function returning 2 values, while second one is always `401`?

packages/adminrouter/extra/src/common/auth/open.lua
```diff
@@ -0,0 +1,181 @@
...
+        return ngx.exit(ngx.HTTP_INTERNAL_SERVER_ERROR)
+    end
+    return uid
+end
+
+
+local function do_authn_and_authz_or_exit()
+    local uid = validate_jwt_or_exit()
+
+    -- Authz using authn :)
```
mhrabovcin: What does this comment means?

packages/adminrouter/extra/src/common/auth/common.lua
vespian: Having two distinct vars returned by the function is I think safer - if for some weird reason we do not catch error condition, and uid ends up nil - it will be immediately visible/easy to spot. Having second value being True/False instead of 401/HTTP code is not as readable I think. Thus I explicitly return "401" so that person reading the code can say - `Ah, so the result of the validation was "unauthenticated"!`. 

If/when we add more error conditions to the `validate_jwt`, we also get the default catch-all which verifies that the error is, in fact, one of the expected ones and thus we get even better error checking.

packages/adminrouter/extra/src/common/auth/open.lua
vespian: Short story: In Open user is authorised if and only if it's identity was transferred into IAM. So in a way - if you manage to authn, then you are also authzed.

Long story: https://mesosphere.slack.com/archives/G0W7PAKSS/p1490372266557793

packages/adminrouter/extra/src/common/auth/common.lua
gpaul: I'm curious, why do we still use exit_401 in some places, then?

vespian: @gpaul Open uses plain `exit_401`; EE also does some audit logging in the process for 401-ing the request. So having error condition in the `validate_jwt` is one thing (-> return value `~=  nil`), responding to it in a correct way is another.

Also, deduplication required introducing some abstraction layers so that code can be reused. Sometimes it makes code less readable, but still, I think that net benefit is positive.

I hope it clarifies a bit.

vespian:
@mesosphere-mergebot bump-ee

orsenthil:
@mesosphere-mergebot  label ship-it

PR #1375, opened 2017-03-17
merges minuteman code to navstar as navstar_l4lb

GoelDeepak:
## High Level Description

Fixes two issues
1. merges the minuteman code to navstar as navstar_l4lb
2. fixes iptables which were mangled by docker 1.12+

## Related Issues

  - [DCOS_OSS-902](https://jira.mesosphere.com/browse/DCOS_OSS-902) move minuteman code to navstar
 - [DCOS_OSS-980](https://jira.mesosphere.com/browse/DCOS_OSS-980) docker 1.12+ breaks virtual network

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [x] Test Results: [link to CI job test results for component]
  - [x] Code Coverage (if available): [link to code coverage report]

https://github.com/dcos/navstar/pull/64
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and Review Process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using github's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

Reviewers should be:
* Developers who understand the code being modified.
* Developers responsible for code that interacts with or depends on the code being modified.

It is best to proactively ask for 2 reviews by @mentioning the candidate reviewers in the PR comments area. The responsibility is on the developer submitting the PR to follow-up with reviewers and make sure a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

PR #1413, opened 2017-03-28
Serialize test_vip

mellenburg:
## High Level Description

Testing all network VIP configuration permutations at once is not a viable test. When it fails, it is near impossible to de-tangle what is broken and the logs are returned in a totally unorganized manner. This PR will:
- consolidate the code for generating consistent test server applications to make the tests calling it far more readable
 - serialize the test so that pytest can report results in an organized manner, thereby allowing others to debug from CI or to simply just focus on problem areas (rather than python threading issues). Unfortunately, this serialization will add ~30 minutes to tests

## Related Issue
https://jira.mesosphere.com/browse/DCOS-14950
https://jira.mesosphere.com/browse/DCOS_OSS-909
https://jira.mesosphere.com/browse/DCOS_OSS-910

## Checklist for all PR's

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -34,13 +34,11 @@ def pytest_collection_modifyitems(session, config, items):
 @pytest.fixture
 def vip_apps(dcos_api_session):
     vip1 = '6.6.6.1:6661'
-    test_app1, _ = get_test_app()
-    test_app1['portDefinitions'][0]['labels'] = {
-        'VIP_0': vip1}
-    test_app2, _ = get_test_app()
-    test_app2['portDefinitions'][0]['labels'] = {
-        'VIP_0': 'foobarbaz:5432'}
-    vip2 = 'foobarbaz.marathon.l4lb.thisdcos.directory:5432'
+    test_app1, _ = get_test_app(vip=vip1)
+    name = 'foobarbaz'
+    port = 5432
+    test_app2, _ = get_test_app(vip='{}:{}'.format(name, port))
+    vip2 = '{}.marathon.l4lb.thisdcos.directory:{}'.format(name, port)
     with dcos_api_session.marathon.deploy_and_cleanup(test_app1):
         with dcos_api_session.marathon.deploy_and_cleanup(test_app2):
             yield ((test_app1, vip1), (test_app2, vip2))
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -35,7 +35,7 @@ def pytest_collection_modifyitems(session, config, items):
 def vip_apps(dcos_api_session):
     vip1 = '6.6.6.1:6661'
     test_app1, _ = get_test_app(vip=vip1)
-    name = 'foobarbaz'
+    name = 'myvipapp'
     port = 5432
     test_app2, _ = get_test_app(vip='{}:{}'.format(name, port))
     vip2 = '{}.marathon.l4lb.thisdcos.directory:{}'.format(name, port)
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -3,7 +3,7 @@ import uuid
 
 import pytest
 
-from test_util.marathon import get_test_app
+from test_util.marathon import Container, get_test_app, Healthcheck, Network
 
 log = logging.getLogger(__name__)
 
@@ -32,12 +32,12 @@ def test_if_docker_app_can_be_deployed(dcos_api_session):
     deployed and accessed as expected.
     """
     dcos_api_session.marathon.deploy_test_app_and_check(
-        *get_test_app(network='BRIDGE', container_type='DOCKER', container_port=9080))
+        *get_test_app(network=Network.BRIDGE, container_type=Container.DOCKER, container_port=9080))
 
 
-@pytest.mark.parametrize("healthcheck", [
-    "HTTP",
-    "MESOS_HTTP",
+@pytest.mark.parametrize('healthcheck', [
+    Healthcheck.HTTP,
+    Healthcheck.MESOS_HTTP,
 ])
 def test_if_ucr_app_can_be_deployed(dcos_api_session, healthcheck):
     """Marathon app inside ucr deployment integration test.
@@ -46,7 +46,7 @@ def test_if_ucr_app_can_be_deployed(dcos_api_session, healthcheck):
     deployed and accessed as expected.
     """
     dcos_api_session.marathon.deploy_test_app_and_check(
-        *get_test_app(container_type='MESOS', healthcheck_protocol=healthcheck))
+        *get_test_app(container_type=Container.MESOS, healthcheck_protocol=healthcheck))
 
 
 def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(dcos_api_session):
@@ -63,7 +63,7 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(dcos_api_sessi
     When port mapping is available (MESOS-4777), this test should be updated to
     reflect that.
     """
-    app, test_uuid = get_test_app(container_type='MESOS')
+    app, test_uuid = get_test_app(container_type=Container.MESOS)
     dcos_api_session.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -3,7 +3,7 @@ import uuid
 
 import pytest
 
-from test_util.marathon import get_test_app, get_test_app_in_docker, get_test_app_in_ucr
+from test_util.marathon import get_test_app
 
 log = logging.getLogger(__name__)
 
@@ -31,7 +31,8 @@ def test_if_docker_app_can_be_deployed(dcos_api_session):
     Verifies that a marathon app inside of a docker daemon container can be
     deployed and accessed as expected.
     """
-    dcos_api_session.marathon.deploy_test_app_and_check(*get_test_app_in_docker(ip_per_container=False))
+    dcos_api_session.marathon.deploy_test_app_and_check(
+        *get_test_app(network='BRIDGE', container_type='DOCKER', container_port=9080))
 
 
 @pytest.mark.parametrize("healthcheck", [
@@ -44,7 +45,8 @@ def test_if_ucr_app_can_be_deployed(dcos_api_session, healthcheck):
     Verifies that a marathon docker app inside of a ucr container can be
     deployed and accessed as expected.
     """
-    dcos_api_session.marathon.deploy_test_app_and_check(*get_test_app_in_ucr(healthcheck))
+    dcos_api_session.marathon.deploy_test_app_and_check(
+        *get_test_app(container_type='MESOS', healthcheck_protocol=healthcheck))
 
 
 def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(dcos_api_session):
@@ -61,19 +63,7 @@ def test_if_marathon_app_can_be_deployed_with_mesos_containerizer(dcos_api_sessi
     When port mapping is available (MESOS-4777), this test should be updated to
     reflect that.
     """
-    app, test_uuid = get_test_app()
-    app['container'] = {
-        'type': 'MESOS',
-        'docker': {
-            # TODO(cmaloney): Switch to an alpine image with glibc inside.
-            'image': 'debian:jessie'
-        },
-        'volumes': [{
-            'containerPath': '/opt/mesosphere',
-            'hostPath': '/opt/mesosphere',
-            'mode': 'RO'
-        }]
-    }
+    app, test_uuid = get_test_app(container_type='MESOS')
     dcos_api_session.marathon.deploy_test_app_and_check(app, test_uuid)
 
 
@@ -121,9 +111,8 @@ def test_octarine(dcos_api_session, timeout=30):
     # reaches the proxy, the port is not used, and a request is made
     # to port 80.
 
-    app, uuid = get_test_app()
+    app, uuid = get_test_app(host_port=80)
     app['acceptedResourceRoles'] = ["slave_public"]
-    app['portDefinitions'][0]["port"] = 80
     app['requirePorts'] = True
 
     with dcos_api_session.marathon.deploy_and_cleanup(app) as service_points:
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -14,7 +14,7 @@ import retrying
 
 from test_helpers import expanded_config
 
-from test_util.marathon import get_test_app, get_test_app_in_docker, get_test_app_in_ucr
+from test_util.marathon import get_test_app
 
 
 log = logging.getLogger(__name__)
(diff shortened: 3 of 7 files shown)


mellenburg:
@spahl @lingmann This PR surfaces the large networking failures

test_util/marathon.py
```diff
@@ -14,91 +16,126 @@
...
+        host_constraint=None):
+    """ Creates an app definition for the python test server which will be
+    consistent (i.e. deployable with green health checks and desired network
+    routability). To learn more about the test server, see in this repo:
+    ../packages/dcos-integration-test/extra/util/python_test_server.py
+
+    Args:
+        host_port: port that marathon will use to route traffic into the
+            test server container. If set to zero, then marathon will assign
+            a port (which is referenced by index). If
```
BenWhitehead: `s/ If$//g`

test_util/marathon.py
```diff
@@ -14,91 +16,126 @@
...
+        assert container_type == 'DOCKER', 'BRIDGE network mode only supported for DOCKER container type'
+        if container_port is None:
+            # provide a dummy value for the bridged container port if user is indifferent
+            container_port = 8080
+    else:
+        assert container_port is None or container_port == host_port, 'Cannot declare a different host and '\
+            'container port outside of BRIDGE network'
+        container_port = host_port
+    if network == 'USER':
+        assert host_port != 0, 'Cannot auto-assign a port on USER network?'
```
BenWhitehead: Is this actually a question, it sounds more like a statement to me.

lingmann:
@mellenburg can you add to the description details about how this change will impact the runtime of these tests? I'd like to make sure that aspect of this change is very clear to networking team and anyone reviewing the PR from a high level. cc: @asridharan

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -31,7 +31,8 @@ def test_if_docker_app_can_be_deployed(dcos_api_session):
     Verifies that a marathon app inside of a docker daemon container can be
     deployed and accessed as expected.
     """
-    dcos_api_session.marathon.deploy_test_app_and_check(*get_test_app_in_docker(ip_per_container=False))
+    dcos_api_session.marathon.deploy_test_app_and_check(
+        *get_test_app(network='BRIDGE', container_type='DOCKER', container_port=9080))
```
nlsun: Can we use constants vs strings

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -40,228 +34,101 @@ def ensure_routable(cmd, host, port):
...
-    return app, uuid
-
-
-def mesos_vip_app(num, network, host, vip, ucr=False):
-    app = None
-    uuid = None
-    port = backend_port_st + num
-    if ucr is False:
-        app, uuid = get_test_app()
+def vip_app(container, network, host, vip, second):
```
nlsun: can we change "second" to something more flexible?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -40,228 +34,101 @@ def ensure_routable(cmd, host, port):
...
-    start_log_level = logging.getLogger('test_util.marathon').getEffectiveLevel()
-    # gotta go up to warning to mute it as its currently at info
-    logging.getLogger('test_util.marathon').setLevel(logging.WARNING)
-    yield
-    logging.getLogger('test_util.marathon').setLevel(start_log_level)
+def generate_vip_app_permutations():
+    """ Generate all possible network interface permutations for applying vips
+    """
+    permutations = []
+    for container in [None, 'MESOS', 'DOCKER']:
```
nlsun: the Big O gods curse us

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -40,228 +34,101 @@ def ensure_routable(cmd, host, port):
...
+    if named_vip:
+        vip = '/namedvip:7000'
+        vipaddr = 'namedvip.marathon.l4lb.thisdcos.directory:7000'
+    else:
+        vip = '1.1.1.7:7000'
+        vipaddr = '1.1.1.7:7000'
+
+    agents = list(dcos_api_session.all_slaves)
+    # make sure we can reproduce
+    random.seed(vip)
```
nlsun: why do we seed it with vip instead of just using 1 constant?

mellenburg: Good question. This was copy/paste from the old one. Will amend

lingmann:
Thanks for taking a look at this @nlsun I know you are swamped!

lingmann:
@nlsun @asridharan what is our path forward on this? Do we want to merge or wait for other improvements to the test suite before merging?

asridharan:
@lingmann I think as per @mellenburg these tests are flaky, so merging this as is won't work. @GoelDeepak or @nlsun will probably have to refactor them based on @mellenburg's comments and hopefully we can get the refactor in the next couple of weeks.

mellenburg:
@asridharan I do not understand your comment. What is currently being run in master is **broken**. It was muted because this test was flaky, but while it was muted, it entered a very bad state. Also, I do not  think you can state that "these tests are flaky". The original test_vip test was flaky because it was trying to launch the absolute maximum number of apps in such a way that lead to cascading failures amonst other reasons. This refactor avoids that problem by making this test painfully reliable with a serialized execution.

The reason we would want to merge this now is so that we can simply mute the few cases that are currently failing and then work on fixing the remaining cases, all the while keeping track of the good cases. Right now, if we do nothing, 99% of this test can be failing hard and it would pass undetected.

GoelDeepak:
These tests if run serially would take a very long time. Do we need to run all these tests? yes and no. We should ideally have two tests suits --- Sanity and Regress. Sanity one should have minimal necessary tests to qualify a build as a success. Regress one should have all the tests that we want to run. One option is to run sanity against each PR and Regress against each Train. We would have to do this anyhow because of the number of tests that keeps growing with each release.

asridharan:
@mellenburg I think I misunderstood the mail you had sent out on this particular PR. I was under the impression that after your re-factoring the test cases we still needed to go in and fix a bunch of tests cases to make them pass. But from your description it looks like only a small subset is failing after the serialization of the tests. I agree with @GoelDeepak and @lingmann a bigger problem is that these tests just take too long to run. We need make the test matrix really sparse. Had a chat with @lingmann me @GoelDeepak and @nlsun can sit down and select 8-10 tests that hopefully will give us a decent coverage so that we can reduce the scope of the tests dramatically. Probably we can do this some time later this week, or worst the coming week.

mellenburg:
@asridharan I absolutely disagree that the biggest problem here is that this takes too long to run. I do not know how this keeps getting mis-communicated so here is a summary:

- This test--as currently exists in master--is an absolute disaster. If you check current master results you will see centos/onprem and ubuntu/azure CONSISTENTLY failing 20 of the permutations.
- When I started this work, the goal was to get the test less flaky as it had begun to flake out at a crippling high rate. At this point the test was muted so that merges could continue
- While I worked on trying to make the test less flaky merges continued and one of those merges in the last month or two actually broke a huge swath of our networking capability and it went unnoticed because of the muted test
- In order for me to distinguish flakiness issues from test parallelization issues, I serialized this test
- Once the test was serialzied, there was no possible room for flakiness. This showed the failures lie squarely within networking.
- The test passes on AWS cloudformation, thus showing that the set of permutation is totally valid and that the refactoring work is not suspect

> it looks like only a small subset is failing after the serialization of the tests
No, **the same set is failing before and after this refactor**. That set is by no means "a small subset" -- its almost a third of the possible networking combinations.

I absolutely fail to see how the scope of the test is the problem: this test has clearly revealed that one of our flagship offereings (onprem + CentOS) is super broken.

If we want to consider making a smoke test and having this be a separate long running test, that is fine, but there are serious, non-test related issues that need to be addressed ASAP

GoelDeepak:
@mellenburg Thanks. That definitely helped in clearly some of the misunderstandings. That also indicate that merging this PR wouldn't help our current situation. However, I could take this PR and debug why the tests are failing. I believe we could take a call whether to merge this PR or not once we debug/fix all the failing tests. Agree?

mellenburg:
@GoelDeepak Considering whether to merge this:
* **DO** Merge This: We will gain 30 minutes to all our tests, but the network combinations that do currently work will not regress because they are now being exposed independently of what was considered a flake-out before
* **DO NOT** Merge This: tests continue to take the same amount of time. Tests will ignore ANY failures in VIP tests and the situation may even get worse while new failures are introduced while you are still debugging and fixing this test

I feel like the pros/cons are a tossup so I defer to whatever networking team (the test owners) think

asridharan:
@mellenburg so basically my earlier assumption was correct. Most of these tests are failing which means that either there is a bug in the networking code or the re-factoring is not problematic. We will look into this. This obviously can't be merged as it stands.

mellenburg:
> This obviously can't be merged as it stands

I do not see what is obvious. What is currently in master is broken and this exposes that by rearranging the results. The question is do you want slow tests that actually stop regression in what currently works or do you want faster tests that are completely silent until this is fixed?

asridharan:
@mellenburg as @GoelDeepak mentioned looks like there has been a misunderstanding. I was under the impression that after the refactoring you weren't sure whether the tests were buggy our it was actually a regression that was being captured in the tests and wanted someone from the networking team to take over the tests to figure out the distinction. From your last message it seems your suspicion is that this a regression in the networking code. @GoelDeepak has kindly volunteered to look into the failures and hopefully it will give us more clarity. As far as merging the PR is concerned, if its passing all the test suites (and we are confident failures are regression) and @lingmann and you are fine with the lag it introduces in PR integration I am fine with the merge.

One reason I suspect the tests are taking so much time are that for each combination we end up deploying a new set of apps, which is a relatively expensive operation. I was thinking that to speed up the test, we can deploy a set of apps for UCR and Docker on host, bridge and overlay, and then initiate commands for generating traffic for each of the 50 odd combination that we want to test, and collate the result. We don't necessarily need to deploy a new set of apps for each test case. That's the only argument I can present for holding of the merge to refactor and speed up the tests.

nlsun:
I am in favor of muting failing test cases now and merging. We get the short term benefit of preventing any regression on presently passing tests once this is merged. I believe that catching regressions is worth the additional test time.

We can then work on fixing the failing test cases as we go.

mellenburg:
Update on this test: I muted all the specific failures cases in Centos/Ubuntu but there were some pods/UCR tests that also failed and did so twice in a row, leading me to believe its not a flakeout. However they pass on AWS, so that makes it unlikely that its my test coding that is wrong.

In other news, I amended to use constants at @nlsun 's suggestion

GoelDeepak:
@lingmann @mellenburg We are good to merge this PR. The tests that are failing here would be fixed in https://github.com/dcos/dcos/pull/1375 PR.

mellenburg:
@GoelDeepak If that is the case, I will go ahead and rebase this PR on #1375 so we can actually verify that

mellenburg:
It looks like #1375 had the intended effect of fixing a lot of the broken cases, but it looks like it can still flake out:
https://teamcity.mesosphere.io/viewLog.html?buildId=624220&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2

GoelDeepak:
@mellenburg Thanks for rebasing this PR on top of #1375. This is strange. All the tests should have passed. Let me take a look at it today...

mellenburg:
The link above was for on onprem deployment, but we do two of those tests simultaneously and only one passed. Very strange flake out

mellenburg:
@nlsun Could I get please get your second pass/approval on this PR?

GoelDeepak:
@mellenburg I have removed the last leg of flakiness. If it is fine with you, could you rebase your PR on top of #1375 and give it a shot?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -40,228 +34,106 @@ def ensure_routable(cmd, host, port):
...
+                            # only DOCKER containers support BRIDGE network
+                            continue
+                        permutations.append((container, named_vip, same_host, vip_net, proxy_net))
+    return permutations
+
+
+@pytest.mark.slow
+@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
+@pytest.mark.parametrize('container,named_vip,same_host,vip_net,proxy_net', generate_vip_app_permutations())
+def test_vip(dcos_api_session, container, named_vip, same_host, vip_net, proxy_net):
```
orsenthil: This is an excellent candidate to to follow the documentation guidelines in code-style guide. The content is already present, please format it according to the style guidelines.

test_util/marathon.py
```diff
@@ -1,5 +1,8 @@
+""" Utilities for integration testing marathon in a deployed DC/OS clsuter
```
orsenthil: `clsuter` ? :)

test_util/marathon.py
```diff
@@ -14,91 +17,141 @@
 log = logging.getLogger(__name__)
 
 
-def get_test_app(custom_port=False):
+class Container(enum.Enum):
```
orsenthil: +1. good use case.

test_util/marathon.py
```diff
@@ -14,91 +17,141 @@
...
+    elif network == Network.BRIDGE:
+        app['healthChecks'][0]['port'] = container_port if \
+            healthcheck_protocol == Healthcheck.MESOS_HTTP else host_port
+    else:
+        # HOST or USER network with non-zero host port
+        app['healthChecks'][0]['port'] = host_port
+    if container_type != Container.NONE:
+        app['container'] = {
+            'type': container_type.value,
             # TODO(cmaloney): Switch to alpine with glibc
```
orsenthil: [request] - Please kill this `TODO` in this one.  My experiments (https://github.com/dcos/dcos/pull/1449) show that using a) docker official alpine image b) alpine+glib image seem to cause failures than the default of `debian:jessie`. I wanted to remove this `TODO` in a separate commit, but I am fine with this being removed as part of this refactor/change.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -34,13 +34,11 @@ def pytest_collection_modifyitems(session, config, items):
...
     vip1 = '6.6.6.1:6661'
-    test_app1, _ = get_test_app()
-    test_app1['portDefinitions'][0]['labels'] = {
-        'VIP_0': vip1}
-    test_app2, _ = get_test_app()
-    test_app2['portDefinitions'][0]['labels'] = {
-        'VIP_0': 'foobarbaz:5432'}
-    vip2 = 'foobarbaz.marathon.l4lb.thisdcos.directory:5432'
+    test_app1, _ = get_test_app(vip=vip1)
+    name = 'foobarbaz'
```
mellenburg: That is more confusing as there is a test_vip and this fixture is not used there. Rather, this fixture is used for resiliency/test_aws_cf.py. This isn't being used generically, but nonetheless let's compromise with `myvipapp`

orsenthil: sgtm.  (a careful name is what I desired, rather than a throwaway name)

mellenburg:
Kicked from train due to flake-outs in strict mode test. Flake-outs exist in current implementation but I will try to lock that down while we are already focused on this issue

amitaekbote:
@mellenburg is this good to go?

mellenburg:
This PR is finally good to go! The test has been refactored to do the named/unamed and different host/ same host cases all at once, and also no longer cares about simple deployment errors stemming from the test stumbling on its way to eventual consistency.

Enterprise PR has been manually verified as the current enterprise testing does not actually get to the networking test in CI

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -34,13 +34,11 @@ def pytest_collection_modifyitems(session, config, items):
...
-    test_app1['portDefinitions'][0]['labels'] = {
-        'VIP_0': vip1}
-    test_app2, _ = get_test_app()
-    test_app2['portDefinitions'][0]['labels'] = {
-        'VIP_0': 'foobarbaz:5432'}
-    vip2 = 'foobarbaz.marathon.l4lb.thisdcos.directory:5432'
+    test_app1, _ = get_test_app(vip=vip1)
+    name = 'myvipapp'
+    port = 5432
+    test_app2, _ = get_test_app(vip='{}:{}'.format(name, port))
```
orsenthil: [non-blocker] - please used {named}.formats.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -31,20 +31,22 @@ def test_if_docker_app_can_be_deployed(dcos_api_session):
...
     """
-    dcos_api_session.marathon.deploy_test_app_and_check(*get_test_app_in_docker(ip_per_container=False))
+    dcos_api_session.marathon.deploy_test_app_and_check(
+        *get_test_app(network=Network.BRIDGE, container_type=Container.DOCKER, container_port=9080))
 
 
-@pytest.mark.parametrize("healthcheck", [
-    "HTTP",
-    "MESOS_HTTP",
+@pytest.mark.parametrize('healthcheck', [
```
orsenthil: I don't know why these are indented.

test_util/marathon.py
```diff
@@ -14,91 +17,140 @@
...
+    Return:
+        (dict, str): 2-Tuple of app definition (dict) and app ID (string)
+    """
+    if network == Network.BRIDGE:
+        assert container_type == Container.DOCKER, 'BRIDGE network mode only supported for DOCKER container type'
+        if container_port is None:
+            # provide a dummy value for the bridged container port if user is indifferent
+            container_port = 8080
+    else:
+        assert container_port is None or container_port == host_port, 'Cannot declare a different host and '\
```
orsenthil: this assert is going to blow up. I assume you wanted log.debug ? (You seem to be falling back on default after this).

test_util/marathon.py
```diff
@@ -14,91 +17,140 @@
...
+        health_check_protocol: can be MESOS_HTTP or HTTP
+        vip: either named or unnamed VIP to be applied to the host port
+        host_constraint: string representing a hostname for an agent that this
+            app should run on
+
+    Return:
+        (dict, str): 2-Tuple of app definition (dict) and app ID (string)
+    """
+    if network == Network.BRIDGE:
+        assert container_type == Container.DOCKER, 'BRIDGE network mode only supported for DOCKER container type'
```
orsenthil: Is the `assert` stmt proper here?

test_util/marathon.py
mellenburg: This is intentional. There is no fallback to default. Only when you are using BRIDGE networking does it make any sense to declare a container port. In any other case, there is only one port to decide on an thus container_port MUST equal host_port.

This is an assertion error because if you are running into this, then that means you do not understand what you are trying to do and it should never work.

test_util/marathon.py
mellenburg: Yes. Absolutely. assert is intended to stop programmer errors last I checked and this is a great example; if a user wants a Network.BRIDGE app, then it MUST be DOCKER. Nothing else is valid or sensible. If DOCKER is not being used, then the programmer must not be knowledgeable about what they are doing or what this test application actually is.

packages/dcos-integration-test/extra/test_applications.py
mellenburg: That's how it was before so this will remain outside of scope in this case, but the next time this code gets touched this marginal style  error can be addressed if anyone cares to notice

packages/dcos-integration-test/extra/conftest.py
mellenburg: `Rationale: If the number of arguments are long. ` Name:port is not a long number of arguments and the format string is trivial. I prefer to not make python tedious when no benefit is provided.

orsenthil: Well, you are conveniently skipping past the first line.

```Prefer '{named}'.format will be over {}.format.```

* Once you get used to this (it is about practice),  then "consistently named" format strings are better.

test_util/marathon.py
GoelDeepak: @mellenburg unless I am missing something ... USER mode should also have a different container_port

orsenthil: @mellenburg - ok.

mellenburg: @GoelDeepak You might be missing something. Why SHOULD user mode have a different container port? That statement doesn't make sense to me. container_port != host_port only makes sense when you are bridging....

packages/dcos-integration-test/extra/conftest.py
mellenburg: I did not conveniently skip that line. The word is "prefer", not "must". The point of using the word "prefer" is that I think we can all agree that splitting hairs and arguing over python string formatting is a giant waste of everyone's time given the variety of python string formatting options available.

This is not a matter of practice. Its just common sense to make things readable. This string only serves to add a SINGLE COLON and you want to have at least 4 variable references and a function call to complete it. If anything, I would rather do:
```host + ':' + port```
Anyone off the street could make sense of that^^
but you think THIS is better?
```'{host}:{port}'.format(host=host, port=port)```
Now this is hopelessly python specific and will appear like chinese to anyone who is not familiar with wacky python paradigms
My point is, these kind of things take nontrival effort (e.g. deciding what silly name to use or making commits and having to rerun tests) and add almost no value whatsoever except for conforming only for the sake of conforming.

Let's try to stick to making this criticism when its more appropriate.

orsenthil: Good arguments. First off, the comment was non-blocker.  I wanted point out since we were using formatted strings.

> This is not a matter of practice. 

Consider this: I think, it just out of practice that anything you see in mergebot/mergebot repo mostly has `{named}`.format().  Even for single arguments. Some code-base just avoid empty `{}`, because named format is considered better. That's just the evolution of how python string formats came about. {0}, {} were predecessors of {named}.

Choosing the names for keywords has been straightforward. Just use the same name. 
It does not feel very odd.  The idea is we read the string entirely like "{host}:{port}" (and do not read the .format part) as opposed to "{}:{}"

>> Its just common sense to make things readable.

Agree.

>> Let's try to stick to making this criticism when its more appropriate.
 
Nod.  So, will avoid it for simpler formats. But when it really is necessary, let's enforce it.

PR #1417, opened 2017-03-29
AMI: XFS - store inode type in directory structure by default

lloesche:
## High Level Description

This turns on an XFS filesystem function where the inode type is stored in the directory structure. This is vital to the use of Overlay and a bug it wasn't there before.
Without it the Filesystem in containers can corrupt.

This setting is default beginning with EL 7.3.

## Related Issues

  - [DCOS-14872](https://jira.mesosphere.com/browse/DCOS-14872) Filesystem corruption in dcos-ui container.

### DIFF
--- a/cloud_images/centos7/dcos_vol_setup.sh
+++ b/cloud_images/centos7/dcos_vol_setup.sh
@@ -44,7 +44,7 @@ function main {
   until test -b "$device"; do sleep 1; echo -n .; done
   echo
   local formated
-  mkfs.xfs $device > /dev/null 2>&1 && formated=true || formated=false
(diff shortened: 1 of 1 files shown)


lloesche:
FYI - this fix is mainly so folks building their own AMIs who can't update to 7.3 yet get an updated volume creation script. For DC/OS 1.9+ this is already implicitly fixed by the AMI update to 7.3.

lingmann:

lingmann:
`integration-test/vagrant-bash` and `continuous-integration/jenkins/pr-head` definitely don't rely on any of this code, so can be ignored.

PR #1420, opened 2017-03-29
Add AWS bare cluster template

mellenburg:
## High Level Description
These templates are currently pulled from a magic bucket on S3. There is no reason why these should not just live in the repo where they can be updated in sync with DC/OS.

Additionally, the arbitrary names in the template are switched from CentOS to Bare as the intention of this template is to provide homogenous AMIs without DC/OS for use with the onprem installer.

## Related Issues
https://jira.mesosphere.com/browse/DCOS-11534

## Checklist for all PR's

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -13,12 +13,13 @@ DcosZenCfStack: Represents DC/OS  deployed from a zen template
 MasterStack: thin wrapper for master stack in a zen template
 PrivateAgentStack: thin wrapper for public agent stack in a zen template
 PublicAgentStack: thin wrapper for public agent stack in a zen template
-VpcCfStack: Represents a homogeneous cluster of hosts with a specific AMI
+BareClusterCfStack: Represents a homogeneous cluster of hosts with a specific AMI
 """
 import logging
 import time
 
 import boto3
+import pkg_resources
 import retrying
 from botocore.exceptions import ClientError
 
@@ -26,15 +27,13 @@ from test_util.helpers import Host, retry_boto_rate_limits, SshInfo
 
 log = logging.getLogger(__name__)
 
-VPC_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-cluster-template.json'
-VPC_EBS_ONLY_TEMPLATE_URL = 'https://s3.amazonaws.com/vpc-cluster-template/vpc-ebs-only-cluster-template.json'
-
 
 def template_by_instance_type(instance_type):
     if instance_type.split('.')[0] in ('c4', 't2', 'm4'):
-        return VPC_EBS_ONLY_TEMPLATE_URL
+        template = pkg_resources.resource_string('test_util', 'templates/vpc-ebs-only-cluster-template.json')
     else:
-        return VPC_TEMPLATE_URL
+        template = pkg_resources.resource_string('test_util', 'templates/vpc-cluster-template.json')
+    return template.decode('utf-8')
 
 
 def param_dict_to_aws_format(user_parameters):
@@ -57,7 +56,7 @@ def fetch_stack(stack_name, boto_wrapper):
             log.debug('Using Basic DC/OS Cloudformation interface')
             return DcosCfStack(stack_name, boto_wrapper)
     log.debug('Using VPC Cloudformation interface')
-    return VpcCfStack(stack_name, boto_wrapper)
+    return BareClusterCfStack(stack_name, boto_wrapper)
 
 
 class BotoWrapper():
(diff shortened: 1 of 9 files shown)


mellenburg:
@orsenthil @branden PTAL

test_util/aws.py
```diff
@@ -99,6 +97,18 @@ def create_stack(self, name, template_url, parameters, deploy_timeout=60):
             # so that we may pass parameters directly from yaml (which parses numbers as non-strings)
             Parameters=param_dict_to_aws_format(parameters))
 
+    def create_stack_from_string(self, name, template, parameters, deploy_timeout=60):
```
branden: This duplicates a lot of `create_stack()`. IMO it'd make more sense to add a `template` arg to `create_stack()`, since that better reflects what the actual CloudFormation API looks like.

```
def create_stack(self, name, parameters, template_url=None, template=None, deploy_timeout=60):
    ...
```

I _think_ you can pass all those params straight through to boto's `create_stack()` and it'll validate that you passed exactly one of the template args.

mellenburg:
@orsenthil What exactly do you want tested locally? The PR checks test this exact thing....

orsenthil:
@mellenburg  - I missed the part that it is exercised already as part of this unit-test run.  For some reason, I thought, this change will start getting exposed and exercised only after it lands in master.

Ignore my previous comment. LGTM.

test_util/aws.py
branden: By "pass straight through" I meant defer choosing/validating `TemplateBody` and `TemplateURL` to boto. See https://github.com/boto/boto/blob/abb38474ee5124bb571da0c42be67cd27c47094f/boto/cloudformation/connection.py#L229

branden: > You could literally pass the args straight through with **kwargs, and then the caller of this method has to choose between providing the TemplateURL arg or the TemplateBody arg. That kind of defeats the purpose of the method.

Not sure how -- that's the API as provided by CloudFormation. One function that takes either `TemplateBody` or `TemplateURL`.

> Alternatively, you can't just default both TemplateBody and TemplateURL as None and then expect this to work when you fill one of them in.

I'm doubtful, since that's how boto's `create_stack()` is implemented: https://github.com/boto/boto/blob/abb38474ee5124bb571da0c42be67cd27c47094f/boto/cloudformation/connection.py#L306

mellenburg: Is it because this is all boto3 code and you are linking to boto?

mellenburg: The above error message was literally just copy pasted from my terminal as I tried implementing this by providing both those args

branden: Ah my bad. Default to empty string then? Trying to locate the implementation of boto3's `create_stack()`.

mellenburg: Nope, cannot default to empty string either, because if TemplateBody is given (even as an empty string) it will be taken over TemplateURL:
```
Invalid length for parameter TemplateBody, value: 0, valid range: 1-inf
```

boto3 can just be viewed as a wrapper for the XML HTTP API. Or rather, the documentation, and the library itself are just wrappers around schemas, so all paths will likely just lead back here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html

branden: Ah, so they're generating the library from schema definitions. That's slick. Thanks for the explanation. 👍

test_util/aws.py
```diff
@@ -83,21 +82,26 @@ def delete_key_pair(self, key_name):
...
+            'Capabilities': ['CAPABILITY_IAM'],
             # this python API only accepts data in string format; cast as string here
             # so that we may pass parameters directly from yaml (which parses numbers as non-strings)
-            Parameters=param_dict_to_aws_format(parameters))
+            'Parameters': param_dict_to_aws_format(parameters)}
+        if template_body is not None:
+            assert template_url is None, 'tempate_body and template_url cannot be supplied simultaneously'
+            args['TemplateBody'] = template_body
+        else:
+            args['TemplateURL'] = template_url
```
branden: Should we check that one of these args passed? Looks like if I don't pass `template_url` or `template_body`, it'll fail with a type error for `template_url`, which could be confusing. "I didn't even pass template_url!"

orsenthil:
The integration test failure are real.

```
test setup failed
boto_wrapper = <test_util.aws.BotoWrapper object at 0x7f94d2d21ba8>

    @pytest.fixture(scope='session')
    def dcos_stack(boto_wrapper):
        """ Works with either Zen or Simple cloud formations
        """
        stack = test_util.aws.fetch_stack(os.environ['AWS_STACK_NAME'], boto_wrapper)
>       if isinstance(stack, test_util.aws.VpcCfStack):
E       AttributeError: module 'test_util.aws' has no attribute 'VpcCfStack'

```

orsenthil:
@mesosphere-mergebot label work in progress

falseperfection:
Looks like `integration-test/aws-deploy` is broken, not flaky. Code removed `VpcCfStack`, but test still relies on it.

mellenburg:
@falseperfection my bad for not doing due diligence there. Sorry for having to deal with that. Fixed now and can be shipped on the next train after.

PR #1432, opened 2017-04-04
[DCOS_OSS-888] Document running integration tests with DC/OS Docker

adamtheturtle:
## High Level Description

This change provides clear instructions for using DC/OS Docker to run integration tests. It is aimed in particular at developers wanting to run integration tests, and for my needs over the last few days - to test changes to integration tests.

## Related Issues

  - [DCOS_OSS-888](https://jira.dcos.io/browse/DCOS_OSS-888) Document how to run integration tests with DC/OS Docker.

## Checklist for all PR's

  - [X] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: There is no test because there is no code, this is a docs only change.

## Notes

This could perhaps be two PRs. I would be happy to split this up if requested.

I tried to make this as similar as possible to the above section.
I wanted to use `test_env.export` but I couldn't find it in the Docker container.

I am wary that this is not tested and it might not be stable.
That is, the integration tests run on the CI server might change without this documentation changing.


adamtheturtle:
I'm not quite sure who to tag for a review, so I will tag the reviewers of the related https://github.com/mesosphere/dcos-enterprise/pull/786: @vespian , @mhrabovcin .

adamtheturtle:
Closing to implement changes suggested in https://github.com/mesosphere/dcos-enterprise/pull/786.

adamtheturtle:
Huh, this doesn't get all of the tests to pass. I will debug later. For example, `test_enterprise_if_marathon_pods_can_be_deployed_with_mesos_containerizer` fails.

README.md
```diff
@@ -85,6 +85,12 @@ Integration tests can be run on any deployed DC/OS cluster. For installation ins
 
 Integration tests are installed via the [dcos-integration-test](./packages/dcos-integration-test/) Pkgpanda package.
 
+Integration test files are stored on the DC/OS master node at `/opt/mesosphere/active/dcos-integration-test`.
+Therefore, in order to test changes to test files, move files from `packages/dcos-integration-test/extra/` in your checkout to `/opt/mesosphere/active/dcos-integration-test` on the master node.
+
+The canonical source of the test suite's results is the continuous integration system.
```
adamtheturtle: This note is the result of a discussion with @jgehrcke .

I wrote this documentation as I used the new instructions to run the IAM tests for the enterprise
After writing this documentation, I found that some tests failed.
For example `test_enterprise_if_marathon_pods_can_be_deployed_with_mesos_containerizer`.

I discussed the following options with @jgehrcke:

* Find and fix the problems with the tests now
* Submit as is
* Don't add documentation now as the tests aren't working, and document this at a later date

The conclusion was that:
* This work was not scheduled, and I did it as a nice-to-have, so the first option (which may take quite some time) is not best
* The third option leaves new developers like me in a worse position than with these docs. For context, it seems that most of the security team is using a similar but custom system maintained by each developer.

I am open to the idea of putting this in the DC/OS Enterprise private repository until this is a fully-working solution, and in fact I have a corresponding PR there.

adamtheturtle: FWIW I plan to take a brief look with @mellenburg to see if we can at least roughly scope the problems, probably in order to create a follow-up JIRA.

adamtheturtle:

adamtheturtle:

README.md
vespian: You may also want to have a look at TeamCity's configuration - some tests are muted, some are disabled for particular builds. I think that we should be able to filter out at least some of the flaky tests this way.

README.md
```diff
@@ -85,6 +85,12 @@ Integration tests can be run on any deployed DC/OS cluster. For installation ins
...
+There may be differences between the results of running the integration tests as described in this document and the results given by the continuous integration system.
```
vespian: What kind of differences? As a reader/consumer of this doc - I find this puzzling :)

README.md
adamtheturtle: Thank you - my preference is that this is merged as-is with a follow up JIRA issue created to make this match TeamCity. Are you OK with that?

README.md
adamtheturtle: What would you like it to say? I guess that tests will fail here but not on CI, but I didn't want to rule out the opposite.

vespian: In my mind, this may be also:
* results are presented differently
* results are incomplete
* results differ as per some stuff fails in CI while it does not in case of local run
* results differ as per some tests are not run in the CI

Do you see what I mean?

adamtheturtle: Yes and I will change as appropriate.

amitaekbote:

orsenthil:
@adamtheturtle  - thanks, good work on this documentation.

PR #1434, opened 2017-04-05
Ignore task failures for viplisten in upgrade test

mellenburg:
## High Level Description

Although the necessary VIP should have been ensured to exist
by polling marathon deployments for the app to be healthy (which
requires the VIP to be established), the viptalk app still fails
due to inability to connect to the VIP. This is a relatively minor
error if the VIP eventually works within the app deployment timeout.
Networking latency relating to app deployment should not cause
upgrade tests to fail out.

## Related Issues
https://jira.mesosphere.com/browse/DCOS-14660

## Checklist for all PR's

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -307,7 +307,9 @@ class VpcClusterUpgradeTest:
 
             dcos_api.marathon.deploy_app(viplisten_app)
             dcos_api.marathon.ensure_deployments_complete()
-            dcos_api.marathon.deploy_app(viptalk_app)
+            # viptalk app depends on VIP from viplisten app, which may still fail
+            # the first try immediately after ensure_deployments_complete
+            dcos_api.marathon.deploy_app(viptalk_app, ignore_failed_tasks=True)
             dcos_api.marathon.ensure_deployments_complete()
 
             dcos_api.marathon.deploy_app(healthcheck_app)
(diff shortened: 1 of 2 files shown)


test_util/test_upgrade_vpc.py
```diff
@@ -307,7 +307,9 @@ def setup_cluster_workload(self, dcos_api: DcosApiSession, healthcheck_app: dict
 
             dcos_api.marathon.deploy_app(viplisten_app)
             dcos_api.marathon.ensure_deployments_complete()
-            dcos_api.marathon.deploy_app(viptalk_app)
+            # viptalk app depends on VIP from viplisten app, which may still fail
```
orsenthil: If ignore any task failure now, and proceed with the upgrade (due to latency), we have to make sure that the task failure is not persistent after the upgrade.  

```
        self.verify_apps_state(dcos_api_upgrade, dns_app)
```

could be modified to include the viplisten and talkapps?

mellenburg: Huh? Those apps are already tested by that method if I am not mistaken. Look again. dns_app is just the only one passed as an arg. Everything else doesnt need to have the app definition passed because the test harness gets the original task id's at the beginning and checks them at the end for all 4 apps. With the dns_app there are some special value that shouldn't be redeclared so the definition is passed

orsenthil: Yup!.  My comment was a mistake. I verified the apps and it's state are verified after the upgrade.

```
            def marathon_app_tasks_survive_upgrade():
            def test_mesos_task_state_remains_consistent():
```

PR #1436, opened 2017-04-06
Add ConnectionError retrying Mixin

mellenburg:
## High Level Description
Modifies the Marathon API client in test_util to ignore Connection and Timeout errors

## Related Issues
https://jira.mesosphere.com/browse/DCOS-14969

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: DEFLAKE WORK

### DIFF
--- a/test_util/helpers.py
+++ b/test_util/helpers.py
@@ -22,6 +22,8 @@ from botocore.exceptions import ClientError, WaiterError
 Host = namedtuple('Host', ['private_ip', 'public_ip'])
 SshInfo = namedtuple('SshInfo', ['user', 'home_dir'])
 
+log = logging.getLogger(__name__)
+
 
 # Token valid until 2036 for user albert@bekstil.net
 #    {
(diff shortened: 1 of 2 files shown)


test_util/helpers.py
```diff
@@ -170,6 +170,27 @@ def options(self, *args, **kwargs):
...
+        if isinstance(exception, ex):
+            return True
+    return False
+
+
+class RetryCommonHttpErrorsMixin:
+    """ Mixin for ApiClientSession so that random disconnects from
+    network instability do not derail entire scripts
+    """
+    def api_request(self, *args, retry_timeout=60, **kwargs):
```
colin-msphere: I am not sure the unit for `retry_timeout`, could this be added to the method doc or rename to `retry_timeout_UNIT`?

orsenthil: @colin-msphere  - the readability of this code comes from  the context. The default unit of the `@retrying` library works with is `milliseconds`. So, we multiply by `1000` for the seconds unit.

That said, I am in favor of more explicit units too. My preference will be in comments rather than in variable in names.Or using some sort of Units Library, [like this](https://github.com/twitter/commons/blob/master/src/python/twitter/common/quantity/__init__.py#L22)

Here, if we rename the variable to having `_UNIT`, it might stand out as an odd duck.

mellenburg: +1 for adding docstring

mellenburg:
@lingmann I will go ahead and add a logging statement. Good tip

PR #1437, opened 2017-04-06
Only run rexray tests when config shows aws env

mellenburg:
## High Level Description
Right now, people need to declare `-m 'not ccm'` when running tests that are outside of AWS. This is wrong for two reasons:
- we shouldnt assume that we are in an AWS environment
- we shouldnt have anything with the label "ccm" as it does not really add any description other than it might be involved with CI

This PR also cleans up the expanded config helper by consolidating it across the codebase

## Checklist for all PR's

### DIFF
--- a/packages/dcos-integration-test/extra/test_endpoints.py
+++ b/packages/dcos-integration-test/extra/test_endpoints.py
@@ -6,7 +6,7 @@ import pytest
 from requests.exceptions import ConnectionError
 from retrying import retry
 
-from pkgpanda.util import load_json
+from test_helpers import expanded_config
 
 
 def test_if_dcos_ui_is_up(dcos_api_session):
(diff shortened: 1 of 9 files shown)


packages/dcos-integration-test/extra/test_endpoints.py
```diff
@@ -6,7 +6,7 @@
 from requests.exceptions import ConnectionError
 from retrying import retry
 
-from pkgpanda.util import load_json
+from test_helpers import expanded_config
```
falseperfection: +1 for centralizing the config over load_json() calls throughout

packages/dcos-integration-test/extra/test_misc.py
```diff
@@ -17,9 +17,6 @@ def test_load_user_config():
 
 
 def test_load_expanded_config():
-    with open("/opt/mesosphere/etc/expanded.config.json", "r") as f:
-        expanded_config = json.load(f)
-
```
branden: This test's name is a little misleading now that it's no longer actually loading the config. Removing "load" from the name would help clarify.

packages/dcos-integration-test/extra/test_rexray.py
```diff
@@ -4,8 +4,12 @@
 
 import pytest
 
+from test_helpers import expanded_config
 
-@pytest.mark.ccm
+
+@pytest.mark.skipif(
+    not (expanded_config['provider'] == 'aws' or expanded_config['platform'] == 'aws'),
```
orsenthil: Yeah, this is better.

PR #1439, opened 2017-04-07
Proposal: Python Style Guide for DCOS

orsenthil:
## High Level Description

Python Style Guide proposal for DCOS Code base.

Readable Version: https://github.com/orsenthil/dcos/blob/style_guide/docs/styleguides/python.md 

## Related Issues

  - [DCOS_OSS-886](https://jira.dcos.io/browse/DCOS_OSS-886) Create style guide documentation for community in dcos/dcos
  - [DCOS-14924](https://jira.mesosphere.com/browse/DCOS-14924) Define and adopt a python style guideline for dcos/dcos

---
This change is [Reviewable](https://reviewable.io/reviews/dcos/dcos/1439)

### DIFF
--- a/docs/styleguides/python.md
+++ b/docs/styleguides/python.md
@@ -1,0 +1,259 @@
+# DC/OS Python Style Guide
+###### Version 0.1
+###### April-6-2017
+
+For Python source code within DC/OS project, we prefer to keep [Google Python Style Guide Revision 2.59] as the base and
+**extend** it with some changes that will make DC/OS codebase look modern, consistent and easy to follow.
+
+##### Rationale
+
+With any style guide, the primary requirement is to adopt one and follow it consistently. With our choices below, we
+can use tools available in the python ecosystem to our advantage, enforce consistency in our codebase, while creating
+and publishing automatically generated documentation from the code itself.
+
+##### Versioned document
+
+Styleguide _conventions_ are not set in stone. We can improve upon this or remove inconvenient styles by creating a PR,
+discussing the merits or demerits to arrive at a consensus. Developers who contribute code to the DC/OS repo are the
+main stake holders of this style guide.
+
+##### Python Version Requirement
+
+This style guide is written for Python3.5 and above. As new features of the language get standardized, this style guide
+can be revised to adopt new features.
+
+#### Base Style Guide
+
+* [Google Python Style Guide Revision 2.59]
+
+
+#### DC/OS Specific extensions to Base Style Guide
+
+* Import Requirements
+    * Adopt the base style guide preferences.
+
+        ```
+        Use import x for importing packages and modules.
+        Use from x import y where x is the package prefix and y is the module name with no prefix.
+        Use from x import y as z if two modules named y are to be imported or if y is an inconveniently long name.
+        ```
+    * Use `__init__.py` only to represent python packages.
+        * Prefer empty `__init__.py` modules. Keep definitions in submodules.
+
+    _Rationale_: Keeping `__init__.py` empty helps in avoid unnecessary imports when you import a submodule from
+    package. Plus,it helps in navigating the code base you know that an object is strictly defined in a submodule and
+    not aliased in `__init__.py`
+
+* Import formatting
+
+    * The types of imports should be separated by single line and order should be.
+
+        1. standard library imports.
+        2. third-party imports.
+        3. application specific imports.
+
+    * Each group of these imports should be sorted.
+
+    _Rationale_:  Readablity is the reason for this separation.
+
+For Example
+
+```
+import json
+import logging
+from subprocess import CalledProcessError
+from typing import Callable, List
+
+import retrying
+from retrying import RetryError, retry
+
+import test_util.aws
+import test_util.cluster
+from pkgpanda.util import logger, write_string
+from teamcity.messages import TeamcityServiceMessages
+from test_util.cluster import run_bootstrap_nginx
+from test_util.helpers import (CI_CREDENTIALS,
+                               marathon_app_id_to_mesos_dns_subdomain,
+                               random_id)
+```
+
+
+* [Packages](https://google.github.io/styleguide/pyguide.html?showone=Packages#Packages)
+    * Import each module using the full pathname location of the module.
+    * No Relative Imports.
+
+  _Rationale_ - Relative imports are hard to follow and hard to write in consistent manner. And if you are using IDEs
+  to navigate the code base, it is much easier to get your code navigation right with absolute imports from the source
+  root.
+
+* Maximum line length is 120 characters.
+    * Following are the exceptions to this rule. These types can be of arbitrary length.
+        * URLs in comments.
+        * Any sequence of chars, like embedded tokens which would better be in a single line.
+
+   _Rationale_: At the time this style guide was written, the existing code base followed 120 column width guideline.
+   We are keeping up with that.
+
+* Indentation
+
+    * Use 4-space indentation.
+    * For long list of arguments, always use double-indented
+      hanging-indentation. See example below.
+    * The hanging indentation should be followed by a single blank line.
+
+    _Rationale_: This improves the readability of the code. White space is significant in Python, the double
+    indentation is to signify that this code block is indented for visual purpose and is continuation segment. This is
+    separate from the 4-space indented context-block of program flow.
+
+
+    * exceeds 120 chars # BAD.
+
+    ```
+    def method():
+
+        installer.setup_remote(tunnel=bootstrap_host_tunnel, installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh', download_url=installer_url)
+    ```
+
+    * single indented - Not recommended. # BAD.
+    * Note hanging indentation aligns with the logical indentation of the program.
+
+    ```
+    def method():
+        if installer.prereqs.ok:
+            print("pre-reqs satisfied")
+
+        installer.setup_remote(
+            tunnel=bootstrap_host_tunnel,
+            installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh',
+            download_url=installer_url)
+
+        if installer.setup.ok:
+            print("remote setup successful.")
+        else:
+            print("remote setup successful.")
+    ```
+
+   * double indented - Recommended # GOOD. Note the visual clarity.
+
+    ```
+    def method():
+        if installer.prereqs.ok:
+            print("pre-reqs satisfied")
+
+        installer.setup_remote(
+                tunnel=bootstrap_host_tunnel,
+                installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh',
+                download_url=installer_url)
+
+        if installer.setup.ok:
+            print("remote setup successful.")
+        else:
+            print("remote setup successful.")
+    ```
+
+
+* Adding Types. Since our code-base is Python3.5+ compatible, we have the choice of introducing types, validating
+  with mypy.
+
+```
+        class VpcClusterUpgradeTestDcosApiSessionFactory:
+
+            def apply(
+                    self,
+                    dcos_url: str,
+                    masters: List[str],
+                    public_masters: List[str],
+                    slaves: List[str],
+                    public_slaves: List[str],
+                    default_os_user: str) -> DcosApiSession:
+```
+
+* Reviewers will request the author to write types for the method parameters.
+
+    _Rationale_: We will start running mypy type checker as part of build process.
+
+
+* Shebang Rule.
+    * We will consistently use the shebang line in our scripts.
+        * `#!/usr/bin/env python3`
+        OR
+        * `#!/usr/bin/env python2` in our scripts.
+
+    _Rationale_ : Across platforms, and different python versions in a given platform, this is the most consistent way
+    to indicate the interpreter required for your python program.
+
+* Use the `format` method
+    * Prefer `'{named}'.format` will be over `{}.format`.
+        * Rationale: If the number of arguments are long. `{named}.format` helps in readability.
+    * Use your best judgement to decide between + and % (or format) when applicable.
+    * This style guide does not recommend [f-strings](https://www.python.org/dev/peps/pep-0498/) as `f-strings` are
+      still very early feature in Python, yet to be seen wide adoption, and it is available only from
+      Python 3.6 onwards.
+
+     _Rationale_: When there are multiple format specifiers used in a string, `{name1} {name2}` formats are easier
+    to read than `{}{}` (empty) or `{0}{1}` (indexed) formats.
+
+
+
+* TODO Comment Format
+    * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
+    * Format required will be `# TODO(active-developer): DCOS_OSS-1234 - Brief one line description.`
+
+    ```
+    # TODO(skumaran): DCOS_OSS-952 - Create a flake8 plugin to enforce TODO format.
+    ```
+
+    _Rationale_: With every todo, there is active developer associated and JIRA ticket possibly has more discussion
+    for the TODO.
+
+* [Comments and Docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).
+    * All public classes, methods, functions should be documented.
+    * Private classes, which should be indicated by `_leading_underscore` in the name need not be documented.
+    * Follow the base style guidelines to adopt the style that explains arguments, return value and exceptions.
+
+```
+class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
+    def __init__(
+            self,
+            dcos_url: str,
+            masters: List[str],
+            public_masters: List[str],
+            slaves: List[str],
+            public_slaves: List[str],
+            default_os_user: str,
+            auth_user: Optional[DcosUser]) -> None:
+
+        """Proxy class for DC/OS clusters.
+
+        This class provides facilities to interact with the DC/OS Cluster launched in the integration tests.
+
+
+        Args:
+            dcos_url: address for the DC/OS web UI.
+            masters: list of Mesos master advertised IP addresses.
+            public_masters: list of Mesos master IP addresses routable from
+                the local host.
+            slaves: list of Mesos slave/agent advertised IP addresses.
+            default_os_user: default user that marathon/metronome will launch tasks under
+            auth_user: use this user's auth for all requests
+                Note: user must be authenticated explicitly or call self.wait_for_dcos()
+
+        Returns:
+            None
+
+        Raises:
+            AssertionError: On empty values.
+            Exception: When a given node is not recognized within the DC/OS Cluster.
+        """
+```
+
+   _Rationale_: We can use sphinx documentation generator. Tools like [sphinx-napolean plugin](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) can be used to automatically generate api documentation with the above chosen convention.
+
+#### Style Enforcements.
+
+* tox will be configured to run pycodestyle/ pylint to enforce the style requirements.
+* With python code type-hint specified, we will use `mypy` in the test suite to check for type incompatibility problems
+  before build process.
+* We will utilize tools like isort, yapf that assist in adherence to coding style guidelines.
+
+[Google Python Style Guide Revision 2.59]: https://google.github.io/styleguide/pyguide.html


orsenthil:
@mesosphere-mergebot label this as ready for review

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Import Requirements
+    * Adopt the base style guide preferences.
+
+        ```
+        Use import x for importing packages and modules.
+        Use from x import y where x is the package prefix and y is the module name with no prefix.
+        Use from x import y as z if two modules named y are to be imported or if y is an inconveniently long name.
+        ```
+    * Use `__init__.py` only to represent python packages.
```
orsenthil: Here is a very good explaination written for adopting this style: https://github.com/Khan/style-guides/blob/master/style/python.md#__init__py

jsrodman: Not sure if I fully understand this suggestion or the referenced article.  I feel the smallest __init__.py is usually the best one, but that it can be a convenient place to declare the primary interfaces to a package constructed out of many submodules that aren't interesting to the consumer.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+    * Reviewers will request the author to write types for the method parameters.
+    * We will start running mypy type checker as part of build process.
+
+* Shebang Rule.
+    * We will consistently use the shebang line in our scripts.
+        * `#!/usr/bin/env python3`
+        OR
+        * `#!/usr/bin/env python2` in our scripts.
+
+* Use the `format` method
```
jsrodman: One caveat with .format is that user-controlled strings can execute arbitrary code.  % interpolation is somewhat more restrained.  If .format is the default, we should take steps to be sure the base strings do not ever contain externally provided data.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+    * Prefer `'{named}'.format` will be over `{}.format`.
+        * Rationale: If the number of arguments are long. `{named}.format` helps in readability.
+    * Use your best judgement to decide between + and % (or format) when applicable.
+
+* TODO Comment Format
+    * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
+    * Format required will be `TODO (active-developer) [JIRA] - Brief one line description`
+    * A plugin can be used to enforce the TODO format requirement.
+
+* [Comments and Docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).
```
adamtheturtle: The biggest issue that I have with the code documentation that I have seen in DC/OS is not that the docstrings are inconsistent. It is that the comments and docstrings are missing. A guideline on when these are needed would be useful. My personal preference: Every non-nested function, class, method should have a docstring - and that includes tests.

docs/styleguides/python.md
adamtheturtle: Perhaps comment on f-strings and their place in codebases which support them (i.e. 3.6+).

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+        ```
+        Use import x for importing packages and modules.
+        Use from x import y where x is the package prefix and y is the module name with no prefix.
+        Use from x import y as z if two modules named y are to be imported or if y is an inconveniently long name.
+        ```
+    * Use `__init__.py` only to represent python packages.
+        * Avoid importing of objects in `__init__.py` for readability.
+* Import formatting
+
+    * The types of imports should be separated by single line and order is should be.
```
adamtheturtle: I would prefer to not have this burden on authors and reviewers as there is a good tool which deals with it - [isort](https://pypi.python.org/pypi/isort). One pass of `isort` can be run on each codebase to automatically sort imports. Then editor plugins and CI can enforce sorting.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+                    masters: List[str],
+                    public_masters: List[str],
+                    slaves: List[str],
+                    public_slaves: List[str],
+                    default_os_user: str) -> DcosApiSession:
+
+    * Reviewers will request the author to write types for the method parameters.
+    * We will start running mypy type checker as part of build process.
+
+* Shebang Rule.
```
adamtheturtle: It would be useful to explain why this is. I believe it is the only consistent way across platforms to guarantee which Python version is required. In general the "why" behind changes is useful because later we can know whether that reason is still valid / outdated / it was always dubious, and we can change the guide accordingly.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+* [Comments and Docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).
+    * Follow the base style guidelines to adopt the style that explains arguments, return value and exceptions.
+
+```
+class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
+    def __init__(
+            self,
+            dcos_url: str,
+            masters: list,
```
adamtheturtle: FWIW these types would be better as `List[<some_type>]` e.g. `List[str]`.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+            public_masters: list,
+            slaves: list,
+            public_slaves: list,
+            default_os_user: str,
+            auth_user: Optional[DcosUser]):
```
adamtheturtle: If you are going to have type hints (wahoo!) then I think have them for return values.
In this case `-> None` and don't require `Returns` in the docstring when it is `None`.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+        """
+```
+
+* We can use sphinx documentation generator. Tools like [sphinx-napolean plugin](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) can be used to automatically generate api documentation with the above chosen convention.
+
+#### Style Enforcements.
+
+* Editor configurations can help with indentation and formatting.
+* Run pycodestyle / pylint to enforce the style requirements.
+* When the python code is completely type-hint specified, use `mypy` in the test suite to check for type incompatibility problems before build process.
```
adamtheturtle: You can use `mypy` earlier than that and ignore missing type hints. Then, if the codebase ever aims to be fully type-hinted, you can enable errors for missing type hints.

adamtheturtle:
Thank you for this. I would love to see a consistent style in the Python code and I encourage all efforts towards this.

Also, I personally happen to be fond of the style choices in this document. My personal preference is that the standard line length is 80 characters, and that a standard line lenght is enforced by a tool (say `pycodestyle`) which allows ignores on exceptions.

However, one downside of style guides is the cognitive overhead they involve. This includes:
* Learning the desired style,
* Remembering to write with this style,
* A reviewer has to review against style guidelines.

To avoid some of that burden, tools are really useful as pointed out in this document. Tools generally fall into two buckets:
* Those which warn you that there is a non-conformity
* Those which change the code automatically

The latter is preferred when it works well as it takes the burden away from humans. I mentioned `isort` in https://github.com/dcos/dcos/pull/1439/files#r110355510, and another tool which is very useful is [yapf](https://github.com/google/yapf). This is analogous, I believe, to `gofmt`. In my experience it is as good as a human in 95+% of cases, and to produce slightly worse style in a small minority of cases is far better than having to consider all style rules during authorship and review.

It would be useful to have this document clarify what is desired in the case where new code is against the style guide but it fits in with the surrounding code. Personally, I would choose to ignore conformity in the interest of making the codebase better in the long term. I can imagine this causing tedious review discussions.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Maximum line length is 120 characters.
+    * Following are the exceptions to this rule. These types can be of arbitrary length.
+        * Long import statements.
+        * URLs in comments.
+        * Any sequence of chars which would better be in a single line.
+
+* Indentation
+
+    * Use 4space indentation.
```
adamtheturtle: These rules can easily be enforced by `yapf` - at first codebase-wide and then on every CI run.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+from teamcity.messages import TeamcityServiceMessages
+
+import test_util.aws
+import test_util.cluster
+
+from pkgpanda.util import logger, write_string
+```
+
+* [Packages](https://google.github.io/styleguide/pyguide.html?showone=Packages#Packages)
+    * Import each module using the full pathname location of the module.
```
adamtheturtle: It would be useful to have a "why" here. Even if that "why" is "we had to choose one, so we flipped a coin".

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+    * No Relative Imports.
```
adamtheturtle: It would be useful to have a "why" here. Even if that "why" is "we had to choose one, so we flipped a coin".

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Use the `format` method
+    * Prefer `'{named}'.format` will be over `{}.format`.
+        * Rationale: If the number of arguments are long. `{named}.format` helps in readability.
+    * Use your best judgement to decide between + and % (or format) when applicable.
+
+* TODO Comment Format
+    * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
+    * Format required will be `TODO (active-developer) [JIRA] - Brief one line description`
+    * A plugin can be used to enforce the TODO format requirement.
```
adamtheturtle: Please give an example plugin.

docs/styleguides/python.md
adamtheturtle: Typo 4space > "4 space" / "4-space".

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+            public_masters: list of Mesos master IP addresses routable from
+                the local host.
+            slaves: list of Mesos slave/agent advertised IP addresses.
+            default_os_user: default user that marathon/metronome will launch tasks under
+            auth_user: use this user's auth for all requests
+                Note: user must be authenticated explicitly or call self.wait_for_dcos()
+        Returns:
+            None
+        Raises:
+            AssertionErrors on empty values.
```
adamtheturtle: This is confusing because one example exception is plural and one is singular.

adamtheturtle: This also is not the style of the Google guide which uses colons as such:

> IOError: An error occurred accessing the bigtable.Table object.

docs/styleguides/python.md
orsenthil: @jsrodman , smallest `__init__.py` is difficult to define.  The suggestion is to have empty `__init__.py` and not have function definitions or write code in `__init__.py`

* Use `__init__.py` only to represent that the dir is a python package. Have the definitions in the submodule.

Some argue that when you declare/include `ClassObj` from `submodule1 ` inside `pkg1/__init__.py` you could write it like this.

``` from pkg import ClassObj```

Instead of

```from pkg1.submodule1 import ClassObj```

* The point of the rule is, while former might be easier to read, it is ambiguous on where something is defined. And if you are using tools to navigate to definitions of the object, then former will lead to `__init__.py` and then, you have to search go to `submodule1` for actual definition. It becomes confusing for the minor benefit the re-declaration provides.

Comment (orsenthil) [inline docs/styleguides/python.md]: I can change the wording of the style guide document to be more explicit and say.

* Prefer empty `__init__.py` modules.  Keep definitions in submodules.

File: docs/styleguides/python.md (hunk 2 above)
Comment (orsenthil) [inline docs/styleguides/python.md]: My assumption is you are referring to this: http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/ 

* Agree that `.format` can be prone to attacks as described above. It's especially true if you are webapplication templates' author.
* For general scenarios, .format actually helps with readability. I think, many of us have evolved out of  "C-style %s" styles.
* In our current code base, we are using `"{}={}".format`. The general recommendation here is `"{key}={value}".format(key="k1", value="v1")"` as the second one is more readable.

File: docs/styleguides/python.md (hunk 3 above)
Comment (orsenthil) [inline docs/styleguides/python.md]: @adamtheturtle , thanks for all the comments. I will addrss them now.

* Yes, the docstrings in dcos codebase is not consistent. This is an attempt to define the recommendation / protocol and adopt a standard.

I agree with you, we were discussing along these lines:

    * Every public class, method should be documented, should be accompanied with unittest.
    * Private methods, private functions, which are defined by using `_infront` need not be documented or have unittest.
 
I guess, this engulfs the requirement of "every non-nested function, class, method", but also gives an extra flexibility to have undocumented "_private" method.

What's your say on this?

Comment (adamtheturtle) [inline docs/styleguides/python.md]: For the former, maybe you missed "not" in my sentence, doesn't matter now 😄 

The latter sounds good to me!

File: docs/styleguides/python.md (hunk 2 above)
Comment (orsenthil) [inline docs/styleguides/python.md]: * I thought about `f-strings`, but did not mention in the style-guide as it could seem we are too early adopters.
* The recommendation for `"{named}.format"` is more for readability.  But if we see `f-strings` helpful (and not magical), perhaps we to try?  
I am hesistant to declaring "Let's use f-strings in our codebase" as a style guideline.

Comment (adamtheturtle) [inline docs/styleguides/python.md]: I'd be happy wth something like "This revision of the style guide was written before `f-strings` were available and so they were not considered...". This guide seems to say "don't use f-strings" because it says "Use the `format`method".

File: docs/styleguides/python.md (hunk 4 above)
Comment (colin-msphere) [inline docs/styleguides/python.md]: I could see import ordering becoming a hassle for reviewers, especially as the amount of imports increases.

**Comment by colin-msphere** [comment]
This looks good. I especially like that it covers expectations around `__init__.py` and provides a clear example for Types.

File: docs/styleguides/python.md [hunk 14]
```diff
@@ -0,0 +1,173 @@
...
+import test_util.cluster
+
+from pkgpanda.util import logger, write_string
+```
+
+* [Packages](https://google.github.io/styleguide/pyguide.html?showone=Packages#Packages)
+    * Import each module using the full pathname location of the module.
+    * No Relative Imports.
+
+* Maximum line length is 120 characters.
```
vespian: Could you please provide some more context for that decision?

Also, at the very beginning, we state that we "extend" the google style guideline (https://github.com/dcos/dcos/pull/1439/files#diff-a301a44012e0a06325b181e5835c6ce1R5), yet the google style guideline clearly states:
```
Maximum line length is 80 characters.
```

Personally, I think that if your code extends beyond 80 columns then you are either:
* nesting too much and should think about refactoring your code into smaller functions
* your functions have too many arguments which is bad for testability anyway
* should break the line somehow

In my opinion, having hard limit of 90-100 (flake8 starts to complain) and "soft" limit (the one we try to follow) of 80 seems like a good compromise. 

We can have a hard limit of 120 so that we can handle/lint old code while encouraging people to adhere to 80 lines.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+                    public_masters: List[str],
+                    slaves: List[str],
+                    public_slaves: List[str],
+                    default_os_user: str) -> DcosApiSession:
+
+    * Reviewers will request the author to write types for the method parameters.
+    * We will start running mypy type checker as part of build process.
+
+* Shebang Rule.
+    * We will consistently use the shebang line in our scripts.
```
vespian: :+1:

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+        * Rationale: If the number of arguments are long. `{named}.format` helps in readability.
+    * Use your best judgement to decide between + and % (or format) when applicable.
+
+* TODO Comment Format
+    * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
+    * Format required will be `TODO (active-developer) [JIRA] - Brief one line description`
+    * A plugin can be used to enforce the TODO format requirement.
+
+* [Comments and Docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments).
+    * Follow the base style guidelines to adopt the style that explains arguments, return value and exceptions.
```
vespian: :heart:

docs/styleguides/python.md
vespian: @colin-msphere isort solves this problem, Developer can set up a pre-commit git hook and have everything formatted automatically. `flake8-isort` can just enforce that all developers do that (or at least manually).

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
+# DCOS Python Style Guide
+###### Version 0.1
+###### April-6-2017
+
+For Python source code within DCOS, we prefer to keep Google Python Style Guide Revision 2.59 as the base and **extend** it with
```
vespian: But this is just me ;)

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+                    default_os_user: str) -> DcosApiSession:
+
+    * Reviewers will request the author to write types for the method parameters.
+    * We will start running mypy type checker as part of build process.
+
+* Shebang Rule.
+    * We will consistently use the shebang line in our scripts.
+        * `#!/usr/bin/env python3`
+        OR
+        * `#!/usr/bin/env python2` in our scripts.
```
BenWhitehead: We probably don't need the python2 line since we don't ship python2 code.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+    * Use `__init__.py` only to represent python packages.
+        * Avoid importing of objects in `__init__.py` for readability.
+* Import formatting
+
+    * The types of imports should be separated by single line and order is should be.
+        * standard library imports.
+        * third-party imports.
+        * application specific imports.
+
+    * The `from` style import and `import` statements should be separated by single line too.
```
mellenburg: I strongly support the latter of these two suggestions. Separating the 'from' imports doesnt seem like a useful requirement and is visually distracting from the organization

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+from pkgpanda.util import logger, write_string
+```
+
+* [Packages](https://google.github.io/styleguide/pyguide.html?showone=Packages#Packages)
+    * Import each module using the full pathname location of the module.
+    * No Relative Imports.
+
+* Maximum line length is 120 characters.
+    * Following are the exceptions to this rule. These types can be of arbitrary length.
+        * Long import statements.
```
mellenburg: How is this an exception? One should just use parenthesis:
```
from foobar import (
    baz,
    qux,
    fizzbuzz)
```

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+        * URLs in comments.
+        * Any sequence of chars which would better be in a single line.
```
mellenburg: I have one in mind: https://github.com/dcos/dcos/blob/master/test_util/helpers.py#L39

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Indentation
+
+    * Use 4space indentation.
+    * For long list of arguments, use double-indented hanging-indentation.
+ For e.g.
+    * The hanging indentation should be followed by a single blank line.
```
mellenburg: I don't understand the importance of this. There is a rule that says if your hanging indent matches up with the next line beginning, then that must be changed (either with double indent or blank line).

However, if there is no offending line following (like in your sample below), this just seems like a needless rule.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+* Maximum line length is 120 characters.
+    * Following are the exceptions to this rule. These types can be of arbitrary length.
+        * Long import statements.
+        * URLs in comments.
+        * Any sequence of chars which would better be in a single line.
+
+* Indentation
+
+    * Use 4space indentation.
+    * For long list of arguments, use double-indented hanging-indentation.
```
mellenburg: I would appreciate if we could define a "long" list of arguments. Perhaps line length exceeding 80?

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+
+         # ...
+         super().__init__(Url.from_string(dcos_url))
+         self.masters = sorted(masters)
+ ```
+
+
+* Adding Types. Since our code-base is Python3.5+ compatible, we have the choice of introducing types, validating with mypy.
+
+        class VpcClusterUpgradeTestDcosApiSessionFactory:
```
mellenburg: This needs to be in a code-style'd box

docs/styleguides/python.md
mellenburg: This already says "in our scripts" but could be just be a little more specific and say "callable"? The reason being that I feel like a lot of people might still consider non-callable python to be "scripts"

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+            AssertionErrors on empty values.
+            Exception when a given node is not recognized within the DCOS Cluster.
+        """
+```
+
+* We can use sphinx documentation generator. Tools like [sphinx-napolean plugin](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/) can be used to automatically generate api documentation with the above chosen convention.
+
+#### Style Enforcements.
+
+* Editor configurations can help with indentation and formatting.
```
mellenburg: I feel this statement is kinda useless without specifics. Maybe we could add a short list of suggestions? E.G. I love syntastic vim plugin

docs/styleguides/python.md
orsenthil: If the flake8-import-order does not help us with asserting of import-order needs, I think, I we can just switch to isort. We should not be limiting our design decision based on the tools that we plan to use.  

The design rule is about unambiguously defining what the import order should be. And let's use the applicable tool to asserrt that. (isort is fine, IMHO).

docs/styleguides/python.md
orsenthil: Sure, I will add why/ rationale statement in the guide for the design decisions. I agree that it is going to be useful for future reference and we decide to change to something.

docs/styleguides/python.md
orsenthil: I agree. I will change the corresponding code too (that was taken from the actual code from master).

docs/styleguides/python.md
orsenthil: That's a good idea. We will include this.

docs/styleguides/python.md
orsenthil: Yup, I will add why / rationale statements to every decision in this doc.

docs/styleguides/python.md
orsenthil: This needs to written, but we have some pre-cursor / prior-art in our code base here: https://github.com/dcos/dcos/blob/master/flake8_dcos_lint/checker.py

docs/styleguides/python.md
orsenthil: Agree. I will link it.

docs/styleguides/python.md
orsenthil: Our existing code base is 120 col width. I think, we should not change it. We have gotten used to that width.

docs/styleguides/python.md
orsenthil: Tokens and Password hashes come to my mind.

docs/styleguides/python.md
orsenthil: I'd rather remove this statement.  I'd not like to suggest what plugins that developers should use. Instead, the style guide should state the requirements and the developers can configure their editors to meet those requirements.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
+# DCOS Python Style Guide
```
orsenthil: Fixed.

docs/styleguides/python.md
```diff
@@ -0,0 +1,173 @@
...
+###### Version 0.1
```
orsenthil: Yeah, better to call out version. We can feel free to update / enhance / clarify the style guide at the given version. But if we make any rule changes, than a version update will be desirable.

docs/styleguides/python.md
orsenthil: That's what I've tried. That's the reason this style guide is small. :) If you see any repetition (that can be avoided), please point out. :)

docs/styleguides/python.md
orsenthil: I imagine some package and script within package needing python2 for some reason (like a cryptographic library with c-extension not ported to python3 yet).  
 
I agree that more than 99% of the code that we might end up writing for DC/OS will be python3.

docs/styleguides/python.md
orsenthil: @vespian  - yes, the extreme case you pointed out holds true.
The median case, seems more like

```
# stdlib imports
import stdlibA
import stdlibB
import stdlibC

from stdlibC import fooA

# thirdparty
import thirdpartyA
import thirdpartyB
import thirdpartyC

from thirdpartyD import fooB
from thirdpartyE import fooB as fooBE

# app specific
import appspecificA

from appspecificB import fooC

```

In those cases, it seems to me, that separating types of import gives greater benefits for reading (and collapsing with your editor configuration).

* Readability and collapsibility of similar import lines it the reason for newline between types.
* isort supports this too: It calls this as [lines_between_types](https://github.com/timothycrosley/isort/blob/develop/isort/settings.py#L96)

docs/styleguides/python.md
orsenthil: `from packageA.moduleB import SomeVeryLongClassName` (if that extends beyond 120 chars)

docs/styleguides/python.md
orsenthil: 120 chars. I have updated the guide.

docs/styleguides/python.md
orsenthil: * I see that hanging indents are used when there a long list args either in the function call, class instantiation or in the definition. Have a separator after this gives a clear demarcation.

The guideline given here: http://legacy.python.org/dev/peps/pep-0008/#indentation is a good suggestion on how we go about defining hanging indentation.

```
When using a hanging indent the following should be considered; there should be no arguments on the first line and further indentation should be used to clearly distinguish itself as a continuation line.
```

* I think, if unambiguously define that there should be a separator after the hanging indent, than it is easy to remember and follow as a guideline.

orsenthil:
**All**:

I've addressed your review comments.  I have added a _Rationale_ for every decision in the guide. Please see if those make sense and you agree with them.

The readable version of the document is here: https://github.com/orsenthil/dcos/blob/style_guide/docs/styleguides/python.md

Please give this another pass.

1) If there is any improvement that you can see in terms of being more explicit, please add a comment.

2) If you have any strong objection to any rule, please create a discussion item in the comments below and we can vote on it.

docs/styleguides/python.md
orsenthil: Fixed.

docs/styleguides/python.md
mellenburg: Why should that be a `#noqa` exception when there is a perfectly valid method for declaring long import statements? I do not see the value and very much disagree with allowing that to be an exception

docs/styleguides/python.md
```diff
@@ -0,0 +1,231 @@
...
+        * URLs in comments.
+        * Any sequence of chars, like embedded tokens which would better be in a single line.
+
+   _Rationale_: At the time this style guide was written, the existing code base followed 120 column width guideline.
+   We are keeping up with that.
+
+* Indentation
+
+    * Use 4-space indentation.
+    * For long list of arguments, when the number of arguments exceed the column limit 120 chars, use double-indented
```
mellenburg: This is not stating anything unique. Anyline that exceeds 120 chars needs to be indented. In any case where the hanging indent matches the next logical indent, you need to double indent. Stating this just adds confusion IMO

docs/styleguides/python.md
orsenthil: @mellenburg  - I see your point.

I was thinking of the first scenario here and trying to the 2nd.
Using hanging indent we can handle long imports too, so  I will remove this line.

```
from packageA.moduleB.submoduleC.submoduleD.submoduleE.submoduleF import SomeVeryLongClassNameThatExceedsHundredTwentyChars #noqa

from packageA.moduleB.submoduleC.submoduleD.submoduleE.submoduleF import \
    SomeVeryLongClassNameThatExceedsHundredTwentyChars

from packageA.moduleB.submoduleC.submoduleD.submoduleE.submoduleF import (
    SomeVeryLongClassNameThatExceedsHundredTwentyChars)

```

orsenthil: Well, there is situation where the rule does make sense. This will be applicable in mono-repo code base and not in ours.

The situation is

```
import packageA.moduleB.submoduleC.submoduleD.submoduleE.submoduleF.submoduleG.submoduleH.longNamedSubModuleExceedingHundredTwentyChars  #noqa
```

docs/styleguides/python.md
orsenthil: > In any case where the hanging indent matches the next logical indent, you need to double indent.

The statement in the style guide is: For long list of arguments, use double-indented hanging-indentation.

i.e, irrespective of any situation, if you are using hanging indent, use double indent. The rationale is that, it gives a visual significance to indentation and indicates that it is used for wrapping and not as a context block.

Here is a comparison for clarity

* exceeds 120 chars # BAD.

```
def method():
    
    installer.setup_remote(tunnel=bootstrap_host_tunnel, installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh', download_url=installer_url)

```
* single indented - Not recommended. # BAD.
Note hanging indentation aligns with the logical indentation of the program.

```
def method():
    if installer.prereqs.ok:
        print("pre-reqs satisfied")

    installer.setup_remote(
        tunnel=bootstrap_host_tunnel,
        installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh',
        download_url=installer_url)

    if installer.setup.ok:
        print("remote setup successful.")
    else:
        print("remote setup successful.")
```
 
* double indented - Recommended # GOOD. 
   Note the visual clarity.



```
def method():
    if installer.prereqs.ok:
        print("pre-reqs satisfied")

    installer.setup_remote(
            tunnel=bootstrap_host_tunnel,
            installer_path=cluster.ssher.home_dir + '/dcos_generate_config.sh',
            download_url=installer_url)

    if installer.setup.ok:
        print("remote setup successful.")
    else:
        print("remote setup successful.")
```

* Even if the `installer.setup_remote` is the last line of the program, I think, it is better served with double indented hanging indent.

If the statement of the rule is not-clean / ambiguous, please suggest a improvement. I can also consider adding this example to the style guide itself.

docs/styleguides/python.md
mellenburg: @orsenthil Even in that case, this rule does not make sense. Just use `\`:
```
import packageA.moduleB.submoduleC.submoduleD.submoduleE.submoduleF.submoduleG. \
    submoduleH.longNamedSubModuleExceedingHundredTwentyChars  #noqa
```

docs/styleguides/python.md
mellenburg: Thanks for the examples. I'm a little on the fence, but I am comfortable with the position you present in the second example

docs/styleguides/python.md
orsenthil: I've updated the guide and removed the rule of long imports. 
* We might hardly come across such a requirement.
* We could work around using from.
* Finally using unix line continuation. (But if we can prevent the use of unix line continuation I am for it)

docs/styleguides/python.md
mellenburg: You propose a simple new rule to encompass a preexisting complicated rule. Usually this would bother me, but come to think of it, it took me forever to learn how to make the style editor happy with hanging indents, so I can go along with this :+1:

docs/styleguides/python.md
```diff
@@ -0,0 +1,259 @@
...
+
+* Import formatting
+
+    * The types of imports should be separated by single line and order should be.
+
+        1. standard library imports.
+        2. third-party imports.
+        3. application specific imports.
+
+    * The `from` style import and `import` statements should be separated by single line too.
```
mellenburg: I still don't agree that this extra line should exist between import and from sections of the same import "type". I do not find it to add any extra clarity or readability; its just an idiom. At that, its an idiom we don't have; look at our code-- most of it does not follow this.  You pointed out there is an isort setting for this to establish that its a thing, but the default is zero.

My point is:
- I personally abhor needless white-space that limits how much code is on my screen and more often than not our code conflicts with this rule
- This is a v0.1 style guide; let's try and stick to the **essential**
- I fail to see any logical argument other than some IDEs want it like this
- If you remove this, we can just adopt the common, default standard and not be a snow flake

docs/styleguides/python.md
```diff
@@ -0,0 +1,259 @@
...
+    * Each group of these imports should be sorted.
+
+    _Rationale_:  Readablity and code-foldability are reasons for this separation.
```
orsenthil: Valid point. Yes, I am referring to folding features available in all editors. "zf", "zo", "za" in vim.
And [+] in IntelliJ.

A good reasoning to keep `from` and `import` statements separate are they are two different types of statements. The similar members of the type can be grouped together. My argument is, this grouping and separation within groups helps in readability. This separation also helps editors with code-folding and auto-imports features.

I think, this need not be mentioned in the style guide.

We have to resolve our contention on whether to group `from` and `import` styles into a single group or keep them as separate groups.

branden: If we split up imports like that, it becomes ambiguous whether a blank line is splitting by library source or import style. I can't tell at a glance which are first-party vs. third-party vs. stdlib. If I see three groups of imports, I have to read them and know where the libraries come from to figure out the grouping.

```
import foo

import bar

from baz import qux
```

`foo` is stdlib and `bar` is third-party. But is `baz` a library from our application, or is it a third-party library? I can't tell from this grouping alone.

docs/styleguides/python.md
orsenthil: I will remove this line and the requirement. I agree with some of the points made here. (Let's stick to the essential).

Also, there is enough negative votes to this ( @mellenburg , @vespian  and @branden ).

My guess is, I was having this bias from days when `isort` tool was not available and I had to take help of IDE/ editor settings to do automated sorting.

I ran this small experiment.

* Start with unsorted imports
```
import test_util.cluster
import traceback
from pkgpanda.util import logger, write_string
from retrying import retry, RetryError
from subprocess import CalledProcessError

import json
import logging
import os
from teamcity.messages import TeamcityServiceMessages
from test_util.cluster import run_bootstrap_nginx
from test_util.dcos_api_session import DcosApiSession, DcosUser
from test_util.helpers import CI_CREDENTIALS, marathon_app_id_to_mesos_dns_subdomain, random_id
from typing import Callable, List
import pprint
import random
import retrying

import sys
import test_util.aws
import uuid
```
* sort using isort with no separation between types.

```isort -o retrying fixed_with_groups_no_lines.py```

Output

```
import json
import logging
import os
import pprint
import random
import sys
import traceback
import uuid
from subprocess import CalledProcessError
from typing import Callable, List

import retrying
from retrying import RetryError, retry

import test_util.aws
import test_util.cluster
from pkgpanda.util import logger, write_string
from teamcity.messages import TeamcityServiceMessages
from test_util.cluster import run_bootstrap_nginx
from test_util.dcos_api_session import DcosApiSession, DcosUser
from test_util.helpers import (CI_CREDENTIALS,
                               marathon_app_id_to_mesos_dns_subdomain,
                               random_id)
```

* Sort using isort with group and line separation

```
isort -o retrying -lbt 1 fixed_with_group_and_lines.py
```

Output.

```
import json
import logging
import os
import pprint
import random
import sys
import traceback
import uuid

from subprocess import CalledProcessError
from typing import Callable, List

import retrying

from retrying import RetryError, retry

import test_util.aws
import test_util.cluster

from pkgpanda.util import logger, write_string
from teamcity.messages import TeamcityServiceMessages
from test_util.cluster import run_bootstrap_nginx
from test_util.dcos_api_session import DcosApiSession, DcosUser
from test_util.helpers import (CI_CREDENTIALS,
                               marathon_app_id_to_mesos_dns_subdomain,
                               random_id)
```

* It seems. it just fine to adopt the 2nd style - that is one without any separation between `import` and `from` types.

File: docs/styleguides/python.md (hunk 30 above)
Comment (orsenthil) [inline docs/styleguides/python.md]: > If we split up imports like that, it becomes ambiguous whether a blank line is splitting by library source or import style. 

>  I can't tell from this grouping alone.

True. What I've seen this folks using `# comments to describe`  or distinguishing it based on familiarity.

**Comment by orsenthil** [comment]
Thanks for the first approval @mellenburg . I see the shore nearby! 🚢 

@adamtheturtle , @branden , @BenWhitehead , @colin-msphere , @jsrodman , @vespian - This could see your reviews/ approvals too. 

Thank you!  Please remember, this guide is open for us to iterate and it will be much easier once we have one in place.

File: docs/styleguides/python.md [hunk 31]
```diff
@@ -0,0 +1,259 @@
...
+      Python 3.6 onwards.
+
+     _Rationale_: When there are multiple format specifiers used in a string, `{name1} {name2}` formats are easier
+    to read than `{}{}` (empty) or `{0}{1}` (indexed) formats.
+
+
+
+* TODO Comment Format
+    * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
+    * Format required will be `TODO (active-developer) [JIRA] - Brief one line description`
```
branden: Can we keep the TODO notation that's in the base style guide and which we already use in a few places in the codebase? `# TODO(active-developer): DCOS_OSS-1234 - Brief one line description`.

docs/styleguides/python.md
```diff
@@ -0,0 +1,259 @@
...
+
+    ```
+    # TODO: (skumaran) DCOS_OSS-952 - Create a flake8 plugin to enforce TODO format.
+    ````
```
branden: One backtick too many.

docs/styleguides/python.md
orsenthil: Sounds good to me.  I think, it is changing `-` to `:`

I will use the example that you specified.

```
# TODO(active-developer): DCOS_OSS-1234 - Brief one line description.
```

orsenthil:
ping again.

I am looking for approvals from folks who contribute / plan to contribute to DCOS code base.  Please review and approve. Thanks!

docs/styleguides/python.md
```diff
@@ -197,11 +197,11 @@ from test_util.helpers import (CI_CREDENTIALS,
 
 * TODO Comment Format
     * Follow the rules on [when to write TODO comments](https://google.github.io/styleguide/pyguide.html?showone=TODO_Comments#TODO_Comments)
-    * Format required will be `TODO (active-developer) [JIRA] - Brief one line description`
+    * Format required will be `#TODO (active-developer) JIRA_ID: Brief one line description`
 
     ```
-    # TODO: (skumaran) DCOS_OSS-952 - Create a flake8 plugin to enforce TODO format.
-    ````
+    #TODO (skumaran) DCOS_OSS-952: Create a flake8 plugin to enforce TODO format.
```
branden: This is different from what's in the base style guide and what we use currently.

```
#TODO(skumaran): DCOS_OSS-952 Create a flake8 plugin to enforce TODO format.
```

orsenthil: Sorry, I don't understand the comment properly.

_And what we use currently_ - is mis-leading, because we don't use any JIRA reference or use a TODO format consistently.

Here is the screenshot base guide style reference which suggest two formats. 


screenshot 2017-04-17 15 38 19

My proposal and suggestion is that we want to use the JIRA ticket with the active developer name.

Can you please share your desirable format? Please make sure to include the elements
1) TODO term 2) developer name 3) JIRA ID 4) Short Description.

orsenthil: @branden  - please see an update. I copy-pasted the format from the previous discussion as it satisfied all the requirements that were desirable.

branden: Update LGTM. 👍

docs/styleguides/python.md
vespian: > Our existing code base is 120 col width.

Some of the codebase is 80 col (AR test harness code, Bouncer, dcos-security-cli, dcos-ca-pylib to name a few), some of the code is 120 lines (most notably genconf).

>  as I feel 80, in this day and age is far too small give how much lateral screen space is available.

Yes, as long as you working at your desk. As soon as you start travelling and/or working away from your desk, you are stuck with a small laptop screen. One can try to decrease the font size, but this is burdensome for one's eyes.

Also, looking at some other cody style guidelines:
* from pep8(https://www.python.org/dev/peps/pep-0008/#maximum-line-length):
```
Limit all lines to a maximum of 79 characters.
```
* from linux kernel style guideline (https://01.org/linuxgraphics/gfx-docs/drm/process/coding-style.html) (it's my personal favourite):
```
The limit on the length of lines is 80 columns and this is a strongly preferred limit.
```
* Drupal (PHP codebase) coding style guideline (https://www.drupal.org/docs/develop/standards/coding-standards#linelength):
```
In general, all lines of code should not be longer than 80 characters.
```
* and finally Django project has a nice compromise I think:
```
As an exception to PEP 8 is our rules on line lengths. Don’t limit lines of code to 79 characters if it means the code looks significantly uglier or is harder to read. We allow up to 119 characters as this is the width of GitHub code review; anything longer requires horizontal scrolling which makes review more difficult. This check is included when you run flake8. Documentation, comments, and docstrings should be wrapped at 79 characters, even though PEP 8 suggests 72.
```

I was wondering if we could rephrase our rules to follow the ones from Django project. For me personally, it would be great if each project could choose it's own max line length as long as it's smaller than the global limit of 120.

docs/styleguides/python.md
vespian: > Yes, I am referring to folding features available in all editors. "zf", "zo", "za" in vim.
And [+] in IntelliJ.

What if I somebody does not use it/is a die-hard purist and uses plain vim? Does it mean that this rule does not concern him?

> A good reasoning to keep from and import statements separate are they are two different types of statements. The similar members of the type can be grouped together. My argument is, this grouping and separation within groups helps in readability. This separation also helps editors with code-folding and auto-imports features.

I think this is already achieved by sorting your imports correctly:
```
import a.b
import c
from d import z
from e import x
```
I think it's easy to distinguish within a particular group `import` subgroup and `from` subgroup.

> True. What I've seen this folks using # comments to describe or distinguishing it based on familiarity.

Are you suggesting that we start commenting as well if there is ambiguity?

docs/styleguides/python.md
vespian: uhh, I just noticed that @orsenthil already removed this requirement - please disregard my earlier comment.

(GH PR reviews discussions suck :( )

docs/styleguides/python.md
vespian: > My proposal and suggestion is that we want to use the JIRA ticket with the active developer name.

Then why not just a JIRA? Jira owner can change, developer can move to a different project/team.

vespian:
Review status: 0 of 1 files reviewed at latest revision, 36 unresolved discussions, some commit checks failed.

---

*[docs/styleguides/python.md, line 34 at r1](https://reviewable.io:443/reviews/dcos/dcos/1439#-Ki-iSZvueoupsoTQrSP-r1-34:-Ki-lYMliNPav_AFk3XD:b8lxgh) ([raw file](https://github.com/dcos/dcos/blob/ac928b223c010b14560df1cb88fd813382c24282/docs/styleguides/python.md#L34)):*
> We should not be limiting our design decision based on the tools that we plan to use.

NIT: Even the best Style Guideline will not be worth much without proper tooling as we will simply not be able to enforce it. If the tools do not exist may need to write it which in turn may and up to be a difficult and time-consuming process.

IMO we should keep it in mind and think what we can enforce and not only what we want to enforce :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos/1439)*

docs/styleguides/python.md
orsenthil: Hi @vespian , I think, we can definitely aim for an unambiguous consistency here and perhaps simpler than Django rule as we can keep the col width same for both docstrings and code.

The 79 /80 col was width seems like a rule for teletypes, who screens were much smaller than any of our 15-inch, 16-inch and 21-inch laptop screen sizes.

* http://rdingwall.com/2008/05/31/is-the-80-character-line-limit-still-relevant/
* https://softwareengineering.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width
* The projects that follow might have had their origins in 1990s or early 2000's perhaps. They, for good reasons, do not want to change that style requirement.

Our current setting already enforces it the code to 120-chars( https://github.com/dcos/dcos/blob/master/tox.ini#L2) The files codebase within the project that are not with 120 col width, might need a reformat to 120 width. 

I think the tools only assert that you don't exceed 120 col width. It will be the responsibility of the developer and code-reviewer to keep the code appear consistently wrapped across the entire project.

Given current setting of 120 max-limit in flake8,  I thought _it is pretty non-controversial_ to pin that requirement in the style guide. Going back to 80 or 100 will be more controversial and **less desirable** IMO than keeping the current width requirement.

I hope this answers the concern.

orsenthil:
Review status: 0 of 1 files reviewed at latest revision, 36 unresolved discussions, some commit checks failed.

---

*[docs/styleguides/python.md, line 18 at r1](https://reviewable.io:443/reviews/dcos/dcos/1439#-Ki-iSZvueoupsoTQrSP-r1-18:-Ki0_3iGIFUWHADLrhTD:b-896fix) ([raw file](https://github.com/dcos/dcos/blob/ac928b223c010b14560df1cb88fd813382c24282/docs/styleguides/python.md#L18)):*
<details><summary><i>Previously, vespian (Paweł Rozlach) wrote…</i></summary><blockquote>

The 2.59 revision is the current one, but if it changes, we will not have a way to see the original one - the URL you provided is generic. 

Maybe we can link directly to their repo/commit-sha? https://github.com/google/styleguide/blob/gh-pages/pyguide.html
</blockquote></details>

Done.

---

*[docs/styleguides/python.md, line 34 at r1](https://reviewable.io:443/reviews/dcos/dcos/1439#-Ki-iSZvueoupsoTQrSP-r1-34:-Ki0_7DEGFNWe9xZjh0c:b-ym2muv) ([raw file](https://github.com/dcos/dcos/blob/ac928b223c010b14560df1cb88fd813382c24282/docs/styleguides/python.md#L34)):*
> IMO we should keep it in mind and think what we can enforce and not only what we want to enforce :)

Definitely.

* isort
* flake8 for pycodestyle with plugins
* mypy
* yapf (?)

However, there will always be limitation for what a tool can do. For .e.g, it can be enforced that the code is not exceeding 120 cols. but it can be enforced via tool that line is using 120 chars properly and not using less than that. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos/1439)*

vespian:
Reviewed 1 of 1 files at r1.
Review status: 0 of 1 files reviewed at latest revision, 35 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos/1439)*

docs/styleguides/python.md
orsenthil: > Then why not just a JIRA?

Because, it captures the person responsible for the `TODO` inline and it helps both when writing the TODO and for another developer coming across the TODO.

> Jira owner can change, developer can move to a different project/team.

Or a developer might leave the project. 

In that situation, it might be a good idea to evaluate the TODO and take one of the necessary actions, like a) kill it b) re-assign to a different active developer.

vespian:
Review status: 0 of 1 files reviewed at latest revision, 35 unresolved discussions, some commit checks failed.

---

*[docs/styleguides/python.md, line 203 at r5](https://reviewable.io:443/reviews/dcos/dcos/1439#-Ki-iSZvueoupsoTQrSP-r5-203:-Ki0dYv4fXxPf0CztdUw:bl1pyax) ([raw file](https://github.com/dcos/dcos/blob/c749fe89ae1c5ab8b5247f79f06159b73afb5275/docs/styleguides/python.md#L203)):*
Comments lie :) http://www.informit.com/articles/article.aspx?p=1326509

> In that situation, it might be a good idea to evaluate the TODO and take one of the necessary actions, like a) kill it b) re-assign to a different active developer.

Aren't we duplicating the stuff we do in Jira then? If we need to track the author, isn't `git blame` enough? Does it mean that we will now create Jira issue and fix/find owners of all the TODO comments that Cody made? Does it mean that we need to find the owners of that comments now?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos/1439)*

orsenthil:
Review status: 0 of 1 files reviewed at latest revision, 35 unresolved discussions, some commit checks failed.

---

*[docs/styleguides/python.md, line 203 at r5](https://reviewable.io:443/reviews/dcos/dcos/1439#-Ki-iSZvueoupsoTQrSP-r5-203:-Ki0gghZxbI6t53k5g9w:b-xn1w25) ([raw file](https://github.com/dcos/dcos/blob/c749fe89ae1c5ab8b5247f79f06159b73afb5275/docs/styleguides/python.md#L203)):*
Nope. We are not . A high level short line is to avoid visiting JIRA ticket.

> Does it mean that we will now create JIRA issue and fix/find owners of all the TODO comments that Cody made? Does it mean that we need to find the owners of that comments now?

Unfortunately, yes.

Let's  judge what is  better situation to rely upon - Unmaintained TODO/comment or a maintained one?  I think, it is latter. So this style guide statement is trying to suggest/ enforce that.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/dcos/dcos/1439)*

orsenthil:

PR #1442, opened 2017-04-10
Deflake pods test

mellenburg:
## High Level Description

This test is failing due to simple timeouts on an extremely regular basis. The two minute deploy timeout for apps is not a hard limit, but rather an optimization used because few apps ever succeeded after 120 seconds of waiting. We currently have no such data for pods and as such should bump this timeout

## Related Issues
  https://jira.mesosphere.com/browse/DCOS_OSS-912
## Checklist for all PR's

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -85,6 +85,7 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
 
     test_uuid = uuid.uuid4().hex
 
+    # create pod with trivial apps that function as long running processes
     pod_definition = {
         'id': '/integration-test-pods-{}'.format(test_uuid),
         'scaling': {'kind': 'fixed', 'instances': 1},
@@ -94,7 +95,7 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
                 'name': 'ct1',
                 'resources': {'cpus': 0.1, 'mem': 32},
                 'image': {'kind': 'DOCKER', 'id': 'debian:jessie'},
-                'exec': {'command': {'shell': 'touch foo'}},
+                'exec': {'command': {'shell': 'touch foo; while true; do sleep 1; done'}},
                 'healthcheck': {'command': {'shell': 'test -f foo'}}
             },
             {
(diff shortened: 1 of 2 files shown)


mellenburg:
dcos-tweeter failure is a flake due to the test not being run in a clean environment

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -94,7 +94,7 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
                 'name': 'ct1',
                 'resources': {'cpus': 0.1, 'mem': 32},
                 'image': {'kind': 'DOCKER', 'id': 'debian:jessie'},
-                'exec': {'command': {'shell': 'touch foo'}},
+                'exec': {'command': {'shell': 'touch foo; while true; do sleep 1; done'}},
```
orsenthil: A comment will be desirable here stating the reason for this infinite loop.

test_util/marathon.py
```diff
@@ -371,6 +371,6 @@ def deploy_and_cleanup(self, app_definition, timeout=120, check_health=True, ign
         self.destroy_app(app_definition['id'], timeout)
 
     @contextmanager
-    def deploy_pod_and_cleanup(self, pod_definition, timeout=120):
+    def deploy_pod_and_cleanup(self, pod_definition, timeout=180):
```
orsenthil: Is the timeout values arbitrary?  why would 180 be better than 120?

mellenburg: Its an estimate. 120 was decided on by data for single app deployments; not pods. Once we raise it, we might have to raise it again; I don't know.

packages/dcos-integration-test/extra/test_applications.py
mellenburg: Because marathon is made to support long-running processes. This seems inherent to me, but more comments the merrier I guess

test_util/marathon.py
orsenthil: So, this is a value higher than the default value of 120. In that case, a comment here might help too. 
Or we can start with a very high value for the default (300) and stay optimistic that all the tests will be done within that timeframe.

test_util/marathon.py
```diff
@@ -253,7 +253,7 @@ def ensure_deployment_is_finished():
         except retrying.RetryError:
             raise Exception("Deployments were not completed within {timeout} seconds".format(timeout=timeout))
 
-    def deploy_pod(self, pod_definition):
+    def deploy_pod(self, pod_definition, timeout):
```
falseperfection: Every other ```timeout``` in this file has a default value, is this one different for a particular reason?

mellenburg: Nope, but you're right, might as well make it consistent.

PR #1459, opened 2017-04-14
Add dcos checks

darkonie:
## High Level Description

[dcos-enterprise PR ](https://github.com/mesosphere/dcos-enterprise/pull/944)

Add a new pkgpanda package `dcos-checks`. The `dcos-checks` provides node and cluster checks implementation.

- [x] Add dcos-checks skeleton.
- [x] Add basic systemd units check.
- [x] Add integration test

## Related Issues

  - [DCOS-14989](https://jira.mesosphere.com/browse/DCOS-14989) Create `dcos-checks` package
## Checklist for all PR's

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): N/A
  - [ ] Test Results: N/A
  - [ ] Code Coverage (if available): N/A
___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/test_dcos_checks.py
+++ b/packages/dcos-integration-test/extra/test_dcos_checks.py
@@ -1,0 +1,21 @@
+import uuid
+
+
+def test_dcos_checks_components_open(dcos_api_session):
+    # target contains a dcos-checks subcommand as a key and tuple of optional parameters as a value.
+    target = {
+        "components": ()
+    }
+
+    cmd_tpl = "/opt/mesosphere/bin/dcos-checks --role agent {} {}"
+    cmds = [cmd_tpl.format(subcommand, " ".join(arg for arg in args)) for subcommand, args in target.items()]
+    test_uuid = uuid.uuid4().hex
+    check_job = {
+        'id': 'test-dcos-checks-' + test_uuid,
+        'run': {
+            'cpus': .1,
+            'mem': 128,
+            'disk': 0,
+            'cmd': " && ".join(cmds)}}
+
+    dcos_api_session.metronome_one_off(check_job)
--- a/packages/dcos-checks/build
+++ b/packages/dcos-checks/build
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e  # Fail the script if anything fails
+set -x  # Verbose output
+set -u  # Undefined variables
+
+mkdir -p /pkg/src/github.com/dcos
+mv /pkg/src/dcos-checks /pkg/src/github.com/dcos/
+cd /pkg/src/github.com/dcos/dcos-checks
+
+go install -v
+cp -r /pkg/bin/ "$PKG_PATH"
--- a/packages/dcos-checks/buildinfo.json
+++ b/packages/dcos-checks/buildinfo.json
@@ -1,0 +1,11 @@
+{
+  "single_source": {
+    "kind": "git",
+    "git": "https://github.com/dcos/dcos-checks.git",
+    "ref": "42dc43b013fe79c3dbca188e3cb7fcd3ff11d86e",
+    "ref_origin": "master"
+  },
+  "state_directory": true,
+  "username": "dcos_3dt",
+  "group": "systemd-journal"
+}


darkonie:

darkonie:
@mesosphere-mergebot bump-ee

darkonie:
@mesosphere-mergebot bump-ee

packages/dcos-checks/extra/dcos-checks/cmd/root.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/dcos/dcos-go/dcos"
+	"github.com/dcos/dcos/packages/dcos-checks/extra/dcos-checks/client"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
+	statusOK      = 0
+	statusWarning = 1
```
collinvandyck: This is unused in the project.  Also, does it make sense to define these using iota? e.g.

```
const (
  statusOK = iota
  statusWarning
  statusFailure
  statusUnknown
)
```

packages/dcos-checks/extra/dcos-checks/client/http.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"net/http"
+
+	"github.com/dcos/dcos-go/dcos/http/transport"
+)
+
+// NewClient returns a new http client ready to handle DC/OS security.
+// A caller can optionally pass a configured *http.Client (with the right timeout for instance), in this case
+// NewClient will replace Transport.
+func NewClient(client *http.Client, iamConfig, caCert string) (*http.Client, error) {
```
collinvandyck: The `client` argument to this function is always nil in this changeset.

packages/dcos-checks/extra/dcos-checks/cmd/root.go
```diff
@@ -0,0 +1,214 @@
...
+
+// DCOSChecker defines an interface for a generic DC/OS check.
+// ID() returns a check unique ID and RunCheck(...) returns a combined stdout/stderr, exit code and error.
+type DCOSChecker interface {
+	ID() string
+	Run(context.Context, *CLIConfigFlags) (string, int, error)
+}
+
+// RunCheck is a helper function that takes a list of DC/OS checks and runs checks one by one.
+func RunCheck(checks []DCOSChecker) {
```
collinvandyck: Should this be called `RunChecks` instead?

packages/dcos-checks/extra/dcos-checks/cmd/root.go
```diff
@@ -0,0 +1,214 @@
...
+	for _, check := range checks {
+		output, retCode, err := check.Run(nil, DCOSConfig)
+		if err != nil {
+			logrus.Fatalf("Error executing %s: %s", check.ID(), err)
+		}
+		if retCode != statusOK {
+			exitCode = statusFailure
+		}
+		if output != "" {
+			fmt.Printf("[%s]: %s\n", check.ID(), output)
```
collinvandyck: Do we want to use `logrus.Infof` here instead?

packages/dcos-checks/extra/dcos-checks/cmd/components.go
```diff
@@ -0,0 +1,185 @@
...
+/system/health/v1 is the local endpoint. The response structure is the following
+{
+  "units": ["unit1", ...]
+}
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		// it is possible to implement a number of other checks under the same subcommand namespace.
+		// this function can be used as a dispatcher.
+		if len(args) == 0 {
+			RunCheck([]DCOSChecker{&ComponentCheck{
```
collinvandyck: Might be cleaner here to have another function 

```
func NewComponentCheck(name string) *ComponentCheck
```

I think that's what we've been talking about recently with regards to initializing structs.

Alternatively, you could define ComponentCheck as

```
type ComponentCheck string
```

And then just use `ComponentCheck("DC/OS components health check")`

packages/dcos-checks/extra/dcos-checks/cmd/components.go
```diff
@@ -0,0 +1,185 @@
...
+		return "", statusUnknown, fmt.Errorf("unable to create a new HTTP request: %s", err)
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return "", statusUnknown, fmt.Errorf("unable to GET %s: %s", healthURLPrefix, err)
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
```
collinvandyck: Instead of reading the response body into a temporary byte slice, you can do

```
var dr diagnosticsResponse
if err := json.NewDecoder(resp.Body).Decode(&dr); err != nil {
 ...
}
```

packages/dcos-checks/extra/dcos-checks/cmd/components.go
```diff
@@ -0,0 +1,185 @@
...
+
+/system/health/v1 is the local endpoint. The response structure is the following
+{
+  "units": ["unit1", ...]
+}
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		// it is possible to implement a number of other checks under the same subcommand namespace.
+		// this function can be used as a dispatcher.
+		if len(args) == 0 {
```
collinvandyck: If `len(args) > 0` should we return an error here intead of not doing anything?  Maybe `RunE func(cmd *Command, args []string) error` is appropriate in this case?

packages/dcos-checks/extra/dcos-checks/cmd/root.go
darkonie: this is intensional, the output would be a custom format, i need to place TODO here

packages/dcos-checks/extra/dcos-checks/client/http.go
darkonie: this is a very first check, we will have more and it might be needed to pass a constructed client.

packages/dcos-checks/extra/dcos-checks/cmd/components.go
darkonie: this is defined by cobra framework. The idea here is to use 3 level of subcommands, for example:
`dcos-checks components 3dt` this is the place to dispatch

packages/dcos-checks/extra/dcos-checks/cmd/root.go
darkonie: this is a skeleton + 1 check, we shall have more checks. These status codes are per design doc.

darkonie: @collinvandyck iota, TIL

packages/dcos-checks/extra/dcos-checks/cmd/checks.go
```diff
@@ -0,0 +1,44 @@
...
+
+const (
+	statusOK = iota
+	statusWarning
+	statusFailure
+	statusUnknown
+)
+
+// DCOSChecker defines an interface for a generic DC/OS check.
+// ID() returns a check unique ID and RunCheck(...) returns a combined stdout/stderr, exit code and error.
```
collinvandyck: RunCheck -> Run

packages/dcos-checks/extra/dcos-checks/cmd/checks.go
```diff
@@ -0,0 +1,44 @@
...
+type DCOSChecker interface {
+	ID() string
+	Run(context.Context, *CLIConfigFlags) (string, int, error)
```
collinvandyck: I know you commented on this above as to what the return values are, but we should probably name them in the interface definition.

```
Run(context.Context, *CLIConfigFlags) (output string, exitCode int, err error)
```

packages/dcos-checks/extra/dcos-checks/cmd/checks.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	statusOK = iota
```
collinvandyck: 💪

packages/dcos-integration-test/extra/test_dcos_checks.py
```diff
@@ -0,0 +1,25 @@
...
+        output = subprocess.check_output(
+            ["/opt/mesosphere/bin/dcos-checks", "--role", "master", target] + [arg for arg in args],
+            universal_newlines=True)
+        return output, 0
+    except subprocess.CalledProcessError as err:
+        return err.output, err.returncode
+
+
+def test_components(dcos_api_session):
+    dcos_api_session
```
darkonie: oops

packages/dcos-integration-test/extra/test_dcos_checks.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Run the dcos-checks binary.
+
+    :param target: dcos-checks subcommand
+    :param args: a list of target arguments passed to a dcos-checks binary
+    :return: combined stdout/stderr, return code
+    """
+    try:
+        output = subprocess.check_output(
+            ["/opt/mesosphere/bin/dcos-checks", "--role", "master", target] + [arg for arg in args],
```
branden: This role arg slightly concerns me -- it's correct right now only by coincidence, because we happen to run integration tests on a master. I don't know if we have any other tests which rely on being run on a master.

@mellenburg Do you have any guidance here? Is there a facility this test can use to e.g. SSH to a node whose role we know for certain and run this executable?

packages/dcos-integration-test/extra/test_dcos_checks.py
branden: This line isn't doing anything.

darkonie: oops :)

packages/dcos-integration-test/extra/test_dcos_checks.py
mellenburg: For quite a while now its been a requirement that all of these tests are run from a master. I don't plan on this changing any time soon so I do not find this case to be reliant on happenstance. Really, I would like us to have the integration tests have ZERO shelling out, but for now this is an acceptable evil

I do feel that it is less than ideal to add a test that makes assumptions about the current location of the test file rather than only assuming there is a cluster. So, perhaps add a TODO/FIXME/JIRA that says "make this test runnable from any node" or something?

As for SSH facility in tests, we are getting closer but I do not think this case is worth jumping the gun and adding more SSH dependency before its ready

darkonie: @branden what if we add a marathon task to test the binaries for now, and when the ssh infrastructure is ready, we can add few more tests to cover it?

branden: @mellenburg Thanks for the context. I'd be OK with shelling out and leaving a TODO.

@darkonie That SGTM too 👍

packages/dcos-integration-test/extra/test_dcos_checks.py
```diff
@@ -0,0 +1,21 @@
+import uuid
+
+
+def test_components(dcos_api_session):
+    # target contains a dcos-checks subcommand as a key and a list of tulip of optional arguments.
```
branden: `s/tulip/tuple`?

> list of tulip of optional arguments

Shouldn't args be a list of strings?

darkonie: we could use lists, i decided to go with tuples because it is immutable

branden: That works, but the comment says it's a list of tuples.

darkonie: fixed

amitaekbote:

branden:

orsenthil:
Sorry that we are bit late in here. But, I raised some concerns with inclusion of this in dcos/dcos https://github.com/dcos/dcos/pull/1535#issuecomment-300853858

In general, I like mono-repo and prefer the code in single-place. However, the unittests should be the same place too and those needs to be run before the changes get merged.

_All Checks have Passed above in this PR has very limited exercise on the 464 files that are being introduced in this PR._

Is my statement correct?

So, if there are unittests can be exercised for this changed and those are run for every change to this package, in the dcos/dcos repo, that will be better way to introduce the codebase.
That will bring an additional dependency to factor in the testing infrastructure requirements for dcos/dcos.

Let us please address these concerns before moving ahead with this.

/cc @BenWhitehead , @mellenburg , @spahl

amitaekbote:
464 files seems like a lot but in reality its the vendor folder which are dependencies. Thats how go does dependencies. 

In reality the files introduced are the skeleton for adding a new check and one new check (~10 files). The check calls 3dt health and it has a unit test and integration test to verify it. 

In its current state it does not interact with the current dcos installation process which is by design, the next few pr's will add more checks with unit tests.

There will be a pr for a "runner" in 3dt that will be used to run these checks which the integration test will use. 

Following the runner the installer will call the runner as part of the install process which will verify these checks on every install.

amitaekbote:
Scheduled a meeting, will update pr with discussion results

amitaekbote:

This is moved into its own repo, should be good to go

amitaekbote:

mellenburg:
This PR is getting kicked from the train again. There needs to be a bump-ee PR for this. More specifically, the integration test added here fails downstream, so if it needs to be excluded there, then there must a PR to do exactly that vs change.

Also, shouldnt this be coordinated with https://github.com/mesosphere/dcos-enterprise/pull/808 anyway?

darkonie:
@mesosphere-mergebot bump-ee

darkonie:
@mesosphere-mergebot bump-ee

darkonie:
@mesosphere-mergebot bump-ee

amitaekbote:

PR #1461, opened 2017-04-18
Simplify parameter input in dcos-launch

mellenburg:
## High Level Description
When dcos-launch was first implemented there was some difference in the internal data structures 'config' and 'info' and as such, only the create function took the config, which created the info, and all other functions took the info. However, the info has simply become a superset of the config, so these remain identical. Additionally, all of these values can just be set at instantiation for simplification and to avoid needless argument passing, e.g. `get_launcher(info).wait(info)`

## Checklist for all PR's

marginal refactor


mellenburg:
The jenkins/public-dcos-adminrouter-pr can be ignored. There were mass failures for that test at the time and I apparently have no way of re-triggering....

PR #1462, opened 2017-04-18
Add onprem support for dcos-launch

mellenburg:
## High Level Description
-Introduces onprem launcher support for AWS platform
- Removes concept of "bare clusters". Initially, it was designed that a onprem lauch would be a compound launch of a "bare cluster" then an onprem cluster. However, it was much simpler to have the wait function drive the installation and make the creation always be the bare cluster
- Introduces OnpremCluster: a simple abstraction for describing the topology of a cluster (supplants test_util/cluster.py)
- Introduces Ssher: a simple client for binding a ssh key to a user (supplants ssh/tunnel.py)
- Consolidates test command code into the launcher base class

## Related Issues
https://jira.mesosphere.com/browse/DCOS-13299
## Checklist for all PR's


mellenburg:
NOTE: Add a check so that if installer_url is invalid, dcos-launch create will not run

mellenburg:
Integration test results:
AZURE
https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosLaunchTestAzure&buildId=624324
ONPREM-AWS
https://teamcity.mesosphere.io/viewLog.html?buildId=624314&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosLaunchTestAwsOnprem
AWS
https://teamcity.mesosphere.io/viewLog.html?buildId=624312&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosLaunchTest

packages/dcos-launch/extra/launch/aws.py
```diff
@@ -50,6 +48,8 @@ def zen_helper(self, config):
         will be provided (must be done in correct order) and added to the info
         JSON as 'temp_resources'
         """
+        if config['zen_helper'] != 'true':
+            return {}
```
colin-msphere: How likely is this method to be called and this condition being met? Would it be better to throw an error here?

packages/dcos-launch/extra/launch/aws.py
```diff
@@ -105,6 +98,12 @@ def delete_temp_resources(self, temp_resources):
             self.boto_wrapper.delete_vpc(temp_resources['vpc'])
 
     def key_helper(self, config):
+        """ If key_helper is true, then create an EC2 keypair with the same name
+        as the cloudformation stack, update the config with the resulting private key,
+        and amend the cloudformation template parameters to have KeyName set as this key
+        """
+        if config['key_helper'] != 'true':
+            return {}
```
colin-msphere: Same question as `zen_helper` - should throw an error when this condition is met? Is it safe to continue execution?

packages/dcos-launch/extra/launch/aws.py
```diff
@@ -119,18 +118,35 @@ def get_stack(self, info):
...
+        return super().create(config)
+
+    def get_hosts(self, info):
+        return self.get_stack(info).get_host_ips()
+
+    def describe(self, info):
+        return {'host_list': launch.util.convert_host_list(self.get_hosts())}
+
+    def test(self, info, args, env):
+        raise NotImplementedError('Bare clusters cannot be tested!')
```
colin-msphere: Maybe `Testing is not supported on onprem clusters.`

packages/dcos-launch/extra/launch/cli.py
```diff
@@ -95,20 +95,21 @@ def do_main(args):
...
             var_list = args['--env'].split(',')
-            launch.util.check_keys(os.environ, var_list)
-            test_cmd = ' '.join(['{}={}'.format(e, os.environ[e]) for e in var_list]) + ' ' + test_cmd
-        if len(args['<pytest_extras>']) > 0:
-            test_cmd += ' ' + ' '.join(args['<pytest_extras>'])
-        launcher.test(info, test_cmd)
-        return 0
+            missing = [v for v in var_list if v not in os.environ]
+            if len(missing) > 0:
+                raise launch.util.LauncherError(
```
colin-msphere: 👍

packages/dcos-launch/extra/launch/config.py
```diff
@@ -307,11 +311,14 @@ def calculate_instance_count(num_masters, num_private_agents, num_public_agents)
...
+def validate_os_name(os_name, platform, provider):
     # TODO DCOS-14033: [gen.internals] Source validate functions are global only
-    if platform == 'bare_cluster':
+    if provider != 'onprem':
+        # the os_name parameter only applies to homogenous clusters that need
+        # to be created by dcos-launch for an onprem install. In other cases,
+        # the deployment template is parameterized for OS selection
         return
-    elif platform == 'aws':
+    if platform == 'aws':
```
colin-msphere: Could this `if` reuse the `validate_onprem_provider_platform` method? If the `platforms` for `onprem` grew, this place would also need updating.

packages/dcos-launch/extra/launch/config.py
```diff
@@ -323,6 +330,19 @@ def calculate_ssh_private_key(ssh_private_key_filename):
     return load_string(ssh_private_key_filename)
 
 
+def calculate_instance_ami(os_name, aws_region):
+    return test_util.aws.OS_AMIS[os_name][aws_region]
+
+
+def validate_key_name(provider, platform, key_helper):
+    if provider == 'aws' and platform == 'onprem' and key_helper == 'false':
```
colin-msphere: Could this use `validate_onprem_provider_platform` to determine the provider and platform?

packages/dcos-launch/extra/launch/config.py
```diff
@@ -323,6 +330,19 @@ def calculate_ssh_private_key(ssh_private_key_filename):
...
+        raise AssertionError('Either provide `key_name`')
+
+
+def calculate_cluster_size(num_masters, num_private_agents, num_public_agents):
+    return str(1 + int(num_masters) + int(num_private_agents) + int(num_public_agents))
```
colin-msphere: Please add a comment noting the `1` is for the provisioning node.

packages/dcos-launch/extra/launch/onprem.py
```diff
@@ -0,0 +1,138 @@
...
+        return self.get_bare_cluster_launcher(config).create(config)
+
+    def post_state(self, state):
+        self.get_ssher(self.config).command(self.bootstrap_host, ['printf', state, '>', STATE_FILE])
+
+    def get_last_state(self):
+        return self.get_ssher(self.config).command(self.bootstrap_host, ['cat', STATE_FILE]).decode().strip()
+
+    def get_bare_cluster_launcher(self, info):
+        if info['platform'] == 'aws':
```
colin-msphere: Is there a way to re-use `validate_onprem_provider_platform` here?

packages/dcos-launch/extra/launch/onprem.py
```diff
@@ -0,0 +1,138 @@
...
+    ssher.get_home_dir(host, port)
+
+
+class OnpremLauncher(launch.util.AbstractLauncher):
+    def __init__(self, config):
+        # can only be set during the wait command
+        self.bootstrap_host = None
+        self.config = config
+
+    def create(self, config):
```
colin-msphere: Do we need `config` here? Should it be using `self.config` at this point?

packages/dcos-launch/extra/launch/onprem.py
```diff
@@ -0,0 +1,138 @@
...
+
+STATE_FILE = 'LAST_COMPLETED_STAGE'
+
+
+@retrying.retry(wait_fixed=1000, stop_max_delay=120000)
+def check_ssh(ssher, host, port):
+    ssher.get_home_dir(host, port)
+
+
+class OnpremLauncher(launch.util.AbstractLauncher):
```
colin-msphere: This class has a `self.config` set in `__init__`, but several methods expect a separate `config` parameter. Is there a reason these methods (`create`, `post_state`, etc) are not using `self.config`?

test_util/onprem.py
```diff
@@ -0,0 +1,203 @@
...
+    def setup_installer_server(self, installer_url: str, offline_mode: bool):
+        log.info('Setting up installer on bootstrap host')
+        return DcosInstallerApiSession.api_session_from_host(
+            self.ssher, self.bootstrap_host.public_ip, installer_url, offline_mode)
+
+
+@retry(wait_fixed=3000, stop_max_delay=300 * 1000)
+def download_dcos_installer(ssher, host, installer_path, download_url):
+    """Response status 403 is fatal for curl's retry. Additionally, S3 buckets
+    have been returning 403 for valid uploads for 10-15 minutes after CI finished build
```
sschneid: 😒

packages/dcos-launch/extra/launch/README.md
```diff
@@ -8,18 +8,15 @@ across a variety of provider APIs.
...
-- [AWS Bare (no DC/OS) Cluster](sample_configs/aws-bare-cluster.yaml)
 - [Onprem Install on AWS Bare Cluster](sample_configs/aws-onprem.yaml)
-- [Onprem Install on Previously Provisioned Linux Cluster](sample_configs/bare-cluster-onprem.yaml)
 - [Azure Template Deployment](sample_configs/azure.yaml)
 
 ## Keywords and Definitions
 ### Required Fields
 * `launch_config_version`: this is still a tool under active development and as such a strict version specifier must be included
-* `platform`: The provider for hardware on which DC/OS will be deployed. May be one of `aws`, `azure`, or `bare_cluster`. See `deploy_bare_cluster_only` and `provider: onprem`
+* `platform`: The provider for hardware on which DC/OS will be deployed. May be one of `aws` or `azure`
```
mellenburg: This is required for `onprem` and currently it is only supported on `aws`

packages/dcos-launch/extra/launch/aws.py
mellenburg: I don't see why an error would have much pertinence here... The point of this utility is: "If `key_helper: true` in the config YAML file then add a key. Otherwise, do nothing. Doing nothing is the intended default case.

packages/dcos-launch/extra/launch/aws.py
mellenburg: In both cases, the default behavior is to do nothing. If the user didn't supply these parameters/resources, then cloudformation will raise an error.

packages/dcos-launch/extra/launch/aws.py
mellenburg: That is not true. onprem clusters do support being tested, but this is actually the harness for the underlying homogeneous instance group without DC/OS installed so it cannot be tested.

packages/dcos-launch/extra/launch/config.py
mellenburg: That's a good question, but the issue comes down to reporting. When one of these validation checks fail, the error is reported and tied to each of the variables the validate function requires. So, if you were to implement that and provide and invalid configuration value for platform/provider and a valid value for os_name, os_name would show as incorrect with a message about onprem platform support. This would be rather confusing. The goal is to get the errors to be as specific and directed as possible.

The great thing about this implementation is that in ANY of these functions that have platform and provider as arguments will have `validate_onprem_platform_provider` run in so no other validate function needs to worry about if its inputs are valid or not

The way this is set up now is to simply pass on all other situations that are unknown to that validate function. This was necessary because all validation functions are global, which means we cannot conditionally add them.

TL;DR Usually you would be right, but because of how `gen` is implemented, it would make the reported error obfuscated. But also because how `gen` is implemented, your suggestion would not add much functionally

packages/dcos-launch/extra/launch/config.py
mellenburg: See above; we do not want the key_name validator function to reporting when platform and provider are incompatible as it will swamp the output of the validation with meaningless and misleading errors

packages/dcos-launch/extra/launch/onprem.py
mellenburg: Funny that you ask this; I agree and already merged a PR to make all Launchers consume the config and each of the functions use self.config: https://github.com/dcos/dcos/pull/1485/commits/2c75b6868691cde37bc01c4d7f233789633640e9

packages/dcos-launch/extra/launch/onprem.py
mellenburg: no, we don't need it so this was merged: https://github.com/dcos/dcos/pull/1485/commits/2c75b6868691cde37bc01c4d7f233789633640e9
:)

packages/dcos-launch/extra/launch/onprem.py
mellenburg: No need to validate again, nor would we want to validate in this part of the code. The gist is that we want as much validation as possible to occur when the user providers the config to the CLI (or, more specifically whenever the get validated config function is called). Then, we can leave out distracting validation code and have only the bare business logic which can assume that the config can only be valid.

test_util/onprem.py
mellenburg: I can't remember if this has become less of a thing or if we have just tuned this curl command so well it rarely breaks

test_util/onprem.py
```diff
@@ -0,0 +1,203 @@
...
+
+def log_and_raise_if_not_ok(response):
+    if not response.ok:
+        log.error(response.content.decode())
+        response.raise_for_status()
+
+
+class OnpremCluster:
+
+    def __init__(self, ssher, masters, private_agents, public_agents, bootstrap_host):
```
orsenthil: This is a good candidate to document. Please document the args and type of the args for this class.

test_util/onprem.py
```diff
@@ -0,0 +1,203 @@
...
+                      '--create-dirs', '-o', installer_path, download_url])
+    except:
+        log.exception('Download failed!')
+        raise
+
+
+class DcosInstallerApiSession(ApiClientSession):
+    @classmethod
+    def api_session_from_host(
+            cls, ssher: Ssher, host: str, installer_url: str, offline_mode: bool, port: int=9000):
```
orsenthil: style nit: hanging indent will be desirable + docs for this method.

test_util/onprem.py
orsenthil: :(

If this is need not be exposed, you could make this a private.
Else, please consider documenting and typing the args.

packages/dcos-launch/extra/test_onprem.py
```diff
@@ -0,0 +1,20 @@
+import launch.onprem
```
orsenthil: is this import used?

packages/dcos-launch/extra/launch/aws.py
```diff
@@ -120,18 +124,28 @@ def stack(self):
...
-                master_list=[m['private_ip'] for m in details['masters']],
-                agent_list=[a['private_ip'] for a in details['private_agents']],
-                public_agent_list=[a['private_ip'] for a in details['public_agents']],
-                aws_access_key_id=self.config['aws_access_key_id'],
-                aws_secret_access_key=self.config['aws_secret_access_key'],
-                region=self.config['aws_region'],
-                test_cmd=test_cmd)
+
+class BareClusterLauncher(DcosCloudformationLauncher):
+    """ Launches a homogenous cluster of plain AMIs intended for onprem DC/OS
```
orsenthil: s/homogenous/homogeneous

packages/dcos-launch/extra/test_onprem.py
mellenburg: launch is used. launhc.onprem is not. Will amend

mellenburg:
@orsenthil Comments addressed PTAL

mellenburg:
All three configuration (onprem/aws/azure) pass:
https://teamcity.mesosphere.io/viewLog.html?buildId=632129&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosLaunchTest
https://teamcity.mesosphere.io/viewLog.html?build
https://teamcity.mesosphere.io/viewLog.html?buildId=632131&tab=buildResultsDiv&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosLaunchTestAzure

PR #1470, opened 2017-04-19
Reduce dcos-integration-test logging

mellenburg:
## High Level Description
Our testing logs are becoming so large that it takes considerable amount of time to download the logs into the browser, not to mention how cumbersome they become to scroll through. Most of this information is overly verbose and should be accessed on demand by changing the logging level.

This PR does the following:
- Reduce a few loggers from info to debug to reduce clutter
- Increase polling interval to cut down on repeated, uninformative information

## Checklist for all PR's

### DIFF
--- a/test_util/helpers.py
+++ b/test_util/helpers.py
@@ -138,7 +138,7 @@ class ApiClientSession:
             fragment=fragment,
             port=port))
 
-        log.info('Request method {}: {}'.format(method, request_url))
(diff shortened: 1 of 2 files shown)


mellenburg:
Not doing a separate PR; Will update commit message

mellenburg:
@BenWhitehead Ive updated with two commits with a better message. Does that work?

PR #1474, opened 2017-04-20
Add ConnectionError Retrying Mixin to test harness

mellenburg:
## High Level Description
Sporadic RemoteDisconnected / ConnectionError test failures do not illuminate any wider issues not happen at a high enough frequency to warrant dedicated investigation. This change will add a retrying mixin for this specific failure case to the DcosApiSession object used in the integration tests. This is the same treatment applied to marathon polling to reduce test flakiness

## Related Issues
https://jira.mesosphere.com/browse/QUALITY-1426

## Checklist for all PR's

### DIFF
--- a/test_util/dcos_api_session.py
+++ b/test_util/dcos_api_session.py
@@ -15,7 +15,7 @@ import requests
 import retrying
 
 import test_util.marathon
-from test_util.helpers import ApiClientSession, Url
+from test_util.helpers import ApiClientSession, RetryCommonHttpErrorsMixin, Url
 
 
 def get_args_from_env():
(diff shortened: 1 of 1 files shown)


test_util/dcos_api_session.py
```diff
@@ -88,7 +88,7 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
                                    query=query, fragment=fragment, port=port, **kwargs)
 
 
-class DcosApiSession(ARNodeApiClientMixin, ApiClientSession):
+class DcosApiSession(ARNodeApiClientMixin, RetryCommonHttpErrorsMixin, ApiClientSession):
```
BenWhitehead: Is there some style thing here on the order of the mixins/subclassing?

mellenburg: The order of the multiple inheritance is critical. super() in python essentially walks left to right through that list (please no one else kill me for an over simplification)
https://www.youtube.com/watch?v=EiOglTERPEo

BenWhitehead: Sounds good, just wanted to double check since I don't know how the inheritance in python works (haven't learned it yet, since I've trained myself away from most OO inheritance the past couple years.)

orsenthil: well, walks as a diamond.

PR #1476, opened 2017-04-20
Cleanup BareClusterStack

mellenburg:
## High Level Description
- no longer overrides inherited delete function with effectively identical function
- fetch_stack never assumes BareClusterStack; instead assumes generic stack and selectively picks BareClusterStack

## Related Issues
https://jira.mesosphere.com/browse/DCOS-13299

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Code Cleanup

### DIFF
--- a/test_util/aws.py
+++ b/test_util/aws.py
@@ -55,8 +55,11 @@ def fetch_stack(stack_name, boto_wrapper):
         if resource.logical_resource_id == 'MasterServerGroup':
             log.debug('Using Basic DC/OS Cloudformation interface')
             return DcosCfStack(stack_name, boto_wrapper)
-    log.debug('Using VPC Cloudformation interface')
-    return BareClusterCfStack(stack_name, boto_wrapper)
+        if resource.logical_resource_id == 'BareServerAutoScale':
+            log.debug('Using Bare Cluster Cloudformation interface')
+            return BareClusterCfStack(stack_name, boto_wrapper)
+    log.warning('No recognized resources found; using generic stack')
+    return CfStack(stack_name, boto_wrapper)
 
 
 class BotoWrapper():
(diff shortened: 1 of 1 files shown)


test_util/aws.py
```diff
@@ -55,8 +55,11 @@ def fetch_stack(stack_name, boto_wrapper):
         if resource.logical_resource_id == 'MasterServerGroup':
             log.debug('Using Basic DC/OS Cloudformation interface')
             return DcosCfStack(stack_name, boto_wrapper)
-    log.debug('Using VPC Cloudformation interface')
-    return BareClusterCfStack(stack_name, boto_wrapper)
+        if resource.logical_resource_id == 'BareServerAutoScale':
+            log.debug('Using Bare Cluster Cloudformation interface')
+            return BareClusterCfStack(stack_name, boto_wrapper)
+    log.warning('No recognized resources found; using generic stack')
+    return CfStack(stack_name, boto_wrapper)
```
orsenthil: Question (not a blocker) - How does this change in behavior affects the caller (or the methods that currently call fetch_stack)?

mellenburg: The purpose of this function is pretty narrow: "here is a stack ID, I know nothing else about htis cloudformation, but I want an object to poll its creation or delete it". Before this change, if the DC/OS specific-resources were not found, this code would return the barecluster stack (even if it was not sure it was a bare cluster stack). This way, we intentionally get the most vague object.

You might ask "why do we not just tag the metadata for this cluster with the appropriate class type to use in the future" to which I can respond that that would require the user to know about this interface and specify the exact class handler to use. This would be a very poor API for launching dcos. Thus, you would just write this function anyway to guess what the user was providing.

TL;DR This will add a little stability to dcos-launch by not making too specific of an assumption about an unknown resource. This code should perhaps live with dcos-launch but I feel dcos-launch shouldnt be overly concerned with the cloudformation wrapper internals, so that question is out of scope for this PR

PR #1486, opened 2017-04-24
Admin Router: Authz/authn should be done before the rewrite.

vespian:
## High Level Description

This PR enforces authn/authz *before* rewrite stage by calling `auth` code inside `rewrite_by_lua_block` before calling the lua code responsible for rewrite. This concerns following endpoints:
* `/service`
* `/agent`
* `/system/v1/leader/marathon`
* `/system/v1/agent/`

Some context:
OpenResty processes directives in a specific order: https://pbs.twimg.com/media/Ci3mA9VXIAAOC6K.jpg Unfortunately for our use-case, the rewrite stage goes before auth stage and thus the Lua code executed in rewrite stage takes precedence. So if rewrite methods decide that i.e. there is no such agent, then the 404 is returned even though JWT of the client sending the request should be evaluated first.

We cannot change the order in which OpenResty processes data but basing on the docs, there is no real difference between [rewrite](url) and [auth](url) stages (except subrequests which we do not use in these cases) so we just move the auth calls into the `rewrite_by_lua` block, e.g.
```
           rewrite_by_lua_block {
                auth.access_agent_endpoint();
                agent.resolve();
￼            }
```

## Related Issues

  - [DCOS-5191](https://jira.mesosphere.com/browse/DCOS-5191) Adminrouter should return 404 when accessing /service/unknownservice
  - [DCOS-14008](https://jira.mesosphere.com/browse/DCOS-14008) Admin Router: do authN/authZ before executing other endpoint logic (do not leak info)

## Dependencies

Related EE PR: https://github.com/mesosphere/dcos-enterprise/pull/831
AR NEXT PR: https://github.com/mesosphere/dcos/pull/17


vespian:
Superseeded by https://github.com/mesosphere/dcos/pull/17

PR #1492, opened 2017-04-25
Bump master to 1.10-dev.

adam-mesos:
1.9.0 has been branched off and shipped. Master is now working on 1.10, so the version should be reflected to say as much.

## Checklist for all PR's


orsenthil:
`dcos_installer/exhibitor_migrate.py`  has this statement. This could be removed as part of this PR.
```
When a user is updating to DC/OS 1.9 from an older version we need to update several values in exhibitor to use the new.```

**Comment by spahl** [comment]
Good catch Adam!

Agreed on the exhibitor comment @orsenthil

**Comment by adam-mesos** [comment]
That comment at the top of `dcos_installer/exhibitor_migrate.py` makes me wonder if we should delete/revert the entirety of https://github.com/dcos/dcos/commit/8518c60ccf49844b8680e5f49ef62e567173df67 if it is no longer needed. @BenWhitehead ?
That seems unrelated to this version bump PR.

**Comment by adam-mesos** [comment]
@mesosphere-mergebot bump-ee

**Comment by BenWhitehead** [comment]
I think it's probably okay to revert that commit. We will have to update the upgrade script I think to no longer try and call it as well as update the dcos for dcos/dcos and mesosphere/dcos-enterprise to no longer reference the script.

File: setup.py [hunk 1]
```diff
@@ -69,8 +69,6 @@ def get_advanced_templates():
             'pkgpanda=pkgpanda.cli:main',
             'mkpanda=pkgpanda.build.cli:main',
             'dcos_installer=dcos_installer.cli:main',
```
BenWhitehead: Not sure if python allows trailing commas, this may need to be cleaned up.

BenWhitehead:
It would be a good idea to get @branden or @amitaekbote to review the changes to the upgrade script.

setup.py
adam-mesos: Nothing new. There were trailing commas before and it seemed to work alright.
Happy to remove a comma though, if requested.

BenWhitehead: If it's allowed I'm okay leaving it there.

orsenthil: It allows. No Harm and we don't define this in our style guide.

adam-mesos:
@mesosphere-mergebot bump-ee

PR #1496, opened 2017-04-26
test_upgrade_vpc: Change the address that DNS test resolves

nlsun:
Since the previous address won't work, we'll switch to testing one that
should.

## High Level Description

Changes the upgrade test to use the "agentip" address

## Related Issues

  - [DCOS-15382](https://jira.mesosphere.com/browse/DCOS-15382)

## Checklist for all PR's

___

Reviewers should be:

### DIFF
--- a/test_util/test_upgrade_vpc.py
+++ b/test_util/test_upgrade_vpc.py
@@ -173,6 +173,11 @@ def create_marathon_dns_app(app_id: str, healthcheck_app_id: str) -> dict:
     # DNS resolution app to make sure DNS is available during the upgrade.
     # Periodically resolves the healthcheck app's domain name and logs whether
     # it succeeded to a file in the Mesos sandbox.
+
+    dns_parsed_app_name = marathon_app_id_to_mesos_dns_subdomain(healthcheck_app_id)
+    resolve_name = '{}.marathon.agentip.dcos.thisdcos.directory'.format(dns_parsed_app_name)
+    mesos_dns_name = '{}.marathon.mesos'.format(dns_parsed_app_name)
+
     return {
         "id": '/' + app_id,
         "cmd": """
@@ -189,7 +194,8 @@ do
 done
 """,
         "env": {
-            'RESOLVE_NAME': marathon_app_id_to_mesos_dns_subdomain(healthcheck_app_id) + '.marathon.mesos',
+            'RESOLVE_NAME': resolve_name,
+            'MESOS_DNS_NAME': mesos_dns_name,
             'DNS_LOG_FILENAME': 'dns_resolve_log.txt',
             'INTERVAL_SECONDS': '1',
             'TIMEOUT_SECONDS': '1',
@@ -317,7 +323,7 @@ class VpcClusterUpgradeTest:
             dcos_api.marathon.ensure_deployments_complete()
             # This is a hack to make sure we don't deploy dns_app before the name it's
             # trying to resolve is available.
-            self.wait_for_dns(dcos_api, dns_app['env']['RESOLVE_NAME'])
+            self.wait_for_dns(dcos_api, dns_app['env']['MESOS_DNS_NAME'])
             dcos_api.marathon.deploy_app(dns_app, check_health=False)
             dcos_api.marathon.ensure_deployments_complete()
 


nlsun:
@branden How do we run the upgrade test on this?

orsenthil:
Trigger on this: https://teamcity.mesosphere.io/viewType.html?buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssUpgradeTestOnPre

nlsun:
@GoelDeepak @drewkerrigan @branden ready for review.

nlsun:
Oh actually the success was because someone muted the failing test (this test). We still have to wait for this test to run it seems.

test_util/test_upgrade_vpc.py
```diff
@@ -173,9 +172,19 @@ def create_marathon_dns_app(app_id: str, healthcheck_app_id: str) -> dict:
...
+    resolve_name = '{}.marathon.agentip.dcos.thisdcos.directory'.format(dns_parsed_app_name)
+
     return {
         "id": '/' + app_id,
         "cmd": """
+until host -W $TIMEOUT_SECONDS $RESOLVE_NAME
+do
+    echo "Waiting to resolve $RESOLVE_NAME for the first time..."
+    sleep 1
+done
```
branden: Why is this change necessary? Seems unrelated to changing the name that this test resolves.

If `$RESOLVE_NAME` can never be resolved, this app will produce an empty log, which `test_app_dns_survive_upgrade()` will interpret as success (no resolution failures in an empty list). If we go this route, we'll need to add an assertion that the log has at least one entry.

nlsun: Previously the wait mechanism was query the mesos dns http api (see the code i deleted). Since this will contain the incorrect info, we need the logic to be in the task

branden: Sgtm. In that case you'll need to update the test to cover the scenario where the app never succeeds in resolving the name.

nlsun:
Added check for empty dns_log

test_util/test_upgrade_vpc.py
branden: Thinking about it some more -- we're losing the guarantee that the name was resolvable _before_ the upgrade. If the name isn't resolvable for some reason until after the upgrade, the test will still pass, even though it's not covering what we expect. So if the app itself is responsible for waiting, we have to do something gross like record the time when we start the upgrade and then compare it with the first timestamp in the log. :(

Is there any way for us to maintain that guarantee without comparing timestamps recorded on different machines?

nlsun: Hmm another option is to keep the mesos dns http api check, but just check a different dns address than the one that we're resolving (both dns addresses point to the same app).

The reason I originally removed the check was because i didn't want to create 2 separate dns addresses, one that is in mesos-dns and another that we're actually testing. But it seems that that way is actually preferable to this.

nlsun:
Switched to having a separate mesos dns name and resolve name

nlsun:
I just realized that whatever method we choose, we'll some some form of dns downtime as at some point in time the dns resolver itself is upgraded. Should we just remove this test?

GoelDeepak:
Yeah, we should remove this test. It is like broken hacks to support this versus a simple retry of DNS query.

nlsun:
@branden With this change we should be able to handle master upgrades without DNS downtime. However, upon an agent upgrade, that agent will temporarily not have a functioning DNS resolver. Are there hooks to only run this test while masters are upgrading?

branden:
@nlsun There aren't hooks for testing something only while we're upgrading masters. The DNS test during upgrades is currently muted since it doesn't pass on master, so we could merge these changes and leave that as follow-up work.

nlsun:
@branden Assigning jira to you but feel free to reassign it: https://jira.mesosphere.com/browse/DCOS-15546

nlsun:
Should we ship this for now then?

jgehrcke:
@nlsun @branden what is the status of this? Should this be revived or is it superseded?

nlsun:
We'll revisit this later

PR #1501, opened 2017-04-27
Switch EC2 instance type default from m3 to m4

lloesche:
## High Level Description

This changes the default AWS EC2 instance type from m3 to m4 in the generated Cloudformation Templates.
m4 are about 20% cheaper than m3 and also a bit faster. However they are EBS only.
m3 will eventually be phased out.

### DIFF
--- a/gen/aws/templates/advanced/advanced-master.json
+++ b/gen/aws/templates/advanced/advanced-master.json
@@ -44,8 +44,8 @@
     },
     "MasterInstanceType": {
       "Type": "String",
-      "Default": "m3.xlarge",
-      "Description" : "\nRegion-specific instance type. E.g. m3.xlarge"
+      "Default": "m4.xlarge",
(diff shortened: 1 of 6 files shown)


lloesche:

jgehrcke:
@lloesche what is the status of this? Do we have a related JIRA ticket that we should link to?

amitaekbote:
can you check with the infinity team to make sure this doesnt break their tests? Last I checked it didnt work 

@gabrielhartmann @benclarkwood

MrMarvin:
@jgehrcke Yes, this is linked to https://jira.mesosphere.com/browse/OPS-684
I'll also link this PR from there and we can continue the discussion if and what breaks in the jira.

eastlondoner:
Don't you need to change the cloud init code which formats and mounts the ephemeral disks (since there are no longer ephemeral disks?)

lloesche:
Closing in favour of new PR.

PR #1507, opened 2017-04-28
Remove public master host argument in integration tests

mellenburg:
## High Level Description
This parameter is essentially useless as it doesnt need to be a public address and in any circumstance where it is actually used, the same private IP is provided for both this and the MASTER_HOSTS argument

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [ ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #1508, opened 2017-04-28
Allow running integration tests without arguments

mellenburg:
## High Level Description
Modifies the integration test suite to be triggerable from the commandline with nothing more than:
```
source /opt/mesosphere/environment.export
cd /opt/mesosphere/active/dcos-integration-test/
pytest
```
The cluster topology arguments can still be provided via the same method, but now they only need to be provided when one wants to verify that provisioning has been successful. 

Note: This PR depends on https://github.com/dcos/dcos/pull/1507
## Related Issues
https://jira.mesosphere.com/browse/DCOS-13464

## Checklist for all PR's


test_util/dcos_api_session.py
```diff
@@ -89,31 +68,92 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
...
         """
         super().__init__(Url.from_string(dcos_url))
-        self.masters = sorted(masters)
-        self.public_masters = sorted(public_masters)
-        self.slaves = sorted(slaves)
-        self.public_slaves = sorted(public_slaves)
-        self.all_slaves = sorted(slaves + public_slaves)
+        self.master_list = masters
+        self.slave_list = slaves
+        self.public_slave_list = public_slaves
```
BenWhitehead: Does losing the `sorted` on these lists result in different behavior that people might be expecting right now?

BenWhitehead: Nvm, I see the `sorted` on the getters below now. 

Though, I do wonder why we don't just sort them to begin with and return that?

test_util/dcos_api_session.py
```diff
@@ -89,31 +68,92 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
...
         self.default_os_user = default_os_user
         self.auth_user = auth_user
 
-        assert len(self.masters) == len(self.public_masters)
+    @staticmethod
+    def get_args_from_env():
+        """ Provides the required arguments for a unauthenticated cluster
+        """
+        masters = os.getenv('MASTER_HOSTS')
```
BenWhitehead: Do we want to add "pretty" error messages here in the case any of these env variables aren't defined (it seems like a cluster without any nodes isn't really valid)

test_util/dcos_api_session.py
mellenburg: We do it as a property so the agent list can be reset during the test and still have the guarantee that the list is always sorted. See: https://github.com/dcos/dcos/pull/1508/files#diff-0ba5c9c4483f599539d64bc5a299545dR93

test_util/dcos_api_session.py
mellenburg: The point of this PR is that you CAN have valid tests without declaring ANYTHING

The only reason we want the user to input the master list, agent list, and public agent list is so that we can verify that provisioning was 100% successful. This will still be done in all of our PRs. However, when you just have an arbitrary cluster and want to test it, you don't want to have to provide ground-truth when you just want your current cluster tested.

Anyway, I already added logging for when these params are not provided and are set by the harness: https://github.com/dcos/dcos/pull/1508/files#diff-f50a13699d4598c4a33fe8b1701335e8R137

I can bump it up from debug to warning as that might be more appropriate

BenWhitehead: If it's something that could catch someone off guard I would set the log level to warning. If it's not going to catch off guard, info would be good. Debug sounds a little too sneaky for derivation stuff.

test_util/dcos_api_session.py
```diff
@@ -278,6 +318,7 @@ def _wait_for_metronome(self):
     def wait_for_dcos(self):
         self._wait_for_adminrouter_up()
         self._authenticate_default_user()
+        self.set_node_lists_if_unset()
```
mellenburg: I was considering making a separate entrypoint that allows you to go on without declaring args, but I think now it is just better to make this function private

test_util/dcos_api_session.py
```diff
@@ -67,8 +48,6 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
         node argument. node must be a host in self.master or self.all_slaves. If given,
         the request will be made to the Admin Router endpoint for that node type
         """
-        assert self.masters
-        assert self.all_slaves
```
mellenburg: This was removed because it was too generic of an assert. What you recommend should work well, so I will add that.

test_util/dcos_api_session.py
mellenburg: I flirted with making this public, but that no longer has a use case; will make it private as well.

test_util/dcos_api_session.py
```diff
@@ -89,31 +68,92 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
...
+
+    @property
+    def public_slaves(self):
+        return sorted(self.public_slave_list)
+
+    @property
+    def all_slaves(self):
+        return sorted(self.slaves + self.public_slaves)
+
+    def set_node_lists_if_unset(self):
```
mellenburg: The expected behavior is to fill in whatever is unset, i.e. the former option. The intention is that perhaps you only care about making sure the public agnents provisioned, then you can provide just that value and the other nodes will be taken from DC/OS so you can run your test. I think the logging statements already communicate enough but I do not mind bumping them up a level

test_util/dcos_api_session.py
```diff
@@ -89,31 +72,92 @@ def api_request(self, method, path_extension, *, scheme=None, host=None, query=N
...
+    @property
+    def all_slaves(self):
+        return sorted(self.slaves + self.public_slaves)
+
+    def _set_node_lists_if_unset(self):
+        """ Sets the expected cluster topology to be the observed cluster
+        topology from exhibitor and mesos. I.E. if masters, slave, or
+        public_slaves were not provided, accept whatever is currently available
+        """
+        if self.master_list is None:
```
orsenthil: Is it intentional that you want allow users to specific master only and leave to slaves and public_slave blank?
Is there a good reason to allow that?

If it is going to be either a) Specify masters, slaves and public slaves OR b) specify None, we will figure it out using mesos, then, this code should see some change.

mellenburg: See above comments. There is a great reason to allow it and I believe there is sufficient logging around this potential discrepancy.

orsenthil: Which comment?  Is this ?

```
The only reason we want the user to input the master list, agent list, and public agent list is so that we can verify that provisioning was 100% successful. This will still be done in all of our PRs. However, when you just have an arbitrary cluster and want to test it, you don't want to have to provide ground-truth when you just want your current cluster tested.

```
I didn't think, it addresses the choice between a) Provide complete information vs Provide partial information (of master list).

Is that, we don't care what the customer provides, if he provides master list, that's fine, we will figure out the slave list?

orsenthil: Michael and I discussed this offline.

* There was some value in letting users specify a partial list.
* If we have to gather information on slaves using mesos end-point , we get only if he/she has not set it already.

mellenburg:
NOTE: This PR must be shipped with its downstream counterpart:
https://github.com/mesosphere/dcos-enterprise/pull/866

PR #1516, opened 2017-05-02
Fixed: Partial download of dcos_generate_config.sh results in unrecoverable error that requires manual intervention.

lloesche:
## High Level Description

Currently if `dcos_generate_config.sh` is truncated or otherwise corrupted and run it'll extract a corrupted tarball that can't be loaded into Docker.
Then when `dcos_generate_config.sh` is redownloaded it'll fail to run as it'll find a tarball with matching name which causes it to not extract the image tarball and load it into Docker.

Resulting in:
```
Unable to find image 'mesosphere/dcos-genconf:0ce03387884523f026-58fd0833ce81b6244f' locally
Trying to pull repository docker.io/mesosphere/dcos-genconf ...
/usr/bin/docker-current: unauthorized: authentication required.
See '/usr/bin/docker-current run --help'.
```

This PR adds an additional check that, if a tarball of the same name already exists it confirms that the image was actually loaded into Docker. If not it removes the tarball causing it to be extracted and loaded from the current `dcos_generate_config.sh` installer.

PR also contains some minor syntax and style fixes. I also added two stdout to stderr redirects when loading the image. Otherwise the `--version` json output will be corrupted when `dcos_generate_config.sh` is run for the first time.

## Related Issues

  - [DCOS_OSS-1047](https://jira.mesosphere.com/browse/DCOS_OSS-1047) Partial download of dcos_generate_config.sh results in unrecoverable error that requires manual intervention.

### DIFF
--- a/gen/build_deploy/bash/dcos_generate_config.sh.in
+++ b/gen/build_deploy/bash/dcos_generate_config.sh.in
@@ -12,6 +12,17 @@ backout(){{
     exit 1
 }}
 trap 'backout' INT
+
+# if the tarball was previously extracted
+if [ -f "{genconf_tar}" ]; then
+    # but not successfully loaded into Docker
+    if [ -z "$(docker images -q {docker_image_name} 2>/dev/null)" ]; then
+        # remove the potentially corrupted tarball and
+        # cause it to be re-extracted in the next step
+        rm -f {genconf_tar}
+    fi
+fi
+
 # extract payload and load into docker if not extracted
 if [ ! -f "{genconf_tar}" ]; then
     echo Extracting image from this script and loading into docker daemon, this step can take a few minutes
--- a/gen/build_deploy/bash/dcos_generate_config.sh.in
+++ b/gen/build_deploy/bash/dcos_generate_config.sh.in
@@ -1,5 +1,10 @@
 #!/bin/bash
 # variant: {variant}
+#
+# All logging and tool output should be redirected to stderr
+# as the Docker container might output json that would
+# otherwise be tainted.
+#
 set -o errexit -o nounset -o pipefail
 
 # preflight checks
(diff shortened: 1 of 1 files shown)


lloesche:

lloesche:
@mesosphere-mergebot bump-ee

lloesche:

lloesche:
@mesosphere-mergebot bump-ee

lloesche:
@mesosphere-mergebot bump-ee

amitaekbote:
@lloesche can you add a test for this? unit test should do. Also the tests are failing

lloesche:
@amitaekbote 64 tests passed which means DC/OS was installed. No idea what's up with the other 25 failed tests, they all received a "502 Bad Gateway" error. Shouldn't really be an issue with this PR though, it only touches the installer's bash bootstrap part. I.e. either DC/OS will be installed or it won't. Nothing's been touched once the installer runs.

I'll rebase on current master to see if that makes any difference.

Re: integration test, where/how is the bash code currently being tested, if at all? I'm not sure how to go about testing it. The test would need to know the tarball name that's going to be extracted from the dcos_generate_config.sh script. It could then touch an empty file with the same name. If everything works as expected the installer would remove the empty file and extract the correct tarball instead of throwing an error as it does now.

jgehrcke:
@lloesche I'd appreciate if you could rebase this onto current master again and resolve the conflicts that GitHub seems to have detected. Then we can have another look at the CI reports (the current report is not useful at all, with `create-release-pr` having failed).

jgehrcke:
Thanks for rebasing, @lloesche. 

https://teamcity.mesosphere.io/viewLog.html?buildId=665232 failed for unknown reasons, restarting.

jgehrcke:
@mesosphere-mergebot bump-ee

lloesche:
@mesosphere-mergebot bump-ee

PR #1521, opened 2017-05-04
Admin Router: make history service upstream go through another Admin Router, conditionally. 

vespian:
## High Level Description

PR may seem big, but at its core there is a very simple change
* use `ip_detect_script` during AR start to set HOST_IP variable. All Mesos Masters bind to this IP and advertise it in `leader.mesos` when they are leading the cluster.
* make AR cache periodically resolve `leader.mesos` variable and compare it to HOST_IP var described in the previous paragraph
  * if they differ - local mesos master is not the cluster leader - proxy it to the `leader.mesos`
  * if they are the same - proxy it to local mesos master
* in order to prevent proxy loop during failovers/when the `leader.mesos` alias is unstable, a header is added to proxied request (`DCOS-Forwarded`). AR checks it before proxying the request again and if it turns out that header is present *AND* local mesos master is not the leader - request is dropped.
* the path of the request is adjusted dynamically - if it's proxied to another AR, it's left as-is, if the upstream is local `dcos-history-service` process - the `/dcos-history-service` prefix is trimmed.
* make history-service listen on localhost

The idea to resolve local IP using ip detect scripts is not new - Marathon and Metronome do it already. The decision to use DNS resolving instead of querying local master and checking if we got redirected stems from the fact that it's more reliable this way - Spartan does dual-dispatch and MesosDNS uses local ZK state. So even if local Mesos is down, AR will still be able to proxy to correct node.

The client IP is properly transferred to the backend as well, e.g.
```
X-Forwarded-For: 213.61.89.166, 10.0.6.11
```

The rest is just boilerplate and testing:
* add DNS mock, basing on @mhrabovcin work in https://github.com/dcos/adminrouter/pull/50, yet heavily modified. Additionally, I removed the old mock and it's config.
* adjust AR start scripts in order to provide HOST_IP var at start
* add support for `leader.mesos` caching in a reusable manner - we can build on top of this infrastructure and replace Nginx reload timer.

## Related Issues

  - [DCOS-15335](https://jira.mesosphere.com/browse/DCOS-15335) Admin Router: make history service upstream go through another Admin Router, conditionally
  - [DCOS-15033](https://jira.mesosphere.com/browse/DCOS-15033) Make history service listen on localhost, enable access through Admin Router only

## Related PRs:
 EE PR: https://github.com/mesosphere/dcos-enterprise/pull/885


packages/adminrouter/extra/src/common/util.lua
```diff
@@ -37,6 +37,31 @@ function util.get_file_content(path)
 end
 
 
+function util.verify_ip(ip)
+  -- Based on http://stackoverflow.com/a/16643628
+  -- Return True if ip is a valid IPv4 IP, False otherwise.
+  if type(ip) ~= "string" then return false end
+
+  -- check for format 1.11.111.111 for ipv4
+  local chunks = {ip:match("(%d+)%.(%d+)%.(%d+)%.(%d+)")}
```
gpaul: Probably silly but I think to be perfectly correct we want ^ in front and $ at the back to anchor the pattern. Otherwise `aaa.10.0.20.3` probably passes this test.

packages/adminrouter/extra/src/master/cache.lua
```diff
@@ -293,6 +294,82 @@ local function fetch_and_store_state_mesos(auth_token)
...
+    -- We want to use Navstar's dual-dispatch approach and get the response
+    -- as fast as possible from any operational MesosDNS. If we set it to local
+    -- instance, its failure will break this part of the cache as well.
+	--
+	-- Also, see https://github.com/openresty/lua-resty-dns#limitations
+    r, err = resolver:new{
+        nameservers = {{"198.51.100.1", 53},
+                       {"198.51.100.2", 53},
+                       {"198.51.100.3", 53}},
+        retrans = 3,  -- 5 retransmissions on receive timeout
```
gpaul: the comment says 5, the code says 3?

packages/adminrouter/extra/src/master/cache.lua
```diff
@@ -293,6 +294,82 @@ local function fetch_and_store_state_mesos(auth_token)
...
+        timeout = 2000,  -- 2 sec
+    }
+
+    if not r then
+        ngx.log(ngx.ERR, "Failed to instantiate the resolver: " .. err)
```
gpaul: i like this whole error handling strategy

packages/adminrouter/extra/src/common/util.lua
vespian: Totally! Thx!

packages/adminrouter/extra/src/master/cache.lua
vespian: I have changed it to:
```
retrans = 3,  -- retransmissions on receive timeout
timeout = 2000,  -- msec
```

So that we do not have to keep comments in sync. Thx!

packages/adminrouter/extra/src/master/cache.lua
vespian: I think this is your inner-gopher speaking ;)

packages/adminrouter/extra/src/master/history-service.lua
```diff
@@ -0,0 +1,50 @@
...
+        ngx.req.set_uri(string.sub(ngx.var.uri, 22))
+        ngx.var.historyservice_upstream = "http://127.0.0.1:15055"
+    else
+        -- Let's prevent infinite proxy loops during failovers when the `leader.mesos`
+        -- can't be reliably determined. Prefixing custom headers with `X-`
+        -- is no longer recommended:
+        -- http://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions
+        if ngx.req.get_headers()["DCOS-Forwarded"] == "true" then
+            ngx.status = ngx.HTTP_SERVICE_UNAVAILABLE
+            ngx.say("503 Service Unavailable: Mesos leader election in progress")
```
mhrabovcin: Maybe we shouldn't add specific message as we have no idea whether the election is in progress. Remove `Mesos leader election in progress` ?

vespian: True.

`Mesos leader is unknown'?

packages/adminrouter/extra/src/master/cache.lua
```diff
@@ -293,6 +294,82 @@ local function fetch_and_store_state_mesos(auth_token)
...
+        retrans = 3,  -- retransmissions on receive timeout
+        timeout = 2000,  -- msec
+    }
+
+    if not r then
+        ngx.log(ngx.ERR, "Failed to instantiate the resolver: " .. err)
+        return
+    end
+
+    answers, err = r:query("leader.mesos")
```
mhrabovcin: Should be `err` handled as well?

packages/adminrouter/extra/src/master/cache.lua
```diff
@@ -387,6 +464,10 @@ local function refresh_cache(from_timer, auth_token)
         fetch_and_store_marathon_leader(auth_token)
     end
 
+    if refresh_needed("mesos_leader_last_refresh") then
+        fetch_and_store_mesos_leader_state(auth_token)
```
mhrabovcin: No need to pass `auth_token`.

packages/adminrouter/extra/src/master/history-service.lua
mhrabovcin: 👍

packages/adminrouter/extra/src/master/cache.lua
vespian: It's only a string describing the error if/when it happens, answer being 'nil' is the real error "flag".

If you are curious, please have a look at https://github.com/openresty/lua-resty-dns#query

packages/adminrouter/extra/src/master/cache.lua
vespian: Truth has been spoken.

packages/adminrouter/extra/src/common/util.lua
```diff
@@ -37,6 +37,31 @@ function util.get_file_content(path)
...
 
+function util.verify_ip(ip)
+  -- Based on http://stackoverflow.com/a/16643628
+  -- Return True if ip is a valid IPv4 IP, False otherwise.
+  if type(ip) ~= "string" then return false end
+
+  -- check for format 1.11.111.111 for ipv4
+  local chunks = {ip:match("^(%d+)%.(%d+)%.(%d+)%.(%d+)$")}
+  if #chunks == 4 then
+    for _,v in pairs(chunks) do
```
mhrabovcin: Nit: `for _, v`

vespian: yep, fixing!

packages/adminrouter/extra/src/master/cache.lua
```diff
@@ -293,6 +294,82 @@ local function fetch_and_store_state_mesos(auth_token)
...
+    if answers[1].address == HOST_IP then
+        res_body = '{"is_local": "yes", "leader_ip": "'.. HOST_IP ..'"}'
+        ngx.log(ngx.INFO, "Mesos Leader is local")
+    else
+        res_body = '{"is_local": "no", "leader_ip": "'.. answers[1].address ..'"}'
+        ngx.log(ngx.INFO, "Mesos Leader is non-local: `" .. answers[1].address .. "`")
+    end
+
+    ::store_cache::
+    if not cache_data("mesos_leader", res_body) then
```
mhrabovcin: Maybe we can add some comment that we're on purpose ignoring `TTL` of the record and going with own refresh cycle period. I am also wondering what should we do in case if record TTL is shorter than our cache period.

packages/adminrouter/extra/src/common/http.conf
```diff
@@ -35,6 +46,7 @@ init_by_lua_block {
     auth = require "common.auth.open".init(os.getenv("ADMINROUTER_ACTIVATE_AUTH_MODULE"));
     -- Initialize cache:
     cache = require "master.cache".init(nil);
+    historyservice = require "master.history-service".init();
```
mhrabovcin: unused variable?

packages/adminrouter/extra/src/master/cache.lua
vespian: > Maybe we can add some comment that we're on purpose ignoring TTL of the record and going with own refresh cycle period.

Will do.

> I am also wondering what should we do in case if record TTL is shorter than our cache period.
Few things:
* At some point, we may split update coroutines anyway - one per cache type. This way DNS "cache" can be updated even every few seconds.
* I do not think that we will go below 30s anytime soon. We already discussed decreasing `mesos.leader` TTL to something shorter than 60s, my guess is that we will not go lower than 30s (right?). Let's cross this bridge when we get there :)
* try looking at it this way: assuming that Navstar and MesosDNS do not do caching on their own, we just treat DNS as an interface to obtain current mesos leader data. How long we cache it is IMO just an internal implementation detail of AR.

packages/adminrouter/extra/src/common/http.conf
vespian: No, this module/library is later used in `rewrite_by_lua` to establish the destination for `proxy_pass` directive in `location /dcos-history-service/` block.

vespian:

vespian:
Merged in #1534

jgehrcke:
Nice work Pawel!

PR #1458, opened 2017-04-14
Update cosmos to version 0.2.1, for 1.8.9

cruhland:
## High Level Description

This updates cosmos to version 0.2.1, which backports the bugfix to the `/package/list` endpoint from this PR: https://github.com/dcos/cosmos/pull/563.

This should be merged at the same time as the corresponding downstream Enterprise PR: https://github.com/mesosphere/dcos-enterprise/pull/807

## Related Issues

  - [DCOS-15077](https://jira.mesosphere.com/browse/DCOS-15077) Backport handling the case when a service from a package is running but the package doesn't exist

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: this is a backported feature; the test is in mainline cosmos
  - [X] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [X] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [X] Change log from the last version integrated (this should be a link to commits for easy verification and review): [change log with diff](https://github.com/dcos/cosmos/releases/tag/v0.2.1)
  - [X] Test Results: [here](https://teamcity.mesosphere.io/viewLog.html?buildId=618203&tab=buildResultsDiv&buildTypeId=DcosIo_Cosmos_PublishRelease)
  - [ ] Code Coverage (if available):
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and Review Process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using github's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

Reviewers should be:
* Developers who understand the code being modified.
* Developers responsible for code that interacts with or depends on the code being modified.

It is best to proactively ask for 2 reviews by @mentioning the candidate reviewers in the PR comments area. The responsibility is on the developer submitting the PR to follow-up with reviewers and make sure a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

PR #1456, opened 2017-04-13
Backporting the fix for DCOS_OSS-683 to 1.8

surdy:
## High Level Description

What features does this change enable? What bugs does this change fix? High-Level description of how things changed.
In newer versions of CoreOS, libcurl shipped with DC/OS is incompatible(old) with system curl, leading to `curl: (48) An unknown option was passed in to libcurl` error. This has laready been fixed in 1.9 but 1.8.8 had already been released before the fix so backporting the fix to 1.8.9

## Related Issues

  - [DCOS_OSS-683](https://jira.dcos.io/browse/DCOS_OSS-683) Latest Stable version of CoreOS break dcos-adminrouter

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [X ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ X] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

PR #1445, opened 2017-04-11
Bumped Mesos to include fixes for leakage of sensitive data - Mesos 1.0.4

tillt:
## High Level Description

Generally fixes sensitive data leakage to task sandbox. 

Updates Mesos package to be based on Apache/Mesos 1.0.4 (WIP). 

## Related Issues

  - [DCOS-11709](https://jira.mesosphere.com/browse/DCOS-11709) Secrets printed directly to the logs.
   -  [MESOS-2537] - AC_ARG_ENABLED checks are broken
   -  [MESOS-6606] - Reject optimized builds with libcxx before 3.9
   -  [MESOS-7008] - Quota not recovered from registry in empty cluster.
   -  [MESOS-7265] - Containerizer startup may cause sensitive data to leak into sandbox logs.
   -  [MESOS-7366] - Agent sandbox gc could accidentally delete the entire persistent volume content.
   -  [MESOS-7383] - Docker executor logs possibly sensitive parameters.
   -  [MESOS-7422] - Docker containerizer should not leak possibly sensitive data to agent log.

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): [dcos-mesos-1.0.3-rc1...dcos-mesos-1.0.x-4154f66](https://github.com/mesosphere/mesos/compare/dcos-mesos-1.0.3-rc1...dcos-mesos-1.0.x-4154f66)
  - [x] Test Results: https://jenkins.mesosphere.com/service/jenkins/view/Mesos/job/mesos/job/Mesos_CI-build/722/
  - [ ] Code Coverage (if available): [link to code coverage report]

PR #1522, opened 2017-05-04
Train 159 (targeting 1.8.9)

lingmann:
## High Level Description

Train 159 targeting 1.8.9 branch

#1494 Backporting https://jira.mesosphere.com/browse/DCOS-10327 to 1.8.9
#1458 Update cosmos to version 0.2.1
#1456 Backporting the fix for DCOS_OSS-683 to 1.8
#1445 Bumped Mesos to include fixes for leakage of sensitive data - Mesos 1.0.4

### DIFF
--- a/packages/adminrouter/extra/dcos-adminrouter.service
+++ b/packages/adminrouter/extra/dcos-adminrouter.service
@@ -22,7 +22,7 @@ ExecStartPre=/bin/ping -c1 marathon.mesos
 ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/usr/bin/test -f /var/lib/dcos/cluster-id
 # Wait for auth backend to seed the secret key
-ExecStartPre=/usr/bin/curl --fail -sS -o /dev/null 127.0.0.1:8101/acs/api/v1/groups
+ExecStartPre=/opt/mesosphere/bin/curl --fail -sS -o /dev/null 127.0.0.1:8101/acs/api/v1/groups
 ExecStart=$PKG_PATH/nginx/sbin/nginx -c $PKG_PATH/nginx/conf/nginx.master.conf
 ExecReload=/usr/bin/kill -HUP $MAINPID
 KillSignal=SIGQUIT
--- a/packages/cosmos/buildinfo.json
+++ b/packages/cosmos/buildinfo.json
@@ -5,8 +5,8 @@
   ],
   "single_source": {
     "kind": "url",
-    "url": "https://downloads.dcos.io/cosmos/0.2.0-4/cosmos-server-0.2.0-4-one-jar.jar",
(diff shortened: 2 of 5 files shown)


lingmann:
Both `integration-test/deploy-vpc-cli` and `integration-test/upgrade-onprem` are failing with `AssertionError: Number of hosts must be equal to sum of masters and agents plus the bootstrap host`. The failure signature is different though in each.

* integration-test/deploy-vpc-cli shows: `host.num_masters: 1, num_agents: 2, num_public_agents: 1` followed by **2 hosts**
* integration-test/upgrade-onprem shows: `host.num_masters: 3, num_agents: 2, num_public_agents: 1` followed by **6 hosts**

I've reproduced this 2x so far.

There is a possibility this is related to https://jira.mesosphere.com/browse/DCOS-13915 cc: @orsenthil

lingmann:
@mellenburg do you have any insight on the AssertionError above? I think there might be fixes related to this currently on master.

mellenburg:
@lingmann Yes, that is an old issue. That has since been fixed. Specifically, we used to ask AWS to describe the hosts by getting the instances with the cloudformation tag, however, what we discovered is that the tagging process is likely asynchronous to the cloudformation claiming that the the stack is "complete". Thus, what happens is that the number of desired instances is in the stack, but the tags are not, so we only get a subset.

See: https://github.com/mesosphere/dcos/blob/train-159/test_util/aws.py#L352

The solution to this was to do a painful refactor where we use the resource model to extract the instances instead of tags, but then there is the issue that boto does not have a resource for AutoscalingGroups to cover this use case, so then this ugly function had to be invented:
https://github.com/dcos/dcos/blob/master/test_util/aws.py#L154

TL;DR
- this is a known problem that was fixed on master quite a long while ago
- the fix required for this was a pretty heavy refactor
- our test code is tied into dc/os proper so fixing this will be a giant pain as there can be no independent bump
- In an ideal world I would have split out test_util into a separate repo ages ago and now we could just bump it and be done with it.
- keep on retrying :/

lingmann:
Ok, thank you @mellenburg for the detailed explanation! Is there any way we could have a simple patch which just disables these tests for the 1.8.x branch? It seems unlikely that we will be backporting the full fix / refactor, but it would be nice if we could still reliably run some of our tests on the 1.8.x series. WDYT?

spahl:
@mellenburg @lingmann what is the decision on this?

amitaekbote:
Changed the stable installer url to 1.8.8 for upgrade tests, once that passes this should be good to go

amitaekbote:
@mesosphere-mergebot ready for review

amitaekbote:
@spahl this is good to go for 1.8.9

PR #1532, opened 2017-05-10
Admin Router: Make AR route to history-service backend, conditionally

vespian:
This is just a backport of https://github.com/dcos/dcos/pull/1521 into 1.8 Please check https://github.com/dcos/dcos/pull/1521 description for details.

As 1.8 did not have ANY AR testing infra, it's omitted here.

EE PR: https://github.com/mesosphere/dcos-enterprise/pull/891

PR #1453, opened 2017-04-12
Bump networking components for 1.8.9

GoelDeepak:
## High Level Description

This patch backports some of the mesos state parser fixes that many customers have reported.

## Related Issues

  - [DCOS_OSS-927](https://jira.mesosphere.com/browse/DCOS_OSS-927) Backport fix for mesos state parser to 1.8.x

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [x] Test Results: [link to CI job test results for component]
  - [x] Code Coverage (if available): [link to code coverage report]

https://github.com/dcos/minuteman/pull/104
https://github.com/dcos/minuteman/pull/109
https://github.com/dcos/navstar/pull/62

___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and Review Process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using github's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

Reviewers should be:
* Developers who understand the code being modified.
* Developers responsible for code that interacts with or depends on the code being modified.

It is best to proactively ask for 2 reviews by @mentioning the candidate reviewers in the PR comments area. The responsibility is on the developer submitting the PR to follow-up with reviewers and make sure a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

PR #1550, opened 2017-05-16
[1.8.9] Train 162

adam-mesos:
Train 162 for 1.8.9 :steam_locomotive: 
Includes:
#1453 Bump networking components
#1532 Make AdminRouter route to history-service backend, conditionally

### DIFF
--- a/packages/adminrouter/build
+++ b/packages/adminrouter/build
@@ -31,6 +31,11 @@ cp -r /pkg/src/adminrouter/* .
 rm -f win-utf uwsgi* scgi* nginx.conf* mime.types.default* koi-* fastcgi* README* *.png
 popd
 
+# AR start script:
+ar_wrapper="$PKG_PATH/nginx/sbin/adminrouter.sh"
+envsubst '$PKG_PATH' < /pkg/extra/adminrouter.sh > "$ar_wrapper"
+chmod a+x $ar_wrapper
+
 # Copy in needed libraries.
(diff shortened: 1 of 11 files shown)


jgehrcke:
This looks good, but I have do not have enough experience with the failed upgrade test to understand if it is "just" flaky or actually negatively affected by the patch in this PR.

Broken test log (`integration-test/upgrade-onprem`): https://teamcity.mesosphere.io/viewLog.html?buildId=651418&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssUpgradeTestOnPre&tab=buildLog

GoelDeepak:
@jgehrcke The default upgrade test tries to upgrade from stable to PR. Since stable is 1.9.0 the default upgrade test will always fail for 1.8.9. You need to change the parameter in upgrade test to 1.8.8

jgehrcke:
> You need to change the parameter in upgrade test to 1.8.8

How is that done?

amitaekbote:
@jgehrcke for reference - https://github.com/dcos/dcos/pull/1532

I updated the upgrade test with correct params in that pr

jgehrcke:
@GoelDeepak and @nlsun can you please have a look at the new upgrade failure? It seems to be networking component related.

nlsun:
@jgehrcke seems like timeout problems, just re-run them, some tests should just have longer timeouts

PR #1551, opened 2017-05-16
Remove useless known hosts warnings

mellenburg:
## High Level Description
This gets rid of this annoying error message: ```Warning: Permanently added '52.59.62.9' (ECDSA) to the list of known hosts.```
This is worthless because we set the known hosts list to /dev/null

## Checklist for all PR's

area. The responsibility is on the developer submitting the PR to follow-up with reviewers and make sure a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

### DIFF
--- a/ssh/ssher.py
+++ b/ssh/ssher.py
@@ -91,6 +91,7 @@ def open_tunnel(user: str, key: str, host: str, port: int=22) -> Tunnelled:
             '-oControlPath=' + temp_paths[0],
             '-oStrictHostKeyChecking=no',
             '-oUserKnownHostsFile=/dev/null',
+            '-oLogLevel=ERROR',
             '-oBatchMode=yes',
             '-oPasswordAuthentication=no',
             '-p', str(port)]


ssh/ssher.py
```diff
@@ -91,6 +91,7 @@ def open_tunnel(user: str, key: str, host: str, port: int=22) -> Tunnelled:
             '-oControlPath=' + temp_paths[0],
             '-oStrictHostKeyChecking=no',
             '-oUserKnownHostsFile=/dev/null',
+            '-oLogLevel=ERROR',
```
falseperfection: This is just one additional log level above the default, yeah? (Default being INFO.) The fact that it says "Warning" in front of the message doesn't mean we'll be missing messages at the WARNING level, because no such level exists. (I'm going off one single man page review, but wanted to make sure this doesn't hide any actual warning messages that we may care about in the future....)

mellenburg: This is my understanding as well. In general, I think its OK to stray on the side of less logging information here because this client is intended for high-level application only. That being said, we will still get all critical information in the event of a considerable issue

falseperfection: :+1:

PR #1552, opened 2017-05-16
Remove gen from dcos launch

mellenburg:
## High Level Description
gen required many hacks to integrate in at all. The only reason it was used was the convenience of the shared code-base. However, given that the shared code-base has now crippled dcos-launch with its custom-SSL, dcos-launch must be split off.

gen and the rest of the "DC/OS Image" library is a long ways away from being converted into a legitimately importable library, so it must be removed to make the dcos-launch packaging feasible.

## Related Issues
https://jira.mesosphere.com/browse/DCOS_OSS-1060
https://jira.mesosphere.com/browse/DCOS_OSS-1041
https://jira.mesosphere.com/browse/DCOS_OSS-1084

## Checklist for all PR's

### DIFF
--- a/packages/dcos-launch/build
+++ b/packages/dcos-launch/build
@@ -1,15 +1,11 @@
 #!/bin/bash
 source /opt/mesosphere/environment.export
 
-# Copy from the ro /pkg/extra to a rw folder.
 mkdir -p /build
 cp -r /pkg/extra/* /build
-# install the lib for export
+cp -r /pkg/src/cerberus/cerberus /build
 cd /build
-pip3 install --no-deps --install-option="--prefix=$PKG_PATH" /pkg/extra/
-# verify the build
-py.test
-# create the installer
+pytest
 python3 /pkg/src/pyinstaller/pyinstaller.py dcos-launch.spec
 cd dist/
 # sanity check the installer
--- a/packages/dcos-launch/extra/dcos-launch.spec
+++ b/packages/dcos-launch/extra/dcos-launch.spec
@@ -6,10 +6,8 @@
 #     with the final exe. Ensure build system has OpenSSL 1.0.2g or greater
 a = Analysis(['launch/cli.py'],
              hiddenimports=['html.parser'],
-             datas=[('/opt/mesosphere/active/dcos-image/lib/python3.5/site-packages/gen/ip-detect/*.sh',
-                    'gen/ip-detect'),
-                    ('/opt/mesosphere/active/dcos-image/lib/python3.5/site-packages/gen/*.yaml',
-                    'gen/'),
+             datas=[('launch/ip-detect/aws.sh', 'launch/ip-detect'),
+                    ('launch/ip-detect/aws_public.sh', 'launch/ip-detect'),
                     ('/opt/mesosphere/active/dcos-image/lib/python3.5/site-packages/test_util/templates/*.json',
                     'test_util/templates'),
 ])
--- a/packages/dcos-launch/extra/launch/aws.py
+++ b/packages/dcos-launch/extra/launch/aws.py
@@ -1,8 +1,6 @@
 import copy
 import logging
 
-import yaml
-
 import launch.util
 import test_util.aws
 import test_util.runner
@@ -30,7 +28,7 @@ class DcosCloudformationLauncher(launch.util.AbstractLauncher):
         try:
             stack = self.boto_wrapper.create_stack(
                 self.config['deployment_name'],
-                yaml.load(self.config['template_parameters']),
+                self.config['template_parameters'],
                 template_url=self.config.get('template_url'),
                 template_body=self.config.get('template_body'))
         except Exception as ex:
@@ -48,9 +46,9 @@ class DcosCloudformationLauncher(launch.util.AbstractLauncher):
         will be provided (must be done in correct order) and added to the info
         JSON as 'temp_resources'
         """
-        if self.config['zen_helper'] != 'true':
+        if self.config['zen_helper'] is not True:
             return {}
-        parameters = yaml.load(self.config['template_parameters'])
+        parameters = self.config['template_parameters']
         temp_resources = {}
         if 'Vpc' not in parameters:
             vpc_id = self.boto_wrapper.create_vpc_tagged('10.0.0.0/16', self.config['deployment_name'])
@@ -70,7 +68,7 @@ class DcosCloudformationLauncher(launch.util.AbstractLauncher):
                 vpc_id, '10.0.128.0/20', self.config['deployment_name'] + '-public')
             parameters['PublicSubnet'] = public_subnet_id
             temp_resources.update({'public_subnet': public_subnet_id})
-        self.config['template_parameters'] = yaml.dump(parameters)
+        self.config['template_parameters'] = parameters
         return temp_resources
 
     def wait(self):
@@ -107,14 +105,15 @@ class DcosCloudformationLauncher(launch.util.AbstractLauncher):
         as the cloudformation stack, update the config with the resulting private key,
         and amend the cloudformation template parameters to have KeyName set as this key
         """
-        if self.config['key_helper'] != 'true':
+        if self.config['key_helper'] is not True:
             return {}
+        if 'KeyName' in self.config['template_parameters']:
+            raise launch.util.LauncherError('KeyHelperError', 'KeyName cannot be set in '
+                                            'template_parameters when key_helper is true')
         key_name = self.config['deployment_name']
         private_key = self.boto_wrapper.create_key_pair(key_name)
         self.config.update({'ssh_private_key': private_key})
-        template_parameters = yaml.load(self.config['template_parameters'])
-        template_parameters.update({'KeyName': key_name})
-        self.config['template_parameters'] = yaml.dump(template_parameters)
+        self.config['template_parameters'].update({'KeyName': key_name})
         return {'key_name': key_name}
 
     @property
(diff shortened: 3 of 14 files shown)


mellenburg:
When reviewing this PR, specifically this part: https://github.com/dcos/dcos/pull/1552/files#diff-870f931cca2a43e3f3b4881046e29c57

I recommend you just ignore the diff and look at the completed file

packages/dcos-launch/build
```diff
@@ -1,15 +1,11 @@
...
 mkdir -p /build
 cp -r /pkg/extra/* /build
-# install the lib for export
+cp -r /pkg/src/cerberus/cerberus /build
 cd /build
-pip3 install --no-deps --install-option="--prefix=$PKG_PATH" /pkg/extra/
-# verify the build
-py.test
-# create the installer
+pytest
```
colin-msphere: Previous line was `py.test` (with `.`), is this line correct?

packages/dcos-launch/extra/dcos-launch.spec
```diff
@@ -6,10 +6,8 @@
 #     with the final exe. Ensure build system has OpenSSL 1.0.2g or greater
 a = Analysis(['launch/cli.py'],
              hiddenimports=['html.parser'],
-             datas=[('/opt/mesosphere/active/dcos-image/lib/python3.5/site-packages/gen/ip-detect/*.sh',
-                    'gen/ip-detect'),
-                    ('/opt/mesosphere/active/dcos-image/lib/python3.5/site-packages/gen/*.yaml',
-                    'gen/'),
+             datas=[('launch/ip-detect/aws.sh', 'launch/ip-detect'),
+                    ('launch/ip-detect/aws_public.sh', 'launch/ip-detect'),
```
colin-msphere: Not sure on the tuple values here. `launch/ip-detect` is used on the line above as well.

packages/dcos-launch/extra/launch/aws.py
```diff
@@ -131,14 +130,18 @@ class BareClusterLauncher(DcosCloudformationLauncher):
     def create(self):
         """ Amend the config to add a template_body and the appropriate parameters
         """
+        template_parameters = {
+            'AllowAccessFrom': self.config['admin_location'],
+            # cluster size is +1 for the bootstrap node
```
colin-msphere: 👍

packages/dcos-launch/extra/launch/ip-detect/aws_public.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -o nounset -o errexit
+
```
colin-msphere: Should this script also load `/etc/environment` like `aws.sh`?

packages/dcos-launch/build
mellenburg: Yes. pytest installs both entrypoints. There should actually be no testing in this file, but running in line with the unit tests is not currently possible given the additional dependencies

packages/dcos-launch/extra/dcos-launch.spec
mellenburg: First item in tuple is local file path. Second item is `pkg_resources` path

packages/dcos-launch/extra/launch/ip-detect/aws_public.sh
mellenburg: Yeah, this was copy/paste but we should take this moment to add some consistency

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
     return os.path.abspath(os.path.join(relative_dir, path))
```
orsenthil: Not your change. But  expand_path's scope is within LaunchValidator. This can be moved inside that class or it could be made a _private function.

mellenburg: I have a unit test for just this function so it made sense to keep it out. It would probably be a better test if I just instantiated that Validator class and made this private

orsenthil: Yeah, I noticed the test later.
And Yes, I agree, that it will be better to test the Validator class.

orsenthil: Also, it's interesting to see much simpler (and straight-forward) config.py with cerebrus.

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
...
+    def _normalize_coerce_expand_local_path(self, value):
+        return expand_path(value, self.config_dir)
 
-def calculate_ssh_user(os_name, platform):
-    if platform == 'aws':
-        return test_util.aws.OS_SSH_INFO[os_name].user
-    else:
-        raise Exception('Cannot yet calculate user for {} platform'.format(platform))
 
+def expand_error_dict(errors: dict) -> str:
```
orsenthil: this could be made private.

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
...
-    if key_helper == 'false':
-        return
-    if provider == 'aws':
-        assert 'KeyName' not in yaml.load(template_parameters), 'key_helper will '\
-            'automatically calculate and inject KeyName; do not set this parameter'
-    if provider == 'azure':
-        assert 'sshRSAPublicKey' not in yaml.load(template_parameters), 'key_helper will '\
-            'automatically calculate and inject sshRSAPublicKey; do not set this parameter'
 
+def raise_errors(validator: LaunchValidator):
```
orsenthil: can be made private.

packages/dcos-launch/extra/launch/util.py
```diff
@@ -24,6 +24,11 @@
 NO_TEST_FLAG = 'NO PRIVATE SSH KEY PROVIDED - CANNOT TEST'
 
 
+def load_string(filename: str):
```
orsenthil: `read_file` instead of `load_string` ?

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
...
+        raise NotImplementedError()
+
+    # create a strict validator with our final schema and process it
+    final_validator = LaunchValidator(basic_validator.schema, config_dir=config_dir, allow_unknown=False)
+    if not final_validator.validate(config):
+        raise_errors(final_validator)
+    return final_validator.normalized(config)
+
+
+COMMON_SCHEMA = {
```
colin-msphere: Would it be beneficial to pull the `_SCHEMA` dictionaries to another file? This would provide separation between functionality (`config.py`) and rules (schema definitions).

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
...
+    """ Returns validated a finalized argument dictionary for dcos-launch
+    Given the huge range of configuration space provided by this configuration
+    file, it must be processed in three steps (common, provider-specifc,
+    platform-specific)
+    """
+    config = load_config(config_path)
+    config_dir = os.path.dirname(config_path)
+    # validate against the fields common to all configs
+    basic_validator = LaunchValidator(COMMON_SCHEMA, config_dir=config_dir, allow_unknown=True)
+    if not basic_validator.validate(config):
```
colin-msphere: These two lines (`if not validate / raise_errors`) are repeated. Could the `validate(config)` method take a function as an optional second parameter (`validate(config, raise_errors)`) and call it on validation failures?

packages/dcos-launch/extra/launch/config.py
```diff
@@ -49,375 +21,256 @@ def expand_path(path: str, relative_dir: str) -> str:
...
+    """ Needs to use unintuitive pattern so that child validator can be created
+    for validated the nested dcos_config
+    """
+    def __init__(self, *args, **kwargs):
+        super(LaunchValidator, self).__init__(*args, **kwargs)
+        if 'config_dir' in kwargs:
+            self.config_dir = kwargs['config_dir']
 
+    def _normalize_coerce_expand_local_path(self, value):
+        return expand_path(value, self.config_dir)
```
mellenburg: The only way that would happen is if someone made a programming error. Config path (or alternatively, the info path) must always exist for any dcos-launch command to work. Thus, a committer would have to add logic that could allow the config or dir to be None. I could add an assertion to the class instantiator to make that clear

packages/dcos-launch/extra/launch/config.py
mellenburg: I liked keeping the code this way because basic_validator.validate() is a state-mutating function and so I wanted to keep all the mutation at one level. Also, I really dislike functions that do too much and to put this into a single function would be something like:
```def validate_and_raise_if_errors(validator: LaunchValidator, config:dict) -> None```
which I thought just confused thing more.
Thoughts?

File: packages/dcos-launch/extra/launch/config.py (hunk 9 above)
Comment (mellenburg) [inline packages/dcos-launch/extra/launch/config.py]: If we tried pulling this out, then we would also have to pull out the custom validator so that the custom schema functions are clearly defined with their schema and we would also have to pull in the validation code. At that point, the config module would then just be the single function that does the three-stage validation process. At which point we would probably just want that function to live with the associated schemas.

So, if this schema was a little more statically defined and didnt require function references and object declaration I would find this beneficial.

File: packages/dcos-launch/extra/launch/config.py (hunk 10 above)
Comment (colin-msphere) [inline packages/dcos-launch/extra/launch/config.py]: Good points!

File: packages/dcos-launch/extra/launch/config.py (hunk 9 above)
Comment (colin-msphere) [inline packages/dcos-launch/extra/launch/config.py]: Thanks.

**Comment by mellenburg** [comment]
@orsenthil Comments addressed (and I even conformed to the python style guidelines I am not fond of)

File: packages/dcos-launch/extra/launch/ip-detect/aws_public.sh [hunk 12]
```diff
@@ -1,4 +1,11 @@
 #!/bin/sh
 set -o nounset -o errexit
 
+if [ -e /etc/environment ]
+then
+  set -o allexport
```
orsenthil: I had to look around to know the meaning of this `bash` set options. It will be great it a comment was provided for the reader. Enable exporting from file; export properties from file, and then disable exporting from file).

mellenburg:
No bump-ee required for this PR as the launch component isnt even built for that variant. :ship:

PR #1557, opened 2017-05-17
Increase timeout for apps to be healthy

mellenburg:
## High Level Overveiew
apps usually take less than 120 seconds to deploy
and achieve a healthy status, but in test_vip there
is no blocking on the previous test cleaning up state.
This speeds up execution to allow marathon to optimize
the deployments, but also adds a small increase to the
time any given app spends deploying

## Related Issues
Timeouts occurring on random test cases such as https://teamcity.mesosphere.io/viewLog.html?buildId=653598&

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: N/A This PR is stabilizing a test

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -3,6 +3,8 @@ import uuid
 
 import pytest
 
+from test_helpers import expanded_config
+
 from test_util.marathon import Container, get_test_app, Healthcheck, Network
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


mellenburg:
@mesosphere-mergebot  bump-ee

adamtheturtle:
Not a blocker but a preference (this isn't a full review) - Usually I like to see timeouts like this annotated with the reasoning which led to them nearby as a comment. That might be a link to a JIRA or a note about previous experience, or a note that a choice is arbitrary.

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@BenWhitehead @asridharan After you approved this PR, I did the EE testing on it and there was still a failure and so I made some changes:
(1) I added another minute to the timeout
(2) added a fixture to ensure deployments are complete before and after all permutations of test_vip have run I.E. if other tests deployed a flapping app that is stuck and hoarding resources before this test starts, then these cases will fail because this test assumes a clean cluster to pass. Also, its a good idea to wait for these apps to be destroyed so that the next test doesn't suffer from this issue
(3) The aforementioned fixture caused a failure in strict mode because the test_octarine case fails and manages to leave around a bunch of flapping apps. This test is currently muted for strict and has no hope of passing in EE strict. Thus, I skip this test when security is strict.

PR #1561, opened 2017-05-18
dcos-test-utils package

mellenburg:
## High Level Description
This PR removes the test toolchain from this repo and places it in its own repo. This was done because the test toolchain is reverse compatible and is in no way linked to a particular build.

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Same tests; different place


mellenburg:
@spahl @lingmann @orsenthil PTAL. I want to remove this code before people try to commit changes here.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -3,10 +3,9 @@
 
 import pytest
 
+from dcos_test_utils.marathon import Container, get_test_app, Healthcheck, Network
```
orsenthil: [minor] - sorted imports in the from import. Does our tox (isort) allow this?

mellenburg:
@lingmann Sure can do! run-all is used by the vagrant-bash integration test, so I don't want to touch that as its been a consistent entrypoint since forever.

Once this was finally merged, I was planning on checking in a few simple launch configs or scripts that would closely resemble what it is that we are using in CI (but now we have the flexibility to dynamically override that config at any time)

Would you prefer that I add that with the README or rather that I do that later when there are less things to consider?

lingmann:
Ok cool. I think all we need with this PR is a README in test_utils telling people where the new repo lives. The rest of the details can come later. :)

mellenburg:
Last push was docs only change, so no need to repoint EE

mellenburg:
@lingmann Are you OK with the README added here?

lingmann:
@mesosphere-mergebot bump-ee

jgehrcke:
Is there a JIRA ticket that needs to be updated now that this is merged?

PR #1528, opened 2017-05-08
Add HTTP basic auth for Exhibitor

branden:
## High Level Description

This adds an optional config param `exhibitor_admin_password`. If provided, requests to Exhibitor must be authenticated with the userid `admin` and the password.

## Related Issues

  - [DCOS_OSS-1064](https://jira.mesosphere.com/browse/DCOS_OSS-1064) Add config param for Exhibitor basic HTTP auth 

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: CLI installer test for this PR was run with `exhibitor_admin_password: "deleteme"`: https://teamcity.mesosphere.io/viewLog.html?buildId=650776&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_DcosOssCcmIntegration_2. After this is merged, I'll update the job to use that config param with every test run.
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]

PR #1568, opened 2017-05-19
Fix broken reference to public_master

mellenburg:
## High Level Description
#1508 and #1528 conflicted but this case silently slipped by because it was a new usage of the variable being removed. This was not caught during testing as the code was not exercised until a specific setting was enabled in CI

## Related Issues
- https://jira.mesosphere.com/browse/DCOS_OSS-1125

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: N/A Test Fix


orsenthil:
@mesosphere-mergebot bump-ee

PR #1595, opened 2017-05-29
Admin Router: Fix /metadata endpoint plus add regression tests

vespian:
## High Level Description

This PR:
* fixes /metadata endpoint, adds regression tests

## Related Issues
* https://jira.mesosphere.com/browse/DCOS_OSS-1150 - SSH access to DCOS Open broken


vespian:

packages/adminrouter/docker/detect_ip_public
```diff
@@ -0,0 +1,73 @@
...
+    if control_string.startswith("break with "):
+        try:
+            exit_status = int(control_string.split()[2])
+        except (IndexError, ValueError):
+            msg_fmt = "Cannot determine exit status basing on `{}` string"
+            logger.error(msg_fmt.format(control_string))
+            return 1
+
+        msg = "Exiting with `{}` status".format(exit_status)
+        print(msg)
```
mhrabovcin: Why do we need to print here?

vespian: I wanted it to simulate script breaking in some realistic way (error message plus exit status), plus it may make debugging easier (e.g. if AR tries to parse output despite non-zero exit status it will break much more verbosely).

packages/adminrouter/docker/detect_ip_public
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
```
jgehrcke: The error handling in this script is over-complex, assuming that there is a way to access `stderr`. If AR does not expose `stderr` of this script then I think we should add this (because this is important in real-world scenarios).

A few things to think about:

* `DATAFILE` is a constant (in fact as well as by convention). Why check for `None`?

* This script defines an interface with a single caller: the test harness. If the test harness uses this interface in a wrong way, then it is OK to just let this script blow up (Python has built-in error handling and emits complete and beautiful error information to stderr in this case, and also sets the exit code appropriately), look at the traceback, and fix the test harness. I don't see value added by catching `IOError`, `IndexError`, `ValueError` and emitting error message tailored for humans. 

* Details about a failure while executing the IP detection script(s) in the _real world_ will most likely be emitted to stderr. Why go through syslog here? I think emitting error details to stderr enables you to test if error details will actually be visible in the Admin Router log (which can be highly useful).

In summary, I think you should remove all exception handlers as well as the `DATAFILE` check for `None` and then make sure that AR actually emits `stderr` of this script.

vespian: Yep, you are right. 

Done, please have another look!

PR #1604, opened 2017-06-01
Explicitly require six

mellenburg:
## High Level Description
Something changed with one of: setuptools wheel or virtualenv and the result was that six needs to be explicitly required for the library tools to work

## Checklist for all PR's


mellenburg:
And great. Now this magically fixed it self. Waste of time I guess.

PR #1616, opened 2017-06-03
Fixed bug in 'test_mesos.test_if_ucr_app_runs_in_new_pid_namespace'.

klueska:
Previously, there was a race between the `app_task` starting and it
writing its output to the `ps_output_file`. We now wait up to 10 seconds
for this file to appear before giving up and throwing an exception.

## High Level Description

## Related Issues

  - [DCOS_OSS-862](https://jira.mesosphere.com/browse/DCOS_OSS-862) Vagrant integration test test_mesos.test_if_ucr_app_runs_in_new_pid_namespace is flaky.

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a fix to a test itself

___

### DIFF
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -2,6 +2,7 @@ import json
 import logging
 import uuid
 
+import retrying
 
 from test_util.marathon import get_test_app, get_test_app_in_ucr
 from test_util.recordio import Decoder, Encoder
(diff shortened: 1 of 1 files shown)


adam-mesos:

karlkfi:
I'm having deja vu... how is this PR different than https://github.com/dcos/dcos/pull/1617 ?

karlkfi:
Shouldn't we wait for the one on master then and then backport it? Why review it twice?

adam-mesos:
Ideally, yes, we would land it in master first, then backport it.
But a) sometimes the patch release has more urgency, so we end up landing
it there first.
b) any cherry-pick conflicts when backporting would require a nominal
review to make sure the backporting didn't screw anything up.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -157,7 +158,12 @@ def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
         marathon_framework_id = dcos_api_session.marathon.get('/v2/info').json()['frameworkId']
         app_task = dcos_api_session.marathon.get('/v2/apps/{}/tasks'.format(app['id'])).json()['tasks'][0]
 
-        content = dcos_api_session.mesos_sandbox_file(
-            app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
+        # There is a short delay between the `app_task` starting and it writing
+        # its output to the `pd_output_file`. Because of this, we wait up to 10
+        # seconds for this file to appear before throwing an exception.
+        @retrying.retry(wait_fixed=0.1, stop_max_delay=10000, retry_on_exception=lambda x: True)
```
mellenburg: wait fixed is set to .1 milliseconds? The function being called here will do an HTTP GET, so I would suggest something like 1000ms or 2000ms.

Also, the default is to retry on all exceptions so you can just completely omit `retry_on_exception=lambda x: True`

klueska: Yikes, that's a major oversight. Thanks for catching it! My intention was to wait for 100ms or 0.1s. The file should appear almost immediately, so I only expect to have to wait between 0 and 100ms for it to show up (i.e. only error out once), but I can set it at 1s if you think that's preferable.

PR #1617, opened 2017-06-03
Fixed bug in 'test_mesos.test_if_ucr_app_runs_in_new_pid_namespace'.

klueska:
Previously, there was a race between the `app_task` starting and it
writing its output to the `ps_output_file`. We now wait up to 10 seconds
for this file to appear before giving up and throwing an exception.

## High Level Description

## Related Issues

  - [DCOS_OSS-862](https://jira.mesosphere.com/browse/DCOS_OSS-862) Vagrant integration test test_mesos.test_if_ucr_app_runs_in_new_pid_namespace is flaky.

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a fix to a test itself

___

### DIFF
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -2,6 +2,7 @@ import json
 import logging
 import uuid
 
+import retrying
 
 from test_util.marathon import Container, get_test_app
 from test_util.recordio import Decoder, Encoder
@@ -157,7 +158,12 @@ def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
         marathon_framework_id = dcos_api_session.marathon.get('/v2/info').json()['frameworkId']
         app_task = dcos_api_session.marathon.get('/v2/apps/{}/tasks'.format(app['id'])).json()['tasks'][0]
 
-        content = dcos_api_session.mesos_sandbox_file(
-            app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
+        # There is a short delay between the `app_task` starting and it writing
+        # its output to the `pd_output_file`. Because of this, we wait up to 10
+        # seconds for this file to appear before throwing an exception.
+        @retrying.retry(wait_fixed=1000, stop_max_delay=10000)
+        def get_ps_output():
+            return dcos_api_session.mesos_sandbox_file(
+                app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
 
-        assert len(content.split()) <= 4, 'UCR app has more than 4 processes running in its pid namespace'
+        assert len(get_ps_output().split()) <= 4, 'UCR app has more than 4 processes running in its pid namespace'


klueska:

klueska:
@mellenburg @karlkfi Could you please review this PR for me?

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -157,7 +158,11 @@ def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
         marathon_framework_id = dcos_api_session.marathon.get('/v2/info').json()['frameworkId']
         app_task = dcos_api_session.marathon.get('/v2/apps/{}/tasks'.format(app['id'])).json()['tasks'][0]
 
-        content = dcos_api_session.mesos_sandbox_file(
-            app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
+        # There is a race between the `app_task` starting and it writing its output to the `ps_output_file`.
```
karlkfi: This comment doesn't really seem accurate. There's no race condition. It's just that:
1. task health just means the process it running. It's more of a "health check" than a "readiness check". If you were testing a readiness check you could make sure the file was written before the app responded as ready.
2. the task has to write to disk _after_ it starts running. So expecting it to be immediately available is a poor assumption.
3. in dcos-vagrant the host resources are often more oversubscribed and slower than on an AWS cluster.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -157,7 +158,11 @@ def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
...
+        # We wait up to 10 seconds for this file to appear before giving up and throwing an exception.
+        @retrying.retry(wait_fixed=0.1, stop_max_delay=10000, retry_on_exception=lambda x: True)
```
karlkfi: You could avoid retrying and polling if you had the task notify the test via network call. Ex: curl from the task to a server fixture. This might make the test more responsive.

packages/dcos-integration-test/extra/test_mesos.py
klueska: Yeah, I guess `race` was a poor choice of words. The underlying problem is that the current test assumes that the file is there immediately after the app reports that it is alive. This is a bad assumption because the app needs time to write the file before the test can read it.

packages/dcos-integration-test/extra/test_mesos.py
klueska: Let me see what I can come up with.

packages/dcos-integration-test/extra/test_mesos.py
karlkfi: maybe just update the comment for more clarity for our future selves.

klueska: Updated

klueska:
I agree, and have filed the following issue to address it:
https://jira.mesosphere.com/browse/DCOS_OSS-1214

klueska:
@mesosphere-mergebot milestone "1.10"

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -157,7 +158,12 @@ def test_if_ucr_app_runs_in_new_pid_namespace(dcos_api_session):
         marathon_framework_id = dcos_api_session.marathon.get('/v2/info').json()['frameworkId']
         app_task = dcos_api_session.marathon.get('/v2/apps/{}/tasks'.format(app['id'])).json()['tasks'][0]
 
-        content = dcos_api_session.mesos_sandbox_file(
-            app_task['slaveId'], marathon_framework_id, app_task['id'], ps_output_file)
+        # There is a short delay between the `app_task` starting and it writing
+        # its output to the `pd_output_file`. Because of this, we wait up to 10
+        # seconds for this file to appear before throwing an exception.
+        @retrying.retry(wait_fixed=0.1, stop_max_delay=10000, retry_on_exception=lambda x: True)
```
mellenburg: wait fixed is set to .1 milliseconds? The function being called here will do an HTTP GET, so I would suggest something like 1000ms or 2000ms.

Also, the default is to retry on all exceptions so you can just completely omit retry_on_exception=lambda x: True

klueska: Yikes, that's a major oversight. Thanks for catching it! My intention was to wait for 100ms or 0.1s. The file should appear almost immediately, so I only expect to have to wait between 0 and 100ms for it to show up (i.e. only error out once), but I can set it at 1s if you think that's preferable. I guess it depends on file system / network latencies.

jgehrcke:
It looks like we better understand what happened in https://teamcity.mesosphere.io/viewLog.html?buildId=676782. CC @vespian (can you maybe create a corresponding JIRA ticket?)

adam-mesos:
@mesosphere-mergebot bump-ee

vespian:
@jgehrcke RE: https://github.com/dcos/dcos/pull/1617#issuecomment-307091513

https://jira.mesosphere.com/browse/DCOS_OSS-1228

PR #1627, opened 2017-06-06
Increase hacky default timeout

mellenburg:
## High Level Description
This timeout is set to 10000 seconds when using the web installer
but in the CLI, this value gets used and deployments fail due to
unnecessarily strict timeouts. 10 minutes is much more reasonable
base timeout for the actions being performed here

## Checklist for all PR's

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -17,7 +17,7 @@ log = logging.getLogger(__name__)
 
 def get_async_runner(config, hosts, async_delegate=None):
     # TODO(cmaloney): Delete these repeats. Use gen / expanded configuration to get all the values.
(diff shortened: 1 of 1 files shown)


mellenburg:
@branden @lingmann PTAL small, impactful improvement as I'm sure plenty of people do not experience everything working in less than 2 minutes

branden:
The current default timeout used to work (I know firsthand through lots of manual testing of past releases), at some point it stopped working, and we missed that point because the tests were configured to use something different. I'm not comfortable with increasing the default timeout without an explanation for why it stopped working.

See https://jira.mesosphere.com/browse/DCOS-11853

mellenburg:
:+1:  That tickets tracks what I am concerned with so I will just close this out

PR #1641, opened 2017-06-08
Increase timeout in test_vip

mellenburg:
## High Level Description
test_networking tests have been failing by timing out while waiting for apps to remain healthy. So far, this has not been indicative of any actual failure

## Related Issues
https://jira.mesosphere.com/browse/DCOS-14108
https://jira.mesosphere.com/browse/DCOS-15989

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: N/A test improvement
a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

### DIFF
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -109,11 +109,10 @@ def test_vip(dcos_api_session, container: Container, vip_net: Network, proxy_net
     failure_stack = []
     for test in setup_vip_workload_tests(dcos_api_session, container, vip_net, proxy_net):
         cmd, origin_app, proxy_app, proxy_net = test
-        log.info('ORIGIN APP INFO')
-        log.info(wait_for_tasks_healthy(dcos_api_session, origin_app))
-        log.info('PROXY APP INFO')
+        log.info('Deploying origin app')
+        wait_for_tasks_healthy(dcos_api_session, origin_app)
+        log.info('Origin app successful, deploying proxy app..')
         proxy_info = wait_for_tasks_healthy(dcos_api_session, proxy_app)
-        log.info(proxy_info)
         proxy_task_info = proxy_info['app']['tasks'][0]
         if proxy_net == Network.USER:
             proxy_host = proxy_task_info['ipAddresses'][0]['ipAddress']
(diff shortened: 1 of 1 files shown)


mellenburg:
Woah, all the tests except upgrade (which is at a clossal 50% flakiness) passed. Thats a good sign...

mellenburg:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -109,11 +109,10 @@ def test_vip(dcos_api_session, container: Container, vip_net: Network, proxy_net
     failure_stack = []
     for test in setup_vip_workload_tests(dcos_api_session, container, vip_net, proxy_net):
         cmd, origin_app, proxy_app, proxy_net = test
-        log.info('ORIGIN APP INFO')
-        log.info(wait_for_tasks_healthy(dcos_api_session, origin_app))
-        log.info('PROXY APP INFO')
+        log.info('Deploying origin app')
+        wait_for_tasks_healthy(dcos_api_session, origin_app)
```
adam-mesos: Do you want to retrieve+log the result of `wait_for_tasks_healthy` like we used to?

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -109,11 +109,10 @@ def test_vip(dcos_api_session, container: Container, vip_net: Network, proxy_net
...
+        log.info('Origin app successful, deploying proxy app..')
         proxy_info = wait_for_tasks_healthy(dcos_api_session, proxy_app)
-        log.info(proxy_info)
```
adam-mesos: No longer worth logging? Maybe debug-level instead of info?

packages/dcos-integration-test/extra/test_networking.py
mellenburg: I actually didn't intend to leave these statements in the first time I wrote this. They provide a large, hard-to-parse output and they only output that information when things are working smoothly. Thus, they were not even useful for debugging when I was actively working on it. In general, I think we should avoid logging output unless we have some idea of why we want to see it at that time. Or else we will be back where we were a few months ago where we logged the complete marathon app status every 2 seconds on all deploys, leading to logs so large that my browser would crash trying to view them.

PR #1645, opened 2017-06-08
Admin Router: Improve /service endpoint tests

vespian:
## High Level Description

This PR:
* makes dns mock a dependency for mocker. Without it the tests can
stall for >60s, depending on the order of fixtures in the function signature.
* adds missing tests for `/service` endpoint:
  * verify that resolving a service waits for Marathon if the cache is empty
  * verify that resolving a service waits for Mesos if the cache is empty
  * verify that resolving a service returns 503 in case when Marathon was borked AND make sure that Mesos data is ignored in this case
  * verify that resolving a service returns 503 in case when Mesos was borked AND resolving via Marathon did not succed
  * verify that resolving a service returns 200 in case when Mesos was borked AND Marathon is OK

## Related Issues

  - https://jira.mesosphere.com/browse/DCOS-16108 `Inspect integration test failures as of Admin Router's service endpoint emitting a 404 response`

### DIFF
--- a/packages/adminrouter/extra/src/test-harness/tests/conftest.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/conftest.py
@@ -32,6 +32,8 @@ def repo_is_ee():
     return generic_test_code.common.repo_is_ee()
 
 
+# We explicitly need dns_server_mock_s fixture here as Mock HTTP servers
+# require DNS to resolve their server_names.
 @pytest.fixture(scope='session')
 def mocker_s(repo_is_ee, syslog_mock, extra_lo_ips, dns_server_mock_s):
     """Provide a gc-ed mocker instance suitable for the repository flavour"""
--- a/packages/adminrouter/extra/src/test-harness/tests/conftest.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/conftest.py
@@ -33,7 +33,7 @@ def repo_is_ee():
 
 
 @pytest.fixture(scope='session')
-def mocker_s(repo_is_ee, syslog_mock, extra_lo_ips):
+def mocker_s(repo_is_ee, syslog_mock, extra_lo_ips, dns_server_mock_s):
     """Provide a gc-ed mocker instance suitable for the repository flavour"""
     if repo_is_ee:
         from mocker.ee import Mocker
--- a/packages/adminrouter/extra/src/test-harness/tests/test_cache.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/test_cache.py
@@ -809,7 +809,8 @@ class TestCacheMarathon:
             resp = requests.get(url,
                                 allow_redirects=False,
                                 headers=valid_user_header)
-            assert "cache state is invalid" in resp.content.decode('utf-8')
+            expected = "503 Service Unavailable: invalid Marathon svcapps cache"
+            assert expected == resp.content.decode('utf-8').strip()
             assert resp.status_code == 503
 
             lbf.scan_log_buffer()
--- a/packages/adminrouter/extra/src/test-harness/tests/test_service.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/test_service.py
@@ -1,3 +1,4 @@
+
 # Copyright (C) Mesosphere, Inc. See LICENSE file for details.
 
 import time
@@ -651,7 +652,7 @@ class TestServiceStateful:
             t_spent = time.time() - t_start
             assert resp.status_code == 200
             data = resp.json()
-            assert data['endpoint_id'] == 'http://127.0.0.1:16000'
+            data['endpoint_id'] == 'http://127.0.0.15:16001'
 
         assert t_spent > backend_request_timeout * 0.5
 
--- a/packages/adminrouter/extra/src/test-harness/tests/test_service.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/test_service.py
@@ -23,6 +23,7 @@ from mocker.endpoints.mesos_dns import (
     EMPTY_SRV,
     SCHEDULER_SRV_ALWAYSTHERE_DIFFERENTPORT,
 )
+from util import GuardedSubprocess
 
 
 class TestServiceStateful:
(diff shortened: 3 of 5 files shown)


vespian:

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -809,7 +809,7 @@ def test_upstream_wrong_json(
             resp = requests.get(url,
                                 allow_redirects=False,
                                 headers=valid_user_header)
-            assert "cache state is invalid" in resp.content.decode('utf-8')
+            assert "invalid Marathon svcapps cache" in resp.content.decode('utf-8')
```
adamtheturtle: ❓ Optional because not a regression but ideally check for the whole message.

packages/adminrouter/extra/src/test-harness/tests/conftest.py
```diff
@@ -33,7 +33,7 @@ def repo_is_ee():
 
 
 @pytest.fixture(scope='session')
-def mocker_s(repo_is_ee, syslog_mock, extra_lo_ips):
+def mocker_s(repo_is_ee, syslog_mock, extra_lo_ips, dns_server_mock_s):
```
adamtheturtle: Comment that you need `dns_server_mock_s` explicitly because else the order of fixture loading matters.

❓ If we used a linter like `pylint` this would warn that `dns_server_mock_s` is an unused parameter. See https://docs.pytest.org/en/latest/fixture.html#using-fixtures-from-classes-modules-or-projects for a way to say "this is needed but not accessed in the test".

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+            data = resp.json()
+            assert data['endpoint_id'] == 'http://127.0.0.1:16000'
+
+            # If AR wait for cache during resolve, then time spent should be
+            # greater than the stall time that has been set. Due to the fact
+            # that update coroutines are not separated yet, this will be
+            # slightly higher than: 2 * (backend_request_timeout * 0.5)
+            assert t_spent > backend_request_timeout * 0.5
+
+    def test_if_marathon_borkerage_prevents_resolving_via_mesos_state_summary(
```
adamtheturtle: ❓ Borkerage is not a word. Perhaps a better term would be `test_if_broken_marathon...`. Optional partly because it is not new to this codebase.

❓ Personally I prefer to organise tests into classes, and would maybe go for `TestEmptyCache.test_waits_for_marathon_during_service_resolve` and `TestBrokenMarathon.test_resolving_via...`. Probably not actionable now but I thought I may say it.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+
+            # If AR wait for cache during resolve, then time spent should be
+            # greater than the stall time that has been set. Due to the fact
+            # that update coroutines are not separated yet, this will be
+            # slightly higher than: 2 * (backend_request_timeout * 0.5)
+            assert t_spent > backend_request_timeout * 0.5
+
+    def test_if_ar_with_empty_cache_waits_for_mesos_during_service_resolve(
+            self, mocker, nginx_class, valid_user_header):
+        # Make service endpoint resolve only Marathon-related data:
```
adamtheturtle: `s/Marathon/Mesos` I think.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+        mocker.send_command(
+            endpoint_id='http://127.0.0.1:8080',
```
adamtheturtle: Looking at `docs/admin-ports.md` is this not saying that Marathon will return data but Mesos will return `EMPTY_SRV`?

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+            aux_data=EMPTY_SRV)
+
+        # Make Mock endpoint stall a little, make sure AR cache update timeouts
+        # are big enough to swallow it:
+        backend_request_timeout = 6
+        refresh_lock_timeout = backend_request_timeout * 2
+
+        # Make period cache refreshes so rare that they do not get into
+        # picture:
+        ar = nginx_class(cache_first_poll_delay=1200,
```
adamtheturtle: How did you come up with these numbers? If it is "random above threshold" then put the threshold. To confirm that the cache refresh doesn't matter, we need to confirm that that threshold is greater than the time difference checked later, right?

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+            func_name='set_apps_response',
+            aux_data={"apps": []})
+        mocker.send_command(
+            endpoint_id='http://127.0.0.1:8123',
+            func_name='set_srv_response',
+            aux_data=EMPTY_SRV)
+
+        # Make Mock endpoint stall a little, make sure AR cache update timeouts
+        # are big enough to swallow it:
+        backend_request_timeout = 6
```
adamtheturtle: ❓ A bit unfortunate that there aren't units here

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -23,6 +23,7 @@
     EMPTY_SRV,
     SCHEDULER_SRV_ALWAYSTHERE_DIFFERENTPORT,
 )
+from util import GuardedSubprocess
```
adamtheturtle: ❓ Nit - I noticed while looking at this that there is a typo in the docstring `s/Subproces/Subprocess`.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+
+        assert resp.status_code == 503
+        assert '503 Service Unavailable: invalid Mesos state cache' in resp.text
+
+    def test_if_mesos_borkerage_does_not_prevent_resolving_via_marathon(
+            self, mocker, nginx_class, valid_user_header):
+        # Bork Mesos Mock, Make MesosDNS mock respond with no apps, so that AR
+        # is able to resolve only via Marathon/we are certain that it resolved
+        # via Marathon.
+        mocker.send_command(
```
adamtheturtle: ❓ Ideally as a reader there would be variables for the ports, or `pytest` fixtures - else I am referring to https://github.com/dcos/dcos/blob/017dcb5fb1f2fd85b97f10dfa837b2801b1290e8/docs/admin-ports.md.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: It is described in the pydoc of the nginx class: https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/test-harness/modules/runner/common.py#L701-L719

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+            endpoint_id='http://127.0.0.2:5050',
+            func_name='always_bork',
+            aux_data=True)
+        mocker.send_command(
+            endpoint_id='http://127.0.0.1:8123',
+            func_name='set_srv_response',
+            aux_data=EMPTY_SRV)
+
+        # Make period cache refreshes so rare that they do not get into
```
adamtheturtle: What could a cache refresh do here? Is it just one of these variables necessary to delay the first cache refresh?

❓ Perhaps move this out to a helper or fixture which is `ar_rare_refresh` or similar - and add a detailed description there.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: 1200s = 20minutes, we just assume that the test run takes much less than that.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Fixed.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Not sure I follow.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: > What could a cache refresh do here?
We want to be sure that the cache update happened only because of the request. This makes test more robust and repeatable. 

> Is it just one of these variables necessary to delay the first cache refresh?
All of them. I have described the inner workings of these variables in README.md.

> Perhaps move this out to a helper or fixture which is ar_rare_refresh or similar - and add a detailed description there.

It makes no sense here as we need raw AR instance, with a clear cache. I also try to avoid having dozens of fixtures that are used rarely - people will forget about them or misuse them.  I think it is better to have a generic one which covers all the use cases.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
adamtheturtle: In the below code when reading ` endpoint_id='http://127.0.0.1:8123',`, to confirm that this is the Mesos DNS port, I check that reference. If there were a variable or fixture e.g. `mesos_dns_endpoint` or `mesos_dns_port` that would help me personally to read this.

packages/adminrouter/extra/src/test-harness/tests/conftest.py
vespian: Comment added

> If we used a linter like pylint this would warn that dns_server_mock_s is an unused parameter.
We do not use pylint, this is a bigger effort to enforce it, and without it it will just bitrot (try grepping the source for pylint, I already played a bit with it here).

packages/adminrouter/extra/src/test-harness/tests/test_service.py
adamtheturtle: As you wish - there is a trade-off of DRY vs a mess of fixtures and my thinking from reading this code was that DRY could provide some benefit here. However, as you are more familiar with the codebase I will go with your judgement.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+        url = ar.make_url_from_path("/service/scheduler-alwaysthere/")
+
+        with GuardedSubprocess(ar):
+            resp = requests.get(
+                url,
+                allow_redirects=False,
+                headers=valid_user_header)
+
+        assert resp.status_code == 503
+        assert '503 Service Unavailable: invalid Mesos state cache' in resp.text
```
adamtheturtle: ❓ If you can change `in` to `==` and it'd still work, please do that. Else leave it.

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
vespian: fixed.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Yep, fixing.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Both will return empty results.

vespian: Empty result for Marathon in `{"apps": []})`, for MesosDNS: `EMPTY_SRV`

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+                                headers=valid_user_header)
+
+            t_spent = time.time() - t_start
+            assert resp.status_code == 200
+            data = resp.json()
+            assert data['endpoint_id'] == 'http://127.0.0.1:16000'
+
+            # If AR wait for cache during resolve, then time spent should be
+            # greater than the stall time that has been set. Due to the fact
+            # that update coroutines are not separated yet, this will be
```
adamtheturtle: I don't know what update coroutines are. Would it be correct to remove the comment about this and change the assertion to `>=`?

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+
+        with GuardedSubprocess(ar):
+            resp = requests.get(
+                url,
+                allow_redirects=False,
+                headers=valid_user_header)
+
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data['endpoint_id'] == 'http://127.0.0.1:16000'
```
adamtheturtle: ❓ As a reader, understanding that this corresponds to `/scheduler-alwaysthere` would be helpful.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+            t_spent = time.time() - t_start
+            assert resp.status_code == 200
+            data = resp.json()
+            assert data['endpoint_id'] == 'http://127.0.0.1:16000'
+
+            # If AR wait for cache during resolve, then time spent should be
+            # greater than the stall time that has been set. Due to the fact
+            # that update coroutines are not separated yet, this will be
+            # slightly higher than: 2 * (backend_request_timeout * 0.5)
+            assert t_spent > backend_request_timeout * 0.5
```
adamtheturtle: I can't quite connect the comment and the code. The comment says that because of X, the time spent will be higher than `2 * (backend_request_timeout * 0.5)`, but the assertion doesn't include the `2 *`. Please update the code or the comment.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: > I don't know what update coroutines are.
NGINX lightweight threads, I had documentation somewhere but I cannot remember where now :(

> Would it be correct to remove the comment about this and change the assertion to >=?
No, because we configure Marathon mock to delay for `backend_request_timeout * 0.5`, if the time spent is exaclty `backend_request_timeout * 0.5` then it means that the request returned earlier than `backend_request_timeout * 0.5` as nothing in instantaneous and there is still some context switching and data processing. It is much more reliable this way.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: > Borkerage is not a word. Perhaps a better term would be test_if_broken_marathon.... Optional partly because it is not new to this codebase.

Sure, fixing.

> Personally I prefer to organise tests into more granular classes, and would maybe go for 
This is more complext than that - e.g. AR fixtures used across test harness may have per-module, per-class or per-test scope. You *could* organise some of the classes here, but others (esp. the ones that use per-class AR fixture) would either have to be kept as-is, or we would have to accept longer tests execution times.

adamtheturtle: Understood

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: changed.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: I was thinking about it at the very begining. The problem is that we have ~38 mock endpoints, giving name to each one of them would be even more confusing. This is why we have this idea of `mocker` which is basically a control plane on top of all these mock endpoints (mock HTTP servers to be precise).

Also, we need to be very carefull to not to overdoo and not to introduce too many terms. People from outside are familiar with the Mesos socket, but not necessarily with `mesos_dns_endpoint` term. 

Even more - I think I have three different Marathons, all of them used for different tests. What you are suggesting would require being extra creative with naming or using `marathon_endpoint_1`, marathon_endpoint_2`, `marathon_endpoint_default`, etc...

adamtheturtle: 😢 - I understand and empathise and can't think of a suitable approach which would alleviate this issue for me.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: > However, as you are more familiar with the codebase I will go with your judgement.

I am really trying to keep them at the minimum. Think also about people from the outside who have a problem with error message that is longer than 10 lines that are trying to contribute to AR, having too many fixtures will make it impossible for them to contribute effectively.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
adamtheturtle: ❓ Re test names, I tend to prefer "test that" names vs "test if". And personally I leave the "that" out. E.g. `test_marathon_brokerage_(prevents|does_not_prevent)`.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: See above for the discussion about using symbolic names with endpoints.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: This is because there is more than one call to marathon from the cache atm (two to be precise).

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -553,3 +554,198 @@ def test_if_scheme_is_honoured_in_mesos_scheduler_entry(
...
+                            aux_data=backend_request_timeout * 0.5)
+
+        # Measure the time it took and the results:
+        with GuardedSubprocess(ar):
+            t_start = time.time()
+            resp = requests.get(url,
+                                allow_redirects=False,
+                                headers=valid_user_header)
+
+            t_spent = time.time() - t_start
```
adamtheturtle: At this point in this and other tests you can leave the context manager.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
adamtheturtle: Thank you. If you think that this means that the test is fragile, please make that clear so that a future maintainer can know when to fix it.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Personally, I think that the name of the test should be a question that is answered in the code. This way potential test failures are very clear to decipher:

`test_if_foo_is_bazzing`: FAIL vs `test_foo_is_bazzing`: FAIL

packages/adminrouter/extra/src/test-harness/tests/conftest.py
adamtheturtle: I simply meant - would be nice to make clear that this isn't used by using `usefixture`.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: fixed.

BenWhitehead:
@mesosphere-mergebot bumb-ee

BenWhitehead:
Holding for `enterprise-pr` status to go green

BenWhitehead:
@mesosphere-mergebot bump-ee

BenWhitehead:
@vespian can you please rebase this pr, then run bump-ee to get a green enterprise verification?

vespian:
@mesosphere-mergebot bump-ee

vespian:
@BenWhitehead rebased & green - both open and EE. Please, have another look.

lingmann:
Cool thank you @vespian ! Removing the holding label, this will be on the train today.

PR #1649, opened 2017-06-12
Upgrade checks

amitaekbote:
## High Level Description

Run the node and cluster checks before and after upgrade

## Related Issues

https://jira.mesosphere.com/browse/DCOS-15924

## Checklist for all PR's

___

Reviewers should be:

### DIFF
--- a/dcos_installer/upgrade.py
+++ b/dcos_installer/upgrade.py
@@ -71,6 +71,23 @@ if [ "$found_version" != "{{ installed_cluster_version }}" ]; then
     exit 1
 fi
 
+# Check if the node has node/cluster checks and run them
+if [ -f /opt/mesosphere/etc/dcos-3dt-runner-config.json ]; then
+   # command exists
+   if !(output=$(3dt check node-poststart)); then
+      echo "Cannot proceed with upgrade, node checks failed"
+      echo $output
+      exit 1
+   fi
+
+   if !(clusteroutput=$(3dt check cluster)); then
+      echo "Cannot proceed with upgrade, cluster checks failed"
+      echo $clusteroutput
+      exit 1
+   fi
+
+fi
+
 # Determine this node's role.
 ROLE_DIR=/etc/mesosphere/roles
 
@@ -96,6 +113,17 @@ echo "Upgrading DC/OS $role_name {{ installed_cluster_version }} -> {{ installer
 pkgpanda fetch --repository-url={{ bootstrap_url }} {{ cluster_packages }} > /dev/null
 pkgpanda activate --no-block {{ cluster_packages }} > /dev/null
 
+T=300
+until OUT=$(3dt check node-poststart && 3dt check cluster) || [[ T -eq 0 ]]; do
+    sleep 1
+    let T=T-1
+done
+RETCODE=$?
+if [ $RETCODE -ne 0 ]; then
+    echo "Node upgrade not successful, checks failed"
+    echo $OUT
+fi
+exit $RETCODE
 """
 
 


amitaekbote:
https://github.com/mesosphere/dcos-test-utils/pull/32 -> upgrade test

dcos_installer/upgrade.py
```diff
@@ -71,6 +71,25 @@
     exit 1
 fi
 
+# Check if the node has node/cluster checks and run them
+if 3dt check; then
+   # command exists
+   if 3dt check node-poststart; then
+      echo "Node checks passed, proceeding with upgrade"
```
branden: I'd recommend being silent on success, since that should be the typical case and the user shouldn't have to think about it.

dcos_installer/upgrade.py
```diff
@@ -71,6 +71,25 @@
...
+   else
+      echo "Cannot proceed with upgrade, node checks failed"
+      exit 1
```
branden: We should print runner output on failure, otherwise failures will be a pain to debug.

dcos_installer/upgrade.py
```diff
@@ -71,6 +71,25 @@
     exit 1
 fi
 
+# Check if the node has node/cluster checks and run them
+if 3dt check; then
```
branden: We should be careful with this test, if `3dt` is broken then this'll cause us to silently skip checks. It'd be more reliable to test for the existence of the check config file.

amitaekbote:
PTAL @darkonie @branden

dcos_installer/upgrade.py
```diff
@@ -96,6 +115,19 @@
...
+do
+  if 3dt check node-poststart && 3dt check cluster; then
+     echo "Node and cluster checks passed"
+     exit 0
+  fi
+  echo "Waiting for node and cluster checks to pass"
+  sleep 5
+done
+
+exit 1
```
branden: If this fails, there'll be no indication aside from the exit code. This should print something on failure.

amitaekbote:
TEST_USE_CHECKS set to false: https://teamcity.mesosphere.io/viewLog.html?buildId=683222&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_UpgradeTest&tab=buildParameters 

TEST_USE_CHECKS set to true: https://teamcity.mesosphere.io/viewLog.html?buildId=683234&buildTypeId=ClosedSource_Dcos_IntegrationTests_CloudIntegrationTests_UpgradeTest&tab=buildResultsDiv

amitaekbote:
env.TEST_UPGRADE_USE_CHECKS is the new parameter to enable testing new method of upgrades - https://github.com/mesosphere/dcos-test-utils/pull/32

dcos_installer/upgrade.py
amitaekbote: The output of node and cluster checks gets printed out, check the integration test, its actually noisy

dcos_installer/upgrade.py
amitaekbote: I think relying on 3dt is ok, if 3dt is broken we have bigger problems

dcos_installer/upgrade.py
amitaekbote: I can change this to output to /dev/null

dcos_installer/upgrade.py
amitaekbote: Currently the way it is it will output everything

dcos_installer/upgrade.py
branden: Nothing else calls `3dt check` except for this script. If that command exits 1, the only symptom would be this script silently skipping checks before running an upgrade, potentially breaking a cluster.

amitaekbote: Would checking if gen has check_config work?

amitaekbote:
@branden two things are still missing from the previous pr:

1. verifying if DCOS_3DT_CONFIG_PATH exists, the path is not stored in gen calc , check_config is. I am not sure how to get this value from dcos-config.yaml 

2. Storing the value of the output for the error, here's a few issues - https://serverfault.com/questions/259339/bash-variable-loses-value-at-end-of-while-read-loop . I am not sure what the right way to do it is yet, still exploring. For now I just call the command again incase it fails and that should give the output, now unless something happens between two runs ( maybe the checks pass right before we exit) this will give the output. 

Do you have suggestions on how to do 1? 

The rest of the comments are addressed.

branden:
@amitaekbote You can just copy the config file's path from `dcos-config.yaml` since it won't change. We could also turn that location into a gen parameter and reference it both there and here, but that feels like overkill.

amitaekbote:

amitaekbote:
@branden @mellenburg ptal

amitaekbote:
Upgrade test passed with TEST_UPGRADE_USE_CHECKS set to true and false

amitaekbote:
@mesosphere-mergebot bump-ee

dcos_installer/upgrade.py
```diff
@@ -71,6 +71,23 @@
     exit 1
 fi
 
+# Check if the node has node/cluster checks and run them
+if [ -f /opt/mesosphere/etc/dcos-3dt-runner-config.json ]; then
```
branden: I added a ticket for removing this test and always running checks after we branch 1.10: https://jira.mesosphere.com/browse/DCOS_OSS-1274

amitaekbote:

PR #1655, opened 2017-06-14
Create state directory for packages activated via the Pkgpanda API

branden:
## High Level Description

This causes the Pkgpanda API to create state directories as appropriate for the packages it activates, making it consistent with the CLI's current behavior.

## Related Issues

  - [DCOS_OSS-1212](https://jira.mesosphere.com/browse/DCOS_OSS-1212) Pkgpanda API does not create package state dirs in /var/lib/dcos/

## Checklist for all PR's

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -23,7 +23,8 @@ from subprocess import CalledProcessError, check_call, check_output
 from typing import Union
 
 from pkgpanda.constants import (DCOS_SERVICE_CONFIGURATION_FILE,
-                                RESERVED_UNIT_NAMES)
+                                RESERVED_UNIT_NAMES,
+                                STATE_DIR_ROOT)
 from pkgpanda.exceptions import (InstallError, PackageError, PackageNotFound,
                                  ValidationError)
 from pkgpanda.util import (download, extract_tarball, if_exists, load_json, write_json, write_string)
@@ -591,7 +592,8 @@ class Install:
             skip_systemd_dirs=False,
             manage_users=False,
             add_users=False,
-            manage_state_dir=False):
+            manage_state_dir=False,
+            state_dir_root=STATE_DIR_ROOT):
 
         assert type(rooted_systemd) == bool
         assert type(fake_path) == bool
@@ -622,6 +624,9 @@ class Install:
         self.__manage_users = manage_users
         self.__add_users = add_users
         self.__manage_state_dir = manage_state_dir
+
+        assert not state_dir_root.endswith('/')
+        self.__state_dir_root = state_dir_root
 
     def _get_dcos_configuration_template(self):
         return {"sysctl": {}}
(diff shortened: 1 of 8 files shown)


branden:
@mesosphere-mergebot bump-ee

branden:

branden:
@amitaekbote @orsenthil @gpaul PTAL

orsenthil:
@branden  - the JIRA talks about the ability to manage state_dir using Pkgpanda HTTP API and this PR is about ability to set to a customizable state dir.

I think, it is good idea to provide customization. I don't see the rationale on need in the JIRA or PR. Would you like to mention that in the JIRA?

orsenthil:
I've reviewed the code. 
I am fine with this change. When there is a customizable --root, why not --state-dir.

You might update the JIRA on the background behind this (if any) so that folks don't wonder why did this come up in 1.10. 

LGTM.

branden:
@orsenthil This PR teaches the Pkgpanda API to manage the state dir. Parameterizing the state dir was necessary for integration tests.

pkgpanda/__init__.py
```diff
@@ -792,10 +797,10 @@ def _get_service_names(_dir):
             if package.username is not None:
                 sysusers.add_user(package.username, package.group)
 
-            # Ensure the state directory in `/var/lib/dcos` exists
+            # Ensure the state directory exists
             # TODO(cmaloney): On upgrade take a snapshot?
             if self.__manage_state_dir:
-                state_dir_path = '/var/lib/dcos/{}'.format(package.name)
+                state_dir_path = self.__state_dir_root + '/' + package.name
```
gpaul: you might want to consider using
```
state_dir_path = os.path.join(self.__state_dir_root, package.name)
```
And perhaps use normpath, too:
```
state_dir_path = os.path.normpath(os.path.join(self.__state_dir_root, package.name))
```
That way it's a bit more robust against extra slashes, etc.

branden: +1, but I'll leave this since this module is already inconsistent about how it joins filesystem paths, as well as the entire repo. I filed a ticket to track cleaning that up: https://jira.mesosphere.com/browse/DCOS_OSS-1270

branden:
@mesosphere-mergebot label shipit

pkgpanda/__init__.py
orsenthil: [not a review comment] 

I can understand when when we use  + '/' + for doing path constructions.
When using os.path.join, we have to be careful with the path join protocols, like 
```
>>> r = '/root'
>>> a = '/a'
>>> b = 'b'
>>> os.path.join(r, a)
'/a'
>>> os.path.join(r, b)
'/root/b'
```

* By doing '/' +, we might as mistake construct a path that has '//' in it. Technically, all programs will traverse that properly.


Also, we should try the pathlib '/' fancy :) https://docs.python.org/3/library/pathlib.html#operators

gpaul: Oh wow! I expected `os.path.join` to match golang's path/filepath but it doesn't:
https://play.golang.org/p/EK5Tk27Zhn

branden: Yeah, that's the big reason that we don't use `os.path.join()` everywhere. I'm sure we could write our own helper function around it that does what we like. Raising an error if `os.path.sep` is in a path segment before we call `os.path.join()` would be a good start.

orsenthil: Or use pathlib. It has many useful path related functions. It is worth exploring IMO.

PR #1683, opened 2017-06-23
Bump ZooKeeper to release 3.4.10

jgehrcke:
This addresses https://jira.mesosphere.com/browse/DCOS_OSS-739 and bumps ZooKeeper in DC/OS from version 3.4.8 to version 3.4.10.

Official changelog:
* http://zookeeper.apache.org/doc/r3.4.9/releasenotes.html
* http://zookeeper.apache.org/doc/r3.4.10/releasenotes.html

Particularly relevant changelog items:
* "Zookeeper service becomes unavailable when leader fails to write transaction log"
* "Back-port ZOOKEEPER-1460 to 3.4 for IPv6 literal address support."
* "CancelledKeyException in zookeeper"
* "Potential memory leak in recordio.c"
* "Large databases take a long time to regain a quorum"
* "Correct DataNode.getChildren() inconsistent behaviour."

### DIFF
--- a/packages/exhibitor/buildinfo.json
+++ b/packages/exhibitor/buildinfo.json
@@ -9,8 +9,8 @@
     },
     "zookeeper": {
       "kind": "url_extract",
-      "url": "https://downloads.mesosphere.com/zookeeper/zookeeper-3.4.8.tar.gz",
-      "sha1": "51b61611a329294f75aed82f3a4517a4b6ff116f"
+      "url": "http://www-us.apache.org/dist/zookeeper/zookeeper-3.4.10/zookeeper-3.4.10.tar.gz",
+      "sha1": "eb2145498c5f7a0d23650d3e0102318363206fba"
     },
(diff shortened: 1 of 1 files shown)


jgehrcke:
@mesosphere-mergebot bump-ee

neilconway:
Why 3.4.9 rather than 3.4.10?

jgehrcke:
@mesosphere-mergebot bump-ee

spahl:
mesos is still using the 3.4.8 client. Concerns?

adam-mesos:
@neilconway Concerns about Mesos using the older client? Are we going to upgrade ZK in Mesos too?

neilconway:
@adam-mesos 3.4.8 and 3.4.10 should interoperate fine. We should definitely upgrade the vendored copy of Zk in Mesos, but I wouldn't consider that a blocker for upgrading Zk in DC/OS.

PR #1701, opened 2017-06-29
Adminrouter support for marathon HTTP basic auth

ts3ng:
## High Level Description

Admin router DC/OS ui breaks when you provide backend services with Basic HTTP Auth such as marathon or when you try to use the admin proxy endpoints (/exhibitor /marathon).  There is alos breakage when you configure jwt token based authentication you will get booted out of UI since authentication headers are set to BASIC vs token.  This change is to fix that behavior with the adminrouter UI.

The admin router service will look for the environment variable MESOSPHERE_HTTP_CREDENTIALS being set and inject the proper auth headers for request to marathon.  For example if you load marathon with the environment variable MESOSPHERE_HTTP_CREDENTIALS (related: https://jira.mesosphere.com/browse/MARATHON-7310), you activate http_credentals flag.  As-is this breaks adminrouter ui behavior as calls from from adminrouter will not get the proper auth headers it needs to make calls against marathon; with this fix we hope to resolve the problem.

Note with this fix all services exposed via the admin router proxy must use the the same HTTP basic auth credentials loaded into MESOSPHERE_HTTP_CREDENTIALS environment variable. 

## Related Issues
https://jira.mesosphere.com/browse/MARATHON-7310
https://jira.mesosphere.com/browse/DCOS_OSS-1064

## Checklist for all PR's

___

Reviewers should be:


orsenthil:
@vespian  could you please shepherd this.
This is currently targeting 1.9.0 and that's wrong

This should either be against `master` or  `1.9`

jgehrcke:
Thank you @ts3ng for this patch, your thinking and suggestion is much appreciated!

Please excuse the long time it took us to respond. I could not fully understand the architecture that you are trying to get to work, so the following response might be based on bad assumptions, which you will hopefully be able to point out.

We believe that the patch is not coherent with the decentralized DC/OS authentication architecture that we strive for. In that architecture, every DC/OS-native component (but also services deployed on top of DC/OS) can be a DC/OS authenticator. These authenticators can verify the canonical DC/OS authentication proof, the "DC/OS authentication token" (which happens to be a JWT). Admin Router is such a component. In fact, it currently acts as an authenticator _on behalf of_ other components.

In that architecture it does not really make sense to strive for enabling authentication schemes _through_ Admin Router that are different from the DC/OS authentication method.

If required for various reasons (such as custom authentication schemes for services run on top of DC/OS) there always are ways to access services _directly_ or through a TCP proxy instead of through an HTTP proxy that could interfere with the authentication scheme expected by the service that is being proxied to. For example, in many DC/OS deployments it makes sense to expose services through something like Marathon-LB (which would -- to my knowledge -- support your use case). Notably, to give a specific recommendation we must clearly distinguish the DC/OS-native "root Marathon" and Marathons launched on top of that, which we usually refer to as "Marathon on Marathon" (MoM).

Hence, BASIC auth against Admin Router (nginx) upstreams is not something that we think we should support in the long term. I hope that this is not too bad news. What's hopefully promising is that at least in my mind Marathon on DC/OS should be a DC/OS authentictor by itself (instead of relying on Admin Router to perform this task for Marathon). I am not sure why you want to "provide backend services with Basic HTTP Auth such as marathon", but this will hopefully obliterate this motivation.

Please let us know what you think!

Thanks and cheers

JP

ts3ng:
@jgehrcke I appreciate the feedback,  Here are my replies per your comments inline:

We believe that the patch is not coherent with the decentralized DC/OS authentication architecture that we strive for. In that architecture, every DC/OS-native component (but also services deployed on top of DC/OS) can be a DC/OS authenticator. These authenticators can verify the canonical DC/OS authentication proof, the "DC/OS authentication token" (which happens to be a JWT). Admin Router is such a component. In fact, it currently acts as an authenticator on behalf of other components.

**_Sorry, i was not clear about the premise of what my PR is suppose to support. We currently are concerned about authentication for each internal traffic component (such as marathon).  i.e if you are able to get access into the internal network via container you do not have to broker requests through the adminrouter.  The use case is when direct access to http://marathon.mesos:8080 is exposed they are bypassing the centralized admin router authentication.  To tighten up security internal we are enabling HTTP basic auth but this essentially breaks the requests between adminrouter and marathon as admin router continues passing the JWT tokens downstream to the root marathon, causing weird UI behavior.  My PR addresses this weird behavior._**

In that architecture it does not really make sense to strive for enabling authentication schemes through Admin Router that are different from the DC/OS authentication method.

**_I might of mis-titled my PR it should be adminrouter support for marathon http basic authentication.  We are not enabling a different scheme for the admin router but simply downstream in marathon.  But this breaks some of the admin router functionality._**

**_If required for various reasons (such as custom authentication schemes for services run on top of DC/OS) there always are ways to access services directly or through a TCP proxy instead of through an HTTP proxy that could interfere with the authentication scheme expected by the service that is being proxied to. For example, in many DC/OS deployments it makes sense to expose services through something like Marathon-LB (which would -- to my knowledge -- support your use case). Notably, to give a specific recommendation we must clearly distinguish the DC/OS-native "root Marathon" and Marathons launched on top of that, which we usually refer to as "Marathon on Marathon" (MoM).

We have only tested our PR with the root marathon.  But i would expect the same issue to arise if you enable HTTP Basic Auth on the the child marathons.  We wouldn’t want people to access marathon outside of the admin router if we can hence adding downstream support for http basic auth in marathon through the admin router.  I.e we use Oauth in admin router with jwt tokens.  But once they are logged in be able to access the marathon via the admin router proxy since they have already auth’d in the admin router they will not need to provide any additional auth via the proxy to downstream applications such as root marathon._**

Hence, BASIC auth against Admin Router (nginx) upstreams is not something that we think we should support in the long term. I hope that this is not too bad news. What's hopefully promising is that at least in my mind Marathon on DC/OS should be a DC/OS authentictor by itself (instead of relying on Admin Router to perform this task for Marathon). I am not sure why you want to "provide backend services with Basic HTTP Auth such as marathon", but this will hopefully obliterate this motivation.

**_Hopefully the initial premise in the beginning of my response will provide adequate info on why we enable HTTP basic auth for internal traffic. Perhaps I’m a little thrown off when you mean upstream?  do you mean requests 

Here are some documents I was looking at when I was researching:
How to enable http authentication on marathon: https://mesosphere.github.io/marathon/docs/ssl-basic-access-authentication.html
Mesos has a smilar authentication method which is essential basic http auth but with storing secret files on the filessytem: https://mesosphere.github.io/marathon/docs/framework-authentication.html
This seems to only supported in enterprise edition: https://mesosphere.github.io/marathon/docs/auth-access-ctrl.html

If you have more information please feel free to pass it my way.  Perhaps there might be some clarity that i need on my side or information that i'm lacking.  perhaps a chat via mesosphere slack might clear up any assumption on both sides? You can DM me at ts3ng.  Apologies if there was any confusion as this PR is still a WIP and was something i wasn't planning on getting back til until after this week_**

ts3ng:
FYI, a similar PR will be opened against the master once we have some time.  This PR is purely informational only for now.

ts3ng:
Swapping this PR out for https://github.com/dcos/dcos/pull/1950/files

PR #1747, opened 2017-07-20
Admin Router: Remove periodic AR reloading in favour of DNS queries

vespian:
## High Level Description

This PR:
* performs some minor cleanup of docker files - CA bundle location sync
* removes $upstream_marathon variable definition as it is no longer needed - all requests go through /service endpoint and the upstream is determined basing on Mesos' `/state-summary` endpoint
* makes devkit as per-flavour. This enables running both EE and Open devkits side-by-side and permits differences between EE and Open devkits
* exposes test harness logs in a log file, Jenkins should automatically pick it up and treat it as yet another artefact
* removes period reloading of AR in favour of DNS queries, see DCOS_OSS-1446 for details
* removes TTL override for AR DNS resolver. See the comments in the configuration for more details.

## Related Issues

  - https://jira.mesosphere.com/browse/DCOS_OSS-1446 - Admin Router: Dynamic DNS resolution of upstreams

## Failing tests
This PR has already been once removed from a Train due to tests constantly failing. Please have a look at https://jira.mesosphere.com/browse/DCOS_OSS-1467 where I tried to prove that the failures are unrelated to the PR itself. The PR just exposes bugs and race conditions that were already present in DC/OS.

The tests that are failing and will need to be muted after merging this PR:
* `test_octarine`
* `test_service_discovery_docker_bridge`
* `test_service_discovery_docker_overlay`
* `test_service_discovery_docker_overlay_port_mapping`
* `test_ip_per_container`

DC/OS Vagrant issues that need to be addressed to unmute them:
* https://jira.mesosphere.com/browse/DCOS_OSS-1523 `DC/OS Vagrant: Improve how memory resources are defined for agents`
* https://jira.mesosphere.com/browse/DCOS_OSS-1524 `dcos-diagnostics --diag` returns false positives during DC/OS install.

## Related PRs:
Base PR from old AR repo: https://github.com/dcos/adminrouter/pull/50
Enterprise PR: https://github.com/mesosphere/dcos-enterprise/pull/1205

### DIFF
--- a/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
@@ -1,6 +1,6 @@
 set $upstream_networking_api http://127.0.0.1:61430;
 
-proxy_ssl_trusted_certificate /run/dcos/pki/CA/certs/ca.crt;
+proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
--- a/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
@@ -4,11 +4,11 @@ proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
-set $upstream_mesos http://127.0.0.2:5050;
+set $upstream_mesos http://leader.mesos:5050;
 set $upstream_secrets http://127.0.0.1:1337;
 set $upstream_cosmos http://127.0.0.1:7070;
 
 # Required by EE Agent:
-set $upstream_iam http://127.0.0.1:8101;
+set $upstream_iam http://master.mesos:8101;
 
 set $adminrouter_agent_port 61001;
--- a/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-ee.conf
@@ -4,10 +4,7 @@ proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
-# For the sake of tests simplicity, we will not be using/testing leader.mesos
-# alias here. This should be done by integration tests.
 set $upstream_mesos http://127.0.0.2:5050;
-set $upstream_marathon http://127.0.0.1:8080;
 set $upstream_secrets http://127.0.0.1:1337;
 set $upstream_cosmos http://127.0.0.1:7070;
 
--- a/packages/adminrouter/docker/adminrouter-upstreams-open.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-open.conf
@@ -1,6 +1,6 @@
 set $upstream_networking_api http://127.0.0.1:61430;
 
-proxy_ssl_trusted_certificate /run/dcos/pki/CA/certs/ca.crt;
+proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
--- a/packages/adminrouter/docker/adminrouter-upstreams-open.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-open.conf
@@ -4,11 +4,11 @@ proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
-set $upstream_mesos http://127.0.0.2:5050;
+set $upstream_mesos http://leader.mesos:5050;
 set $upstream_secrets http://127.0.0.1:1337;
 set $upstream_cosmos http://127.0.0.1:7070;
 
 # Required by EE Agent:
-set $upstream_iam http://127.0.0.1:8101;
+set $upstream_iam http://master.mesos:8101;
 
 set $adminrouter_agent_port 61001;
--- a/packages/adminrouter/docker/adminrouter-upstreams-open.conf
+++ b/packages/adminrouter/docker/adminrouter-upstreams-open.conf
@@ -4,10 +4,7 @@ proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
-# For the sake of tests simplicity, we will not be using/testing leader.mesos
-# alias here. This should be done by integration tests.
 set $upstream_mesos http://127.0.0.2:5050;
-set $upstream_marathon http://127.0.0.1:8080;
 set $upstream_secrets http://127.0.0.1:1337;
 set $upstream_cosmos http://127.0.0.1:7070;
 
--- a/packages/adminrouter/build
+++ b/packages/adminrouter/build
@@ -25,16 +25,12 @@ mkdir -p "$PKG_PATH/lib"
 cp /lib/x86_64-linux-gnu/libpcre.so.3 "$PKG_PATH/lib/libpcre.so.3"
 
 # Copy systemd unit files:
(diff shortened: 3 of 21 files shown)


vespian:

vespian:
@mhrabovcin @gpaul @jgehrcke Please have a look, thanks!

packages/adminrouter/docker/adminrouter-upstreams-ee.conf
```diff
@@ -1,17 +1,14 @@
 set $upstream_networking_api http://127.0.0.1:61430;
 
-proxy_ssl_trusted_certificate /run/dcos/pki/CA/certs/ca.crt;
```
mhrabovcin: Can we now remove this file from this repo and `Dockerfile`?

packages/adminrouter/docker/adminrouter-upstreams-open.conf
```diff
@@ -1,17 +1,14 @@
 set $upstream_networking_api http://127.0.0.1:61430;
 
-proxy_ssl_trusted_certificate /run/dcos/pki/CA/certs/ca.crt;
+proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
```
jgehrcke: Thanks and I hope we never break this again :-).

packages/adminrouter/docker/adminrouter-upstreams-ee.conf
```diff
@@ -1,17 +1,14 @@
...
+proxy_ssl_trusted_certificate /run/dcos/pki/CA/ca-bundle.crt;
 proxy_ssl_verify on;
 proxy_ssl_verify_depth 5;
 
-# For the sake of tests simplicity, we will not be using/testing leader.mesos
-# alias here. This should be done by integration tests.
-set $upstream_mesos http://127.0.0.2:5050;
-set $upstream_marathon http://127.0.0.1:8080;
+set $upstream_mesos http://leader.mesos:5050;
```
mhrabovcin: My concern with this patch is whether we can make things worse in respect of existing `leader.mesos` TTL.  We had 30 seconds refresh interval that forced re-resolving the leader. Our DNS service has set TTL to 60 seconds so we can possibly double the time for new leader detection.

packages/adminrouter/docker/adminrouter-upstreams-ee.conf
vespian: Yes, but not in this PR. There is an issue for that:https://jira.mesosphere.com/browse/DCOS-16017 We can/should probably also remove EE stuf from Open and vice-versa now that devkit containers have been split.

Let's properly pull it into a sprint and create a separate PR.

packages/adminrouter/docker/adminrouter-upstreams-open.conf
vespian: The problem here is that due to pkgpanda limitations there is no easy way to keep docker files DRY :(

packages/adminrouter/docker/adminrouter-upstreams-ee.conf
vespian: Your comment made me thinking... 

In case of upstreams defined using DNS (like e.g. `/mesos`), we can't override DNS validity period with a value higher than the TTL set originally by mesos_dns. This gives us 60s upper limit.

As for the tests, the lower the better, with TTL=1 being the reasonable minimul.

As for the Lua code - this is a place where stuff starts to be interesting. Currently `CACHE_POLL_PERIOD = 25` seconds, we also want to be sure that `leader.mesos` will be re-resolved on every cache refresh, so max TTL of any DNS entry should be lower than 25s.

I think that we, on one hand, should not be using default TTL of the MesosDNS as it is too high, on the other hand, we should not query MesosDNS too often. Five seconds seems like a reasonable value. This shouldn't slow down our tests too much.

What do you think ?

mhrabovcin: 5 seconds is exactly the TTL I had on my mind, but I was thinking about patching `MesosDNS` to allow to configure different TTL for leader record (which is a snowflake already).

vespian: I would like to avoid touching MesosDNS, too many things rely on it, especially on `leader.mesos`. AR has very specific needs when it comes to DNS TTL and overriding it in the config seems OK to me.

jgehrcke:
Thanks for this work! I would very much appreciate if @gpaul could also give a quick review.

packages/adminrouter/docker/adminrouter-upstreams-ee.conf
```diff
@@ -1,17 +1,14 @@
...
 set $upstream_secrets http://127.0.0.1:1337;
 set $upstream_cosmos http://127.0.0.1:7070;
 
 # Required by EE Agent:
-set $upstream_iam http://127.0.0.1:8101;
+set $upstream_iam http://master.mesos:8101;
```
gpaul: The Agent shouldn't be connecting to the IAM directly, we're moving towards proxying all IAM traffic through the master AdminRouter and having that listen on localhost, only.

vespian: We do - in production. 

Test harness cheats a little by pushing requests directly to IAM mock, instead of starting two AR instances (Master + Agent). This helps to keep things simpler without any negative effect on tests (we are testing all IAM endpoints in Master using different/separate tests) :)

jgehrcke:
Thank you, everyone!

BenWhitehead:
@vespian Can you please investigate the failures that were encountered when we tried to add this PR to [train-195](https://github.com/dcos/dcos/pull/1764).

It looks like @jgehrcke already made a ticket to track debugging the failure in `integration-test/vagrant-bash` here: https://github.com/dcos/dcos/pull/1764#issuecomment-318036447

mellenburg:
@vespian @jgehrcke 
We need to resolve the vagrant-bash issue in this PR before this can be given a ship-it label.

I am strongly opposed to simply muting tests as this is an extremely unreliable and error prone process.

Rather, I see at least two ways forward:
(1) We add some pytest commands into the test suite to skip or xfail these test cases when we detect that platform==vagrant (very easy change)
(2) We declare the vagrant tests to be flawed at the vagrant level and then drop dcos-vagrant as a REQUIRED test. To do this, we would need to make sure the coverage provided by dcos-vagrant is handled elsewhere. More specifically, we would need to use one of the onprem integration test clusters (there are currently 2 per open PR) and set it to `num_masters:3` and `lb_enabled: false` (also very easy)

The point is, we need to pick one of these definitive options. Muting in CI is just brushing something under the rug until the mute expires or until we see that the mute logic accidentally allowed other failed tests to be considered valid (this has happened before)

jgehrcke:
> We need to resolve the vagrant-bash issue in this PR before this can be given a ship-it label.

@mellenburg given the technical details that I know your proposal from Slack to "make the tests aware they are running on vagrant and skip" would have been the sanest choice, I think. I am fine with "holding" this, but more for political/organizational reasons rather than technical reasons.

Because we know

> Its a new test failure caused by a change that exposes an old bug.

(Karl's words)

and

> The PR just exposes bugs and race conditions that were already present in DC/OS.

(Pawel's words)

Given all boundary conditions that I am aware of I find it appropriate to not block train 201 until DCOS_OSS-1524 and DCOS_OSS-1523 are addressed, but to move ahead with

This patch needs to land in 1.10 beta2 and if we want to wait for DCOS_OSS-1524 and DCOS_OSS-1523 before we land this patch, then I am fine -- but the release has to wait.

jgehrcke:
I want to add that when @vespian wrote

> The tests that are failing and will need to be muted

he _probably_ meant your option (1), Michael:

> We add some pytest commands into the test suite to skip or xfail these test cases when we detect that platform==vagrant (very easy change)

Did you, @vespian? :)

vespian:
> Did you, @vespian? :)

It was never my intention to mute whole dcos-vagrant test, and I do not know where it stems from. Since the very beginning, I was always talking about five particular tests - can link slack messages if needed :D 

As for the way how exactly they should be muted - I do not know what is the best way to suppress TC failures ATM. By:

> The tests that are failing and will need to be muted after merging this PR:

I initially meant doing it in TC basing on our experience with i.e. strict cluster tests, but this can be anything that disables offending tests until the bugs are fixed. I try to trust that people responsible for maintaining tests will be able to choose the best option, hence I did not investigate how exactly it should be done and left it to domain experts to decide. I am sorry if this caused confusion.

I hope it helps :)

orsenthil:

PR #1761, opened 2017-07-24
Admin Router: improve logging messages related to caching

vespian:
## High Level Description

Improves logging messages related to caching.

## Related Issues

  - https://jira.mesosphere.com/browse/DCOS-16614 - `Admin Router: "Using stale `svcapps` cache" and "Resolved via Mesos state-summary" for the same request`

## Related PRs

EE: https://github.com/mesosphere/dcos-enterprise/pull/1220

### DIFF
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -617,7 +617,7 @@ local function get_cache_entry(name, auth_token)
 
     -- Cache is stale, but still usable:
     if cache_age > _CONFIG.CACHE_MAX_AGE_SOFT_LIMIT then
-        ngx.log(ngx.NOTICE, "Using stale `" .. name .. "` cache entry to fulfill the request")
+        ngx.log(ngx.NOTICE, "Cache entry `" .. name .. "` is stale")
     end
 
     local entry_json = cache:get(name)
--- a/packages/adminrouter/extra/src/lib/service.lua
+++ b/packages/adminrouter/extra/src/lib/service.lua
@@ -203,7 +203,7 @@ local function resolve(service_name, mesos_cache, marathon_cache)
     --    or not.
     --  - err_code, err_text - if an error occured these will be HTTP status
     --    and error text that should be sent to the client. `nil` otherwise
-    ngx.log(ngx.DEBUG, "Resolving service `".. service_name .. "`")
+    ngx.log(ngx.DEBUG, "Trying to resolve service name `".. service_name .. "`")
     res, err_code, err_text = resolve_via_marathon_apps_state(
         service_name, marathon_cache)
 
--- a/packages/adminrouter/extra/src/test-harness/tests/test_cache.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/test_cache.py
@@ -207,7 +207,7 @@ class TestCache:
                 SearchCriteria(1, False),
             'Mesos state cache has been successfully updated':
                 SearchCriteria(2, False),
-            'Using stale `svcapps` cache entry to fulfill the request':
+            'Cache entry `svcapps` is stale':
                 SearchCriteria(1, True),
         }
 
@@ -300,7 +300,7 @@ class TestCache:
                 SearchCriteria(1, False),
             'Marathon apps cache has been successfully updated':
                 SearchCriteria(2, False),
-            'Using stale `mesosstate` cache entry to fulfill the request':
+            'Cache entry `mesosstate` is stale':
                 SearchCriteria(1, True),
         }
 


vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot ready for review

orsenthil:

orsenthil:
@mesosphere-mergebot label  ship-it

vespian:
@jgehrcke I think that with the introduction of nested services we must be a little bit more carefull about how we contstruct log entries. Your proposal is:
```
Trying to resolve service name `<sname>` with stale `svcapps` cache state
```
In order to route the request to `/service/foo/bar/baz/cat.jpg`, the Lua code first fetches the contents of the cache and stores them in a local variable. This is a moment when you can see
```
-        ngx.log(ngx.NOTICE, "Using stale `" .. name .. "` cache entry to fulfill the request")
￼ +        ngx.log(ngx.NOTICE, "Cache entry `" .. name .. "` is stale, using it anyway")
```
Then it iteratively tries to resolve each service name "candidate":
* `foo`
* `foo/bar`
* `foo/bar/baz`
* `foo/bar/baz/cat.jpg`

First match/shortest match wins. At this stage, depending on which "candidate" turns out to be the service name, you can see one, two, three, or four messages like this:
```
-    ngx.log(ngx.DEBUG, "Resolving service `".. service_name .. "`")
￼ +    ngx.log(ngx.DEBUG, "Trying to resolve service name `".. service_name .. "`")
```
We can just add "using stale cache" segment to each and every occurance of this line, e.g.
```
Trying to resolve service name `foo` with stale `svcapps` cache state
Trying to resolve service name `foo/bar` with stale `svcapps` cache state
Trying to resolve service name `foo/bar/baz` with stale `svcapps` cache state
Trying to resolve service name `foo/bar/baz/cat.jpg` with stale `svcapps` cache state
```
But this will:
* inflate logs size a bit
* make logs a bit harder to read (AR already adds lots of extra text to each log line stemming from Lua code)
* make the code unecesarrily complex, as we will need to drag cache freshness information along all the function interfaces (cache fetch&check occurs now only at the begining of the resolving).

So the idea is to have:
```
Cache entry `svcapps` is stale, using it anyway.
Trying to resolve service name `foo`
Trying to resolve service name `foo/bar`
Trying to resolve service name `foo/bar/baz`
Resolved via Marathon, service id: `foo/bar/baz`
```
This way:
* amount of information logged is minimal, there is no duplication between lines
* we solve the original problem of `"Using stale `" .. name .. "` cache entry to fulfill the request"` wording - we no longer implying that we are resolved svcapps (-> vide it is a Marathon app) to resolve service name, only that we are using cache entry for Marathon state that is stale. Same with "Resolving..." vs "Trying to resolve..."
* All log entries tell a story - cache was loaded, service names iterated and finally resolved using ....

I would rather like to keep `...using it anyway` part, as `... is stale` on its own implies negative meaning - we can't use the cache, which is not true.

Does this help?

vespian:
@jgehrcke 
Also, regarding:
```
AR checked it and there was no data relevant to resolved service. It then tried resolving via Mesos state-summary cache and it succeded.
```
This ambiguity is resolved by:
* (via Marathon) https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/lib/service.lua#L104
* (via MesosDNS) https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/lib/service.lua#L134
* ( via Mesos) https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/lib/service.lua#L188

So in my opinion we should conentrate on not implying that `stale svcapps cache entry is used` to resolved an entry, but rather communicate that `svcapps cache entry is stale` and then relying on the final log line which specifies which exactly source was used to route request.

jgehrcke:
> We can just add "using stale cache" segment to each and every occurance of this line, e.g.

This is independent of the reported problem ...

You added this example in your previous comment:
```
Cache entry `svcapps` is stale, using it anyway.
Trying to resolve service name `foo`
Trying to resolve service name `foo/bar`
Trying to resolve service name `foo/bar/baz`
Resolved via Marathon, service id: `foo/bar/baz`
```
That kind of overview is useful for focusing on the problem pointed out in the ticket.

In the case discussed in the ticket it looked like this:
```
Using stale `svcapps` cache entry to fulfill the request
Resolved via Mesos state-summary, service id: `spark`
```
(the original report really just showed these two subsequent lines, no intermediate line)

with your current patch it would now say:
```
Cache entry `svcapps` is stale, using it anyway.
Resolved via Mesos state-summary, service id: `spark`
```

If you put yourself into the position of a human log reader those two are still contradictory, and the confusion remains. What does ", using it anyway" even intend to say, in addition to "Cache entry `svcapps` is stale"? Does it contain additional information?

Maybe the first line should just be "Cache entry `svcapps` is stale".

> ... is stale on its own implies negative meaning - we can't use the cache, which is not true.

I disagree. "Cache entry `svcapps` is stale" is a rather neutral observation.

vespian:
> Maybe the first line should just be "Cache entry svcapps is stale".

If you think that this is more readable for you, as a person looking at this "outside of my rabbit hole", then it WFM! Will adjust the PR tomorrow.

vespian:
@mesosphere-mergebot bump-ee

PR #1772, opened 2017-07-28
Refactor dcos_test_utils.marathon

mellenburg:
## High Level Description
Updates test suite so that the test server marathon app lives with the test server code (in dcos-integration-test). Meanwhile, the methods in dcos-test-utils were untangled

EE PR: https://github.com/mesosphere/dcos-enterprise/pull/1235
## Checklist for all PR's

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): https://github.com/dcos/dcos-test-utils/compare/b4859cf710d4ec1d2b49354dba98a4a7b9d40b5a...b297d1ca110411340fe4adb338588270a19eeb10
  - [x] Test Results: https://teamcity.mesosphere.io/viewLog.html?buildId=721328&buildTypeId=DcosIo_DcosTestUtils_ToxDcosTestUtils

### DIFF
--- a/packages/dcos-integration-test/extra/test_helpers.py
+++ b/packages/dcos-integration-test/extra/test_helpers.py
@@ -1,5 +1,133 @@
+import copy
 import json
+import uuid
 
+from dcos_test_utils import marathon
 
+TEST_APP_NAME_FMT = 'integration-test-{}'
+
+# make the expanded config available at import time to allow determining
+# which tests should run before the test suite kicks off
 with open('/opt/mesosphere/etc/expanded.config.json', 'r') as f:
     expanded_config = json.load(f)
+
+
+def marathon_test_app(
+        host_port: int=0,
+        container_port: int=None,
+        container_type: marathon.Container=marathon.Container.NONE,
+        network: marathon.Network=marathon.Network.HOST,
+        healthcheck_protocol: marathon.Healthcheck=marathon.Healthcheck.HTTP,
+        vip: str=None,
+        host_constraint: str=None):
+    """ Creates an app definition for the python test server which will be
+    consistent (i.e. deployable with green health checks and desired network
+    routability). To learn more about the test server, see in this repo:
+    ../packages/dcos-integration-test/extra/util/python_test_server.py
+
+    Args:
+        host_port: port that marathon will use to route traffic into the
+            test server container. If set to zero, then marathon will assign
+            a port (which is referenced by index).
+        container_port: if network is BRIDGE, then the container can have a
+            port remapped inside the container. In HOST or USER network, the
+            container port must be the same as the host port
+        container_type: can be NONE (default Mesos runtime), MESOS (the UCR),
+            or DOCKER
+        health_check_protocol: can be MESOS_HTTP or HTTP
+        vip: either named or unnamed VIP to be applied to the host port
+        host_constraint: string representing a hostname for an agent that this
+            app should run on
+
+    Return:
+        (dict, str): 2-Tuple of app definition (dict) and app ID (string)
+    """
+    if network == marathon.Network.BRIDGE:
+        assert container_type == marathon.Container.DOCKER, \
+            'BRIDGE network mode only supported for DOCKER container type'
+        if container_port is None:
+            # provide a dummy value for the bridged container port if user is indifferent
+            container_port = 8080
+    else:
+        assert container_port is None or container_port == host_port, 'Cannot declare a different host and '\
+            'container port outside of BRIDGE network'
+        container_port = host_port
+    if network == marathon.Network.USER:
+        assert host_port != 0, 'Cannot auto-assign a port on USER network!'
+
+    test_uuid = uuid.uuid4().hex
+    app = copy.deepcopy({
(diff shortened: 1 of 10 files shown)


packages/dcos-integration-test/extra/util/delete_ec2_volume.py
```diff
@@ -55,7 +92,7 @@ def _get_current_aws_region():
         try:
             return requests.get('http://169.254.169.254/latest/meta-data/placement/availability-zone').text.strip()[:-1]
         except requests.RequestException as ex:
-            logging.warning("Can't get AWS region from instance metadata: {}".format(ex))
+            print("Can't get AWS region from instance metadata: {}".format(ex))
```
margaret: What's the convention around when to use `logging` vs `print`?

packages/dcos-integration-test/extra/util/delete_ec2_volume.py
```diff
@@ -1,15 +1,52 @@
...
 import botocore
 import requests
 import retrying
+from botocore.exceptions import ClientError, WaiterError
 
-from dcos_test_utils.helpers import retry_boto_rate_limits
+
+def retry_boto_rate_limits(boto_fn, wait=2, timeout=60 * 60):
+    """Decorator to make boto functions resilient to AWS rate limiting and throttling.
+    If one of these errors is encounterd, the function will sleep for a geometrically
```
margaret: spelling: `encounterd`

packages/dcos-integration-test/extra/util/delete_ec2_volume.py
```diff
@@ -1,15 +1,52 @@
...
 
 import boto3
 import botocore
 import requests
 import retrying
+from botocore.exceptions import ClientError, WaiterError
 
-from dcos_test_utils.helpers import retry_boto_rate_limits
+
+def retry_boto_rate_limits(boto_fn, wait=2, timeout=60 * 60):
```
margaret: Is the `retrying.retry` decorator too simple to handle this case?

margaret: Looks like you might be able to do something like this?
```python
def retry_on_rate_limiting(exception):
    if exception in [ClientError, Waitererror]:
        if isinstance(e, ClientError):
             error_code = e.response['Error']['Code']
        elif isinstance(e, WaiterError):
             error_code = e.last_response['Error']['Code']
        return error_code in ['Throttling', 'RequestLimitExceeded']
    return False

@retrying.retry(wait_exponential_multiplier=wait, wait_exponential_max=timeout, retry_on_exception=retry_on_rate_limiting)        
```

packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -1,5 +1,133 @@
...
 
+TEST_APP_NAME_FMT = 'integration-test-{}'
+
+# make the expanded config available at import time to allow determining
+# which tests should run before the test suite kicks off
 with open('/opt/mesosphere/etc/expanded.config.json', 'r') as f:
     expanded_config = json.load(f)
+
+
+def marathon_test_app(
```
margaret: imo function would be easier to read if you break it up at the top level conditional statements on the setting sections that are more than one or two lines. Something like

```python
def marathon_test_app(...):
    container_port = set_container_port(network)
    test_uuid = uuid.uuid4().hex
    app = ...
    app.update(set_healthchecks(...)
    app.update(set_container(...))
    app.update(set_networking(...))
    if host_constraint is not None:
        app['constraints'] = [['hostname', 'CLUSTER', host_constraint]]
   return app, test_uuid
```

packages/dcos-integration-test/extra/util/delete_ec2_volume.py
mellenburg: My personal thoughts are:
* stdout (aka print) should be reserved for functional information, like JSON output to be passed to another tool
* when dealing with context/debug information logging should be used whenever possible (i.e. in an environment where the logger will be configured)
* if the script is somewhat trivial and/or single-purposed print can be used

In this case, the script is called in isolation (either manually or by marathon) and only outputs this single message so it seems reasonable to use a print statement and avoid the somewhat verbose logging setup given that the output will just be time-stamped by marathon anyway in the test.

packages/dcos-integration-test/extra/util/delete_ec2_volume.py
mellenburg: Yes! I like that; will do

packages/dcos-integration-test/extra/test_helpers.py
mellenburg: Turns out this is much harder in practice for a few reasons:
* container and networking sections of the definition are split across multiples branches
* container and networking sections require almost all of the main function parameters anyway
* some parts of the definition are lists of dicts which means a top-level update will overwrite the entire list, not update the index elements

So, we shall leave untangling this logic for another day.

mellenburg:
@jgehrcke @margaret could I get an approval on this as well? :D (this is the upstream component of  https://github.com/mesosphere/dcos-enterprise/pull/1235 )

PR #1781, opened 2017-07-30
Configurable Download Command [WIP]

malnick:
## High Level Description
This PR makes the download command for `dcos-dowload.service` configurable for either `curl` (default) or `aws s3`. This PR adds a calc function for determining the command to use based on the URL protocol. If the protocol evaluates to `s3://`, it uses the `aws s3 cp` command. If it's anything else, it defaults to the standard `curl` command. 

This enables users who are on AWS to push and pull the bootstrap of a given configuration to S3. This in turn gives a highly available location of their configuration, which is useful if their clusters leverage autoscaling groups. Without a HA location for their bootstrap contents, ASG becomes less durable. 

### Questions & Caveats:
1. This currently doesn't pass a fully qualified path to `aws` CLI, how should this be determined? Should this be shipped in `/opt/mesosphere/bin` or should it be another parameter in the config.yaml for the user to configure? Or, should we do something like `$(which aws) s3 ...`? 

2. This implementation assumes the S3 bucket and prefix have completely open permissions (public) or that the EC2 instance profile allows access to this resource. It does not assume any local, explicit API key access.

## Checklist for all PR's

___

Reviewers should be:

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -135,9 +135,13 @@ def validate_ip_port_list(json_str: str):
 
 def calculate_bootstrap_download_command(bootstrap_tmp_dir, bootstrap_url, bootstrap_id):
     if 's3://' in bootstrap_url:
-        return 'aws s3 cp {}/bootstrap/{}.bootstrap.tar.xz /{}/bootstrap.tar.xz'.format(bootstrap_url, bootstrap_id, bootstrap_tmp_dir)
+        return 'aws s3 cp {}/bootstrap/{}.bootstrap.tar.xz /{}/bootstrap.tar.xz'.format(
+            bootstrap_url,
+            bootstrap_id,
+            bootstrap_tmp_dir)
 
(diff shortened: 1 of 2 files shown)


malnick:
@mesosphere-mergebot label Request For Comment

spahl:
I don't think we need to do some interpretation and try to use `aws` cli when seeing s3 but rather make it simpler.

Let the user specify the full download command in addition to the bootstrap url. That way is not only simpler code but also more flexible (yes even torrent:-P).

This will also need a new test of course.

malnick:
Sounds good to me. My first pass on this was exactly that, but after looking at the code, I realized the tmp dir is mandatory since the installer must expect that. If we let the user define the entire command, we need to make sure they don't shoot themselves in the foot with that caveat, and document the fact that they need to place the download stuff into the tmp dir where the installer expects it.

gen/calc.py
```diff
@@ -133,6 +133,17 @@ def validate_ip_port_list(json_str: str):
     validate_ipv4_addresses(ip_list)
 
 
+def calculate_bootstrap_download_command(bootstrap_tmp_dir, bootstrap_url, bootstrap_id):
+    if 's3://' in bootstrap_url:
```
vespian: `.startswith()`?

malnick:
I'm only able to see the code quality tests in TC after making an account. Is there public access to the other integration tests? https://www.dropbox.com/s/bjs64nocv4szja1/Screenshot%202017-07-31%2012.46.22.png?dl=0

PR #1827, opened 2017-08-15
Admin Router: Prevent forwarding dcos-* cookies to upstreams

vespian:
## High Level Description

This PR introduces removing `dcos-*` cookies for upstream requests.

For technical motivation - please check the linked Jira issue
For technical implementation details - please check the description of the function `util.clear_dcos_cookies` in `lib/util.lua` file

## Related Issues

  - https://jira.mesosphere.com/browse/DCOS-12015 `Admin Router: remove auth cookie after translating it into Authorization header`

## Related PRs:

EE PR: https://github.com/mesosphere/dcos-enterprise/pull/1298

### DIFF
--- a/packages/adminrouter/extra/src/lib/util.lua
+++ b/packages/adminrouter/extra/src/lib/util.lua
@@ -1,8 +1,64 @@
 local cjson_safe = require "cjson.safe"
+local cookiejar = require "resty.cookie"
 
 
 local util = {}
 
+
+function util.clear_dcos_cookies()
+    -- This function removes dcos-specific cookies from the request header.
+    --
+    -- This prevents forwarding these cookies upstream as they are an internal
+    -- implementation detail and services behind AR must only rely on
+    -- Authorization header.
+    --
+    -- Cookie filtering was implemented in Lua because:
+    -- * the `$cookie_<cookie_name>` variables are read-only
+    --   (http://nginx.org/en/docs/http/ngx_http_core_module.html#variables)
+    -- * reliable editing of Cookie header using Nginx requires using Ifs
+    --   (https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)
+    --   and the regexps themselves are not trivial. E.g.
+    --   ```
+    --   set $new_cookie $http_cookie;
+    --   if ($new_cookie ~ "(.*)(?:^|;)\s*dcos-acs-auth-cookie=[^;]+(.*)") {
+    --      set $new_cookie $1$2;
+    --   }
+    --   if ($new_cookie ~ "(.*)(?:^|;)\s*dcos-acs-info-cookie=[^;]+(.*)") {
+    --      set $new_cookie $1$2;
+    --   }
+    --   proxy_set_header Cookie "new_cookie";
+    --   ```
+    --
+    -- Besides that, the `resty.cookie` library can't remove cookies from the
+    -- request, but has pretty good cookie parser. So the idea is to use the
+    -- parser and re-assemble the header without the dcos-* cookies.
+
+    local filtered_cookies = {}
+    local cookie_obj = cookiejar:new()
+
+    local cookies, err = cookie_obj:get_all()
+    if err then
+        -- Cookie header is not present
+        return
+    end
+
+    for k, v in pairs(cookies) do
+        if k ~= "dcos-acs-auth-cookie" and k ~= "dcos-acs-info-cookie" then
+            filtered_cookies[#filtered_cookies+1] = k .. "=" .. v
+        end
+    end
+
+    -- Upstream request will inherit headers, so lets adjust the Cookie header
+    -- before sending it upstream:
+    -- https://github.com/openresty/lua-nginx-module/issues/437#issuecomment-65697745
+    if filtered_cookies == {} then
+        -- There were only DC/OS specific cookies
+        ngx.req.set_header("Cookie", nil)
+    else
+        -- Forward user's cookies as-is
+        ngx.req.set_header("Cookie", table.concat(filtered_cookies, "; "))
+    end
+end
 
 function util.get_stripped_first_line_from_file(path)
     -- Return nil when file is emtpy.
--- a/packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
+++ b/packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
@@ -4,6 +4,7 @@ import copy
 import logging
 import os
 from contextlib import contextmanager
+from http import cookies
 
 import requests
 
@@ -95,6 +96,67 @@ def generic_response_headers_verify_test(
     if assert_headers_absent is not None:
         for name in assert_headers_absent:
             header_is_absent(resp.headers.items(), name)
+
+
+def generic_upstream_cookies_verify_test(
+        ar,
+        headers,
+        path,
+        cookies_to_send=None,
+        assert_cookies_present=None,
+        assert_cookies_absent=None):
+    """Test if cookies that are passed to the upstream by AR are correct
+
+    Helper function meant to simplify writing multiple tests testing the
+    same thing for different endpoints.
+
+    Arguments:
+        ar: Admin Router object, an instance of runner.(ee|open).Nginx
+        headers (dict): headers dict that contains DC/OS authentication token
+            and cookies. The auth data it contains must be valid.
+        path (str): path for which request should be made
+        cookies_to_send (dict): dictionary containing all the cookies that should
+            be send in the request
+        assert_cookies_present (dict): cookies to test where key is the
+            asserted cookie name and value is expected value of the cookie
+        assert_cookies_absent (list or set): cookies that *MUST NOT* be present
+            in the upstream request
+    """
+    url = ar.make_url_from_path(path)
+    resp = requests.get(url,
+                        cookies=cookies_to_send,
+                        allow_redirects=False,
+                        headers=headers)
+
+    assert resp.status_code == 200
+
+    req_data = resp.json()
+
+    # Let's make sure that we got not more than one 'Cookie' header:
+    # https://tools.ietf.org/html/rfc6265#section-5.4
+    cookie_headers = []
+    for header in req_data['headers']:
+        if header[0] == 'Cookie':
+            cookie_headers.append(header)
+    assert len(cookie_headers) <= 1
+
+    if len(cookie_headers) == 1:
+        jar = cookies.SimpleCookie()
+        # It is a list containing a single tuple (`header name`, `header value`),
+        # we need the second element of it - the value of the header:
+        jar.load(cookie_headers[0][1])
+    else:
+        jar = {}
+
+    if assert_cookies_present is not None:
+        jar_cookies_dict = {x: jar[x].value for x in jar if x in assert_cookies_present}
+        # We only want to check the keys present in cookies_present_dict
+        assert jar_cookies_dict == assert_cookies_present
+
+    if assert_cookies_absent is not None:
+        jar_cookies_set = set(jar.keys())
+        cookies_absent_set = set(assert_cookies_absent)
+        assert jar_cookies_set.intersection(cookies_absent_set) == set()
 
 
 def generic_upstream_headers_verify_test(
--- a/packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
+++ b/packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
@@ -13,6 +13,7 @@ from generic_test_code.common import (
     generic_location_header_during_redirect_is_adjusted_test,
     generic_no_slash_redirect_test,
     generic_response_headers_verify_test,
+    generic_upstream_cookies_verify_test,
     generic_upstream_headers_verify_test,
 )
 
@@ -144,6 +145,9 @@ def _verify_are_upstream_req_headers_ok(t_config):
     assert 'jwt_should_be_forwarded' in t_config
     assert t_config['jwt_should_be_forwarded'] in [True, False, 'skip']
 
+    assert 'skip_authcookie_filtering_test' in t_config
+    assert t_config['skip_authcookie_filtering_test'] in [True, False, 'skip']
+
     assert 'test_paths' in t_config
     for p in t_config['test_paths']:
         assert p.startswith('/')
@@ -224,7 +228,9 @@ def _testdata_to_are_upstream_req_headers_ok_testdata(tests_config, node_type):
         h = x['tests']['are_upstream_req_headers_ok']
 
         for p in h['test_paths']:
-            e = (p, h['jwt_should_be_forwarded'])
+            e = (p,
+                 h['jwt_should_be_forwarded'],
+                 h['skip_authcookie_filtering_test'])
             res.append(e)
 
     return res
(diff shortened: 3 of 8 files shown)


vespian:

vespian:
@mhrabovcin @gpaul please have a look

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
```diff
@@ -204,6 +206,7 @@ endpoint_tests:
   - tests:
       are_upstream_req_headers_ok:
         enabled: true
+        skip_authcookie_filtering_test: false
```
mhrabovcin: Can we have this option `false` by default?

jgehrcke:
Thanks, Pawel.

> as the generic tests syntax has changed and the documentation needs to be updated

Can you please treat the docs change as an independent task and -- with that -- push this PR towards a reviewable state?

I wonder why we do not hook in directly into [this](https://github.com/dcos/dcos/blob/06b089ac266b100503b0b0c71c159b429789e02f/packages/adminrouter/extra/src/lib/auth/common.lua#L69) code path. Note that the goal is to get rid of specifically the `dcos-acs-auth-cookie`. We do not care too much about the info cookie.

If the cookie filter stays implemented outside of the auth module I wonder if we should still trigger the filter logic conditionally from within the code I have pointed towards so that we don't unnecessarily inspect the cookies twice. Was there no easy way with having just one cookie jar? Thanks for clarification, I think addressing these points helps reviewers understanding the choices you have made.

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
```diff
@@ -374,9 +393,22 @@ endpoint_tests:
   - tests:
       are_upstream_req_headers_ok:
         enabled: true
+        skip_authcookie_filtering_test: True
```
mhrabovcin: Should this be lowercase `true`?

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
```diff
@@ -620,6 +661,12 @@ endpoint_tests:
     type:
       - master
   - tests:
+      are_upstream_req_headers_ok:
+        enabled: true
+        skip_authcookie_filtering_test: True
```
mhrabovcin: Lowercase `true`?

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
```diff
@@ -413,23 +484,16 @@ def test_if_upstream_headers_are_correct(
...
-        generic_upstream_headers_verify_test(
+        _universal_test_if_upstream_headers_are_correct(
+            self,
             master_ar_process_perclass,
             valid_user_header,
             path,
-            assert_headers=headers_present,
-            assert_headers_absent=headers_absent,
+            jwt_forwarded_test,
+            skip_authcookie_filtering_test,
```
mhrabovcin: Other invocation uses named arg format: `skip_authcookie_filtering_test=skip_authcookie_filtering_test`

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
vespian: During the design phase of the syntax for generic tests YAML, I was considering having "silent defaults", but eventually decided against it and this is a general rule for all YAML-based tests. The reason is that we want/need people unfamiliar with AR repo to contribute. Being explicit about what is enabled and what is not is, I think, helpful as they do not need to browse through docs and/or wonder why their copy-pasted test does not work. 

In this particular case, I suppose, it is even better as it directly states the fact that auth-cookie passing test is performed unless explicitly disabled.

This is not written in stone, esp. as the number of generic tests increases and the amount of extra information grows. I do not think this is the moment though.

Happy to hear your thoughts on that.

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
vespian: Fixed - now all of them use positional args. Thx!

vespian:
> Can you please treat the docs change as an independent task and -- with that -- push this PR towards a reviewable state?

https://jira.mesosphere.com/browse/DCOS_OSS-1548

I scheduled it for this/next sprint

> I wonder why we do not hook in directly into this code path. 

Few reasons:
a) because this code is not always called. Please note that `util.clear_dcos_cookies`  is i.e. called here:
* https://github.com/dcos/dcos/pull/1827/files#diff-04b4f1568a453a99cae518ab0e07d0baR102
* https://github.com/dcos/dcos/pull/1827/files#diff-04b4f1568a453a99cae518ab0e07d0baR113
* https://github.com/mesosphere/dcos-enterprise/pull/1298/files#diff-c9d8d4dcd20678c47adb4a2804bf2e28R18
* https://github.com/mesosphere/dcos-enterprise/pull/1298/files#diff-c9d8d4dcd20678c47adb4a2804bf2e28R81
* https://github.com/mesosphere/dcos-enterprise/pull/1298/files#diff-c9d8d4dcd20678c47adb4a2804bf2e28R157

I suppose there will be more places like that in the future. We want to be thorough and never let the cookie through, hence a separate function.

b) because we should rethink how we circulate JWT token in the system, esp. considering the fact that we currently have no way to revoke it (apart from re-keying *all* JWTs, but this is a different thing). Lots of DC/OS services receive token either via an HTTP header or through a cookie. Compromising one of such service would mean compromising DC/OS security in general. 

This is a part of the bigger effort, that started during Admin Router refactoring and is tracked in     `DCOS-17924 Admin Router: JWT should not be forwarded to services that do not need`. 

One day I would really like to have AR pass JWT only to components that need it and nowhere else :)

> Note that the goal is to get rid of specifically the dcos-acs-auth-cookie. We do not care too much about the info cookie.

One of the goals of this PR is to hide `dcos-acs-auth-cookie` cookie from upstreams so that they can not start relying on it, even acidentally. The question is whether `dcos-acs-info-cookie` is an interface that we support and other services can rely upon. I think it is not (please correct me if I am wrong), hence the filtering. This is not written in stone, and as soon as there is a use case for this cookie, I will be more than happy to re-enable it, but let's cross this bridge.... :)

> If the cookie filter stays implemented outside of the auth module I wonder if we should still trigger the filter logic conditionally from within the code I have pointed towards so that we don't unnecessarily inspect the cookies twice. Was there no easy way to having just one cookie jar? 

I gave it a deeper thought. There are few reasons:
* the `Cookie` header parser is IMO pretty efficient: https://github.com/cloudflare/lua-resty-cookie/blob/master/lib/resty/cookie.lua#L62-L91 My *guess* is that it's O(n) but would have to formally analyse it to be sure
* in theory, we can share it but remember that we already need to support overlaying EE on top of Open which makes the code a bit "non-direct" already. I would like to avoid making it even more complex without strong technical reason (e.g. unacceptable increase of CPU usage) but at for now, considering that Lua code is JITed, I do not think it is worth it.
* fun fact: we would have been able to use `ngx.var.cookie_dcos_acs_auth_cookie` variable if it wasn't for the hyphens: http://lists.w3.org/Archives/Public/ietf-http-wg/2016JulSep/0015.html and https://github.com/openresty/lua-nginx-module/issues/19#issuecomment-111686154 My *guess* is that NGINX parser is much more effiicent than any Lua, JITed or not.

> Thanks for the clarification, I think addressing these points helps reviewers understanding the choices you have made.

nw, I hope that I managed to clarify all the points.

packages/adminrouter/extra/src/lib/util.lua
```diff
@@ -1,9 +1,65 @@
...
+
+    local cookies, err = cookie_obj:get_all()
+    if err then
+        -- Cookie header is not present
+        return
+    end
+
+    for k, v in pairs(cookies) do
+        if k ~= "dcos-acs-auth-cookie" and k ~= "dcos-acs-info-cookie" then
+            filtered_cookies[#filtered_cookies+1] = k .. "=" .. v
```
gpaul: wow lua

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
```diff
@@ -97,6 +98,67 @@ def generic_response_headers_verify_test(
...
+
+    Helper function meant to simplify writing multiple tests testing the
+    same thing for different endpoints.
+
+    Arguments:
+        ar: Admin Router object, an instance of runner.(ee|open).Nginx
+        headers (dict): headers dict that contains JWT and cookies. The auth
+            data it contains must be valid.
+        path (str): path for which request should be made
+        cookies_to_send (dic): dictionary containing all the cookies that should
```
gpaul: dict

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
```diff
@@ -97,6 +98,67 @@ def generic_response_headers_verify_test(
...
+            be send in the request
+        assert_cookies_present (dict): cookies to test where key is the
+            asserted cookie name and value is expected value of the cookie
+        assert_cookies_absent (dict): cookies that *MUST NOT* be present in the
```
gpaul: This actually expects a `list()` or `set()` at the moment, which is great, the docstring just needs to get updated.

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
```diff
@@ -97,6 +98,67 @@ def generic_response_headers_verify_test(
...
+    # https://tools.ietf.org/html/rfc6265#section-5.4
+    cookie_headers = []
+    for header in req_data['headers']:
+        if header[0] == 'Cookie':
+            cookie_headers.append(header)
+    assert len(cookie_headers) <= 1
+
+    if len(cookie_headers) == 1:
+        jar = cookies.SimpleCookie()
+        # It is a list containing a single tuple, we need the second element of
```
gpaul: :question: 
Nice comment, consider mentioning the tuple fields: "a single tuple `(name, value)`" if that is in fact what the tuple fields are

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
```diff
@@ -97,6 +98,67 @@ def generic_response_headers_verify_test(
...
+                        allow_redirects=False,
+                        headers=headers)
+
+    assert resp.status_code == 200
+
+    req_data = resp.json()
+
+    # Let's make sure that we got not more than one 'Cookie' header:
+    # https://tools.ietf.org/html/rfc6265#section-5.4
+    cookie_headers = []
```
gpaul: What do you think about using presence or `None` instead of  instead of using a list that has length 0 or 1?

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
```diff
@@ -348,9 +366,21 @@ endpoint_tests:
 ######### /system/v1/logs endpoint
   - tests:
       are_upstream_req_headers_ok:
+        skip_authcookie_filtering_test: True
```
gpaul: not sure if it matters but the false and true values don't have the same case

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
vespian: This list can have more than one element and this is an error. RFC forbids sending more than one `Cookie` header, but the tests just in case check for that before actually verifying cookies. If we switch to `None/Present`, we will miss the duplicated `Cookie` header scenario.

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
vespian:
```
# It is a list containing a single tuple (`header name`, `header value`),
# we need the second element of it - the value of the header:
```

packages/adminrouter/extra/src/test-harness/tests/test_generic.config.yml
vespian: uhhh... somehow I missed that, @mhrabovcin also pointed it out. I will grep the tests once more and try to fix all of them now.

jgehrcke:
> Few reasons:
> a) because this code is not always called. Please note that util.clear_dcos_cookies is i.e. called here:

(a) is concise and convinces me! I was not really able to parse the relation to my question from the remaining point(s?), but that is also not required anymore.

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/common.py
```diff
@@ -97,6 +98,67 @@ def generic_response_headers_verify_test(
...
+        assert_cookies_present=None,
+        assert_cookies_absent=None):
+    """Test if cookies that are passed to the upstream by AR are correct
+
+    Helper function meant to simplify writing multiple tests testing the
+    same thing for different endpoints.
+
+    Arguments:
+        ar: Admin Router object, an instance of runner.(ee|open).Nginx
+        headers (dict): headers dict that contains JWT and cookies. The auth
```
jgehrcke: Please replace "JWT" with what you actually mean. It's probably "the DC/OS authentication token".

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
```diff
@@ -305,6 +311,71 @@ def _testdata_to_redirect_testdata(tests_config, node_type):
...
+    if skip_authcookie_filtering_test:
+        return
+
+    filtered_cookies = ['dcos-acs-info-cookie', 'dcos-acs-auth-cookie']
+    jar = {}
+    # both dcos-* cookies should be removed by AR
+    # the dcos-acs-auth-cookie cookie is simply the DC/OS authentication token:
+    jar['dcos-acs-auth-cookie'] = valid_user_header['Authorization']
+    # the dcos-acs-info-cookie is base64-encoded data for dc/os UI,
+    # currently set to default user and password
```
jgehrcke: > currently set to default user and password

What does that mean, and where does this come from? :)

The info cookie is a cookie accessible by JavaScript code (as opposed to the auth cookie) and just contains the `uid` of the user that is currently logged in, so that the UI can display it.

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
```diff
@@ -305,6 +311,71 @@ def _testdata_to_redirect_testdata(tests_config, node_type):
...
+    jar['dcos-acs-info-cookie'] = (
+        'eyJ1aWQiOiAiYm9vdHN0cmFwdXNlciIsICJkZX'
+        'NjcmlwdGlvbiI6ICJCb290c3RyYXAgc3VwZXJ1c2VyIiwgImlzX3JlbW90ZSI6IGZ'
+        'hbHNlfQ==')
+
+    # First case - cookie contains ONLY dcos-specific cookies, `Cookie` header
+    # should not be send
```
jgehrcke: s/send/sent

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
vespian: > currently set to default user and password
Yep, this needs to be fixed :/

Will take care of it while working on https://jira.mesosphere.com/browse/DCOS_OSS-1572

packages/adminrouter/extra/src/test-harness/modules/generic_test_code/generalised_tests.py
vespian: see above comment.

cdaw:
Merged in Train 208, please close.

vespian:
@cdaw Train 208 is still open/unmerged :)

cdaw:
sorry about that ... i thought we merged it ...

PR #1818, opened 2017-08-10
Fix mount failure handling

lloesche:
## High Level Description

Fixes a bug where the loop would get interrupted by a failed mount command rendering the entire until loop useless. The idea is to keep retrying until the mount succeeds (or systemd times out the service unit and kills the script). Instead a single mount failure would currently abort the script (`set -e`).

No longer makes the script fail on SIGPIPE (reverts to bash default behaviour).

Moves dependency handling to the systemd unit instead of stopping/restarting services in the volume mount script itself.

PR #1836, opened 2017-08-16
Prevent dcos_vol_setup from formatting partitions while services are starting

scareything:
## High Level Description

This PR updates the dcos_vol_setup.service dependencies so that it starts as early as possible in the boot sequence. It also simplifies dcos_vol_setup.sh - the systemd manipulations and /var/log migration are no longer necessary now that dcos_vol_setup.service is running early.

Because dcos_vol_setup.service did not include DefaultDependencies=no, it was implicitly running with After=basic.target which is fairly late in the systemd targets. Many other services are started after basic.target - like dockerd and chrony - which in some cases were left in a bad state because file systems were being formatted while the services were initializing.

Coincidentally, dcos_vol_setup was failing prematurely because it was receiving SIGPIPE when trying to write to stdout/stderr when systemd-journald was restarted (https://bugs.freedesktop.org/show_bug.cgi?id=84923). systemd-journald was being restarted in two places:
1. dcos_vol_setup.sh - this PR removes all systemctl usage from dcos_vol_setup.sh
2. cloud-final.service - the userdata generated by `dcos_generate_config.sh --aws-cloudformation` includes a runcmd directive `systemctl restart systemd-journald`:
```#cloud-config
"runcmd":
- - |-
    systemctl
  - |-
    restart
  - |-
    systemd-journald.service
- - |-
...
```

## Related Issues

  - [DCOS_OSS-1043](https://jira.dcos.io/browse/DCOS_OSS-1043) dcos_vol_setup fails to set up /var/log - breaking cronyd and navstar

## Checklist for all PR's

  - [  ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This PR fixes an intermittent failure that happens early in the boot up of AWS instances and cannot be reproduced at will.
  - [ x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs): I recognize that Before, After, etc are discouraged, but I don't know of another way to ensure early execution of dcos_vol_setup.

___

Reviewers should be:

### DIFF
--- a/cloud_images/centos7/dcos_vol_setup.sh
+++ b/cloud_images/centos7/dcos_vol_setup.sh
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -o errexit -o nounset -o pipefail
 
+# Avoid getting killed due to https://bugs.freedesktop.org/show_bug.cgi?id=84923 if
+# systemd-journald is restarted while we are running.
+trap '' PIPE
+
 export device=${1:-}
 export mount_location=${2:-}
 
@@ -24,10 +28,28 @@ EXAMPLES:
 EOUSAGE
 }
 
+# Try to be resilient to SIGPIPE, incase anyone restarts journald while we are running.
+function noncritical {
+  set +e; ${*}; set -e
+}
+
+function checked_mount() {
+  local dev="$1"
+  local location="$2"
+  noncritical echo -n "Mounting: $dev to $location"
+  until grep "^$dev" /etc/mtab > /dev/null; do
+    noncritical sleep 1
+    noncritical echo -n .
+    # mount might think the device is already mounted, so accept nonzero exit status
+    mount "$location" || :
+  done
+  noncritical echo
+}
+
 for i in "$@"
 do
   case "$i" in                                      # Munging globals, beware
-    -h|--help)                usage                 ;;
+    -h|--help)                noncritical usage     ;;
     --)                       break                 ;;
     *)                        # unknown option      ;;
   esac
@@ -36,50 +58,40 @@ done
 function main {
   if [[ -z "$mount_location" || -z "$device" ]]
   then
-    usage
+    noncritical usage
     exit 1
   fi
 
-  echo -n "Waiting for $device to come online"
-  until test -b "$device"; do sleep 1; echo -n .; done
-  echo
+  noncritical echo -n "Waiting for $device to come online"
+  until test -b "$device"; do noncritical sleep 1; noncritical echo -n .; done
+  noncritical echo
   local formated
   mkfs.xfs -n ftype=1 $device > /dev/null 2>&1 && formated=true || formated=false
   if [ "$formated" = true ]
   then
-    echo "Setting up device mount"
+    noncritical echo "Setting up device mount"
     mkdir -p "$mount_location"
     fstab="$device $mount_location xfs defaults 0 2"
-    echo "Adding entry to fstab: $fstab"
+    noncritical echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
     if [ "$mount_location" = "/var/log" ]; then
-      echo "Preparing $device by migrating logs from $mount_location"
+      noncritical echo "Preparing $device by migrating logs from $mount_location"
       mkdir -p /var/log-prep
-      mount $device /var/log-prep
+      mount "$device" /var/log-prep
       mkdir -p /var/log-prep/journal
-      systemctl is-active chronyd > /dev/null && systemctl stop chronyd || :
-      systemctl is-active tuned > /dev/null && systemctl stop tuned || :
-      # rsyslog shouldn't be active but in case it is stop it as well
-      systemctl is-active rsyslog > /dev/null && systemctl stop rsyslog || :
       cp -a /var/log/. /var/log-prep/
       umount /var/log-prep
       rmdir /var/log-prep
       rm -rf /var/log
       mkdir -p /var/log
-      echo -n "Mounting: $device to $mount_location"
-      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
-      echo
-      systemd-tmpfiles --create --prefix /var/log/journal || :
-      systemctl restart systemd-journald || :
-      systemctl is-enabled tuned > /dev/null && systemctl start tuned || :
-      systemctl is-enabled chronyd > /dev/null && systemctl start chronyd || :
+      checked_mount "$device" "$mount_location"
+      systemd-tmpfiles --create --prefix /var/log/journal
+      systemctl kill --signal=SIGUSR1 systemd-journald
     else
-      echo -n "Mounting: $device to $mount_location"
-      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
-      echo
+      checked_mount "$device" "$mount_location"
     fi
   else
-    echo "Device $device contains a filesystem: no action taken"
+    noncritical echo "Device $device contains a filesystem: no action taken"
     exit
   fi
 }
--- a/cloud_images/centos7/install_prereqs.sh
+++ b/cloud_images/centos7/install_prereqs.sh
@@ -26,6 +26,8 @@ echo ">>> Set up filesystem mounts"
 cat << 'EOF' > /etc/systemd/system/dcos_vol_setup.service
 [Unit]
 Description=Initial setup of volume mounts
+DefaultDependencies=no
+Before=local-fs-pre.target
 
 [Service]
 Type=oneshot
@@ -35,7 +37,7 @@ ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdg /dcos/volume0
 ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdh /var/log
 
 [Install]
-WantedBy=local-fs.target
+RequiredBy=local-fs-pre.target
 EOF
 systemctl enable dcos_vol_setup
 


scareything:
`systemd-analyze plot` charts from before and after the changes in this PR were applied. Note the active period of dcos_vol_setup.service in relation to other systemd services.

systemd-analyze-before

systemd-analyze-after

scareything:
@lloesche, please review. Do you know who else I might add to this review? Thanks.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -52,32 +52,8 @@ function main {
...
-      systemctl is-active rsyslog > /dev/null && systemctl stop rsyslog || :
-      cp -a /var/log/. /var/log-prep/
-      umount /var/log-prep
-      rmdir /var/log-prep
-      rm -rf /var/log
-      mkdir -p /var/log
-      echo -n "Mounting: $device to $mount_location"
-      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
-      echo
-      systemd-tmpfiles --create --prefix /var/log/journal || :
```
lloesche: Without this the journal won't be persisted between reboots.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -52,32 +52,8 @@ function main {
...
-      systemctl restart systemd-journald || :
-      systemctl is-enabled tuned > /dev/null && systemctl start tuned || :
-      systemctl is-enabled chronyd > /dev/null && systemctl start chronyd || :
-    else
-      echo -n "Mounting: $device to $mount_location"
-      until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location"; done
-      echo
-    fi
+    echo -n "Mounting: $device to $mount_location"
+    mount "$mount_location"
```
lloesche: Without the until loop mounts will fail sometimes. We had systems where EBS volumes couldn't be mounted directly after being formatted, a couple seconds later the device was available and mountable. That's why the loop was added in the first place.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -35,10 +37,20 @@ ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdg /dcos/volume0
...
 
 [Install]
-WantedBy=local-fs.target
+RequiredBy=local-fs-pre.target
 EOF
 systemctl enable dcos_vol_setup
 
+echo ">>> Configure tmpfiles.d/chrony"
+cat > /etc/tmpfiles.d/chrony.conf <<EOF
+d /var/log/chrony 0755 chrony chrony -
```
lloesche: Of the top of my head doesn't /var/log/chrony have the setgid bit set?

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -52,32 +52,8 @@ function main {
     fstab="$device $mount_location xfs defaults 0 2"
     echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
-    if [ "$mount_location" = "/var/log" ]; then
-      echo "Preparing $device by migrating logs from $mount_location"
-      mkdir -p /var/log-prep
-      mount $device /var/log-prep
-      mkdir -p /var/log-prep/journal
```
lloesche: Without this the journal won't be persisted between reboots.

lloesche: Also needs the `systemd-tmpfiles` command.

lloesche:
Changes from https://github.com/dcos/dcos/pull/1818 should be merged into this PR. Specifically commit c9fc3769bd6457f6dcd7e70f61c1ce167eb0e577.

cloud_images/centos7/install_prereqs.sh
scareything: From one of our older nodes, before this change:
```
[centos@ip-10-0-3-81 ~]$ ls -ld /var/log/chrony/
drwxr-xr-x. 2 chrony chrony 6 Dec  6  2016 /var/log/chrony/
```

cloud_images/centos7/dcos_vol_setup.sh
scareything: I haven't seen that before, but I'll take your word for it. I'm currently testing a change that incorporates your comment about preserving the journal between reboots. I'll restore the until loop in that push.

scareything:
I think we would need more than `-o pipefail` if dcos_vol_setup is not running well in front of other services that use the filesystems it is formatting.

My understanding of `-o pipefail` is that it changes the way the shell sets the exit status for a pipeline: without pipefail set (the default), ${?} will be the exit status of the last command in the pipeline. With pipefail set, ${?} will be the exit status of the first process in the pipeline with a non-zero exit status.

The SIGPIPE that I observed by running dcos_vol_setup with trace would happen on single-command lines, like 'echo something' when writing to stdout, and would not be affected by `-o pipefail`

In either case, the random dcos_vol_setup deaths are prevented by not bouncing systemd-journald while dcos_vol_setup is active. This PR achieves that partly by removing the restart in the script and partly by running dcos_vol_setup before the cloud-init runcmd executes (in cloud-final.service).

lloesche:
Sorry you're right. What I intended was to `trap '' PIPE` so that when the script loses it's stdout reader it won't get killed. `pipefail` has nothing to do with that.

Like I mentioned we can't get rid of the journald restart because we're creating the /var/log/journal directory which makes it so that the journal is being persisted. Otherwise all previous logs will be gone after a reboot.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o errexit -o nounset -o pipefail
+set -o errexit -o nounset
```
lloesche: As per your comment this is unnecessary, what I intended here was a `trap '' PIPE` instead.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -53,7 +53,12 @@ function main {
     echo "Adding entry to fstab: $fstab"
     echo "$fstab" >> /etc/fstab
     echo -n "Mounting: $device to $mount_location"
-    mount "$mount_location"
+    until grep ^$device /etc/mtab > /dev/null; do sleep 1; echo -n .; mount "$mount_location" || :; done
+    if [ "$mount_location" = "/var/log" ]; then
+      echo "Recreating /var/log/journal"
+      mkdir /var/log/journal
+      systemd-tmpfiles --create --prefix /var/log/journal
+    fi
```
lloesche: We do need to restart systemd-journald to pick up the directory creation at this point. This shouldn't be an issue anymore once we trap SIGPIPE above.

scareything: The journal was preserved when I tested this change, but perhaps it was because journald was eventually restarted by cloud-final (long after docs-vol-setup completed).

I'm trying a journald restart from within the script with a handler installed. Stand by.

cloud_images/centos7/install_prereqs.sh
lloesche: Right, I just checked, it's the journal directory that has setgid bit set. But that's taken care of by `systemd-tmpfiles`.

lloesche:
Thinking about it, we would now mount /var/log into a non-empty directory. It technically works but feels a bit dirty.

I could see users running into issues/it causing confusion in situations where for some reason the EBS is not present on the system, the mount fails but nobody notices and services start writing to the root device.

@lingmann what's your thoughts on this? Also could you be second reviewer please as you wrote the initial version of this script.

cloud_images/centos7/dcos_vol_setup.sh
lloesche: So the test I did on my local system just now (don't have access to the real thing atm.) was:

```
micromi:~ lukas$ time bash -c "echo foo; sleep 5" | :

real	0m0.004s
user	0m0.002s
sys	0m0.002s
micromi:~ lukas$ time bash -c "trap '' PIPE; echo foo; sleep 5" | :
bash: line 0: echo: write error: Broken pipe

real	0m5.013s
user	0m0.003s
sys	0m0.003s
```

The first one returns immediately, the second executes the `sleep`.

cloud_images/centos7/install_prereqs.sh
lloesche: Since we're leaving the `if [ "$mount_location" = "/var/log" ]` case in the script we might as well create the directory directly in there. Removing the dependency to `systemd-tmpfiles`.

cloud_images/centos7/dcos_vol_setup.sh
scareything: I tried a similar test while waiting for an AMI build, and I think we'll need to remove `-e` from the shell options:

```
$ bash -c "trap 'echo OUCH' PIPE; while true; do sleep 10; echo here; done" &
[1] 31715
$ kill -SIGPIPE %1
OUCH
here
here
```

But with `-e`:
```
$ bash -e -c "trap 'echo OUCH' PIPE; while true; do sleep 10; echo here; done" &
[1] 31733
$ kill -SIGPIPE %1
OUCH
[1]+  Exit 141                bash -e -c "trap 'echo OUCH' PIPE; while true; do sleep 10; echo here; done"
```

lloesche: Good catch! I'm a bit hesitant to just remove `-e` though.
Maybe we could wrap the echo in a function so that errexit is turned off while printing to stdout. But then what if it happens during execution of one of the other commands that output something.

I'll have to think a bit more about it with a refreshed mind tomorrow (almost midnight here).

lloesche: So I talked with some colleagues about the issue. We think two things would help the situation:
a) put an explicit `Before=dcos-download.service` so that we can gurantee partitioning happens before our script.

b) create a function ala
```
function noncritical {
  set +e; $*; set -e
}
```
And wrap our echo and sleeps in it like `noncritical echo -n .` and `noncritical sleep 1`.

b) would just make the script a more resilient against other stray services restarting journald for whatever reason. It shouldn't happen in our controlled AMI environment but once users start customising the AMI it might happen again.
For critical commands like `mkfs` or `mount` we still would want the script to abort and fail if the command gets interrupted.

What do you think?

lloesche: Btw. something I just also ran into is https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IgnoreSIGPIPE=
which I'm wondering why it doesn't apply in our case. I did a test on Fedora 25 and there the issue is non-existent, SIGPIPE is being ignored by default. On CentOS 7 however even explicitly setting IgnoreSIGPIPE=true doesn't do anything. Though looking at the `systemd.exec` manpage the option is documented in that version of systemd.

scareything: I also tried IgnoreSIGPIPE, and was disappointed as you are :(.

Regarding the ordering guarantee (before dcos-download), we already have that. `local-fs-pre.target` is the first target that systemd starts (https://www.freedesktop.org/software/systemd/man/bootup.html). dcos-download (and the other dcos services) are effectively started from cloud-final.service (when it runs /var/lib/cloud/instance/scripts/runcmd), which runs in `multi-user.target` (well after `local-fs-pre.target`). There would also be some ambiguity using the dcos-* services or targets as dependencies, since many of them don't exist until dcos-setup.service runs.

The critical section idea is interesting. Another idea that I would like to explore is sending the output somewhere other than the journal so it doesn't matter if/when journald restarts. Ideally this could be done with `StandardOutput` and a tty, but a straightforward option would be to simply set the script's output descriptors with `exec`.

I'm going to try this out, and restore some of the /var/log logic as we've discussed here. Let me know what you think in the meantime.

lloesche: I previously did tests with `Before=` and any non-existing units are being ignored. Remember though that dcos_vol_setup.service is being executed at every bootup, not just on first boot. In most cases the partitions will exist after first boot and most of the code won't ever be executed again. But it's perceivable that some users will destroy an EBS and reattach it to a running cluster. In that case the dcos_vol_setup would be run again to initialize the partition and dcos-setup.service would exist.

Re: sending output somewhere else. I experimented with a simple subshell that accepts the data on stdin and ignores output errors. E.g.
```
bash -c "set -e; trap '' PIPE; echo foo | (cat || :); sleep 3" | :
```
Here the echo output is being piped into the subshell. The cat get's interrupted but the signal is being ignored and because the failed cat results in a noop the script won't get aborted and the following sleep is executed.

scareything: For all of these reasons, I think it's best to keep `Before=local-fs-pre.target`, and no other dependencies. This target is created by the CentOS installation, it is always present, and it is always the first target that systemd activates (at least the first that we can hook into), which is exactly what we want when formatting partitions. If you haven't already looked at them, see the systems-analyze graphs for an illustration of when dcos_vol_setup.service is active with the dependencies as they are in this request.

So to me it seems that adding additional targets or services to the Before directive would only confuse matters for anyone who ends up looking at this systemd unit in the future. :)

Also, it is very important to be aware that `Before=local-fs-pre.target` _must_ be accompanied by `DefaultDependencies=no`; without this, systemd implicitly assigns `After=basic.target` which is much later in the sequence than we want to be.

lloesche: Sounds good to me! I added a summary of next code changes below. Let's add those and then see how we like the resulting script. Once we're satisfied I'll spin up a large cluster with those changes (default CI is only using a handful of nodes).

scareything:
My latest commit incorporates everything that we discussed.

- I did omit the explicit directory creation for chrony and audit, since these directories are created by the yum package. Now that we are migrating the original /var/log contents again, those directories and any others that were there are left intact.
- I will admit that I'm not especially fond of the `noncritical` addition, although I do appreciate that it adds resiliency for activity that may be out of our control. Before this PR there were two places where journald was restarted, both of which we are now insulated from:
  1. Within dcos_vol_setup. The latest commit uses `systemctl kill --signal SIGUSR1` instead to achieve journal persistence.
  2. From cloud-final.service (via userdata, assuming the CF templates created by --aws-cloudformation). The updated systemd dependency (on local-fs-pre) means that dcos_vol_setup is finished before cloud-final.service is activated.

  Any tactics above and beyond these two changes are only there to protect against journald restarts that subsequent image layers might perform, which take place during local-fs-pre.target. I'm just not sure if the additional protection is worth the layer of obscurity and whatever weight may be imposed by toggling `-e` so frequently.

- I discovered a way to more easily reproduce this issue: the SIGPIPE will only happen when the script is writing to stdout/stderr, so make it write often! The script/service failed for me 100% of the time if I added this line immediately after the systemctl restart:
  ```
  for i in {1..100}; do echo "I'm speaking up. what are you going to do about it?"; done
  ```
  And yes, the service did not fail when I replaced the systemctl restart with a SIGUSR1 poke.

So, I would appreciate if you would reconsider the noncritical change, but I'm generally happy with where this request is now. Let me know how your testing goes.

lingmann:
Thanks for the PR @scareything. I likely don't have time to review this week, but would like to expand the net and see if there are some other people that could take a look... @spahl @klueska @amitaekbote @BenWhitehead @mellenburg are all good candidates for a review. Any thoughts ya'll?

lingmann:
@mesosphere-mergebot bump-ee

lingmann:
@scareything can you rebase on master please?

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set -o errexit -o nounset -o pipefail
 
+# Avoid getting killed due to https://bugs.freedesktop.org/show_bug.cgi?id=84923 if
+# systemd-journald is restarted while we are running.
+trap 'echo dodged SIGPIPE' PIPE
```
lloesche: If we get a SIGPIPE it means the reader is gone. Adding the `echo` instead of just `''` shouldn't give any extra output, just another SIGPIPE.

lloesche:
@mesosphere-mergebot bump-ee

lloesche:

scareything:
Hello @lingmann, could I ask that you give this a look this week if you have time? Thanks!

scareything:
Hello to anyone who is watching this PR, I'm hoping that one more approved reviewer can take a look at the changes to these two files.

@lingmann @spahl @klueska @amitaekbote @BenWhitehead @mellenburg

This is an important fix for my team, it causes us a lot of plain when we stand up or scale up DC/OS clusters.

I suspect a fair number of other DC/OS users may not be aware that they are affected by it. For example, when docker fails to start because of this issue it prevents the mesos agent from starting and the node never joins the cluster (even in an unhealthy state), so you end up paying for EC2 instances that are not joined to the cluster. Without a comprehensive monitoring solution it os very easy for this to go unnoticed.

Thanks for taking a look when you get a chance.

lloesche:
@scareything sorry this PR is taking longer than usual. We're currently all hands on deck releasing DC/OS 1.10.0. Which isn't an excuse for our lack of status updates. That said our Director of Test and Release Engineering let me know last week already that this PR will be included in 1.10.1. There's been quite a bit of activity around it in our Slack. Much more than is visible in the public Github. Unfortunately we're not exposing that activity very well.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -26,6 +26,8 @@ echo ">>> Set up filesystem mounts"
 cat << 'EOF' > /etc/systemd/system/dcos_vol_setup.service
 [Unit]
 Description=Initial setup of volume mounts
+DefaultDependencies=no
```
lingmann: A little description on why this setting is necessary would be helpful here. Per [systemd docs](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) under DefaultDependencies they say:
> Generally, only services involved with early boot or late shutdown should set this option to false. It is highly recommended to leave this option enabled for the majority of common units. If set to false, this option does not disable all implicit dependencies, just non-essential ones.

scareything: dcos_vol_setup is formatting partitions that will be used by services that are to be started by systemd. It is better for this script to run before other systemd services are started rather than after.

Setting `DefaultDependencies=no` removes an implicit dependency `After=sysinit.target` (https://www.freedesktop.org/software/systemd/man/systemd.target.html):

> Service units, unless they set DefaultDependencies=no, automatically get a dependency on sysinit.target.

https://www.freedesktop.org/software/systemd/man/bootup.html shows where sysinit.target is activated in relation to other predefined targets and services.

cloud_images/centos7/install_prereqs.sh
```diff
@@ -35,7 +37,7 @@ ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdg /dcos/volume0
 ExecStart=/usr/local/sbin/dcos_vol_setup.sh /dev/xvdh /var/log
 
 [Install]
-WantedBy=local-fs.target
+RequiredBy=local-fs-pre.target
```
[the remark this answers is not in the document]
scareything: I know you're asking @lloesche, but I feel the need to emphasize that this change is one of the three critical changed lines in this PR (the others are the DefaultDependencies=no, and sending journald a sighup instead of restarting it).

1. Before this PR, the systemd unit was attempting to pull in local-fs.target with `WantedBy`. Coincidentally that was not working correctly because the default implicit dependencies were not disabled (see attached systemd analyze graphs). We could have left this with WantedBy, but we probably don't want the formatting to happen if local-fs.target fails for any reason, so RequiredBy seemed to make more sense.
2. local-fs-pre.target is an earlier hook than local-fs.target in the systemd sequence. Also, local-fs.target is intended for fsck'ing volumes, so we really do want to run before local-fs (https://www.freedesktop.org/software/systemd/man/bootup.html). And  since we are formatting partitions that are going to be used by other services, it is best to do this as early as possible.

cloud_images/centos7/dcos_vol_setup.sh
```diff
@@ -24,10 +28,28 @@ EXAMPLES:
...
+  local dev="$1"
+  local location="$2"
+  noncritical echo -n "Mounting: $dev to $location"
+  until grep "^$dev" /etc/mtab > /dev/null; do
+    noncritical sleep 1
+    noncritical echo -n .
+    # mount might think the device is already mounted, so accept nonzero exit status
+    mount "$location" || :
+  done
+  noncritical echo
```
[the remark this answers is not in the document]
scareything: This echo is here to insert a newline after the previous `echo -n`. It could fail like any other command in this script that emits to stdout after journald is killed/restarted.

I'll just point out that I think all of the changes in this script that involve `noncritical` are only going to matter if someone recreates the scenario where journald is restarted while dcos_vol_setup is running. The critical changes in this PR avoid that scenario.

lloesche:
@lingmann Yes, definitely.

PR #1857, opened 2017-08-18
dcos-mesos-agent: prefer system libraries for exec'd processes

sethp-nr:
## High Level Description

Global modification of the linker's path is confusing, and can result in non-DCOS
binaries being linked against DCOS libraries. This has caused breakage in at
least one instance where a system-provided `curl` linked against an interface-
incompatible older `libcurl` provided by DCOS, causing MESOS_HTTP healthchecks
to fail.

This commit mirrors the path started in dcos/dcos#707, but with the addition of
rpath in 27339e3 that change is no longer necessary. Further, we opt to preserve
the SASL_PATH override, even though it suffers similar concerns, because libsasl2
would require special mesos-only configuration at compilation time in order to
provide for mesos' out-of-the-box authentication, and that library is not
currently part of the mesos build flow.

Fixes DCOS_OSS-1302.

## Related Issues

  - [DCOS_OSS-1302](https://jira.mesosphere.com/browse/DCOS_OSS-1302) `MESOS_HTTP` health-checks are not working
  - dcos/dcos#707  

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: I'm unsure how to write a test case for this that doesn't require a fully running system – if someone could provide a pointer or two, I'd be glad to do so!

cc @rukletsov @bmahler 
h/t to @rukletsov for pointing me to the right configuration file

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -473,7 +473,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "SASL_PATH": "/opt/mesosphere/lib/sasl2",
         "LIBPROCESS_NUM_WORKER_THREADS": "8"
       }
   - path: /etc/dcos-log.env
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -473,7 +473,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "LD_LIBRARY_PATH": "/opt/mesosphere/lib",
         "SASL_PATH": "/opt/mesosphere/lib/sasl2",
         "LIBPROCESS_NUM_WORKER_THREADS": "8"
       }


rukletsov:

rukletsov:
@sethp-nr looks like I can't officially give a ship it here. The change looks good to me. @lingmann Jeremy, could you please have a look?

lingmann:
@mesosphere-mergebot bump-ee

lingmann:
@sethp-nr can you rebase on master please?

sethp-nr:
@lingmann I've merged master – is that good enough or should I pull this down locally and rebase?

lingmann:
@sethp-nr the merge update is fine, thank you!

lingmann:
@mesosphere-mergebot bump-ee

sethp-nr:
As mentioned on the JIRA, the reason for this change is that the current configuration of setting `LD_LIBRARY_PATH` but leaving `PATH` as the default breaks curl (and consequently `MESOS_HTTP` healtchecks) because the dcos vendor'd version, `7.48.0`, apparently presents a different interface than at least `7.51.3`. So any attempt to use the newer `curl` "frontend" binary dynamically linked against the older `libcurl` presents the error: `curl: (48) An unknown option was passed in to libcurl` (option, here, seems to indicate a code-level configuration that can't be controlled by the user, not command line flag).

I'm sensitive to the concern around increasing the maintenance scope, but from my perspective we must choose one of the following:

1. Modify the `PATH` to use the vendor'd curl
2. Modify the `LD_LIBRARY_PATH` to avoid the vendor'd `libcurl` and always use the system libraries from the executor
3. Modify MESOS_HTTP healthchecks to avoid using the `PATH` (or possibly the `curl` external binary) entirely, but this option leaves open the possibility of issues in the future with library compatibility between the host system and `/opt/mesosphere/lib`.

spahl:
This is quite a big scope change though. Suddenly you can use python (with tons of libraries), java and much more by default. Sure it was still accessible, but you would have had to explicitely opt in.This might cause some more support issues with apps using them without knowing, breaking after upgrades.

"
3. Modify MESOS_HTTP healthchecks to avoid using the PATH (or possibly the curl external binary) entirely, but this option leaves open the possibility of issues in the future with library compatibility between the host system and /opt/mesosphere/lib.
"

This seems safer.

jgehrcke:
(another option, just as a brain dump, would be to create another directory with a limited set of executables in DC/OS for exactly this purpose, and then still use `PATH`)

sethp-nr:
@spahl If the goal is to protect the executor process from DC/OS packages, then it sounds like the right change here is to delete the `LD_LIBRARY_PATH` override.

As it stands, right now processes will be running with system binaries (per `PATH`) but linking against the DC/OS shared objects. We can adjust either side of that, and it's quite possible that either change will be breaking, but the current situation of linking system binaries against DC/OS libraries seems untenable.

Option 3 is definitely something that's possible to pursue, but it's a different set of work and would mean implementing DC/OS-specific code in Mesos – if that's a PR you feel you can land, I'm happy to provide as much information as I can to help you do so.

sethp-nr:
@jgehrcke An explicit path would have dodged this issue were it present here: https://github.com/apache/mesos/blob/6b3e4b72a0b8e67359ce2bbcd890c4e6c96289f4/src/checks/checker_process.cpp#L88 , but my understanding is that due to the diversity of deployment environments it's not guaranteed that a `curl` binary will be available at `/opt/mesosphere/bin/curl`?

rukletsov:
I'm strongly against option 3. Mesos has a dependency on `curl` and I'm convinced Mesos can assume that there is version mismatch between `libcurl` and `curl` in its environment. Using the default `curl` seems to be the best option: an operator or external tooling can decide what this default is.

DC/OS should set up consistent environment for Mesos components: option 1 or 2.

rukletsov:
FWIW, `/opt/mesosphere/environment` (used by, e.g., mesos master and agent) sets `PATH` an `LD_LIBRARY_PATH` consistently. Why should executor environment do it differently?

jgehrcke:
> I'm convinced Mesos can assume that there is version mismatch between libcurl and curl in its environment.

Agree (with an imaginary additional "no").

There are two more options.

Option *4*, as mentioned above, could be setting up dedicated directories in dcos.

Option *5* is to add a configuration parameter to Mesos that allows for pointing to a specific curl executable via a file system path. (This kind of configuration interface is not uncommon in configuration interfaces for programs that need to call other programs in a controlled fashion -- but it's probably not what's done in Mesos, yes?)

rukletsov:
Option 4 is probably ok.

Option 5: you're right, I'm reluctant to introduce a flag that specifies which curl to use.

sethp-nr:
I'm still not quite clear what intention is reflected by separating the library and executable paths? Under what scenarios is it the desired behavior to use the system binaries linked against DC/OS's libraries?

sethp-nr:
@jgehrcke @spahl I've updated the PR to remove the `LD_LIBRARY_PATH` override in lieu of modifying the `PATH` – for the `MESOS_HTTP` healthchecks this assumes some suitable version o f `curl` will be present on the host system, but those checks currently require both that 1) there is a `curl` on the system path, and 2) that `curl` is link-compatible with whatever version DC/OS ships.

Dropping requirement #2 seems to strictly increase the set of systems on which `MESOS_HTTP` checks function correctly.

cdaw:
This needs to also be tagged for 1.9.5.

rukletsov:
FYI: https://github.com/dcos/dcos/pull/707

sethp-nr:
@rukletsov excellent find! Looks like this work is superseded by that change – I'll close this one out.

sethp-nr:
Reopening as the source repo in #707 seems defunct

sethp-nr:
Pulled in c13fe6e, but it looks like ever since 27339e3 the rpath has been set on mesos binaries so that portion of the change was unnecessary.

adam-mesos:
> This needs to also be tagged for 1.9.5.

Due to reasons I expressed in #707 , we cannot backport this to 1.9.x, and probably not even to 1.10.x. Let's leave this based on master/1.11 and see what kind of backlash we get from that before we consider backporting further.
cc: @judithpatudith for community package impact

rukletsov:
Agree with @adam-mesos, no back porting for now.

gen/dcos-config.yaml
```diff
@@ -473,8 +473,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "LD_LIBRARY_PATH": "/opt/mesosphere/lib",
-        "SASL_PATH": "/opt/mesosphere/lib/sasl2",
```
tillt: It appears that our bundle is lacking the SASL plugins - this means that IFF Mesos was trying to do SASL based authentication (e.g. CRAM-MD5), it would fail as it was unable to locate the needed plugin within the bundle and due to this line also on the DCOS system itself.
However, we commonly do not use SASL at all within DCOS, hence this may be entirely fine. We just need to be aware.

sethp-nr: This change I inherited from #707 without context – from what I am able to gather, though, it seems like we it will break two reasons:

1. [SASL is the default authentication strategy for Mesos](http://mesos.apache.org/documentation/latest/authentication/)
2. Unlike the linker's support for library-scoped pathing (i.e. `rpath` and friends), Cyrus SASL appears to look only in the single directory identified by [`SASL_PATH`](https://www.cyrusimap.org/sasl/sasl/reference/manpages/library/sasl_client_init.html?highlight=sasl_path), falling back to `/usr/lib/sasl2` if that's not set.

The alternative would be to keep this change and add a build phase to DC/OS that packages libsasl2 directly, as it looks like the configure script has `--with-plugindir` and `--with-configdir` that would allow a DC/OS-packaged libsasl2 to do the right thing without setting the environment variable.

I'm looking into building a DC/OS vendored sasl2, but I wouldn't mind a pointer or two if they're handy.

tillt: re 1: Yes it is.
re 2: Correct, these mechanism plugins are dynamically loaded by libsasl itself and hence are not a hard dependency. They are only used when the specific mechanisms (e.g. CRAM-MD5) are used/requested

Let's take a step back here :) .... (ow how I commonly dislike such statements) ....

SASL is being used by Mesos for two things, framework- and agent-authentication. In the open-source world, currently only CRAM-MD5 is being used as a mechanism.

The Mesos agent is a core part of DC/OS and hence it has to be relying on the components installed within DC/OS - think of Mesos like the DC/OS kernel. `libsasl` while being a dependency of Mesos itself however is not considered a core component but only a dependency.

Frameworks on the other hand are not a core part of DC/OS, they are what I would describe as user-land software. Frameworks should not rely on any dependency other than the kernel (libmesos) within DC/OS. 

If a Framework now wants to use SASL for authenticating against the master, this framework should bring its own set of SASL plugins. While this may sound a bit picky or even non-sensical, this is what we commonly strife for -- one core argument here is that we don't want our user-land software to make assumptions about non kernel components as that would restrict us a lot in upgrading kernel dependencies. 

In the end, SASL as being a dependency of Mesos itself needs to work for both agents and masters without further user intervention. As such, we need `libsasl` to be able to locate its plugins and hence removing `SASL_PATH` is not a good idea unless of cause we build `libsasl` the way you are suggesting (`--with-plugindir`).

This still means that framework authors should not rely on any functioning SASL components on the DC/OS platform itself -- they should bundle their own libsasl + plugins.

In the end, this means I am suggesting to:
- leave `SASL_PATH` in or compile SASL the way you are suggesting
- tell our users that they should bundle `libsasl` and at least the CRAM-MD5 SASL mechanism plugin as part of their Mesos frameworks, if they intend to rely on this authentication variant

karya0: Regarding "framework authors should not rely on any functioning SASL ..." -- does it make sense to bundle SASL libraries in the libmesos-bundle and have them use it? Would it be reasonable?

sethp-nr: many thanks @tillt for the additional information! As you suggest, I've backed out the SASL_PATH change for now; if I discover a way to vendor in libsasl in my copious free time I'll be sure to open a new PR.

tillt: @karya0 it would make sense in my opinion, yes. There will likely be very very few users of that functionality but given that Mesos itself offers it and DC/OS does not prevent it, we should enable it via the libmesos-bundle.

tillt:
@mesosphere-mergebot bump-ee

rukletsov:
@mesosphere-mergebot bump-ee

tillt:
First we need to make sure our framework libmesos bundle contains the CRAM-MD5 mechanism plugins for libsasl. This is mandatory before we can change anything on the library path via this PR.
THEN we need to make sure our executors stop overriding `SASL_PATH` as we want the workload (e.g. a framework) to stop using the hosted SASL mechanism plugins.

So for now, lets hold this until we fixed our bundles and then remove `SASL_PATH` again .... sorry for this back-and-forth.

tillt:
We have updated our libmesos bundle build scripts accordingly - new versions of that will contain the needed SASL mechanism plugins. 
Now we can safely remove SASL_PATH from the executor environment and finally merge this PR.

@sethp-nr please remove SASL_PATH again - once again sorry for this back-and-forth.

sethp-nr:
@tillt All done! Thanks for the quick follow-up.

spahl:
@mesosphere-mergebot bump-ee

gen/dcos-config.yaml
```diff
@@ -473,8 +473,6 @@ package:
       {
         "PATH": "/usr/bin:/bin",
         "SHELL": "/usr/bin/bash",
-        "LD_LIBRARY_PATH": "/opt/mesosphere/lib",
```
gkleiman: Cody tried to remove this in this PR: https://github.com/dcos/dcos/pull/707/

That PR isn't merged because the SDK frameworks weren't ready at that time. Are they ready now? =)

sethp-nr:
It looks like I can't yet merge this PR – it seems @tillt doesn't have write access to this repo, and therefore your review doesn't count towards my required two minimum approvers?

Additionally, I'm seeing a "has ship-it" job that's awaiting status, is there a label we should add?

BenWhitehead:
@mesosphere-mergebot bump-ee

adam-mesos:
@sethp-nr The green/grey nature of the approvals doesn't matter. What's blocking it now is that there was another (merge) commit after the previous approvals, so mergebot doesn't count them anymore.
We're bumping EE to see if that gets EE green, at which point we may need some reviewers to re-approve.

lingmann:
💥 🎉 💥

sethp-nr:
Thank you so much DCOS crew for your help getting this one through! It felt real good to mark [DCOS_OSS-1302](https://jira.mesosphere.com/browse/DCOS_OSS-1302) as resolved :)

PR #1881, opened 2017-08-28
Install jq because it is useful to have as an essential tool

GabrielNicolasAvellaneda:
jq is a useful tool to have on all nodes because:
1. It is a useful companion for the dcos-cli tool
2. It is required by the dcos-postflight script from the dcos-docker project that you maybe want to run on your cloud instances during cluster deployment.


orsenthil:
Our cloud images are maintained at https://github.com/dcos/dcos-images repo.

PR #1902, opened 2017-09-05
dcos_installer: Don't run genconf validation twice

mhrabovcin:
## High Level Description

When a user runs `./dcos_generate_config.sh` they'll get messages produced by gen package twice, i.e.

```
====> EXECUTING CONFIGURATION GENERATION
Generating configuration files...
Generating configuration files...
```

This patch is a minimal change that prevents dupes in the logging output. This PR addresses the issue discovered in https://github.com/mesosphere/dcos-enterprise/pull/1391#issuecomment-327164903

## Related Issues

  - [DCOS-18206](https://jira.mesosphere.com/browse/DCOS-18206) gen_extra: Add installer warning for private key on node when installing cluster with custom CA cert

## Checklist for all PR's

### DIFF
--- a/dcos_installer/backend.py
+++ b/dcos_installer/backend.py
@@ -15,8 +15,13 @@ import release
 import release.storage.aws
 import release.storage.local
 from dcos_installer import config_util, upgrade
-from dcos_installer.config import Config, normalize_config_validation
+from dcos_installer.config import (
+    Config,
+    normalize_config_validation,
+    normalize_config_validation_exception,
+)
 from dcos_installer.constants import CONFIG_PATH, GENCONF_DIR
+from gen.exceptions import ValidationError
 
 log = logging.getLogger()
 
@@ -24,17 +29,6 @@ log = logging.getLogger()
 def print_messages(messages):
     for key, error in messages.items():
         log.error('{}: {}'.format(key, error))
-
-
-def validate_gen(config):
-    """ Returns True on success
-    """
-    validate = config.do_validate(include_ssh=False)
-    if len(validate) > 0:
-        for key, error in validate.items():
-            log.error('{}: {}'.format(key, error))
-        return False
-    return True
 
 
 def do_configure(config_path=CONFIG_PATH):
@@ -45,12 +39,14 @@ def do_configure(config_path=CONFIG_PATH):
     """
     config = Config(config_path)
 
-    validation = validate_gen(config)
-
-    if not validation:
+    try:
+        gen_out = config_util.onprem_generate(config)
+    except ValidationError as e:
+        validation = normalize_config_validation_exception(e)
+        print_messages(validation)
         return 1
 
-    config_util.do_configure(config)
+    config_util.make_serve_dir(gen_out)
 
     return 0
 
@@ -62,10 +58,13 @@ def generate_node_upgrade_script(installed_cluster_version, config_path=CONFIG_P
         return 1
 
     config = Config(config_path)
-    validation = validate_gen(config)
-    if not validation:
+    try:
+        gen_out = config_util.onprem_generate(config)
+    except ValidationError as e:
+        validation = normalize_config_validation_exception(e)
+        print_messages(validation)
         return 1
-    gen_out = config_util.onprem_generate(config)
+
     config_util.make_serve_dir(gen_out)
 
     # generate the upgrade script
--- a/dcos_installer/config.py
+++ b/dcos_installer/config.py
@@ -45,6 +45,20 @@ def normalize_config_validation(messages):
     return validation
 
 
+def normalize_config_validation_exception(error):
+    """
+    Accepts ValidationError which is transofmed to dict and processed by
+    `normalize_config_validation` function.
+
+    :param exception: An exception raised during the config validation
+    :type exception: ValidationError
+    """
+    messages = {}
+    messages['errors'] = error.errors
+    messages['unset'] = error.unset
+    return normalize_config_validation(messages)
+
+
 def make_default_config_if_needed(config_path):
     if os.path.exists(config_path):
         return
--- a/dcos_installer/config.py
+++ b/dcos_installer/config.py
@@ -7,6 +7,7 @@ import yaml
 import gen
 import ssh.validate
 from gen.build_deploy.bash import onprem_source
+from gen.exceptions import ValidationError
 from pkgpanda.util import load_yaml, write_string, YamlParseError
 
 log = logging.getLogger(__name__)
@@ -45,13 +46,13 @@ def normalize_config_validation(messages):
     return validation
 
 
-def normalize_config_validation_exception(error):
+def normalize_config_validation_exception(error: ValidationError) -> dict:
     """
-    Accepts ValidationError which is transofmed to dict and processed by
+    A ValidationError is transformed to dict and processed by
     `normalize_config_validation` function.
 
-    :param exception: An exception raised during the config validation
-    :type exception: ValidationError
+    Args:
+        exception: An exception raised during the config validation
     """
     messages = {}
     messages['errors'] = error.errors
--- a/dcos_installer/test_backend.py
+++ b/dcos_installer/test_backend.py
@@ -1,6 +1,8 @@
 import json
+import logging
 import os
 import subprocess
+import textwrap
 import uuid
 
 import passlib.hash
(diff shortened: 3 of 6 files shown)


mhrabovcin:
@gpaul @adamtheturtle @vespian @jgehrcke would you please comment if you think that this approach is OK?

dcos_installer/backend.py
```diff
@@ -26,10 +27,34 @@ def print_messages(messages):
...
+        with muted_logging():
+            log.info("Message")
+
+    """
+    logger = logging.getLogger()
+    logger.disabled = True
+
+    yield
+
+    logger.disabled = False
```
adamtheturtle: What if the logger was disabled before entering this context manager? Does this then enable it? If so, that seems like unexpected behaviour.

jgehrcke:
This is almost certainly not OK. From within the installer we need to have a way to emit data to stdout in a controlled fashion. What is the problem to solve, really?

mhrabovcin:
Thanks @jgehrcke. I've changed approach and removed duplicate validation of configuration.

dcos_installer/config.py
```diff
@@ -45,6 +45,13 @@ def normalize_config_validation(messages):
     return validation
 
 
+def normalize_config_validation_exception(exception):
+    messages = {}
+    messages['errors'] = exception.errors
```
adamtheturtle: As discussed - this won't work and it should be something like `exception.errors`.

I'd like it if a test in this repository would catch that but I'm happy to discuss the trade-offs involved there.

adamtheturtle: Ah! Race condition between comment and change 😄

dcos_installer/backend.py
```diff
@@ -45,13 +39,13 @@ def do_configure(config_path=CONFIG_PATH):
     """
     config = Config(config_path)
 
-    validation = validate_gen(config)
-
-    if not validation:
+    try:
+        config_util.do_configure(config)
```
adamtheturtle: ❓ Separate really but it looks like `config_util.do_configure` is only called here. If that's right, what do you think of removing the function for consistency with https://github.com/dcos/dcos/pull/1902/files#diff-d66881dc86f4e8947b79ac9c556d0fe1R60 ?

adamtheturtle:
After Martin described this to me, I am on board.

`config.do_validate` logs some information and returns a list of validation errors.

```
config_util.do_configure
```

is equivalent to

```
config_util.onprem_generate(config)
config_util.make_serve_dir(gen_out)
```

Each of these calls `config.do_validate` and so logs some information. Each of these raises a `ValidationError` if there is a validation error.

Before, in the case that there were no errors, the code ran `do_validate` twice for each of the functions changed here. Once directly, once indirectly.

I am happy with the approach. If someone else is too, then it is appropriate to get this ready for submission.

I think that it is appropriate to have at least the following tests in this repository:
* An error is shown in the right place (stdout / stderr / the logger default) when there is a validation error.
* The logs are not duplicated either when there is or when there is not a validation error.

I understand that some of this functionality is tested indirectly by Enterprise DC/OS's tests, but any errors will be found much more quickly by tests in this repository. Also, relying on tests from Enterprise DC/OS to catch bugs in DC/OS OSS narrows the contributors who can effectively contribute to DC/OS OSS.

Separately, and for posterity, I would like to note that this issue has made it very clear to me that this package is unfriendly to develop. @mhrabovcin  has done a great job of unravelling a part of this package for a small fix. In my eyes there is no way that this configuration system can be maintained in the long term without a substantial amount of rewriting.

mhrabovcin:
@adamtheturtle thanks for the review. This PR now includes unit tests for new functionality.

Ping @jgehrcke

dcos_installer/test_backend.py
```diff
@@ -345,6 +349,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
         s3_bucket.delete()
 
 
+def test_do_configure_valid_config_no_duplicite_logging(tmpdir, monkeypatch, caplog):
+    """
+    Log messages aren't printed twice.
```
adamtheturtle: ❓ Don't do yourself a disservice! This test tests more than just that the log messages aren't printed twice. It tests that they aren't printed thrice, four times, and even seven times 😆 

I'd go with something like "Log messages are logged exactly once".

If not, fix "duplicite" in the test name to say "duplicate".

dcos_installer/test_backend.py
```diff
@@ -345,6 +349,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
...
+    """
+    monkeypatch.setenv('BOOTSTRAP_VARIANT', 'test_variant')
+    create_config(simple_full_config, tmpdir)
+    create_fake_build_artifacts(tmpdir)
+    with tmpdir.as_cwd():
+        assert backend.do_configure(config_path='genconf/config.yaml') == 0
+
+    # The message comes from gen.get_dcosconfig_source_target_and_templates()function
```
adamtheturtle: Space after `)`.

dcos_installer/test_backend.py
```diff
@@ -345,6 +349,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
...
+    expected_message = 'Generating configuration files...'
+    filtered_messages = [rec.message for rec in caplog.records if rec.message == expected_message]
+    assert [expected_message] == filtered_messages
+
+
+def test_do_configure_prints_validation_errors(tmpdir, monkeypatch, caplog):
+    """
+    Configuration validation errors are printed as `error` messages.
```
adamtheturtle: That logs are "print"ed is, I think, a poor choice of wording from the implementation `print_messages`. Let's not keep that bad terminology in the tests. Instead, use "logged". Note that this applies to the test function name as well.

dcos_installer/test_backend.py
```diff
@@ -345,6 +349,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
...
+    create_config(invalid_config, tmpdir)
+    create_fake_build_artifacts(tmpdir)
+    with tmpdir.as_cwd():
+        assert backend.do_configure(config_path='genconf/config.yaml') == 1
+
+    expected_error_message = (
+        'exhibitor_storage_backend: Must set exhibitor_storage_backend, '
+        'no way to calculate value.'
+        )
+    error_logs = list([rec for rec in caplog.records if rec.message == expected_error_message])
```
adamtheturtle: ❓ Why wrap in `list`?

dcos_installer/test_backend.py
```diff
@@ -345,6 +347,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
         s3_bucket.delete()
 
 
+def test_do_configure_valid_config_no_duplicite_logging(tmpdir, monkeypatch, caplog):
```
adamtheturtle: Typo `s/duplicite/duplicate`

dcos_installer/test_backend.py
```diff
@@ -345,6 +347,52 @@ def aws_cf_configure(s3_bucket_name, config, config_aws, tmpdir, monkeypatch):
...
+    create_config(invalid_config, tmpdir)
+    create_fake_build_artifacts(tmpdir)
+    with tmpdir.as_cwd():
+        assert backend.do_configure(config_path='genconf/config.yaml') == 1
+
+    expected_error_message = (
+        'exhibitor_storage_backend: Must set exhibitor_storage_backend, '
+        'no way to calculate value.'
+    )
+    error_logs = list([rec for rec in caplog.records if rec.message == expected_error_message])
```
adamtheturtle: ❓ no need for `list`

dcos_installer/config.py
```diff
@@ -45,6 +45,13 @@ def normalize_config_validation(messages):
     return validation
 
 
+def normalize_config_validation_exception(exception):
```
adamtheturtle: Please add a docstring for this function.

❓ I think that it is worth having a unit test for this function.

dcos_installer/backend.py
```diff
@@ -62,10 +58,13 @@ def generate_node_upgrade_script(installed_cluster_version, config_path=CONFIG_P
         return 1
 
     config = Config(config_path)
-    validation = validate_gen(config)
-    if not validation:
+    try:
```
branden: This stanza is identical to the one in `do_configure()`, and these functions are intended to do the same thing, except that `generate_node_upgrade_script()` outputs an additional artifact. So I think it'd be worth it to factor this bit of code into a helper function and call it in both places.

dcos_installer/config.py
```diff
@@ -45,6 +45,20 @@ def normalize_config_validation(messages):
     return validation
 
 
+def normalize_config_validation_exception(error):
+    """
+    Accepts ValidationError which is transofmed to dict and processed by
+    `normalize_config_validation` function.
+
+    :param exception: An exception raised during the config validation
+    :type exception: ValidationError
```
branden: Please use Python's type hints instead of docstrings for type annotation. We use docstrings in a lot of places because DC/OS predates Python's support for type hints, but we're trying to favor type hints going forward.

https://github.com/dcos/dcos/blob/master/docs/styleguides/python.md
https://www.python.org/dev/peps/pep-0484/
https://docs.python.org/3/library/typing.html

jgehrcke:
> Separately, and for posterity, I would like to note that this issue has made it very clear to me that this package is unfriendly to develop. 

:heart:

dcos_installer/backend.py
mhrabovcin: If you don't mind I'd like to keep it as it is. I don't see any good and useful abstraction I could build out of this code block.

mhrabovcin:

jgehrcke:
Let's push this forward.

PR #1911, opened 2017-09-07
Updated the default disk watch interval to 5 seconds in Mesos

chhsia0:
## High Level Description

This change makes the Mesos agents to check disk usage every 5 seconds
and GC completed top-level containers when there's not enough headroom.
It alleviates CORE-1289, and together with the following patch the problem
should be fixed:
https://issues.apache.org/jira/browse/MESOS-7939

## Related Issues

  - [CORE-1289](https://jira.mesosphere.com/browse/CORE-1289) Agent unable to recover due to no space left on disk.

## Checklist for all PR's

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: No test included since it is a configuration change. Manually tested on soak110 and this change made the problematic agent recovered from an out-of-disk state.

___

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -419,6 +419,7 @@ package:
       MESOS_MODULES_DIR=/opt/mesosphere/etc/mesos-slave-modules
       MESOS_CONTAINER_LOGGER=com_mesosphere_mesos_JournaldLogger
       MESOS_ISOLATION={{ mesos_isolation }}
+      MESOS_DISK_WATCH_INTERVAL=5secs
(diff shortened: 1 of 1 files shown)


vinodkone:
Does this result in a log line every second? If so, that might be excessive.

chhsia0:
@vinodkone Would it sound better if we make it 2 seconds? I don't want to introduce some special logic to not always output logs though. Or maybe modify Mesos to only do a per-second check only if we have not enough headroom?

chhsia0:
Changed the interval to 5 seconds to prevent excessive logs. @vinodkone PTAL!

gkleiman:
@vinodkone ping =)

PR #1921, opened 2017-09-12
Removed leader.mesos prerequisite for dcos-mesos-slave.

ArmandGrillet:
## High Level Description

Since DC/OS 1.8+ agents do not depend on `leader.mesos` (e.g. we can start a dcos-mesos-slave with `--master=zk1:2181`), we should not have this prerequisite as it can result in outages.

### DIFF
--- a/packages/mesos/extra/dcos-mesos-slave-public.service
+++ b/packages/mesos/extra/dcos-mesos-slave-public.service
@@ -19,7 +19,6 @@ EnvironmentFile=-/var/lib/dcos/mesos-slave-common
 EnvironmentFile=-/var/lib/dcos/mesos-resources
 EnvironmentFile=-/run/dcos/etc/mesos-slave-public
 ExecStartPre=/bin/ping -c1 ready.spartan
-ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-mesos-slave-public
 ExecStartPre=/opt/mesosphere/bin/make_disk_resources.py /var/lib/dcos/mesos-resources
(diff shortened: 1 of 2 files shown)


ArmandGrillet:
Hi @lingmann. Thanks for your comment, I have also updated `dcos-mesos-slave-public.service` in a new commit. The other services having this prerequisite are:

- `dcos-adminrouter.service`
- `dcos-adminrouter-agent.service`
- `dcos-history.service`
- `dcos-signal.service`

I am not sure if removing this `ping -c1 leader.mesos` on other services is advisable, I will ask the persons knowing about these services soon to have an answer.

ArmandGrillet:
I have discussed with @vespian about this prerequisite for `dcos-adminrouter.service` and it needs to stay. I would thus suggest not changing other services in that PR.

bernadinm:
Thanks @ArmandGrillet. LGTM

lingmann:
@mesosphere-mergebot bump-ee

lingmann:
@spahl can you add your review to the enterprise half (just created it and waiting for tests to go green). I don't anticipate any problems just trying to follow the process. :)

jgehrcke:
Huhu. I came and expected to see a link to a JIRA ticket. Is there one? I would hope that the JIRA ticket provides a bit of background on

> as it can result in outages.

ArmandGrillet:
@jgehrcke We only have a private ticket for it: COPS-575. To give more context in this conversation:

Some users ran into an issue where mesos agents died because they lost access to the zookeeper quorum. The dcos-mesos-slave.service main process exited, code=killed, status=6/ABRT. Once it restarted, it was unable to start properly as the host `leader.mesos` was not set (due to an issue with the masters that was not related to the agents). 

The reason for this PR is that we wants agents to restart properly in such a case as `leader.mesos` is not required.

PR #1922, opened 2017-09-12
Increase recovery timeout in agent config.

lava:
## High Level Description

With the latest partition awareness changes, we can increase
the recovery timeout to a much higher value to avoid outages
due to the agent being down for whatever reason.

## Related Issues

  - [CORE-1291](https://jira.mesosphere.com/browse/CORE-1291) Increase the value for `--recovery_timeout` agent flag in DC/OS

## Checklist for all PR's

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -25,6 +25,7 @@ import collections
 import ipaddress
 import json
 import os
+import re
 import socket
 import string
 import textwrap
@@ -618,6 +619,20 @@ def validate_mesos_max_completed_tasks_per_framework(
                                  "parameter as an integer: {}".format(ex)) from ex
 
 
+def validate_mesos_recovery_timeout(mesos_recovery_timeout):
+    units = ['ns', 'us', 'ms', 'secs', 'mins', 'hrs', 'days', 'weeks']
+
+    match = re.match("([\d\.]+)(\w+)", mesos_recovery_timeout)
+    assert match is not None, "Error parsing 'mesos_recovery_timeout' value: {}.".format(mesos_recovery_timeout)
+
+    value = match.group(1)
+    unit = match.group(2)
+
+    assert value.count('.') <= 1, "Invalid decimal format."
+    assert float(value) <= 2**64, "Value {} not in supported range.".format(value)
+    assert unit in units, "Unit '{}' not in {}.".format(unit, units)
+
+
 def calculate_check_config_contents(check_config, custom_checks, check_search_path, check_ld_library_path):
 
     def merged_check_config(config_a, config_b):
@@ -888,6 +903,7 @@ entry = {
         lambda adminrouter_tls_1_0_enabled: validate_true_false(adminrouter_tls_1_0_enabled),
         lambda gpus_are_scarce: validate_true_false(gpus_are_scarce),
         validate_mesos_max_completed_tasks_per_framework,
+        validate_mesos_recovery_timeout,
         lambda check_config: validate_check_config(check_config),
         lambda custom_checks: validate_check_config(custom_checks),
         lambda custom_checks, check_config: validate_custom_checks(custom_checks, check_config),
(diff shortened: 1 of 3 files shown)


lava:
@mesosphere-mergebot bump-ee

lava:
@vinodkone , @adam-mesos Could you review this?

adam-mesos:
@lava Please fill in the rest of the PR template, especially issue links for more context.
A few questions:
1. What is the default value we're changing from?
2. Are we certain that increasing this value is strictly a good thing? What are the possible negative consequences? (e.g. truly down agents appear "unreachable" much longer, and Marathon will not move its tasks for an entire day)
2a. If it's always a good idea, why not change the default value in Apache Mesos?
2b. If there are some users/use-cases which wouldn't want this change, please add a dcos-config.yaml option for them to override with their own custom value.

lava:
@adam-mesos
It used to be that there was no point in keeping executors alive for a long time, because if an agent failed for longer than `agent_reregister_timeout` (10m), all tasks running on this node would be killed anyways as soon as the agent tries to reconnect.

With the new partition-awareness changes, this is no longer the case, so now it can happen that an executor will suicide even if the scheduler didn't give up on the task yet.

1. The default values of this setting is 15 minutes.
2. For truly down agents, nothing should change since the statuses they receive are still controlled by master configuration, i.e. `agent_reregister_timeout`, which I don't touch in this review. For partition-aware frameworks, nothing should change as well because they couldn't have assumed that the tasks were killed even before this change. The one case where we would see different behaviour is when a non-partition-aware framework is running a task on a host where the agent fails, it would be likely to reschedule the while the old one is still running, and the two would run in parallel until the agent reregisters or 24 hours elapse.
2a. The problem surfaced in DC/OS due to the systemd service file keeping the agent from restarting for more than 10 minutes, so it seems logical to change the Mesos configuration of DC/OS to handle this situation better rather than change the default configuration of all Mesos users.
2b. I would be happy to do so, but I'm not quite sure I understand your request? I thought I *am* adding an option to dcos-config.yaml with this request already.

vinodkone:
Good explanation @lava . Sorry I didn't give you more context when I filed the ticket, but looks like you got it :)

Note that for the NPA task running on a truly down agent scenario, the framework will not be able to distinguish this scenario from a partitioned agent scenario; so it was/is always possible for a NPA scheduler to relaunch parallel instances. So this change is not that much of a breaking change for NPA frameworks.

vinodkone:
LGTM

vinodkone:
Regarding configurability, look at how for example `gc_delay` is made configurable in config.yaml. I think Adam is looking for something similar.

lava:
@vinodkone , @adam-mesos I updated the review to allow users to override this setting with custom values.

gen/calc.py
```diff
@@ -919,6 +919,7 @@ def node_check_names(config):
         'mesos_log_retention_mb': '4000',
         'mesos_container_log_sink': 'logrotate',
         'mesos_max_completed_tasks_per_framework': '',
+        'mesos_recovery_timeout': '24hrs',
```
adam-mesos: Can we get a validate function on this input, so users don't accidentally try "24hours" or "24h" or "24" or "1day" or something else that won't be parsed properly by Mesos?

lava: @adam-mesos Sure, I just pushed a new commit that adds validation.

lava:
@mesosphere-mergebot bump-ee

gen/tests/test_validation.py
```diff
@@ -141,6 +141,26 @@ def test_invalid_mesos_dns_set_truncate_bit():
         true_false_msg)
 
 
+def test_validate_mesos_recovery_timeout():
+    validate_success(
+        {'mesos_recovery_timeout': '24hrs'})
```
adam-mesos: What about '24.5hrs'? Does it have to be an integer?

adam-mesos:
@mesosphere-mergebot bump-ee

gen/tests/test_validation.py
lava: Not anymore with the latest commit.

lava:
@adam-mesos I added an additional test.
@mesosphere-mergebot bump-ee

adam-mesos:
@mesosphere-mergebot bump-ee

adam-mesos:
@rukletsov @vinodkone Can you please re-review (and "approve") after the recent update?
The EE bump PR also needs an approval, even though it's a trivial bump.

PR #1946, opened 2017-09-20
[1.9.5] Removed leader.mesos prerequisite for dcos-mesos-slave(-public).

ArmandGrillet:
## High Level Description

Since DC/OS 1.8+ agents do not depend on `leader.mesos` (e.g. we can start a dcos-mesos-slave with `--master=zk1:2181`), we should not have this prerequisite for the services dcos-mesos-slave and dcos-mesos-slave-public as it can result in outages.

This PR is a backport of #1921 for DC/OS 1.9.


jgehrcke:
I think this PR should have a link to a corresponding JIRA ticket :) As mentioned in https://github.com/dcos/dcos/pull/1921#issuecomment-330795305.

PR #1975, opened 2017-09-27
Use m4.large instead of m4.xlarge as default in CF

mellenburg:
## Checklist for all PR's


mellenburg:
@mesosphere-mergebot bump-ee

PR #1984, opened 2017-09-29
Added exhibitor TLS conf

rdeliallisi:
Exhibitor modifications. See https://github.com/mesosphere/dcos-enterprise/pull/1422.

## High Level Description

## Related Issues

  - [DCOS-17894](https://jira.mesosphere.com/browse/DCOS-17894) Encrypt & authenticate cross-Exhibitor communication (1.11).
## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: No new functionality.


rdeliallisi:
@mesosphere-mergebot Ready For Review

gen/dcos-config.yaml
```diff
@@ -798,9 +798,8 @@ package:
 {% case "static" %}
   - path: /etc_master/dns_config_master
     content: |
-      MASTER_SOURCE=exhibitor_uri
-      EXHIBITOR_URI=http://127.0.0.1:8181/exhibitor/v1/cluster/status
-      EXHIBITOR_ADDRESS=127.0.0.1
+      MASTER_SOURCE=master_list
+      RESOLVERS={{ resolvers_str }}
```
jgehrcke: Nice to see that.

Can you link to the corresponding JIRA ticket here as a reply to this comment? At least for posterity.

packages/dcos-integration-test/extra/test_composition.py
```diff
@@ -52,7 +52,7 @@ def test_if_all_exhibitors_are_in_sync(dcos_api_session):
     correct_data = sorted(r.json(), key=lambda k: k['hostname'])
 
     for zk_ip in dcos_api_session.masters:
-        resp = requests.get('http://{}:8181/exhibitor/v1/cluster/status'.format(zk_ip))
+        resp = requests.get('http://{}/exhibitor/exhibitor/v1/cluster/status'.format(zk_ip), verify=False)
```
jgehrcke: Please add a comment that this test method relies on the fact that the Admin Router instance being talked to here proxies the HTTP request to the _local_ Exhibitor instance, on the same node.

For posterity: the goal here is to test a specific Exhibitor instance. The patch makes sure that we reach that through Admin Router, instead of directly reaching it. But we must still ensure that we test _all_ Exhibitor instances. That is currently guaranteed precisely because Admin Router proxies to the local Exhibitor (as opposed to proxying to e.g. `https://master.mesos:8181` which would contact a "random" Exhibitor.

packages/exhibitor/extra/start_exhibitor.py
```diff
@@ -175,7 +175,24 @@ def invoke_detect_ip():
...
+truststore_path = '/var/lib/dcos/exhibitor-tls-artifacts/truststore.jks'
+clientstore_path = '/var/lib/dcos/exhibitor-tls-artifacts/clientstore.jks'
+serverstore_path = '/var/lib/dcos/exhibitor-tls-artifacts/serverstore.jks'
+
+exhibitor_env = os.environ.copy()
+if os.path.exists(truststore_path) and \
+   os.path.exists(clientstore_path) and \
+   os.path.exists(serverstore_path):
+    exhibitor_env['EXHIBITOR_TLS_TRUSTSTORE_PATH'] = truststore_path
+    exhibitor_env['EXHIBITOR_TLS_TRUSTSTORE_PASSWORD'] = 'changeme'
```
jgehrcke: Please, in your entire system, change `changeme` to `not-relevant-for-security` or something shorter but comparable.

packages/exhibitor/extra/start_exhibitor.py
```diff
@@ -175,7 +175,24 @@ def invoke_detect_ip():
...
+    exhibitor_env['EXHIBITOR_TLS_SERVER_KEYSTORE_PATH'] = serverstore_path
+    exhibitor_env['EXHIBITOR_TLS_SERVER_KEYSTORE_PASSWORD'] = 'changeme'
+    exhibitor_env['EXHIBITOR_TLS_REQUIRE_CLIENT_CERT'] = 'true'
+    exhibitor_env['EXHIBITOR_TLS_VERIFY_PEER_CERT'] = 'true'
+
 # Start exhibitor
 print("Running exhibitor as command:", exhibitor_cmdline)
 sys.stdout.flush()
-os.execv('/opt/mesosphere/bin/java', exhibitor_cmdline)
+os.execve('/opt/mesosphere/bin/java', exhibitor_cmdline, exhibitor_env)
```
jgehrcke: Interesting. You could also modify the current environment (the environment of the CPython process by mutating `os.environ` directly, and then keep `os.execv()` as-is.

Do you prefer one of the other and if yes, why?

packages/dcos-integration-test/extra/test_composition.py
jgehrcke: Let's please also change the variable name `zk_ip` to `master_node_ip`.

gen/dcos-config.yaml
rdeliallisi: I would prefer that this change was made by the networking team since I still don't know that this does not break anything. It is present in this diff for the sole purpose of passing the tests.

packages/dcos-integration-test/extra/test_composition.py
rdeliallisi: Done.

packages/exhibitor/extra/start_exhibitor.py
rdeliallisi: I would prefer to pass the new environment variables explicitly to the new process. Since the parent process does not need them why change its environment (even though it will be cleanly replaced by it's child).

gen/dcos-config.yaml
jgehrcke: Still it would be nice if you could link the JIRA ticket here in this comment stream so that the related discussion in JIRA can be easily discovered by reviewers of your pull request!  :)

rdeliallisi: https://jira.mesosphere.com/browse/DCOS-18703

jgehrcke:
Ship-it was applied prematurely. See DCOS-19252.

packages/exhibitor/extra/start_exhibitor.py
gpaul: The parent process' purpose is as a launcher for the child, I think changing the parent's environment in that case makes complete sense. I wouldn't block this PR on that change though.

gpaul:
It sounds like this PR is blocked by https://jira.mesosphere.com/browse/DCOS-18703? It switches 
`MASTER_SOURCE=exhibitor_uri` to `MASTER_SOURCE=master_list`.

Doing so seems wrong: where does it get `master_list` from if the cluster is not using the static backend?

gen/dcos-config.yaml
urbanserj: Hi @rdeliallisi!

Could you please remove `RESOLVERS` from this config? It seems like it's already configured in `/opt/mesosphere/etc/dns_config`.

Please check that this patch will work for dcos-enterprice. As I know bootstrap also [uses](https://github.com/mesosphere/dcos-enterprise/blob/master/packages/bootstrap/extra/dcos_internal_utils/bootstrap.py#L1529-L1553) this configuration on agent nodes.

Also I can make it by myself as part of [DCOS-18703](https://jira.mesosphere.com/browse/DCOS-18703) if you prefer.

gpaul: @urbanserj @rdeliallisi's internship at Mesosphere has ended but I'm happy to take his work further. I'll remove the `RESOLVERS` config and check that the upstream bump passes

gpaul: Not all units that source the `dns_config_master` file also source `dns_config`. For example 
https://github.com/dcos/dcos/blob/e61cfc2038146a58614a58989ff68f64dbdc8038/packages/dcos-net/extra/dcos-dns-watchdog.service
https://github.com/dcos/dcos/blob/e61cfc2038146a58614a58989ff68f64dbdc8038/packages/dcos-net/extra/dcos-l4lb-watchdog.servicehttps://github.com/dcos/dcos/blob/e61cfc2038146a58614a58989ff68f64dbdc8038/packages/dcos-net/extra/dcos-overlay-watchdog.service

I suggest we keep the `RESOLVERS=` variable in `dns_config_master` and address that refactoring in a separate PR.

gpaul:
@mesosphere-mergebot bump-ee

gpaul:
@jgehrcke please close this now that https://github.com/dcos/dcos/pull/2098 was merged.

PR #1992, opened 2017-10-04
Add Mesos work dir config params

branden:
## High Level Description

This adds two `config.yaml` parameters that can override the location of the work directory for Mesos masters and agents: `mesos_master_work_dir` and `mesos_agent_work_dir`.

## Related Issues

  - [DCOS_OSS-1586](https://jira.mesosphere.com/browse/DCOS_OSS-1586) Configurable Mesos work dir location

## Checklist for all PR's

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -113,6 +113,11 @@ def validate_url(url: str):
         ) from ex
 
 
+def validate_absolute_path(path):
+    if not path.startswith('/'):
+        raise AssertionError('Must be an absolute filesystem path starting with /')
+
+
 def validate_ip_list(json_str: str):
     nodes_list = validate_json_list(json_str)
     check_duplicates(nodes_list)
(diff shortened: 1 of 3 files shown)


branden:
@mesosphere-mergebot bump-ee

branden:

gen/calc.py
```diff
@@ -113,6 +113,11 @@ def validate_url(url: str):
         ) from ex
 
 
+def validate_absolute_path(path):
```
jgehrcke: Mhm. Can we rename this to `validate_startswith_slash(text)`?

jgehrcke: Ha well, I see that the error message should be adjusted to the case where this is a filesystem path. But `/` as well as `/etc` both pass this validation, and I wonder if we're actually checking for whatever matters most here.

jgehrcke: Should we maybe additionally confirm that
* this does not have a trailing slash
* this results in a non-emtpy dirname and non-empty basename when using `os.path.split()` on the text?

spahl:
Why do we want to expose this to the user? What is the use case?

branden:
@spahl Sent you an email. Also see discussion on https://jira.mesosphere.com/browse/DCOS_OSS-1586.

gen/calc.py
branden: Here's a successful integration test run that specifies the work dirs with trailing slashes: https://teamcity.mesosphere.io/viewLog.html?buildId=808547&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsOnpremWStaticBackend&tab=buildResultsDiv

Agreed that we should catch as many problems as possible prior to deployment, since that's when they're easier to debug. But I don't think it's worth it to validate the location itself, since the idea behind this feature is to allow users to adapt their installation to meet requirements imposed by their environment that we can't predict. I think it'd make more sense to explain in documentation what these params do, that they're intended for advanced users, and that they should be left at their defaults unless there's a specific reason to change them.

branden:
Enterprise bump PR isn't reviewed but is only for testing, and it passes CI, so I'll give this a shipit.

branden:
@mesosphere-mergebot label shipit

PR #1991, opened 2017-10-03
Update CentOS Docker packages

sschneid:
Update the `docker-engine` packages pulled by CentOS to `17.05.0.ce` (latest).

PR #1961, opened 2017-09-25
Update centos AMIs to version 7.4

margaret:
## High Level Description

Doing this via changing the build scripts in the cloud_images dcos package to read in environment variable for `CENTOS_VERSION`.

- Defaults to 7.3.1611 (initial hard-coded value) in scripts, but I set it to 7.4 in TeamCity.

Manually-kicked-off CI job that created the DC/OS-ready AMIs:
https://teamcity.mesosphere.io/repository/download/DcosIo_Dcos_BuildCloudBaseImages_AwsBuildAndPublishCentOS7baseImage/798303:id/cloud_images/centos7/artifacts.txt

GovCloud DC/OS-ready AMIs generated with https://teamcity.mesosphere.io/viewLog.html?buildId=816581&tab=buildResultsDiv&buildTypeId=DcosIo_Dcos_BuildCloudBaseImages_AwsBuildAndPublishCentOS7baseImagesForGovcloud

Details of testing are in the comments below, but the images were tested manually by bringing up different types of clusters and then running the integration tests.

## Related Issues

This is in support of
 - https://jira.mesosphere.com/browse/DCOS_OSS-1646
Generating images that will be added to dcos-launch configuration.

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [ ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and Review Process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using github's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

[...]

PR #2015, opened 2017-10-11
Enhance install-prereqs

karlkfi:
Fixes multiple bugs and UX issues with the GUI and CLI installer.

- Enable errexit, nounset, and pipefail (exit when there's a problem)
- Add progress messages for easier debugging
- Validate CentOS/RHEL >= 7.2 (or CoreOS)
- Validate Kernel >= 3.10.x (for minimal overlayfs support)
- Validate overlayfs kernel module (or enable and load)
- Validate Docker >= 1.11 (if installed)
- Validate Docker Data Root (default /var/lib/docker) was provisioned with `ftype=1` if FS is XFS (to avoid known bug with ftype=0)
- Only yum install utilities if not found (print version if already installed)
- Only install Docker (17.05.0.ce) if not found (print version if already installed)
- Validate nogroup linux group (or create)
- Remove obsolete `/opt/dcos-prereqs.installed` file (all steps skipped if validated)

## Related Issues

  - [DCOS-14036](https://jira.mesosphere.com/browse/DCOS-14036) yum update considered harmful
  - [DCOS_OSS-926](https://jira.mesosphere.com/browse/DCOS_OSS-926) 1.9 - Install prereqs always fails on the first attempt

## Checklist for all PR's

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: `Existing CLI install tests should validate this works, but we don't have enough test coverage of OS permutations to really cover all the use cases.`

### DIFF
--- a/dcos_installer/action_lib.py
+++ b/dcos_installer/action_lib.py
@@ -317,18 +317,13 @@ exit $RETCODE"""
         [node_check_script],
         stage='Executing node postflight checks')
 
-    node_cleanup_chain = ssh.utils.CommandChain('postflight_cleanup')
-    node_cleanup_chain.add_execute(
-        ['sudo', 'rm', '-f', '/opt/dcos-prereqs.installed'],
-        stage='Removing prerequisites flag')
-
     cluster_postflight_chain = ssh.utils.CommandChain('cluster_postflight')
     cluster_postflight_chain.add_execute(
         [cluster_check_script],
         stage='Executing cluster postflight checks')
 
     node_check_result = yield from node_runner.run_commands_chain_async(
-        [node_postflight_chain, node_cleanup_chain],
+        [node_postflight_chain],
         block=block,
         state_json_dir=state_json_dir,
         delegate_extra_params=nodes_count_by_type(config))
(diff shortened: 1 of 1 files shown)


karlkfi:
Note: It's easier to review in `unified` diff mode.

karlkfi:
Fixed to use Docker 17, per https://github.com/dcos/dcos/pull/1991

karlkfi:
`teamcity/dcos/test/aws/onprem/static` is the only test that uses the install-prereqs script.

karlkfi:
Well, now we're failing for a new reason...
```
Validating distro...
Distro: CentOS
Validating distro version...
Validating kernel version...
Validating kernel modules...
Validating file system...
File System: xfs
Error: /var/lib/docker must use XFS provisioned with ftype=1 to avoid known issues with OverlayFS.
```

Turns out the AMI we're using is XFS, but has `type=0`, not `type=1` (unless my test is broken). That probably means it was installed with the CentOS 7.2 installer, which uses `type=0` by default. The 7.3 installer uses `type=1` by default. So we should upgrade to 7.3 or 7.4.

Technically we still support 7.2, but the user will need to modify their kickstart script to configure provisioning to use `type=1`. If they _really_ want to use `type=0` they can still do so by skipping `--install-prereqs` or using the advanced installer.

Using `type=0` apparently passes all our existing tests, but `type=1` is recommended by RedHat: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/7.2_Release_Notes/technology-preview-file_systems.html

We might want to also add a test that validates this... like deploying and using the nginx package.

karlkfi:
@sschneid I think `teamcity/dcos/test/aws/onprem/static` is failing now because my new test checks that the root volume uses XFS ftype=1, but the AMIs being used by the `teamcity-dcos-test-aws-onprem-static` CloudFormation template are using `ftype=0`. That probably means they were provisioned with the CentOS 7.2 installer. Can you update the AMI that CF template uses to be one provisioned with the CentOS 7.3 installer? Doing `yum update` on 7.2 wont do the trick. The volume provisioning needs to be done with `ftype=1` which is optional in 7.2 and default in 7.3.

dcos_installer/action_lib.py
```diff
@@ -371,38 +371,129 @@ def uninstall_dcos(config, block=False, state_json_dir=None, async_delegate=None
...
+  if [[ ${docker_major_version} -lt 1 || ${kernel_minor_version} -lt 11 ]]; then
+    echo "Docker version ${docker_server_version} not supported. Please uninstall Docker and try again." >&2
+    exit 1
+  fi
+
+  install_docker='false'
+fi
+
+if [[ "${install_docker}" == 'true' ]]; then
+  echo "Installing Docker..."
```
branden: It might be helpful if this output included the version of Docker we're installing.

karlkfi:
Blocked by https://github.com/dcos/dcos/pull/1961 which should fix the ftype=1 issue by using the CentOS 7.4 installer

dcos_installer/action_lib.py
```diff
@@ -371,38 +371,129 @@ def uninstall_dcos(config, block=False, state_json_dir=None, async_delegate=None
...
+  exit 1
 fi
 
-version=$(cat /etc/*-release | sed -n 's@^VERSION_ID="\([0-9]*\)\([0-9\.]*\)"$@\1@p')
-if [ $version -lt 7 ]; then
-  echo "$version is not supported. Only >= 7 version is supported" >&2
-  exit 0
+echo "Validating distro version..."
+# CentOS & RHEL < 7 have inconsistent release file locations
+distro_major_version="$(cat /etc/*elease | sed -n 's/^VERSION_ID="\([0-9][0-9]*\).*"$/\1/p')"
```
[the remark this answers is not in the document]
karlkfi: Yes, to catch different capitalization that occurs on older versions and different distros

dcos_installer/action_lib.py
karlkfi: I had that originally but then there was more than one place to change the version... the yum install output should say which version is installed

karlkfi:
> It'd be nice if the supported version ranges were more readable and easier to update
I have a bash semver lib but not all of these use semver and the lib is larger than this whole install script. 
I could maybe abstract a function but then I have to do version parsing twice for each instance because usually it's finding the version from a bigger string...

karlkfi:
Waiting for https://github.com/dcos/dcos/pull/1961

dcos_installer/action_lib.py
```diff
@@ -371,38 +371,129 @@ def uninstall_dcos(config, block=False, state_json_dir=None, async_delegate=None
...
+
+if [[ -f /opt/dcos-prereqs.installed ]]; then
+  echo "install_prereqs has been already executed on this host, exiting..."
   exit 0
 fi
 
-if ([ x$dist != 'xrhel' ] && [ x$dist != 'xcentos' ]); then
-  echo "$dist is not supported. Only RHEL and CentOS are supported" >&2
+echo "Validating distro..."
+distro="$(cat /etc/os-release | sed -n 's/^ID="\(.*\)"$/\1/p')"
```
asridharan: @karlkfi I am running the CoreOS `alpha` build, and the above `sed` doesn't seem to work:
```
dist=$(cat /etc/os-release | sed -n 's@^ID="\(.*\)"$@\\1@p')
```
These are the content of the `/etc/os-release`:
```
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1562.1.0
VERSION_ID=1562.1.0
BUILD_ID=2017-10-12-0536
PRETTY_NAME="Container Linux by CoreOS 1562.1.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```
However, when I do:
```
 cat /etc/os-release | sed -n 's@^ID=\(.*\)@\1@p'
coreos
```
It works fine. The quotes shouldn't be present in the `sed`?

karlkfi: Sounds like the quotes need to be optional in the sed, because CentOS definitely has them. I guess these are env vars. So if there's no space in the string, the quotes are optional.

branden: Since these are env vars, how about we skip the `sed` and just source the file? One less regex to think about.

```
dist=$(. /etc/os-release && echo $ID)
```

karlkfi: We might be better off with `distro="$(source /etc/os-release && echo $ID)"`

karlkfi: done. I also changed the other one later on since the VERSION_ID is always in the same file on the 3 OSs we support.

karlkfi:
rebased against master

karlkfi:
rebased again... maybe I'll get CentOS 7.4 now..?

karlkfi:
@margaret I'm still getting CentOS 7.2 on the `teamcity/dcos/test/aws/onprem/static` test. I thought we upgraded to CentOS 7.4 recently. Did that not get propagated to the test suites?

karlkfi:
Blocked on https://github.com/dcos/dcos-launch/pull/37

karlkfi:
More blockers...

- Apparently the old centos-only AMI included Docker and the docker user group, but the new centos-only AMI does not.
- dcos-installer requires Docker on the bootstrap node.
- dcos-launch expected the docker user group on the bootstrap node

So @mellenburg will be attempting to remove that last requirement. 

Follow up:
- Make a bootstrap-specific AMI that just has the bootstrap requirements on it, not the full DC/OS node requirement set.
- Make the bootstrap AMI configurable in dcos-launch
- Remove the docker user group from the centos w/ prereqs image builder: https://github.com/margaret/dcos/blob/master/cloud_images/centos7/install_prereqs.sh#L63
- Verify that no one else is using the docker user group in their DC/OS tests (@mellenburg mentioned that some of the data services might use it for testing)

karlkfi:
Latest issue seems to be bailing after unzip is detected... no idea why, yet...

```
[16:43:55][Step 5/7]              'Detecting unzip...\r',
[16:43:55][Step 5/7]              'UnZip 6.00 of 20 April 2009, by Info-ZIP.  Maintained by C. '
[16:43:55][Step 5/7]              'Spieler.  Send\r',
[16:43:55][Step 5/7]              '']}]Failed to execute script cli
```

karlkfi:
rebased

karlkfi:
I never did figure out why the unzip version printing was erroring. It would error int he script but not when I ran it manually.

Anyway, I changed the script to use `rpm -q` to print the version instead, which should work, as long as rpm or yum were used to install it. If it was installed manually or from source the prereqs installer might error, but that's probably fine.

karlkfi:
Rebased to fix Jenkins

karlkfi:

spahl:
@mesosphere-mergebot bump-ee

spahl:
@branden @amitaekbote I think a lot of the checks in this script should be added to dcos-checks at some point.

orsenthil:
@branden  - could you please review this one again. I see there are update since your last approval.

karlkfi:
This is done as soon as we get the enterprise CIs to be happy. 5 of them failed, but none of them required changes here.

karlkfi:
Squashed and rebased. Needs new approvals!
@spahl @branden

karlkfi:
@mesosphere-mergebot bump-ee

PR #2055, opened 2017-10-27
QUALITY-1607: `pkgpanda fetch --force-reinstall` to recover from modified local packages

MrMarvin:
### `fetch --force-reinstall` feature
This change will add a new flag to `pkgpanda fetch`, which allows an operator to forcefully overwrite an already fetched package, without manually deleting it from the nodes disk. This is especially useful for the python package, which cannot be (temporarily) removed to re-fetch it, without breaking `pkgpanda` and other tools relying on that exact DC/OS provided python installation. 
For a actual use-case in which this is desirable, please see the linked (internal) [SOAK-28](https://jira.mesosphere.com/browse/SOAK-28) ticket. Basically it allows operators to recover  a modified DC/OS installation and not have to necessarily install a node from scratch, in some cases.

### reproducible `activate` runs
Also this PR adds an order on pkgpanda Package objects, so these can be sorted, which is useful to allow (a failing) `pkgpanda activate` to be reproducible across multiple runs.
The key on which the package objects are ordered doesn't really matter, as long as they do not change across invocations of the function, so `name` is doing fine.

## Related Issues

  - [SOAK-28](https://jira.mesosphere.com/browse/SOAK-28) while investigation this internally, we needed to debug pkgpandas behavior
  - [QUALITY-1607](https://jira.mesosphere.com/browse/QUALITY-1607) tracks the issue on DC/OS  to allow easier detection and resolution of these kind of errors by operators
  - [DCOS_OSS-1644)](https://jira.mesosphere.com/browse/DCOS_OSS-1644) tracks the improvements on DC/OS  installer or upgrade scripts to allow easier detection and resolution of these kind of errors by operators

## Example output
```
[root@int-master1-test110 mesosphere]# touch /opt/mesosphere/packages/python--6fd09648eded002b46ab08542af7cd7910f60f46/bin/chardetect
[root@int-master1-test110 mesosphere]# bash /tmp/dcos_node_upgrade.sh
Upgrading DC/OS master 1.10.1 -> 1.10.1
Validation Error: Two packages are trying to install the same file /opt/mesosphere/bin.new/chardetect or two roles in the set of roles ['master'] are causing a package to try activating multiple versions of the same file. One of the package files is /opt/mesosphere/packages/python--6fd09648eded002b46ab08542af7cd7910f60f46/bin/chardetect. This might also be caused by modified content of an installed package. You can try to fix it by using `pkgpanda fetch --force-reinstall ...` on the packge(s) stated in this message.
[root@int-master1-test110 mesosphere]# pkgpanda fetch --repository-url=http://172.31.4.196:8080 python--6fd09648eded002b46ab08542af7cd7910f60f46
Did not fetch, already there: python--6fd09648eded002b46ab08542af7cd7910f60f46
[root@int-master1-test110 mesosphere]#
[root@int-master1-test110 mesosphere]#
[root@int-master1-test110 mesosphere]# pkgpanda fetch --force-reinstall --repository-url=http://172.31.4.196:8080 python--6fd09648eded002b46ab08542af7cd7910f60f46
Fetched: python--6fd09648eded002b46ab08542af7cd7910f60f466
[root@int-master1-test110 mesosphere]# bash /tmp/dcos_node_upgrade.sh
Upgrading DC/OS master 1.10.1 -> 1.10.1
[root@int-master1-test110 mesosphere]#
```

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -382,13 +382,13 @@ class Repository:
     # Add the given package to the repository.
     # If the package is already in the repository does a no-op and returns false.
     # Returns true otherwise.
-    def add(self, fetcher, id, warn_added=True):
+    def add(self, fetcher, id, warn_added=True, force_reinstall=False):
         # Validate the package id.
         PackageId(id)
 
         # If the package already exists, return true
         package_path = self.package_path(id)
-        if os.path.exists(package_path):
+        if os.path.exists(package_path) and not force_reinstall:
             if warn_added:
                 print("Package already added.")
             return False
@@ -403,13 +403,24 @@ class Repository:
         # number however so other code doing directory scans will be fine with
         # the temp folders.
         tmp_path = pkg_path + '_tmp'
+        old_path = pkg_path + '_old'
 
         # Cleanup artifacts (if any) laying around from previous partial
         # package extractions.
         check_call(['rm', '-rf', tmp_path])
 
         fetcher(id, tmp_path)
-        os.rename(tmp_path, pkg_path)
+        try:
+            if os.path.exists(pkg_path):
+                if warn_added:
+                    print("Package already added, moving old installation to {}".format(old_path))
+                os.rename(pkg_path, old_path)
+            os.rename(tmp_path, pkg_path)
+        finally:
+            try:
+                shutil.rmtree(old_path)
+            except OSError:
+                pass
         return True
 
     def remove(self, id):
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -188,7 +188,7 @@ class Package:
     def __repr__(self):
         return str(self.__id)
 
-    def __lt__(self,other):
+    def __lt__(self, other):
         return (self.name < other.name)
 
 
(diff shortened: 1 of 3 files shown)


MrMarvin:
cc @branden @cmaloney Is this something you would like to give a review on?

MrMarvin:

branden:
+1 on clearer error messages. What's the benefit of `pkgpanda fetch --force-reinstall` vs doing a `pkgpanda remove` and then a `pkgpanda fetch`?

MrMarvin:
>What's the benefit of pkgpanda fetch --force-reinstall vs doing a pkgpanda remove and then a pkgpanda fetch?

It tries to be more atomic and allows to reinstall a modified python installation itself.

> This is especially useful for the python package, which cannot be (temporarily) removed to re-fetch it, without breaking pkgpanda and other tools relying on that exact DC/OS provided python installation.

orsenthil:
@MrMarvin  - I am +1 on this change ` for package in sorted(packages)`


On `--force-reinstall` itself, I am not sure if we want that supported. The entire idea behind pkgpanda is to bring up new system from scratch (by setting things in /opt/mesosphere) and **only once**, this option of doing a `--force-install` of packages doesn't suggest  it's a fresh install. 

It might be useful when we did a mistake of installing packages. But brings us to question the system environment and `--force-install` might not be good answer there.

MrMarvin:
Thanks @orsenthil for your comments!

> The entire idea behind pkgpanda is to bring up new system from scratch (by setting things in /opt/mesosphere) and only once [...]

We do support in-place upgrades via `dcos_node_upgrade.sh`, which makes use of pkgpanda to fetch and then activate potentially newer package versions. The problems I linked surfaced by us using this supported feature, which currently has no (clear) error handling.

In my opinion if we support upgrading a DC/OS installation via `pkgpanda fetch`, we also need to support reinstalling already installed packages.

> It might be useful when we did a mistake of installing packages. But brings us to question the system environment and --force-install might not be good answer there.

Agreed, however having to completely reinstall doesn't seem to be a good answer either, especially if it is easily solved by reinstalling a single component.

jgehrcke:
@MrMarvin do you find time to respond to the review feedback?

MrMarvin:

PR #2068, opened 2017-11-01
Bump used sec-ci-libs to v2, adjust Jenkinsfiles

karlkfi:
## High Level Description
(clone of https://github.com/mesosphere/dcos/pull/49)

This PR adjust project's Jenkinsfiles in order to support the new way in which `sec-ci-lib` is called.

In order to enable other teams to reuse `sec-ci-lib`, both slack channel and slack token ID must be passed as parameter. This also requires a new approach on versioning `sec-ci-lib` libs - see the discussion in https://github.com/mesosphere/sec-ci-libs/pull/5 for more details.

## Related PRs
PR that introduces the changehttps://github.com/mesosphere/sec-ci-libs/pull/5
AR-NEXT #5 PR: <TBD>
DC/OS EE PR: https://github.com/mesosphere/dcos-enterprise/pull/1408


Jenkinsfile
```diff
@@ -11,9 +11,9 @@ if (master_branches.contains(env.BRANCH_NAME)) {
     ])
 }
 
-task_wrapper('mesos-sec', master_branches){
+task_wrapper('mesos-sec', master_branches, '8b793652-f26a-422f-a9ba-0d1e47eb9d89', '#dcos-security-ci') {
```
asridharan: @karlkfi is it possible to add a comment here to explain where we got the token `8b793652-f26a-422f-a9ba-0d1e47eb9d89` from?

vespian: @asridharan It is described in the `sec-ci-libs` documentation: https://github.com/mesosphere/sec-ci-libs#standardised-task-wrapper

karlkfi:
`continuous-integration/jenkins/pr-head` failure is the tests this PR fixes, but it uses the one on master to test with. That's why it's not passing here.

karlkfi:
Updating branch to fix `teamcity/dcos/test/aws/cloudformation/simple`

Jenkinsfile
karlkfi: Seems like it's a token passed to jenkins so that jenkins knows which credentials to use from its secret store. But it's not very maintenance-friendly as a magic string....
It should probably be either a string variable ro an environment variable injected by jenkins.

karlkfi:
- Since this PR only changed the Jenkinsfile and `continuous-integration/jenkins/pr-head` is the only thing affected by the Jenkinsfile we can be assured that this change wont break anything else.
- Since `continuous-integration/jenkins/pr-head` wont pass until this fix is on master, there's no way to get it to pass here.
- @vespian has confirmed that this is will fix the breakage caused by https://github.com/mesosphere/sec-ci-libs/pull/5

So I will merge after the `teamcity/dcos/test/docker` (dcos-docker integration tests) pass, rather than waiting for the rest of the tests.

orsenthil:
Looks like there a separate Jenkinsfile in dcos-enterprise : https://github.com/mesosphere/dcos-enterprise/blob/master/Jenkinsfile  so percolating this to dcos-enterprise's dcos-image is strictly isn't necessary.

karlkfi:
dcos-docker integration tests passed. I'm going to merge now.

PR #2145, opened 2017-11-30
Admin Router: more robust devkit building

vespian:
## High-level description

This PR enables more robust devkit building.

As DCOS-19707 has shown, in some rare cases it is possible that the intermediate devkit container still lingers while the next CI is launched on the Agent. Changes made here should prevent it from happening for both gen_extra and Admin Router devkits.

Even though this PR only touches/changes Admin Router Makefile, I have named both EE and Open PRs the same for clarity. The EE PR adjusts the `gen_extra` devkit, the EE Admin Router's Makefile inherits the changes from the Open Adminrouter's one.

See the Jira and its comments for more details.

 * https://jira.mesosphere.com/browse/DCOS-19707 `continuous-integration/jenkins/pr-head fails on seemingly unrelated changes`

## Related PRs:

EE DC/OS PR: https://github.com/mesosphere/dcos-enterprise/pull/1784

### DIFF
--- a/packages/adminrouter/extra/src/Makefile.common
+++ b/packages/adminrouter/extra/src/Makefile.common
@@ -41,7 +41,7 @@ DEVKIT_DOCKER_OPTS := $(DEVKIT_BASE_DOCKER_OPTS) \
 .PHONY: clean-devkit-container
 clean-devkit-container:
 	@# Let's suppress ALL the output:
-	@docker rm -vf $(DEVKIT_NAME) &> /dev/null || true
+	@docker rm -vf $(DEVKIT_NAME) || true
 
 .PHONY: clean-containers
 clean-containers: clean-devkit-container
@@ -56,26 +56,11 @@ devkit:
 	@if $$(docker images | grep mesosphere/$(DEVKIT_NAME) | grep -q full); then \
 		echo "Devkit image is present, not rebuilding it."; \
 	else \
-		echo "Building devkit image"; \
-		docker build \
-			--rm --force-rm \
-			-t mesosphere/$(DEVKIT_NAME):noresty \
-			-f ../../docker/Dockerfile \
-				../../docker/ && \
-		docker run \
-			$(DEVKIT_BASE_DOCKER_OPTS) \
-			mesosphere/$(DEVKIT_NAME):noresty \
-				/bin/bash -c "\$$OPENRESTY_COMPILE_SCRIPT" && \
-		docker commit $$(docker ps -a -q -f name=$(DEVKIT_NAME)) \
-			mesosphere/$(DEVKIT_NAME):full && \
-		docker rm $(DEVKIT_NAME) && \
-		docker rmi -f mesosphere/$(DEVKIT_NAME):noresty; \
+		$(MAKE) update-devkit; \
 	fi
 
 .PHONY: update-devkit
 update-devkit: clean-devkit-container
-	@# Let's suppress ALL the output:
-	@docker rm $(DEVKIT_NAME) &> /dev/null || true
 	@docker build \
 		--rm --force-rm \
 		-t mesosphere/$(DEVKIT_NAME):noresty \
@@ -87,18 +72,18 @@ update-devkit: clean-devkit-container
 			/bin/bash -c "\$$OPENRESTY_COMPILE_SCRIPT" && \
 	docker commit $$(docker ps -a -q -f name=$(DEVKIT_NAME)) \
 		mesosphere/$(DEVKIT_NAME):full
-	@docker rm $(DEVKIT_NAME)
+	@docker rm -f $(DEVKIT_NAME)
 	@docker rmi -f mesosphere/$(DEVKIT_NAME):noresty
 
 .PHONY: shell
-shell: clean-devkit-container devkit
+shell: devkit clean-containers
 	@docker run --rm -it \
 		$(DEVKIT_DOCKER_OPTS) \
 		--privileged \
 		mesosphere/$(DEVKIT_NAME):full /bin/bash
 
 .PHONY: test
-test: clean-devkit-container devkit
+test: devkit clean-containers
 	@docker run \
 		$(DEVKIT_DOCKER_OPTS) \
 		--privileged \
@@ -107,7 +92,7 @@ test: clean-devkit-container devkit
  		"
 
 .PHONY: api-docs
-api-docs: clean-devkit-container devkit
+api-docs: devkit clean-containers
 	@mkdir -p $(SRC_DIR)/$(API_DOCS_DIR)
 	@for prefix in nginx.master nginx.agent ; do \
 		echo "Generating $(API_DOCS_DIR)/$${prefix}.yaml" >&2 && \
@@ -120,7 +105,7 @@ api-docs: clean-devkit-container devkit
 
 .PHONY: check-api-docs
 DOCS_TEXT := See https://github.com/dcos/dcos/tree/master/packages/adminrouter/extra/src\#endpoints-documentation for more info.
-check-api-docs: api-docs
+check-api-docs: api-docs clean-containers
 ifneq ($(shell git update-index --refresh; echo $$?),0)
 	$(error "Found local changes. Please run `make api-docs` and commit the result. $(DOCS_TEXT)")
 endif
@@ -133,7 +118,7 @@ endif
 	@echo 'No changes found -- Admin Router API docs are up to date.' >&2
 
 .PHONY: lint
-lint: clean-devkit-container devkit
+lint: devkit clean-containers
 	@docker run \
 		$(DEVKIT_DOCKER_OPTS) \
 		mesosphere/$(DEVKIT_NAME):full /bin/bash -x -c "flake8"


vespian:
@mesosphere-mergebot bump ee

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot label ready-for-review

jgehrcke:
@vespian can you inspect the CI failures here for known instabilities?

adamtheturtle:
Does this mean that we will see "Error response from daemon: No such container:" in the logs when there's no problem? I say that because it seems like we're being extra cautious as of "I do not know what exactly causes that".

If so, perhaps send that error to `/dev/null` and add a comment.

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -68,14 +69,14 @@ devkit:
...
-		docker rm $(DEVKIT_NAME) && \
+		docker rm -f $(DEVKIT_NAME) && \
 		docker rmi -f mesosphere/$(DEVKIT_NAME):noresty; \
 	fi
 
 .PHONY: update-devkit
 update-devkit: clean-devkit-container
 	@# Let's suppress ALL the output:
-	@docker rm $(DEVKIT_NAME) &> /dev/null || true
+	@docker rm -f $(DEVKIT_NAME) &> /dev/null || true
```
adamtheturtle: Is this not handled by `clean-devkit-container`?

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -57,6 +57,7 @@ devkit:
 		echo "Devkit image is present, not rebuilding it."; \
 	else \
 		echo "Building devkit image"; \
+		docker rm -f $(DEVKIT_NAME) ; \
```
adamtheturtle: My understanding from JIRA is that we have not encountered any issues with the container existing at this point. Is this extra caution or is there a case we can get here?

Can we have container `$(DEVKIT_NAME)` without image `mesosphere/$(DEVKIT_NAME)`? If not, is there a race condition between getting to `else` and building something? I'd really like a clarifying comment.

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -68,14 +69,14 @@ devkit:
...
 		docker commit $$(docker ps -a -q -f name=$(DEVKIT_NAME)) \
 			mesosphere/$(DEVKIT_NAME):full && \
-		docker rm $(DEVKIT_NAME) && \
+		docker rm -f $(DEVKIT_NAME) && \
 		docker rmi -f mesosphere/$(DEVKIT_NAME):noresty; \
 	fi
 
 .PHONY: update-devkit
 update-devkit: clean-devkit-container
 	@# Let's suppress ALL the output:
```
adamtheturtle: ❓ Existing: This says that all output will be suppressed, but `docker commit` is not.

adamtheturtle:
Please comment on the failing `teamcity/dcos/test/docker`.

packages/adminrouter/extra/src/Makefile.common
vespian: > My understanding from JIRA is that we have not encountered any issues with the container existing at this point. Is this extra caution or is there a case we can get here?

This is me being extra cautious. We have not fully understood the root cause, hence we put the removal code even in the places where (in theory) it should not be needed.

> Can we have container $(DEVKIT_NAME) without image mesosphere/$(DEVKIT_NAME)? If not, is there a race condition between getting to else and building something? I'd really like a clarifying comment.

It depends. CI runs both the intermediate container (i.e. `mesosphere/$(DEVKIT_NAME):noresty`) and the final container (i.e. `mesosphere/$(DEVKIT_NAME):full`) as `$(DEVKIT_NAME)`. No, you can't have `$(DEVKIT_NAME)` running without *both* of these images, you can have it if one or the other (or both) have not been properly cleaned up. The `make devkit` target call may happen after the devkit container has been build (or the build was interrupted while the intermediate stage container was still there) as the Jenkins agent can be reused. I am not sure if I answered your question. Please clarify if I did not.

packages/adminrouter/extra/src/Makefile.common
vespian: It does not have to be explicitly silenced as it is linked to the previous statement with the `&&` and for the `make` command this two count as one.

packages/adminrouter/extra/src/Makefile.common
vespian: It is, I will remove it from here.

packages/adminrouter/extra/src/Makefile.common
adamtheturtle: > No, you can't have $(DEVKIT_NAME) running without both of these images, you can have it if one or the other (or both) have not been properly cleaned up.

If I understand this correctly, the code says:

```
if Docker image mesosphere/$(DEVKIT_NAME):
   return
else:
   remove container $(DEVKIT_NAME)
```

This can be simplified to:

```
if not Docker image mesosphere/$(DEVKIT_NAME):
    remove container $(DEVKIT_NAME)
```

my question was "can `container $(DEVKIT_NAME)` exist if "Docker image `mesosphere/$(DEVKIT_NAME)`" does not exist.

If not, change adds nothing.

adamtheturtle: Separately, maybe we can do:

```
else
    $(MAKE) update-devkit
```

to reduce duplication / bug surface area.

adamtheturtle: or if not:

```
else
    $(MAKE) clean-devkit-container
```

vespian:
@mesosphere-mergebot bump-ee

vespian:
@adamtheturtle Ready for another pass

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -56,26 +56,13 @@ devkit:
...
-				../../docker/ && \
-		docker run \
-			$(DEVKIT_BASE_DOCKER_OPTS) \
-			mesosphere/$(DEVKIT_NAME):noresty \
-				/bin/bash -c "\$$OPENRESTY_COMPILE_SCRIPT" && \
-		docker commit $$(docker ps -a -q -f name=$(DEVKIT_NAME)) \
-			mesosphere/$(DEVKIT_NAME):full && \
-		docker rm $(DEVKIT_NAME) && \
-		docker rmi -f mesosphere/$(DEVKIT_NAME):noresty; \
+		$(MAKE) update-devkit \
```
adamtheturtle: ❤️

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -56,26 +56,13 @@ devkit:
...
 	fi
 
 .PHONY: update-devkit
 update-devkit: clean-devkit-container
 	@# Let's suppress ALL the output:
-	@docker rm $(DEVKIT_NAME) &> /dev/null || true
+	@docker rm -f $(DEVKIT_NAME) || true
```
adamtheturtle: I think that you said that you would remove this, as it is covered by `clean-devkit-container`

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/Makefile.common
vespian: oh, missed that one :( Fixing.

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -87,18 +74,18 @@ update-devkit: clean-devkit-container
 			/bin/bash -c "\$$OPENRESTY_COMPILE_SCRIPT" && \
 	docker commit $$(docker ps -a -q -f name=$(DEVKIT_NAME)) \
 		mesosphere/$(DEVKIT_NAME):full
-	@docker rm $(DEVKIT_NAME)
+	@docker rm -f $(DEVKIT_NAME)
```
adamtheturtle: What do you think of making this `$(MAKE) clean-devkit-container`? To me this would be more clear but also allow us to add changes in just one place.

Saying that - why do we need this at all?

If we do need it:

> This, in turn, stems from the fact that the `update-devkit` target does not call `clean-containers` target just like AR CI does[3]. I will adjust the PR in order to prevent this scenario from happening again.

from JIRA tells me that the problem wasn't that `-f` didn't exist, so maybe this change isn't needed?

packages/adminrouter/extra/src/Makefile.common
```diff
@@ -41,7 +41,7 @@ DEVKIT_DOCKER_OPTS := $(DEVKIT_BASE_DOCKER_OPTS) \
 .PHONY: clean-devkit-container
 clean-devkit-container:
 	@# Let's suppress ALL the output:
-	@docker rm -vf $(DEVKIT_NAME) &> /dev/null || true
+	@docker rm -vf $(DEVKIT_NAME) || true
```
adamtheturtle: BTW - I believe that `-` at the beginning of a line is the `make` way of saying "ignore errors" so you don't need `|| true`, but I'm not a `make` expert.

packages/adminrouter/extra/src/Makefile.common
vespian: > What do you think of making this $(MAKE) clean-devkit-container? To me, this would be more clear but also allow us to add changes in just one place.

I am OK with that if you insist. For me, the mental extra effort of going through another layer of abstraction (i.e. calling different makefile target) is not worth avoiding coping just one line of code. Let me know what suits you best.

> Saying that - why do we need this at all?
I would like the Makefile to clean after itself and not to leave artefacts lying around, it does not seem right. We cannot launch `docker run` (see 3 lines earlier) with `--rm` as we need to commit the container, so the idea is that we clean the container up right after it has been committed to the new image.

> ...from JIRA tells me that the problem wasn't that -f didn't exist, so maybe this change isn't needed?

This is just a part of `making it more robust`. Theoretically, I do not think there could be a situation when the container would still be running and thus the `-f` flag would be still required, but OTOH adding this option does not hurt and *may* at some point prevent false negatives. Let me know if you have a strong opinion on that.

packages/adminrouter/extra/src/Makefile.common
vespian: IIRC `-` and `@` does not play well together, hence I mute the output with `/dev/null` redirection.

packages/adminrouter/extra/src/Makefile.common
adamtheturtle: > I am OK with that if you insist

I don't insist. Changing it is my preference. However, you are the maintainer of this and use it more than me, and it is not a strong preference so I will instead leave it to your preference.

I also don't have a strong opinion on the other points and am happy with you making the judgement call. I also like the Makefile cleaning up after itself. As a reviewer I wanted to separate: "What do we thing is necessary to solve the issue" vs "What is extra safety". This is now clear and I'm happy.

vespian: > I will instead leave it to your preference.

Thank you, let's please leave it as then.

packages/adminrouter/extra/src/Makefile.common
adamtheturtle:
```
Actually, @- and -@ both do work, but will print a make: [target] Error 1 (ignored) warning.

Instead, you can use

@command || true
or, since : is shorthand for true in shell,

@command ||:
```

https://stackoverflow.com/a/11872095 - TIL!

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

PR #2199, opened 2017-12-13
[DCOS-19976] Make outside-facing TLS configuration of Master Admin Router configurable

keith-mcclellan:
## High-level description

This is based on NIST guidance that all government and regulated bodies
should have migrated to TLS v1.2 by May 30, 2018.  The default
handshake is already TLS v1.2 and all major browsers have supported
v1.2 for several years.  

However, since some customers may need legacy support, so this change allows the customer to override the default with a new minimum TLS version (v1, v1.1, or v1.2).

NIST guidance also requires disabling of ciphers below 128 bits, so I added a way to override the default ciphers if required.

[DCOS_OSS-2029](https://jira.mesosphere.com/browse/DCOS_OSS-2029) | Introduce adminrouter_tls_1_x_enabled install-time options
[DCOS_OSS-2030](https://jira.mesosphere.com/browse/DCOS_OSS-2030) | Introduce adminrouter_tls_cipher_suite install-time option

## Related tickets (optional)

[Closed] DCOS-10596 Master Admin Router: disable TLS 1.0 support
[Closed] DCOS-16633 Admin Router: only allow TLS 1.2 for Agent AR (separate master/agent TLS configs)
[Closed] DCOS_OSS-444 Admin Router: disable TLS 1.0 support by default

## Checklist for all PRs

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): [adminrouter](https://github.com/dcos/dcos/pull/2199/files)
  - [x] Test Results: AdminRouter is a core component of dcos, so see CI test results below


keith-mcclellan:
@mesosphere-mergebot label [Work In Progress]

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot label [Ready for Review]

keith-mcclellan:
@jgehrcke I'd appreciate it if someone from your team took a look at this for me. Thanks!!

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@mesosphere-mergebot bump-ee

keith-mcclellan:
@jgehrcke does someone on your team have some time this week to review please?

jgehrcke:
> @jgehrcke does someone on your team have some time this week to review please?

@mhrabovcin is on it :)

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -57,7 +169,7 @@ class TestToggleTLS1:
     def supported_ssl_protocols(self, new_config_arguments) -> List[str]:
```
mhrabovcin: Lets rename this function to include `master` in name to indicate that it checks for protocol version support in AR master configuration.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,25 +29,137 @@ def test_default(self):
...
+class TestSetCipherOverride:
     """
-    Tests for toggling TLS 1.0.
+    Tests for setting ssl_ciphers
+
+    To test manually, either use openssl commands or sslscan
+    [https://github.com/rbsec/sslscan]
+    """
+
+    def supported_ssl_ciphers_master(self, new_config_arguments) -> List[str]:
```
mhrabovcin: Can you add type hint for `new_config_arguments` ?

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,25 +29,137 @@ def test_default(self):
...
+            # We strip whitespace from the beginning of the line as NGINX
+            # configuration lines can start with whitespace.
+            line.lstrip().startswith('ssl_ciphers ')
+        ]
+        ssl_ciphers_line = ssl_ciphers_line.strip(';')
+        ciphers = ssl_ciphers_line.split()[1:]
+        return ciphers
 
-    To manually test that this is, in fact, a working toggle for TLS 1.0, use
+    def supported_ssl_ciphers_agent(self, new_config_arguments) -> List[str]:
```
mhrabovcin: Both `supported_ssl_ciphers_master` and `supported_ssl_ciphers_agent` are almost identical. Can we extract single function and keep existing one calling the new function.

gen/calc.py
```diff
@@ -547,6 +547,37 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+
+
+def calculate_adminrouter_tls_cipher_override(adminrouter_tls_cipher_suite):
+    if adminrouter_tls_cipher_suite != '':
+        return 'true'
+    else:
+        return 'false'
+
+
+def validate_adminrouter_tls_version_override(adminrouter_tls_version_override, adminrouter_tls_1_0_enabled,
```
mhrabovcin: I believe that `adminrouter_tls_1_0_enabled`, `adminrouter_tls_1_1_enabled` and `adminrouter_tls_1_2_enabled` are string variables with `'true'` and `'false'` content. `any()` function returns false only on empty array or all elements of the array are boolean `False`.

```
assert any(['false', 'false']) == True
```

vs

```
assert any([False, False]) == False
```
We need to fix this logic and add appropriate unit test that checks if the message is issued under expected conditions.

gen/calc.py
```diff
@@ -547,6 +547,37 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+                                              adminrouter_tls_1_1_enabled, adminrouter_tls_1_2_enabled):
+    if adminrouter_tls_version_override == '':
+        assert any([adminrouter_tls_1_0_enabled, adminrouter_tls_1_1_enabled, adminrouter_tls_1_2_enabled]) == 'false',\
```
mhrabovcin: Lets break down this single line assert into variables to improve readability.

Pseudocode:
```
msg = (
  "When not explicitly setting a tls_version_override, "
   "at least one tls boolean (tls_1_0, tls_1_1, tls_1_2) must be set."
)
has_some_tls_option = any(... fix logic here ...)
assert has_some_tls_option, msg
```

gen/calc.py
keith-mcclellan: I forgot that I need an in clause in any when using strings - not sure why we don't use booleans in calc.py generally, I just followed the other examples.  Will fix in a follow up commit.

gen/calc.py
keith-mcclellan: I'm happy to do this, but the syntax I used is consistent with other examples in calc.py - see line 541 (validate_exhibitor_storage_master_discovery). Please advise.

mhrabovcin:
@mesosphere-mergebot bump-ee

mhrabovcin:
@mesosphere-mergebot bump-ee

mhrabovcin:
@mesosphere-mergebot bump-ee

mhrabovcin:
@jgehrcke @adamtheturtle could you please review this one?

gen/calc.py
```diff
@@ -920,6 +972,11 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
         'weights': '',
         'adminrouter_auth_enabled': calculate_adminrouter_auth_enabled,
         'adminrouter_tls_1_0_enabled': 'false',
+        'adminrouter_tls_1_1_enabled': 'true',
+        'adminrouter_tls_1_2_enabled': 'true',
+        'adminrouter_tls_version_override': calculate_adminrouter_tls_version_override,
+        'adminrouter_tls_cipher_override': calculate_adminrouter_tls_cipher_override,
+        'adminrouter_tls_cipher_suite': '',
```
jgehrcke: For documentation it is important to note that both `_override` options are fully internal, i.e. they are not supposed to be set by operators in the config YAML document. Correct?

keith-mcclellan: technically the TLS version override could be overwritten by the user (I did this in case someone needed TLSv1.3 at some point so we'd have a workaround before building a formal interface).  The other one shouldn't ever be set by the user.

jgehrcke: Ah, ok, thanks for clarifying the intent.

Having both, `adminrouter_tls_1_x_enabled` as well as `adminrouter_tls_version_override` as public configuration variables seems to create unnecessary complexity.

Especially in configuration interfaces there should be one way to achieve a given thing. Unnecessary complexity is not so good in terms of using, testing, reasoning, documentation.

I think we have (had) two choices:
1) Use `adminrouter_tls_1_x_enabled`, and once we support TLS 1.Y we can introduce `adminrouter_tls_1_Y_enabled` (rare event).
2) Deprecate `adminrouter_tls_1_0_enabled` and have *only* a single configuration parameter for the entire version specification string.

A mixture of both I think is not healthy.

Can we think of a use case where a mixture enables something that (1) or (2) alone would not enable?

mhrabovcin:
@mesosphere-mergebot bump-ee

gen/calc.py
mhrabovcin: I've decided to go with option (1) and remove `adminrouter_tls_version_override` configuration option.

mhrabovcin:
@mesosphere-mergebot bump-ee

gen/dcos-config.yaml
```diff
@@ -985,23 +985,30 @@ package:
...
       ssl_certificate_key includes/snakeoil.key;
-  - path: /etc/adminrouter-tls.conf
+  - path: /etc/adminrouter-tls-master.conf
     content: |
       # Ref: https://github.com/cloudflare/sslconfig/blob/master/conf
       # Modulo ChaCha20 cipher.
-      ssl_ciphers EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5;
+{% switch adminrouter_tls_cipher_override %}
+{% case "false" %}
+      ssl_ciphers EECDH+AES256:RSA+AES256:EECDH+AES128:RSA+AES128:EECDH+3DES:RSA+3DES:!MD5;
```
jgehrcke: I only now saw that [this commit|https://github.com/dcos/dcos/pull/2199/commits/63f60a3321175cf702ff919b44c480dbb7f1c0de] 'sneaks' in a change of the default cipher suite. We change from

> EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5

to 

> EECDH+AES256:RSA+AES256:EECDH+AES128:RSA+AES128:EECDH+3DES:RSA+3DES:!MD5

I do not see this discussed or motivated properly. It also renders the corresponding comment invalid (what we have so far is indeed the Cloudflare config from 2016 with the mentioned diff). @mhrabovcin I'd very much appreciate if you could revert that. That's a matter of following process.

CC @keith-mcclellan that's something to keep in mind for the future: let's treat independent what's independent :-). Overall, that minimizes risk and also smoothens integration.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -104,17 +254,44 @@ def test_default(self, new_arguments):
...
 
-    def test_enable(self):
+    @pytest.mark.parametrize(
+        'enabled,expected_protocols', [
+            (('false', 'false', 'true'), ['TLSv1.2']),
+            (('false', 'true', 'true'), ['TLSv1.1', 'TLSv1.2']),
+            (('true', 'true', 'true'), ['TLSv1', 'TLSv1.1', 'TLSv1.2']),
+            (('true', 'false', 'true'), ['TLSv1', 'TLSv1.2']),
+            (('true', 'false', 'false'), ['TLSv1']),
+        ]
```
jgehrcke: Well, quite a few of these combinations are not technically sound. That must be kept in mind -- not all of these need to "work". That is, this test can be considered overly fine-grained and might get in the way of developers in the future. Fine for now, but important to keep in mind.

gen/dcos-config.yaml
mhrabovcin: Reverted.

mhrabovcin:
@mesosphere-mergebot bump-ee

gen/calc.py
```diff
@@ -547,6 +547,53 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+        adminrouter_tls_1_2_enabled):
+
+    tls_version_flags = [
+        adminrouter_tls_1_0_enabled,
+        adminrouter_tls_1_1_enabled,
+        adminrouter_tls_1_2_enabled,
+    ]
+
+    enabled_tls_flags_count = len(
+        list(filter(lambda x: x == 'true', tls_version_flags)))
```
jgehrcke: List comprehension, for next time :-)
```python
>>> l = [0, 1, 1, 0, 0]
>>> len([i for i in l if i == 1])
2
```

gen/dcos-config.yaml
```diff
@@ -985,23 +985,30 @@ package:
...
-{% switch adminrouter_tls_1_0_enabled %}
-{% case "true" %}
-      ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-{% case "false" %}
-      ssl_protocols TLSv1.1 TLSv1.2;
-{% endswitch %}
+      ssl_protocols {{ adminrouter_tls_version_override }};
+  - path: /etc/adminrouter-tls-agent.conf
+    content: |
+      # Note that Agent Admin Router only servers cluster-internal clients. Hence,
```
jgehrcke: `s/servers/serve`

If that is too much pain to change then let's punt on it

gen/calc.py
mhrabovcin: fixed.

gen/dcos-config.yaml
mhrabovcin: Fixed.

mhrabovcin:
@mesosphere-mergebot bump-ee

adamtheturtle:
This PR is linked to an Epic. This PR does not close out that epic - it still includes documentation work for example. Can this PR please be linked to an issue which describes the exact scope that this aims to address.

jgehrcke:
> This PR is linked to an Epic. This PR does not close out that epic - it still includes documentation work for example. Can this PR please be linked to an issue which describes the exact scope that this aims to address.

I am doing that, please check the descr in 1 min.

gen/calc.py
```diff
@@ -547,6 +547,53 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+        tls_versions.append('TLSv1.2')
+
+    tls_version_string = " ".join(tls_versions)
+    return tls_version_string
+
+
+def calculate_adminrouter_tls_cipher_override(adminrouter_tls_cipher_suite):
+    if adminrouter_tls_cipher_suite != '':
+        return 'true'
+    else:
```
adamtheturtle: ❓ ( <- means "optional") nit: no need for `else`.

gen/calc.py
```diff
@@ -547,6 +547,53 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+        adminrouter_tls_1_2_enabled,
+    ]
+
+    enabled_tls_flags_count = len(
+        [flag for flag in tls_version_flags if flag == 'true'])
+
+    msg = (
+        'At least one tls boolean (adminrouter_tls_1_0_enabled, '
+        'adminrouter_tls_1_1_enabled, adminrouter_tls_1_2_enabled) must '
+        'be set.'
```
adamtheturtle: ```s/be set/be set to `'true` `` or similar - you can set them to "false".

gen/calc.py
```diff
@@ -547,6 +547,53 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+        adminrouter_tls_1_0_enabled,
+        adminrouter_tls_1_1_enabled,
+        adminrouter_tls_1_2_enabled,
+    ]
+
+    enabled_tls_flags_count = len(
+        [flag for flag in tls_version_flags if flag == 'true'])
+
+    msg = (
+        'At least one tls boolean (adminrouter_tls_1_0_enabled, '
```
adamtheturtle: I'm not sure that referring to these as the "tls boolean"s is useful - we could have other TLS booleans which are not part of this. I would say:

> At least one of adminrouter_tls_1_0_enabled, adminrouter_tls_1_1_enabled and adminrouter_tls_1_2_enabled must be set to 'true'.

gen/calc.py
jgehrcke: While we are here:I prefer "At least one of" instead of "At least one tls boolean".

jgehrcke: (can be done in a follow-up!!)

gen/calc.py
jgehrcke: :-)

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -12,15 +12,21 @@ class TestAdminRouterTLSConfig:
...
     """
 
-    def test_default(self):
+    @pytest.mark.parametrize(
+        'adminrouter_tls_1_0_enabled, tls_versions',
+        [('true', 'TLSv1 TLSv1.1 TLSv1.2'), ('false', 'TLSv1.1 TLSv1.2')]
+    )
+    def test_master(self, adminrouter_tls_1_0_enabled, tls_versions):
         """
         By default, the configuration specifies certain TLS settings.
```
adamtheturtle: This docstring is now incorrect - this does not test what the default is with no extra arguments. That is now a missing test.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,35 +35,159 @@ def test_default(self):
...
             # details.
+            ssl_protocols {tls_versions};
+            """.format(tls_versions=tls_versions)
+        )
+        assert config['content'] == expected_configuration
+
+    @pytest.mark.parametrize('adminrouter_tls_1_0_enabled', ['true', 'false'])
+    def test_agent(self, adminrouter_tls_1_0_enabled):
+        """
+        By default, the configuration specifies certain TLS settings.
```
adamtheturtle: This does not test the default, but the docstring says that it does.

mhrabovcin:
@mesosphere-mergebot bump-ee

mhrabovcin:
@adamtheturtle I've addressed your comments.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -12,15 +12,18 @@ class TestAdminRouterTLSConfig:
     Tests for the Admin Router TLS Config creation.
     """
```
adamtheturtle: This is missing test cases for the defaults. As Martin is away, I will create a JIRA issue.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -12,15 +12,18 @@ class TestAdminRouterTLSConfig:
...
+        'adminrouter_tls_1_0_enabled, tls_versions',
+        [('true', 'TLSv1 TLSv1.1 TLSv1.2'), ('false', 'TLSv1.1 TLSv1.2')]
+    )
+    def test_master(self, adminrouter_tls_1_0_enabled, tls_versions):
         """
-        By default, the configuration specifies certain TLS settings.
-
-        This test is a sanity check for the configuration template logic
-        rather than a particularly useful feature test.
+        Test that Master Admin Router config file has the correct content.
```
adamtheturtle: I think that the comment:

```
-        This test is a sanity check for the configuration template logic		
 -        rather than a particularly useful feature test.
```

is useful.

That is - what is "correct content"? How is "correct content" chosen and defined?

What we really mean is that we have tested this at a higher level, and this file happens to contain certain contents when the TLS settings work as promised. This is a regression test which will give us early indication that a larger test will fail, but that's it.

This comment is misleading.

adamtheturtle: As @mhrabovcin is away, I will create a JIRA issue to address this and other non-critical comments.

packages/adminrouter/extra/src/nginx.agent.conf
```diff
@@ -13,6 +13,6 @@ http {
         include includes/server/open/agent.conf;
 
         include /opt/mesosphere/etc/adminrouter-listen-open.conf;
-        include /opt/mesosphere/etc/adminrouter-tls.conf;
+        include /opt/mesosphere/etc/adminrouter-tls-agent.conf;
```
adamtheturtle: This means that any manual changes made to `/opt/mesosphere/etc/adminrouter-tls.conf` will not apply.

@jgehrcke are we sure enough that this is a private-enough interface that we do not need to worry about this?

gen/dcos-config.yaml
```diff
@@ -985,23 +985,30 @@ package:
...
-{% case "true" %}
-      ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-{% case "false" %}
-      ssl_protocols TLSv1.1 TLSv1.2;
-{% endswitch %}
+      ssl_protocols {{ adminrouter_tls_version_override }};
+  - path: /etc/adminrouter-tls-agent.conf
+    content: |
+      # Note that Agent Admin Router only serves cluster-internal clients. Hence,
+      # browser compatibility is not a criterion for the TLS cipher suite selection.
```
adamtheturtle: It would be nice to know what the criteria are for cipher suite selection.

packages/adminrouter/extra/src/nginx.agent.conf
jgehrcke: Everything in /opt/mesosphere is private. In other words:
- `/opt/mesosphere` should be considered immutable
- manual modifications in `/opt/mesosphere` are not supported

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -12,15 +12,18 @@ class TestAdminRouterTLSConfig:
     Tests for the Admin Router TLS Config creation.
     """
 
-    def test_default(self):
+    @pytest.mark.parametrize(
```
adamtheturtle: These tests do not cover what happens when changing `adminrouter_tls_1_1_enabled` or `adminrouter_tls_1_2_enabled`.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,35 +32,156 @@ def test_default(self):
...
+        """
+        new_arguments = {'adminrouter_external_cipher_override': 'false'}
+        ciphers = self.supported_ssl_ciphers_agent(
+            new_config_arguments=new_arguments,
+        )
+        assert ciphers == ['EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:!MD5']
+
+    def test_cipher_master_default(self):
+        """
+        The config variable adminrouter_external_cipher_string must not be set.
```
adamtheturtle: I guess that this is an incorrect docstring. At least, it is not clear.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,35 +32,156 @@ def test_default(self):
...
+        """
+        new_arguments = {'adminrouter_external_cipher_string': ''}
+        ciphers = self.supported_ssl_ciphers_master(
+            new_config_arguments=new_arguments,
+        )
+        assert ciphers == ['EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5']
+
+    def test_cipher_master_custom(self):
+        """
+        The config variable adminrouter_external_cipher_string must be set
```
adamtheturtle: I guess that this is an incorrect docstring. At least, it is not clear.

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -104,17 +248,45 @@ def test_default(self, new_arguments):
...
+        expected_error_msg = (
+            'At least one of adminrouter_tls_1_0_enabled, '
+            'adminrouter_tls_1_1_enabled and adminrouter_tls_1_2_enabled must '
+            "be set to 'true'."
         )
-        assert protocols == ['TLSv1', 'TLSv1.1', 'TLSv1.2']
+        result = gen.validate(arguments=make_arguments(new_arguments))
+        assert result['status'] == 'errors'
+
+        key = 'adminrouter_tls_1_2_enabled'
```
adamtheturtle: Why test just this key and not all of them?

gen/tests/test_adminrouter_tls_conf.py
```diff
@@ -29,35 +32,156 @@ def test_default(self):
...
+        assert ciphers == ['EECDH+AES128:RSA+AES128']
+
+
+class TestToggleTLSVersions:
     """
-    Tests for toggling TLS 1.0.
+    Tests for toggling TLS 1.0/1.1.
 
-    To manually test that this is, in fact, a working toggle for TLS 1.0, use
+    To manually test that this is, in fact, a working toggle for TLS 1.0/1.1, use
```
adamtheturtle: These tests also cover version 1.2.

gen/tests/test_adminrouter_tls_conf.py
adamtheturtle: Or maybe this is covered by `test_enable_custom_tls_versions`, in which case, either add a comment or remove the parametrisation.

gen/calc.py
```diff
@@ -547,6 +547,53 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
...
+        tls_versions.append('TLSv1.1')
+
+    if adminrouter_tls_1_2_enabled == 'true':
+        tls_versions.append('TLSv1.2')
+
+    tls_version_string = " ".join(tls_versions)
+    return tls_version_string
+
+
+def calculate_adminrouter_tls_cipher_override(adminrouter_tls_cipher_suite):
```
adamtheturtle: So we don't validate the cipher suite at all? When we get to the documentation, let's try to consider what could go wrong if I put "foo" in here.

mhrabovcin:
@mesosphere-mergebot ship_it_label_status_check

jgehrcke:
@mesosphere-mergebot ship_it_label_status_check

jgehrcke:
@mesosphere-mergebot pr-status-checks

jgehrcke:
Approval requested. Let's please not mutate this anymore until we have further feedback.

artemharutyunyan:
Signed-off. Let's Ship It!

PR #2204, opened 2017-12-14
Skip the ip_per_container test when the number of slaves is less than 2.

orsenthil:
## High-level description

We have an 'assertion' in tests which check if the number of slaves > 2.

* https://jira.mesosphere.com/browse/DCOS_OSS-1977

Instead of asserting the requirement, we should skip the tests as not applicable.


orsenthil:
@mesosphere-mergebot bump-ee

orsenthil:

PR #2212, opened 2017-12-15
bootstrap: More robust handling of a zk.pid file while waiting for exhibitor

vespian:
## High-level description

DCOS_OSS-1919 has revealed that it is possible for Zookeeper to not to create zk.pid even though Exhibitor reports that all the ZK instances have joined the quorum.

This change does not solve the root cause of the problem, it just makes the bootstrap handle it more gracefully and it adds some extra logging so that debugging of the underlying issue will be easier next time.

See the Jira issue for more details. Tests will be made green after I get all the required thumbs up. Same goes for porting these changes to `master` and `1.9`.

- https://jira.mesosphere.com/browse/DCOS_OSS-1919 `Terminated Master node in 3 HA cluster, Failed to start.`

## Related PRs

EE PR: https://github.com/mesosphere/dcos-enterprise/pull/1888

### DIFF
--- a/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py
@@ -17,7 +17,11 @@ stash_zk_pid_stat_mtime_path = "/var/lib/dcos/bootstrap/exhibitor_pid_stat"
 
 
 def get_zk_pid_mtime():
-    return os.stat(zk_pid_path).st_mtime_ns
+    try:
+        return os.stat(zk_pid_path).st_mtime_ns
+    except FileNotFoundError:
+        log.error("ZK pid file `%s` does not exist.", zk_pid_path)
+        return None
 
 
 def get_zk_pid():
@@ -34,6 +38,9 @@ def try_shortcut():
 
     # Make sure that the pid hasn't been written anew
     cur_pid_stat = get_zk_pid_mtime()
+
+    if cur_pid_stat is None:
+        return False
 
     if stashed_pid_stat != cur_pid_stat:
         return False
@@ -90,20 +97,24 @@ def wait(master_count_filename):
 
     data = response.json()
 
-    serving = 0
-    leaders = 0
+    serving = []
+    leaders = []
     for node in data:
         if node['isLeader']:
-            leaders += 1
+            leaders.append(node['hostname'])
         if node['description'] == 'serving':
-            serving += 1
+            serving.append(node['hostname'])
 
-    if serving != cluster_size or leaders != 1:
+    log.info(
+        "Serving hosts: `%s`, leader: `%s`", ','.join(serving), ','.join(leaders))
+
+    if len(serving) != cluster_size or len(leaders) != 1:
         msg = 'Expected {} servers and 1 leader, got {} servers and {} leaders'.format(cluster_size, serving, leaders)
         raise Exception(msg)
 
     # Local Zookeeper is up. Config should be stable, local zookeeper happy. Stash the PID so if
     # there is a restart we can come up quickly without requiring a new zookeeper quorum.
-    zk_pid_mtime = str(get_zk_pid_mtime())
-    log.info('Stashing zk.pid mtime %s to %s', zk_pid_mtime, stash_zk_pid_stat_mtime_path)
-    write_string(stash_zk_pid_stat_mtime_path, zk_pid_mtime)
+    zk_pid_mtime = get_zk_pid_mtime()
+    if zk_pid_mtime is not None:
+        log.info('Stashing zk.pid mtime %s to %s', zk_pid_mtime, stash_zk_pid_stat_mtime_path)
+        write_string(stash_zk_pid_stat_mtime_path, str(zk_pid_mtime))


vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot label ReadyForReview

vespian:
@mhrabovcin @gpaul Please have a look

@jgehrcke FYI

packages/bootstrap/extra/dcos_internal_utils/exhibitor.py
```diff
@@ -35,6 +39,9 @@ def try_shortcut():
     # Make sure that the pid hasn't been written anew
     cur_pid_stat = get_zk_pid_mtime()
 
+    if cur_pid_stat is None:
+        return False
+
```
jgehrcke: Where this function is called it previously might have raised FileNotFoundError, and now `False` is returned.

In this patch, I however do not see a consumer changed that handles the case where this function returns `False`.

vespian: Please check the line 75, it is outside of the diff.

The FileNotFound exception being unhandled is the main reason why the bootstrap script was failing. The `try_shortcut()` function should be returning only `True/False`.

jgehrcke: I see below we do 
```
     zk_pid = get_zk_pid()
     cmdline_path = '/proc/{}/cmdline'.format(zk_pid)
```

Can it now happen that we construct a `cmdline_path` with the value `/proc/False/cmdline`?

jgehrcke: Oh wow now I get that we did not modify the function `get_zk_pid()`.

When I wrote

> Where this function is called it previously might have raised 

I assumed that we modify `get_zk_pid()` (as shown by the GitHub diff) and not `try_shortcut()` (hidden by the GitHub diff).

jgehrcke:
Pawel, I would appreciate if you could add a comment here to the GitHub conversation describing why precisely this change would help in precisely which scenario. It's a little hard to understand which case _specifically_ this handles more gracefully, and _how_ it's doing it.

vespian:
> I would appreciate if you could add a comment here to the GitHub conversation describing why precisely this change would help in precisely which scenario.

There seems to be a way for the exhibitor to launch ZK in a way that does not result in ZK PID file being created. Despite numerous attempts, I was unable to recreate this situation but removing ZK PID *without* restarting exhibitor service yielded the very same error log entries as the ones reported by the issue submitter.

After a quick debug of the bootstrap.py code, it turns out that exhibitor.py module used by bootstrap script makes a shortcut while waiting for the local ZK by comparing the ZK PID file's mtime with the one stored internally. The problem is though that the shortcut always assumes that there is a PID file as long as there is a stashed mtime value (see [1] -> [2] -> [3]). This was resulting in FileNotFound exception being thrown as reported by the user.

This PR changes that:
* even if the mtime was stashed by the module, it is ignored by it if the ZK PID is missing. In such case, the normal (i.e. long) wait for the ZK quorum is performed and the "shortcut" is terminated early.
* after the normal wait for the ZK quorum has been performed, the module does not try to stash the ZK PID's mtime if the PID does not exist.

This, of course, does not solve the underlying issue - i.e. why the ZK did not create the PID or created it in the wrong place. But it will allow the bootstrap script to handle the error more gracefully and in cases when the ZK quorum, as seen by Exhibitor, is healthy, it will allow the bootstrap script to proceed, even though the ZK PID is missing. The extra logline that was added will maybe allow us to debug the root cause at some point in the future.

@jgehrcke Does this help?

[1] https://github.com/dcos/dcos/blob/98d0ac1f34737540a1f9b60fe9d4db2aeffc20db/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py#L28-L33
[2] https://github.com/dcos/dcos/blob/98d0ac1f34737540a1f9b60fe9d4db2aeffc20db/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py#L36
[3] https://github.com/dcos/dcos/blob/98d0ac1f34737540a1f9b60fe9d4db2aeffc20db/packages/bootstrap/extra/dcos_internal_utils/exhibitor.py#L20

jgehrcke:
Thanks, that describes it really well @vespian.

Can you please add the two bullet points to the commit message? That will make a lovely commit message explaining the gist of this patch in simple, precise terms.

vespian:
@jgehrcke Done, ready for another pass.

vespian:
Re-run the `teamcity/dcos/test/aws/cloudformation/simple ` - seems to be AWS issue - stack was rolled back.

vespian:
@mesosphere-mergebot override-status  teamcity/dcos/test/aws/cloudformation/simple https://jira.mesosphere.com/browse/DCOS-18825

jgehrcke:
I commented in DCOS-18825.

Quoting myself:

> As of today the CI run `teamcity/dcos/test/aws/cloudformation/simple` is comprised of more than 100 individual tests.

> I do not feel comfortable overriding teamcity/dcos/test/aws/cloudformation/simple or other CI jobs when they failed entirely as of a cluster provisioning error.

I am re-running the job now.

vespian:
@mesosphere-mergebot bump-ee

packages/bootstrap/extra/dcos_internal_utils/exhibitor.py
```diff
@@ -90,20 +97,24 @@ def wait(master_count_filename):
...
         if node['description'] == 'serving':
-            serving += 1
+            serving.append(node['hostname'])
+
+    log.info(
+        "Serving hosts: `%s`, leader: `%s`", ','.join(serving), ','.join(leaders))
 
-    if serving != cluster_size or leaders != 1:
+    if len(serving) != cluster_size or len(leaders) != 1:
         msg = 'Expected {} servers and 1 leader, got {} servers and {} leaders'.format(cluster_size, serving, leaders)
```
adam-mesos: Won't this now print the contents of the serving/leaders arrays instead of their sizes?
cc: @colin-msphere @orsenthil

vespian: My bad, sorry about that.

https://github.com/mesosphere/dcos-enterprise/pull/1967
https://github.com/dcos/dcos/pull/2264

vespian: PRs for `1.9` and `master` have been updated as well.

jgehrcke:
Looks like this needs a little patch? See https://github.com/dcos/dcos/pull/2226#issuecomment-355015444

vespian:
> Looks like this needs a little patch? See #2226 (comment)

See https://github.com/dcos/dcos/pull/2212#discussion_r159220807

PR #2238, opened 2017-12-20
Random task selection for adminrouter

stg-:
## High-level description

Web requests to an scaled Marathon App  through adminrouter go always to the same task (the first one). This PR allows to randomly select the task so the load is more "balanced" between them.

  - [DCOS_OSS-2007](https://jira.mesosphere.com/browse/DCOS_OSS-2007) Adminrouter: Randomly select task.

## Related tickets (optional)

  - NONE

## Checklist for all PRs

___

### DIFF
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -166,9 +166,11 @@ local function fetch_and_store_marathon_apps(auth_token)
           end
        end
 
-       -- next() returns nil if table is empty.
-       local i, task = next(tasks)
-       if i == nil then
+       -- Randomly select the task.
+       local i = math.random(1,#tasks)
+       local task = tasks[i]
+
+       if task == nil then
           ngx.log(ngx.NOTICE, "No task in state TASK_RUNNING for app '" .. appId .. "'")
           goto continue
        end
@@ -553,6 +555,9 @@ local function periodically_refresh_cache(auth_token)
     -- This function is invoked from within init_worker_by_lua code.
     -- ngx.timer.at() can be called here, whereas most of the other ngx.*
     -- API is not available.
+
+    -- Seeding the PRNG
+    math.randomseed(os.time())
 
     timerhandler = function(premature)
         -- Handler for recursive timer invocation.


stg-:
Hi @vespian @gpaul @SlevinBE , could you please take a look at this?

vespian:
@jgehrcke FYI

@stg-  Do you think you could write a test for that change?

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -166,9 +166,12 @@ local function fetch_and_store_marathon_apps(auth_token)
           end
        end
 
-       -- next() returns nil if table is empty.
-       local i, task = next(tasks)
-       if i == nil then
+       -- Randomly select the task.
+       math.randomseed(os.time())
```
jgehrcke: A pseudorandom number generator (PRNG) is supposed to be seeded _once_ for many subsequent pseudorandom number generations.

In fact, that's the designed way for a PNRG to yield a uniform distribution of pseudorandom numbers.

Seeting it before every number generation probably yields a non-uniform distribution.

SlevinBE:
@stg- nice catch! PR looks good to me apart from the already mentioned comments.

@jgehrcke  why shouldn't the service endpoints be used in production? We quite heavily rely on them for custom platform administration services, since they are convenient and allow us to reuse the authentication layer of DC/OS.

jgehrcke:
> why shouldn't the service endpoints be used in production? 

For exposing your apps to users -- especially when you need to load-balance requests across app instances. For that we have Marathon-LB and Edge-LB.

>  We quite heavily rely on them for custom platform administration services

That seems to be fine. In general _Admin_ Router is supposed to be used for operational/admin use cases, and not really for routing heavy application traffic to agents :-).

packages/adminrouter/extra/src/lib/cache.lua
stg-: Good. Do you think the function refresh_mesos_state_cache (in cache empty and cache expired) would be a good place to initialize it?

vespian: What do you think about the beginning of `periodically_refresh_cache` function[1]? It runs in `init_worker_by_lua` context [2].

[1] https://github.com/dcos/dcos/blob/5c019cfdee6a17ea4b2252e7c0e98118bc845a06/packages/adminrouter/extra/src/lib/cache.lua#L552
[2] https://github.com/openresty/lua-nginx-module#init_worker_by_lua

stg-: Done. Thanks!

vespian:
Let's try to ship the test together with the change. If it turns out to be a big task, we can split it into a separate task.

As for the test itself, we can try following:
* there are already two mock applications [1] that we can re-use here. The only problem is that the second one (`..._DIFFERENTPORT`) Is not enabled by default. We can fix it by issuing a command to the mock that will set Marathon reply to include both tasks when queried by Admin Router. An example how it can be done can be found in [2].
* in order to control the data that AR uses for `/service` endpoint routing, we should purge the data from Mesos_DNS and Mesos mocks, e.g. [3]
* we also need to make sure that AR has loaded the Mock data we prepared for it and only it. This can be achieved by using `GuardedSubprocess` context manager will launch AR instance only after all the mock configuration has been set [4]
* the tricky part is how we want to make the requests. As @jgehrcke pointed out testing uniformity of a random numbers generator may not be the best idea, but OTOH we want to make the test reliable. With N requests it is possible for all of them to land on a single. In order to better simulate the real-life scenario, we can also try to issue requests in parallel. What do you think about issuing multiple requests in a predefined time window and just checking if both Tasks mock got at least one request and that the ratio of the requests served by first mock to the requests served by the second mock is not greater than 2 (ideally it would equal 2 so we should be fine). As for the requests themselves, the test harness is able to call Vegeta HTTP benchmarking tool [5]. You can find an example use of it in [6].
* the final thing would be to gather the results from the mocks and compare the number of requests. This is done e.g. in [7].

This is of course just a suggestion :) Let me know if we can further improve something or anything is unclear. Also, I recommend the tests writing documentation for AR[8] if you would like to get more complex on the functions/mocks/techniques I mentioned here.

Thanks!

[1] https://github.com/dcos/dcos/blob/3520215eeb1054039f4f94e8e153785dbaa8b5c8/packages/adminrouter/extra/src/test-harness/modules/mocker/endpoints/marathon.py#L200-L203
[2] https://github.com/dcos/dcos/blob/3520215eeb1054039f4f94e8e153785dbaa8b5c8/packages/adminrouter/extra/src/test-harness/tests/test_service.py#L44-L48
[3] https://github.com/dcos/dcos/blob/3520215eeb1054039f4f94e8e153785dbaa8b5c8/packages/adminrouter/extra/src/test-harness/tests/test_service.py#L37-L42
[4] https://github.com/dcos/dcos/blob/3520215eeb1054039f4f94e8e153785dbaa8b5c8/packages/adminrouter/extra/src/test-harness/tests/test_service.py#L647
[5] https://github.com/tsenart/vegeta
[6] https://github.com/dcos/dcos/blob/c1ae4514b07ece64613a833b61f6ac38a259d4c3/packages/adminrouter/extra/src/test-harness/tests/test_cache.py#L671
[7] https://github.com/dcos/dcos/blob/c1ae4514b07ece64613a833b61f6ac38a259d4c3/packages/adminrouter/extra/src/test-harness/tests/test_service.py#L387-L390
[8] https://github.com/dcos/dcos/tree/c1ae4514b07ece64613a833b61f6ac38a259d4c3/packages/adminrouter/extra/src#testing
[9] https://github.com/dcos/dcos/tree/c1ae4514b07ece64613a833b61f6ac38a259d4c3/packages/adminrouter/extra/src#how-to-write-tests

jgehrcke:
> With N requests it is possible for all of them to land on a single node/mock.

If it helps: The parameters can easily be chosen so that there is negligible probability for that to happen. Take four Marathon app instances and 15 independent requests, for example. The probability for all of them to be sent to the app instances then is if I am not mistaken 1/4 ^ 15 which is roughly 10^(-10), i.e. less than one billionth. 

I am not entirely sure what "we can also try to issue requests in parallel" would add, I thought that the idea of this patch is that for every incoming request the Marathon app instance to proxy to would be chosen randomly (i.e. independent from other incoming requests). Is that not true?

vespian:
> If it helps: The parameters can easily be chosen so that there is a negligible probability for that to happen. Take four Marathon app instances and 15 independent requests, for example. The probability for all of them to be sent to the app instances then is if I am not mistaken 1/4 ^ 15 which is roughly 10^(-10), i.e. less than one billionth.

@jgehrcke The intention is to balance the load between instances. Having at least one request go to every instance in the Marathon App is something different. Hence I suggested that we issue multiple (order of hundreds ) requests and check how "balanced" the number of requests is, i.e. the difference between the number of requests received by each task mock is no more than 2 (arbitrary number). This also has a benefit of not having to worry about how accurate the random number generator is for small sample sizes.

> I am not entirely sure what "we can also try to issue requests in parallel" would add, I thought that the idea of this patch is that for every incoming request the Marathon app instance to proxy to would be chosen randomly (i.e. independent from other incoming requests). Based on what you wrote that is probably a wrong assumption of mine. Would love to see clarification.

@jgehrcke Most of all it is the execution time of the test. Secondly - I would like us to be thorough here - distributing the incoming requests in parallel seems like scenario when more things can break now or in the future than the case when we issue the requests one after the another.

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -166,9 +166,11 @@ local function fetch_and_store_marathon_apps(auth_token)
           end
        end
 
-       -- next() returns nil if table is empty.
-       local i, task = next(tasks)
-       if i == nil then
+       -- Randomly select the task.
```
vespian: @stg- I just realised something - the task you pick here will be used by ALL requests until the next cache refresh, i.e. the task AR is proxing requests to will change very 25s, during that period all the requests will go to the same task.

If we want to really load-balance the requests, then we first need to store the routing data for all the task with given ApplicationID, and then we should be picking randomly one in the `/service` endpoint routing code[1].

[1] https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/lib/service.lua#L83-L108

PR #2294, opened 2018-01-10
[Release scripting] Use native boto3 session interface for AWS authentication

mellenburg:
## High-level description
Before this change, the release script had to use explicit environment variables for access key ID and secret access key in order to authenticate with AWS but this is less than ideal and instead we should just use the AWS built-in interface. For example, with this change, using AWS_SESSION_TOKEN is an option

https://jira.mesosphere.com/browse/DCOS_OSS-2064

## Checklist for all PRs


mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@lingmann I think this is ready for review now, so PTAL

mellenburg:
@mesosphere-mergebot bump-ee

dcos_installer/test_backend.py
```diff
@@ -288,7 +265,7 @@ def test_do_configure(tmpdir, monkeypatch):
 """
 
 
-def test_do_aws_configure(tmpdir, monkeypatch):
+def test_do_aws_configure(tmpdir, monkeypatch, release_config_aws):
```
orsenthil: argument order. There could be an intuitive preference for `monkeypatch` arg to be the last one.

dcos_installer/test_backend.py
```diff
@@ -297,54 +274,50 @@ def test_do_aws_configure(tmpdir, monkeypatch):
         assert backend.do_aws_cf_configure() == 0
 
 
-valid_storage_config = """---
+@pytest.fixture
+def valid_storage_config(release_config_aws):
+    """ Uses the settings from dcos-release.config.yaml ['testing'] to create a
+    new upload and then deletes it when the test is over
+    """
+    s3_bucket_name = release_config_aws['bucket']
```
orsenthil: `release_config_aws.get('bucket', '')` ?  Or do you want this to fail with `KeyError` ?

dcos_installer/test_backend.py
```diff
@@ -297,54 +274,50 @@ def test_do_aws_configure(tmpdir, monkeypatch):
...
+    bucket_path = str(uuid.uuid4())
```
orsenthil: `uuid.uuid4().encode('ascii')` - please.

release/storage/aws.py
```diff
@@ -6,32 +6,38 @@
...
-                aws_access_key_id=access_key_id,
-                aws_secret_access_key=secret_access_key,
-                region_name=region_name)
-        else:
-            raise ValueError("boto_profile or explicit AWS credentials (access_key_id, "
-                             "secret_access_key, region_name) must be provided")
+def get_aws_session(access_key_id, secret_access_key, region_name=None):
+    """ This method will replace access_key_id and secret_access_key
+    with None if one is set to '' This allows falling back to the AWS internal
+    logic so that AWS_SESSION_TOKEN or something else can be used
```
orsenthil: What does `something else` refer to here?
`some else like IAM role` ?

orsenthil: or link the `http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials` again.

tox.ini
```diff
@@ -41,17 +41,15 @@ passenv =
...
+  AWS_ACCESS_KEY_ID
+  AWS_SECRET_ACCESS_KEY
+  AWS_SESSION_TOKEN
+  AWS_PROFILE
+  AWS_CONFIG_FILE
   AWS_DEFAULT_REGION
 deps =
   dnspython
   pytest
-  pytest-catchlog
```
orsenthil: Why is this removed?

orsenthil:
@mellenburg  - In the testing done session, could you please share the output behavior of  the `release` script after this change is made. 
```
release create some-version mellenburg/dcos/some-version
```

and verify that the artifacts are uploaded correctly. 

* This way we can ensure that the release behavior hasn't regressed.

conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+@pytest.fixture
+def release_config_testing(release_config):
+    if 'testing' not in release_config:
+        pytest.skip("Skipped because there is no `testing` configuration in dcos-release.config.yaml")
+    return release_config['testing']
+
+
+@pytest.fixture
+def release_config_aws(release_config_testing, monkeypatch):
```
orsenthil: `monkeypatch` arg kind of indicates that only this method is exercised/ tested in code. 
Perhaps others need to be exercised and tested too.

dcos_installer/test_backend.py
```diff
@@ -297,54 +274,50 @@ def test_do_aws_configure(tmpdir, monkeypatch):
...
 
 
-def test_do_aws_cf_configure_valid_storage_config(config_aws, tmpdir, monkeypatch):
-    bucket = str(uuid.uuid4())
-    config_str = valid_storage_config.format(
-        key_id=config_aws["access_key_id"],
-        bucket=bucket,
-        access_key=config_aws["secret_access_key"])
-    assert aws_cf_configure(bucket, config_str, config_aws, tmpdir, monkeypatch) == 0
+def test_do_aws_cf_configure_valid_storage_config(release_config_aws, tmpdir, monkeypatch, valid_storage_config):
```
orsenthil: same here with w.r.t order of args. We could consider keeping the patcher arg to the last one.

dcos_installer/test_backend.py
```diff
@@ -297,54 +274,50 @@ def test_do_aws_configure(tmpdir, monkeypatch):
...
 
-def test_override_aws_template_storage_region_name(config_aws, tmpdir, monkeypatch):
-    bucket = str(uuid.uuid4())
-    config_str = valid_storage_config.format(
-        key_id=config_aws["access_key_id"],
-        bucket=bucket,
-        access_key=config_aws["secret_access_key"])
-    config_str += '\naws_template_storage_region_name: {}'.format(config_aws['region_name'])
-    assert aws_cf_configure(bucket, config_str, config_aws, tmpdir, monkeypatch) == 0
+def test_override_aws_template_storage_region_name(release_config_aws, tmpdir, monkeypatch, valid_storage_config):
```
orsenthil: ditto.

release/storage/aws.py
```diff
@@ -6,32 +6,38 @@
...
+        region_name=region_name)
 
 
 class S3StorageProvider(AbstractStorageProvider):
     name = 'aws'
 
-    def __init__(self, bucket, object_prefix, download_url, boto3_profile=None, region_name=None,
-                 access_key_id=None, secret_access_key=None):
+    def __init__(self, bucket, object_prefix, download_url,
+                 access_key_id=None, secret_access_key=None, region_name=None):
```
orsenthil: What if the `default-region` is None. AFAICS, it takes it from the environment / config file of the user. The code inside boto3 does not have any default region.

So, perhaps it might be useful to state this / having  default region in `our code` set to us-west2?

dcos_installer/test_backend.py
mellenburg: You most definitely want this to fail with a key error. Running this without a pre-configured bucket for writing to not a valid test case (as the advanced aws template doesnt itself create a new bucket, but rather just a bucket path)

tox.ini
```diff
@@ -41,17 +41,15 @@ passenv =
...
-  AWS_TESTING_SECRET_ACCESS_KEY
-  AWS_PROD_ACCESS_KEY_ID
-  AWS_PROD_SECRET_ACCESS_KEY
-  AWS_DEV_ACCESS_KEY_ID
-  AWS_DEV_SECRET_ACCESS_KEY
+  AWS_ACCESS_KEY_ID
+  AWS_SECRET_ACCESS_KEY
+  AWS_SESSION_TOKEN
+  AWS_PROFILE
+  AWS_CONFIG_FILE
```
orsenthil: It will be great, if we have a comment with references in this file. I think, `.ini` files allow it.
If someone starts the read from the `.ini` file, they are going to miss the understanding of the behavior and fallbacks that is used in the code.

dcos_installer/test_backend.py
mellenburg: by that same logic, wouldn't `tmpdir` also have to be behind `release_config_aws`? Seems inconsequential unless we have a rule that says "all built-in pytest fixtures should be listed after custom fixtures"

tox.ini
mellenburg: because when you run pytest/tox you see this:
`pytest-catchlog plugin has been merged into the core, please remove it from your requirements`
https://teamcity.mesosphere.io/viewLog.html?buildId=931198&buildTypeId=DcOs_Open_Build_Tox&tab=buildLog&_focus=1523#_state=268

mellenburg:
@orsenthil I do not understand your request that I do `release create foo bar` our of band and verify the artifacts... This is literally exactly what the CI does...
- teamcity/dcos/build/dcos runs build_teamcity like this https://github.com/dcos/dcos/blob/master/build_teamcity#L80
- teamcity/dcos/test/aws/onprem/static pulls the generated dcos_generate_config.sh from the upload bucket
- teamcity/dcos/test/aws/cloudformation/simple pulls the simple templates and launches them from the upload bucket
- teamcity/dcos/test/azure/arm pulls the ACS templates from the upload bucket and runs them

I fail to see what I should be doing that isn't already covered in CI

release/storage/aws.py
mellenburg: I oppose this. We would *much* rather see an error to set the default region than to quietly set it somewhere buried in release/storage/aws.py

conftest.py
mellenburg: monkeypatch was a holdover from a previous revision and should be removed now. good eye

orsenthil:
@mellenburg  - I might have "release promote" (followed by release create) in my mind. We certainly don't do `release promote` in CI.

dcos_installer/test_backend.py
mellenburg: I'm confused... UUID object does not natively have an 'encode' method, so are you asking me to do `str(uuid.uuid4()).encode('ascii')`? That serves no purpose because that will have to be decoded as a python string when tit is used later on... This seems like a very unnecessary request.

mellenburg:
@mesosphere-mergebot bump-ee

mellenburg:
@orsenthil I verified release promotion with the following:
```
release promote testing/pull/2294 testing/foobar/release
```

dcos_installer/test_backend.py
orsenthil: My mistake. I mistakenly assumed `uuid4()` returned a byte-string.

dcos_installer/test_backend.py
```diff
@@ -297,54 +274,50 @@ def test_do_aws_configure(tmpdir, monkeypatch):
...
+""".format(
+        bucket=release_config_aws['bucket'],
+        bucket_path=bucket_path)
+    session = boto3.session.Session()
+    s3 = session.resource('s3')
+    s3_bucket = s3.Bucket(s3_bucket_name)
+    for o in s3_bucket.objects.filter(Prefix=bucket_path):
+        o.delete()
+    # objects = [{'Key': o.key} for o in s3_bucket.objects.filter(Prefix=bucket_path)]
+    # s3_bucket.delete_objects(Delete={'Objects': objects})
```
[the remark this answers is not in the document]
mellenburg: Oops, thanks. Good eye!

mellenburg:
@mesosphere-mergebot bump-ee

release/storage/aws.py
deric: I don't find it very useful to see an AWS region error when trying to build a local release, that I'm not deploying to AWS. Is there an option to disable boto3 completely?

PR #2293, opened 2018-01-10
Pkgpanda: Copy systemd unit files instead of symlinking

branden:
## High-level description

Systemd requires that all unit files be readable when it boots. If DC/OS unit files are symlinked to a non-root volume mount that's managed by systemd, that mount won't be available when systemd starts, so systemd won't be able to read DC/OS unit files and will mark those units as failed without starting them. This PR changes the behavior of `pkgpanda setup` to copy unit files instead of symlinking, so that systemd can read DC/OS unit files before it mounts the volume that contains the rest of the installation.

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS_OSS-1587](https://jira.mesosphere.com/browse/DCOS_OSS-1587) Support DC/OS install on non-root LVM volume

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Tested manually on CentOS 7.2, see https://github.com/dcos/dcos/pull/2293#issuecomment-358243915.
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]

PR #2383, opened 2018-01-31
(1.11) Pkgpanda: Copy systemd unit files instead of symlinking

branden:
## High-level description

Systemd requires that all unit files be readable when it boots. If DC/OS unit files are symlinked to a non-root volume mount that's managed by systemd, that mount won't be available when systemd starts, so systemd won't be able to read DC/OS unit files and will mark those units as failed without starting them. This PR changes the behavior of Pkgpanda to copy unit files instead of symlinking then to the original package files, so that systemd can read DC/OS unit files before it mounts the volume that contains the rest of the installation.

Unit files are copied to the systemd config directory in two steps: First they're copied to a temporary location, and then they're moved to their intended location. This maintains the invariant that Pkgpanda always moves or symlinks files to their destination, which prevents Pkgpanda from leaving the filesystem in an ambiguous state with partial copies if the package swap doesn't execute to completion.

This is a reimplementation of #2293 that meets the new systemd requirement (added in systemd 233) that files under `dcos.target.wants` must be symlinks.

  - [DCOS_OSS-1587](https://jira.mesosphere.com/browse/DCOS_OSS-1587) Support DC/OS install on non-root LVM volume

## Checklist for all PRs

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -49,7 +49,17 @@ linux_group_regex = "^[a-z_][a-z0-9_-]*$"  # https://github.com/shadow-maint/sha
 
 
 class Systemd:
-    """Manages systemd units and unit files during installation."""
+    """Manages systemd units and unit files during installation.
+
+    This class uses a unit.wants directory as the list of all units it needs to manage. Unit files are copied to the
+    base systemd dir to make sure they're available on the root volume, and thus readable when systemd starts. Symlinks
+    in the unit.wants directory are rewritten to point to the copied unit files.
+
+    """
+
+    # Use "unit.new" to prevent removing unrelated ".new" directories if the install root is being used as the systemd
+    # base dir.
+    new_unit_suffix = ".unit.new"
 
     def __init__(self, unit_directory, active, block):
         self.__unit_directory = unit_directory
@@ -78,6 +88,36 @@ class Systemd:
                 if ex.returncode != 5:
                     raise
 
+    def remove_staged_unit_files(self):
+        """Remove staged unit files created by Systemd.stage_new_units()."""
+        for filename in os.listdir(self.__base_systemd):
+            if filename.endswith(self.new_unit_suffix):
+                os.remove(os.path.join(self.__base_systemd, filename))
+
+    def stage_new_units(self, new_wants_dir):
+        """Prepare new systemd units for activation.
+
+        Unit files targeted by the symlinks in new_wants_dir are copied to a temporary location in the base systemd
+        directory, and the symlinks are rewritten to target the intended final destination of the copied unit files.
+
+        """
+        for unit_name in self.unit_names(new_wants_dir):
+            wants_symlink_path = os.path.join(new_wants_dir, unit_name)
+            package_file_path = os.path.realpath(wants_symlink_path)
+            systemd_file_path = os.path.join(self.__base_systemd, unit_name)
+            tmp_systemd_file_path = systemd_file_path + self.new_unit_suffix
+
+            # Copy the unit file to the systemd directory with a suffix added to the filename.
+            # This file will be moved to systemd_file_path when the new package set is swapped in.
+            shutil.copyfile(package_file_path, tmp_systemd_file_path)
+            shutil.copymode(package_file_path, tmp_systemd_file_path)
+
+            # Rewrite the symlink to point to the copied unit file's destination.
+            # This symlink won't point to the correct file until the copied unit file is moved to its target location
+            # during activate_new_unit_files().
+            os.remove(wants_symlink_path)
+            os.symlink(systemd_file_path, wants_symlink_path)
+
     def remove_unit_files(self):
         if not os.path.exists(self.__unit_directory):
             return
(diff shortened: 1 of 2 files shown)


gpaul:
Wow. Thanks for taking care of this.

branden:
@mesosphere-mergebot bump-ee

branden:
@mesosphere-mergebot bump-ee

branden:

branden:
@mesosphere-mergebot bump-ee

branden:
I tested this PR manually with an onprem deployment on CentOS 7.2 hosts. Prior to deploying, I set up an LVM-managed volume mount on a master and an agent, and created a symlink at `/opt/mesosphere` pointing to it. I then deployed this PR and observed the cluster coming up with all nodes healthy. I rebooted an agent, observed it rejoin the cluster, and did the same to a master. I then did a configuration upgrade (adding a custom check) across the cluster, observed the cluster come up with all nodes healthy, and again rebooted an agent and then a master and observed them rejoining the cluster.

branden:
Tested an onprem deployment on CoreOS 1465.8.0, which includes systemd 233.

branden:
Tested manually (CoreOS 1235.12.0, systemd 231) that the system will still boot if there's a `dcos.target` unit and a `dcos.target.wants` directory with broken symlinks to nonexistent unit files. (This is a potential filesystem state that Pkgpanda may create if the host reboots in the middle of a package swap.) DC/OS units don't start of course, but the system still boots with everything else.

pkgpanda/__init__.py
```diff
@@ -77,6 +88,65 @@ def stop_all(self):
...
+            shutil.copyfile(package_file_path, tmp_systemd_file_path)
+            shutil.copymode(package_file_path, tmp_systemd_file_path)
+
+            # Rewrite the symlink to point to the copied unit file's destination.
+            # This symlink won't point to the correct file until the copied unit file is moved to its target location
+            # during activate_new_unit_files().
+            os.remove(wants_symlink_path)
+            os.symlink(systemd_file_path, wants_symlink_path)
+
+    def remove_unit_files(self):
```
mnaboka: i wonder if `remove_unit_files` should accept a list of files to remove as a parameter instead of calling `self.unit_names(self.__unit_directory):` internally?

pkgpanda/__init__.py
```diff
@@ -77,6 +88,65 @@ def stop_all(self):
...
+        if not os.path.exists(self.__unit_directory):
+            return
+
+        for unit_name in self.unit_names(self.__unit_directory):
+            try:
+                os.remove(os.path.join(self.__base_systemd, unit_name))
+            except FileNotFoundError:
```
mnaboka: it seems that if, for some reason, the file was not found, it'll just silently ignore, would not it be more useful catch an exception in the caller?

pkgpanda/__init__.py
```diff
@@ -77,6 +88,65 @@ def stop_all(self):
...
+    def activate_new_unit_files(self):
+        """Move new unit files to their final locations."""
+        if not os.path.exists(self.__unit_directory):
+            return
+
+        self.remove_unit_files()
+
+        for unit_name in self.unit_names(self.__unit_directory):
+            systemd_file_path = os.path.join(self.__base_systemd, unit_name)
+            os.rename(systemd_file_path + self.new_unit_suffix, systemd_file_path)
```
mnaboka: nit: maybe use `os.path.join` instead of `+` ?

pkgpanda/__init__.py
```diff
@@ -77,6 +88,65 @@ def stop_all(self):
...
+
+    @staticmethod
+    def unit_names(unit_dir):
+        units = os.listdir(unit_dir)
+        for unit_name in units:
+            if unit_name in RESERVED_UNIT_NAMES:
+                raise Exception("Reserved name encountered - {}.".format(unit_name))
```
mnaboka: `Exception` seems to be too broad. Maybe adding a custom exception here to let the caller handle this properly?

pkgpanda/__init__.py
branden: No, we want to use the wants directory as the source of truth. This is existing behavior that I don't want to change in this PR.

pkgpanda/__init__.py
branden: Catching this in the caller would mean exiting the method, which we don't want. This is existing behavior that I don't want to change in this PR.

pkgpanda/__init__.py
branden: This is adding an extension to a filename, not joining parts of a path.

pkgpanda/__init__.py
branden: There's no proper way to handle this, we have to blow up. And this is existing behavior that I don't want to change in this PR.

pkgpanda/__init__.py
mnaboka: sounds good, i thought it would be easier to test if a caller passes units as parameter

pkgpanda/__init__.py
branden: That would mean the `Install` class would have to inspect the wants directory for unit names and then pass them to this method. Knowing how unit files are organized in a wants directory and checking for reserved unit names seem like details that should belong to the `Systemd` class.

pkgpanda/__init__.py
branden: That would mean the `Install` class would have to inspect the wants directory for unit names and then pass them to this method. Knowing how unit files are organized in a wants directory and checking for reserved unit names seem like details that should belong to the `Systemd` class.

pkgpanda/__init__.py
mnaboka: right, but we could set a flag to file not found and blow up at the very end when we finished the loop,
this would allow us to remove all files and signal back to a caller that something went wrong

branden: It's not necessarily an error if we don't find the unit file. We clean up unit files as a sanity check before activating the new set, in which case we don't actually expect to find any.

branden:
Tested a CloudFormation deployment on CoreOS 1632.2.1 (systemd 235), see https://github.com/dcos/dcos/pull/2414.

amitaekbote:
@mesosphere-mergebot ship_it_label_status_check

PR #2411, opened 2018-02-05
fixture for downloading diagnostics bundles

margaret:
## High-level description

Fixture that will download the diagnostics bundles if provided a directory via environment variable. Uses the `dcos_api_session` fixture because otherwise this would require installing and auth for the dcos CLI.

Change to TeamCity `dcos-launch test` template to get this to work:

- add environment variable `env.DIAGNOSTICS_DIRECTORY` which is set to `teamcityWorkingDir/diagnostics`
- add `teamcityWorkingDir/diagnostics` to artifacts paths
- add `env.DIAGNOSTICS_DIRECTORY` to `PYTEST_PASS_ENV` parameter
- add lines in 'run tests' build step to scp the diagnostics bundle back and then upload it as an artifact

Additionally, the 3 enterprise integration tests that inherit from the dcos-test-utils template need to have the `PYTEST_PASS_ENV` updated because they override the default from `dcos-launch test`. The enterprise test Manual trigger project also needs to have that environment variable updated to include `DIAGNOSTICS_DIRECTORY`.

As stated in the conversation with branden below, this is the """""quick""""" update to get this working and after this is merged in, the functionality will be implemented in dcos-test-utils and dcos-launch in a more stable/reusable way. https://jira.mesosphere.com/browse/QUALITY-1819

https://jira.mesosphere.com/browse/QUALITY-1802

## Checklist for all PRs

___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -64,7 +64,8 @@ def _dump_diagnostics(request, dcos_api_session):
     """
     yield
 
-    if os.environ.get('DIAGNOSTICS_DIRECTORY'):
+    make_diagnostics_report = os.environ.get('DIAGNOSTICS_DIRECTORY') is not None
+    if make_diagnostics_report:
         log.info('Create diagnostics report for all nodes')
         check_json(dcos_api_session.health.post('report/diagnostics/create', json={"nodes": ["all"]}))
 
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -1,10 +1,17 @@
+import logging
 import os
 
 import api_session_fixture
 import pytest
 from dcos_test_utils import logger
-
+from test_dcos_diagnostics import (
+    _get_bundle_list,
+    check_json,
+    wait_for_diagnostics_job,
+    wait_for_diagnostics_list
+)
 logger.setup(os.getenv('TEST_LOG_LEVEL', 'INFO'))
+log = logging.getLogger(__name__)
 
 
 def pytest_configure(config):
@@ -42,3 +49,45 @@ def dcos_api_session():
 @pytest.fixture(scope='session')
 def noauth_api_session(dcos_api_session):
     return dcos_api_session.get_user_session(None)
+
+
+@pytest.fixture(scope='session', autouse=True)
+def _dump_diagnostics(request, dcos_api_session):
+    """Download the zipped diagnostics bundle report from each master in the cluster to the home directory. This should
+    be run last. The _ prefix makes sure that pytest calls this first out of the autouse session scope fixtures, which
+    means that its post-yield code will be executed last.
+
+    * There is no official way to ensure fixtures are called in a certain order
+    https://github.com/pytest-dev/pytest/issues/1216
+    * However it seems that fixtures at the same scope are called alphabetically
+    https://stackoverflow.com/a/28593102/1436300
+    """
+    yield
+
+    if os.environ.get('DIAGNOSTICS_DIRECTORY'):
+        log.info('Create diagnostics report for all nodes')
+        check_json(dcos_api_session.health.post('report/diagnostics/create', json={"nodes": ["all"]}))
+
+        last_datapoint = {
+            'time': None,
+            'value': 0
+        }
+
+        log.info('\nWait for diagnostics job to complete')
+        wait_for_diagnostics_job(dcos_api_session, last_datapoint)
+
+        log.info('\nWait for diagnostics report to become available')
+        wait_for_diagnostics_list(dcos_api_session)
+
+        log.info('\nDownload zipped diagnostics reports')
+        bundles = _get_bundle_list(dcos_api_session)
+        for bundle in bundles:
+            for master_node in dcos_api_session.masters:
+                r = dcos_api_session.health.get(os.path.join('report/diagnostics/serve', bundle), stream=True,
+                                                node=master_node)
+                bundle_path = os.path.join(os.path.expanduser('~'), bundle)
+                with open(bundle_path, 'wb') as f:
+                    for chunk in r.iter_content(1024):
+                        f.write(chunk)
+    else:
+        log.info('\nNot downloading diagnostics bundle for this session.')
--- a/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
+++ b/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
@@ -19,10 +19,6 @@ __contact__ = 'dcos-cluster-ops@mesosphere.io'
 LATENCY = 120
 
 
-class NotCriticalException(Exception):
-    """Exception should be raised to continue retry loop"""
-
-
 def check_json(response):
     response.raise_for_status()
     try:
@@ -415,7 +411,7 @@ def _get_bundle_list(dcos_api_session):
 
 
 @retrying.retry(wait_fixed=2000, stop_max_delay=120000,
-                retry_on_exception=lambda e: isinstance(e, NotCriticalException))
+                retry_on_result=lambda x: x is False)
 def wait_for_diagnostics_job(dcos_api_session, last_datapoint):
     response = check_json(dcos_api_session.health.get('report/diagnostics/status/all'))
     # find if the job is still running
@@ -440,8 +436,7 @@ def wait_for_diagnostics_job(dcos_api_session, last_datapoint):
     last_datapoint['value'] = percent_done
     last_datapoint['time'] = datetime.datetime.now()
 
-    if job_running:
-        raise NotCriticalException('Job is still running')
+    return not job_running
 
 
 # sometimes it may take extra few seconds to list bundles after the job is finished.
--- a/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
+++ b/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
@@ -19,6 +19,10 @@ __contact__ = 'dcos-cluster-ops@mesosphere.io'
 LATENCY = 120
 
 
+class NotCriticalException(Exception):
+    """Exception should be raised to continue retry loop"""
+
+
 def check_json(response):
     response.raise_for_status()
     try:
@@ -410,6 +414,42 @@ def _get_bundle_list(dcos_api_session):
     return bundles
 
 
+@retrying.retry(wait_fixed=2000, stop_max_delay=120000,
+                retry_on_exception=lambda e: isinstance(e, NotCriticalException))
+def wait_for_diagnostics_job(dcos_api_session, last_datapoint):
+    response = check_json(dcos_api_session.health.get('report/diagnostics/status/all'))
+    # find if the job is still running
+    job_running = False
+    percent_done = 0
+    for _, attributes in response.items():
+        assert 'is_running' in attributes, '`is_running` field is missing in response'
+        assert 'job_progress_percentage' in attributes, '`job_progress_percentage` field is missing in response'
+
+        if attributes['is_running']:
+            percent_done = attributes['job_progress_percentage']
+            logging.info("Job is running. Progress: {}".format(percent_done))
+            job_running = True
+            break
+
+    # if we ran this bit previously compare the current datapoint with the one we saved
+    if last_datapoint['time'] and last_datapoint['value']:
+        if percent_done <= last_datapoint['value']:
+            assert (datetime.datetime.now() - last_datapoint['time']) < datetime.timedelta(seconds=15), (
+                "Job is not progressing"
+            )
+    last_datapoint['value'] = percent_done
+    last_datapoint['time'] = datetime.datetime.now()
+
+    if job_running:
+        raise NotCriticalException('Job is still running')
+
+
+# sometimes it may take extra few seconds to list bundles after the job is finished.
+@retrying.retry(stop_max_delay=5000)
+def wait_for_diagnostics_list(dcos_api_session):
+    assert _get_bundle_list(dcos_api_session), 'get a list of bundles timeout'
+
+
 def test_dcos_diagnostics_bundle_create(dcos_api_session):
     """
     test bundle create functionality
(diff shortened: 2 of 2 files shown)


margaret:
@mesosphere-mergebot bump-ee

margaret:
@mesosphere-mergebot bump-ee

margaret:
@mesosphere-mergebot bump-ee

margaret:

margaret:
@mesosphere-mergebot bump-ee

margaret:
https://teamcity.mesosphere.io/viewLog.html?buildId=964325&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsCloudFormationSimple&tab=artifacts#!bxqmlynw4w0,-5b2r2guf109n,xgdkfjyoovrt,-1fkh4v9v65vg6

⬆️ Example of diagnostics bundle available as artifact. 

Going to clean up some of the logging statements then rebase.

margaret:
@mesosphere-mergebot bump-ee

margaret:
@mesosphere-mergebot bump-ee

margaret:
https://teamcity.mesosphere.io/viewLog.html?buildId=965278&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsCloudFormationSimple&tab=artifacts#!bxqmlynw4w0,-yucavokbh8w2
✅ diagnostics report artifacts

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,44 @@ def dcos_api_session():
...
+    """Download the zipped diagnostics bundle report from each master in the cluster to the home directory. This should
+    be run last. The _ prefix makes sure that pytest calls this first out of the autouse session scope fixtures, which
+    means that its post-yield code will be executed last.
+
+    * There is no official way to ensure fixtures are called in a certain order
+    https://github.com/pytest-dev/pytest/issues/1216
+    * However it seems that fixtures at the same scope are called alphabetically
+    https://stackoverflow.com/a/28593102/1436300
+    """
+    yield
```
orsenthil: nit - newline after `yield`

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,44 @@ def dcos_api_session():
...
+    if os.environ.get('DIAGNOSTICS_DIRECTORY'):
```
orsenthil: What's a the default for `DIAGNOSTICS_DIRECTORY`? Let's think about the situation when the tests are run locally and this envvar is not set.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,44 @@ def dcos_api_session():
...
+        log.info('\nWait for diagnostics report to become available')
+        wait_for_diagnostics_list(dcos_api_session)
+
+        log.info('\nDownload zipped diagnostics reports')
+        bundles = _get_bundle_list(dcos_api_session)
+        for bundle in bundles:
+            for master_node in dcos_api_session.masters:
+                r = dcos_api_session.health.get(os.path.join('report/diagnostics/serve', bundle), stream=True,
+                                                node=master_node)
+                bundle_path = os.path.join(os.path.expanduser('~/'), bundle)
```
orsenthil: `minor`: expanduser('~'). `/` after `~` is redundant and join should take care of adding `/`.

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,44 @@ def dcos_api_session():
...
+                with open(bundle_path, 'wb') as f:
+                    for chunk in r.iter_content(1024):
```
orsenthil: The constant `1024` could be a CONSTANT `CHUNK_SIZE=1024` declared above.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -407,6 +411,42 @@ def _get_bundle_list(dcos_api_session):
     return bundles
 
 
+@retrying.retry(wait_fixed=2000, stop_max_delay=120000,
+                retry_on_exception=lambda e: isinstance(e, NotCriticalException))
```
orsenthil: The retry_on_exception could be `retry_on_exception=(NotCriticalException,))` - No need to check for the instance. The library does that for us.

packages/dcos-integration-test/extra/conftest.py
margaret: The default is whatever `%env.DIAGNOSTICS_DIRECTORY%` evaluates to, which is a randomly generated directory name ending in `/diagnostics`.

packages/dcos-integration-test/extra/conftest.py
margaret: Since the chunked reads only happens once in this file and is a common pattern anyway, I think it would be clearer just have the size written directly instead of having another level.

margaret:
Addressed 👍-ed comments and rebased

margaret:
@mesosphere-mergebot bump-ee

margaret:
Onprem Strict tests with artifacts (hopefully): https://teamcity.mesosphere.io/viewLog.html?buildId=965345&buildTypeId=DcOs_Enterprise_Test_Inte_AwsOnpremWStaticBackendAndSecurityStrict&tab=artifacts

margaret:
@mesosphere-mergebot bump-ee

margaret:
@orsenthil changing it to tuple results in an error that tuple is not callable. It seems that the version we're using expects a function https://pypi.python.org/pypi/retrying

```
    def should_reject(self, attempt):
        reject = False
        if attempt.has_exception:
>           reject |= self._retry_on_exception(attempt.value[1])
E           TypeError: 'tuple' object is not callable
```
https://teamcity.mesosphere.io/viewLog.html?buildId=965636&buildTypeId=DcOs_Open_Test_DcosDocker_DockerIntegrationTests_IntegrationTestDcosDockerPr&guest=1

margaret:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
orsenthil: This is still not addressed.

orsenthil:
@margaret  - Interesting! Before suggesting, I did look at the examples of retrying library - https://github.com/rholder/retrying/blob/master/test_retrying.py#L265

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,45 @@ def dcos_api_session():
...
+@pytest.fixture(scope='session', autouse=True)
+def _dump_diagnostics(request, dcos_api_session):
+    """Download the zipped diagnostics bundle report from each master in the cluster to the home directory. This should
+    be run last. The _ prefix makes sure that pytest calls this first out of the autouse session scope fixtures, which
+    means that its post-yield code will be executed last.
+
+    * There is no official way to ensure fixtures are called in a certain order
+    https://github.com/pytest-dev/pytest/issues/1216
+    * However it seems that fixtures at the same scope are called alphabetically
+    https://stackoverflow.com/a/28593102/1436300
```
branden: Thanks for this docstring. It looks like we're relying on an undocumented implementation detail of pytest to make sure this fixture gets torn down last, which seems risky. Since pytest doesn't define the ordering of teardowns, meaning it could change, it seems like our best bet for the long term is to handle this outside of pytest. Were there any other options you explored? Perhaps we could implement this as a separate script (or a DC/OS CLI command?) that TC jobs explicitly run after tests. That would give us more confidence that the bundle covers the entire test run. And perhaps we could optimize space a little by limiting bundle collection to failed test runs.

margaret: The long term plan is to get this functionality into https://github.com/dcos/dcos-test-utils and so that it could be used in `dcos-launch` directly, but I had originally hoped to get this out in time to be helpful for 1.11 debugging and reusing the code from the diagnostics tests was the simplest/quickest way.

For the dcos-integration-test pytest-based implementation I also looked into hooks that could be run at the end of a session, but if something isn't a fixture or a test, then it wouldn't be able to use the `dcos_api_session` fixture properly, and at that point I think it would be better to just go forward with the `dcos-test-utils`-based implementation.

The same effect can be done easily through the CLI with the `dcos node diagnostics` [`create`](https://docs.mesosphere.com/1.11/cli/command-reference/dcos-node/dcos-node-diagnostics-create/) and [`download`](https://docs.mesosphere.com/1.11/cli/command-reference/dcos-node/dcos-node-diagnostics-download/) commands (I think these ultimately call the same endpoints used here) which should be called with the [CLI module](https://github.com/dcos/dcos-test-utils/blob/master/dcos_test_utils/dcos_cli.py) in dcos-test-utils. I would have just put in these lines on the TC build script, but we currently don't install and set up the CLI on each integration test CI jobs because it's a lot of overhead.

re: artifact size: I can change that to only copy them over when tests fail.

margaret: I'm currently looking into some of the ee diagnostics integration tests failing on the strict mode CI job. If it looks like those require any more finagling on this end I will just put this PR on hold and try doing it through dcos-test-utils anyway 😞

branden: Sounds good, I'm OK with merging this as a short-term solution (assuming you can get it working in strict security mode) and following up on it later.

margaret:
@mesosphere-mergebot bump-ee

margaret:
After rebasing again:

Open AWS Cloudformation Simple https://teamcity.mesosphere.io/viewLog.html?buildId=967694&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsCloudFormationSimple&tab=artifacts

Enterprise Onprem Strict https://teamcity.mesosphere.io/viewLog.html?buildId=967687&buildTypeId=DcOs_Enterprise_Test_Inte_AwsOnpremWStaticBackendAndSecurityStrict&tab=artifacts

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
mellenburg: Rather than making a whole new exception for this singular purpose, can we just use `retry_on_result`? and then you `return False` in the code whenever you want to retry rather than raising this purpose-built exception

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -19,6 +19,10 @@
 LATENCY = 120
 
 
+class NotCriticalException(Exception):
```
mellenburg: Doesn't seem necessary; see below. We can just return a certain value to retry instead of creating a whole new exception

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -42,3 +49,45 @@ def dcos_api_session():
...
+    """
+    yield
+
+    if os.environ.get('DIAGNOSTICS_DIRECTORY'):
```
colin-msphere: Should documentation be updated with this env var? 

What are the values expected? Should the value returned be verified?

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
margaret: Yeah that does seem better. That's how it was originally from the stuff I refactored but I'll change it.

packages/dcos-integration-test/extra/conftest.py
margaret: `DIAGNOSTICS_DIRECTORY` is an environment variable used in TeamCity or whatever is calling into pytest to indicate where the job should scp the diagnostics report to from the cluster. In the fixture it's just used as an indicator of whether or not the diagnostics job should be run and the value doesn't get fed into anythign else. I didn't want to set an extra environment variable when the directory being set was an indicator of whether to run it, but I can change that to
```
make_diagnostics_report = os.environ.get('DIAGNOSTICS_DIRECTORY') is not None
if make_diagnostics_report:
```
unless you think it's better to just make another env var for `RUN_DIAGNOSTICS`? I think the latter is cleaner from the POV of dcos-integration-tests but from the CI setup it's another variable that we have to remember to update when changing things.

colin-msphere: Thanks for the response.

I agree with keeping it a single env var and not breaking it up.

margaret:
@mesosphere-mergebot bump-ee

margaret:
@mesosphere-mergebot bump-ee

margaret:
@mesosphere-mergebot bump-ee

margaret:
teamcity/dcos/test/docker/smoke failure is the oinker timeout https://jira.mesosphere.com/browse/DCOS_OSS-1902

margaret:
(waiting) check that diagnostics artifacts still work:

[AWS CloudFormation Simple](https://teamcity.mesosphere.io/viewLog.html?buildId=970084&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsCloudFormationSimple&tab=artifacts)

[AWS Onprem With Static Backend Strict Mode](https://teamcity.mesosphere.io/viewLog.html?buildId=970086&buildTypeId=DcOs_Enterprise_Test_Inte_AwsOnpremWStaticBackendAndSecurityStrict&tab=artifacts)

margaret:
@colin-msphere @mellenburg Can you review this again please?

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -410,6 +410,41 @@ def _get_bundle_list(dcos_api_session):
...
 
 
+@retrying.retry(wait_fixed=2000, stop_max_delay=120000,
+                retry_on_result=lambda x: x is False)
+def wait_for_diagnostics_job(dcos_api_session, last_datapoint):
+    response = check_json(dcos_api_session.health.get('report/diagnostics/status/all'))
+    # find if the job is still running
+    job_running = False
+    percent_done = 0
+    for _, attributes in response.items():
```
mellenburg: FYI if you are just throwing away the keys you could use `response.values()`

margaret: Yep. Again, this was part of the stuff I copy-paste-refactored and I didn't bother going through and refining it.

mellenburg:
@mesosphere-mergebot override-status teamcity/dcos/test/docker/smoke https://jira.mesosphere.com/browse/DCOS_OSS-1902

PR #2457, opened 2018-02-13
Patch DC/OS to use Marathon build from PR 6011.

jeschkies:
## DO NOT MERGE


jeschkies:
This was just used for investigations. Let's drop it.

PR #2551, opened 2018-03-02
Coreos update 1.11

mellenburg:
## High-level description
Updates CoreOS to 1632.2.1 in the AWS CloudFormation Templates, which also includes a bump to docker 17

https://jira.mesosphere.com/browse/DCOS_OSS-2130

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: N/A update to cloud-provided underlying operating system

### DIFF
--- a/packages/mesos/extra/start_mesos.sh
+++ b/packages/mesos/extra/start_mesos.sh
@@ -12,12 +12,12 @@ fi
 function coreos_networkd_config() {
  network_config="/etc/systemd/network/dcos.network"
  sudo tee $network_config > /dev/null<<'EOF'
- [Match]
-  Type=bridge
-  Name=docker* m-* d-* vtep*
+[Match]
+Type=bridge
+Name=docker* m-* d-* vtep*
 
- [Link]
-  Unmanaged=yes
+[Link]
+Unmanaged=yes
 EOF
 }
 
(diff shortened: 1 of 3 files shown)


mellenburg:
@mesosphere-mergebot bump-ee

packages/mesos/extra/start_mesos.sh
```diff
@@ -27,7 +27,7 @@ if [[ "${distro}" == 'coreos' ]]; then
        coreos_networkd_config
        echo "Configuring systemd-networkd to ignore docker bridge and DC/OS overlay interfaces..."
 
-       if systemctl is-enabled systemd-networkd > /dev/null; then
+       if systemctl is-active systemd-networkd > /dev/null; then
```
orsenthil: Why is this change?

mellenburg: because the unit is disabled and active in the AMIs published by coreOS

orsenthil: Looks like we only care about the exit code in the condition. but we want the networkd to be **active**

* enabled - a service (unit) is configured to start when the system boots
* active - a service (unit) is currently running.

https://www.digitalocean.com/community/tutorials/how-to-use-systemctl-to-manage-systemd-services-and-units

```
There are also methods for checking for specific states. For instance, to check to see if a unit is currently active (running), you can use the is-active command:

systemctl is-active application.service
This will return the current unit state, which is usually active or inactive. The exit code will be "0" if it is active, making the result simpler to parse programmatically.
To see if the unit is enabled, you can use the is-enabled command:

systemctl is-enabled application.service
This will output whether the service is enabled or disabled and will again set the exit code to "0" or "1" depending on the answer to the command question.
```

PR #2593, opened 2018-03-08
Upgrade REX-Ray to v0.11.2

branden:
## High-level description

This upgrades REX-Ray from v0.9.0 to v0.11.2. The current package can't be built due to the issue fixed by https://github.com/rexray/rexray/pull/995, and v0.11.2 is the latest available release.

  - [DCOS_OSS-2245](https://jira.mesosphere.com/browse/DCOS_OSS-2245) Rexray does not build in 1.11 and master

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Building without cache should now succeed. REX-Ray is covered by an existing integration test.

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): https://github.com/rexray/rexray/compare/2a7458dd90a79c673463e14094377baf9fc8695e...fc8bfbd2d02c2690fc3a755a9560dd12c88e0852
  - [x] Test Results: Official release of an externally managed project.
  - [x] Code Coverage (if available): Official release of an externally managed project.

### DIFF
--- a/packages/rexray/buildinfo.json
+++ b/packages/rexray/buildinfo.json
@@ -1,9 +1,9 @@
 {
-  "docker": "golang:1.7",
+  "docker": "golang:1.9",
   "single_source" : {
     "kind": "git",
     "git": "https://github.com/rexray/rexray.git",
-    "ref": "2a7458dd90a79c673463e14094377baf9fc8695e",
-    "ref_origin": "v0.9.0"
+    "ref": "fc8bfbd2d02c2690fc3a755a9560dd12c88e0852",
+    "ref_origin": "v0.11.2"
   }
 }
--- a/packages/rexray/buildinfo.json
+++ b/packages/rexray/buildinfo.json
@@ -2,7 +2,7 @@
   "docker": "golang:1.7",
   "single_source" : {
     "kind": "git",
-    "git": "https://github.com/codedellemc/rexray.git",
+    "git": "https://github.com/rexray/rexray.git",
     "ref": "2a7458dd90a79c673463e14094377baf9fc8695e",
     "ref_origin": "v0.9.0"
   }
--- a/packages/rexray/extra/dcos-rexray.service
+++ b/packages/rexray/extra/dcos-rexray.service
@@ -13,4 +13,6 @@ Environment=REXRAY_HOME=/
 ExecStartPre=/bin/rm -f /var/run/libstorage/lsx.lock
 # Allow REX-Ray configuration to be changed during upgrade.
 ExecStartPre=/bin/cp /opt/mesosphere/etc/rexray.conf /etc/rexray/config.yml
+# Prevent startup error: "open /run/docker/plugins/rexray.sock: no such file or directory"
+ExecStartPre=/bin/mkdir -p /run/docker/plugins
 ExecStart=__PKG_PATH__/bin/rexray start -f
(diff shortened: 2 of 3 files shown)


branden:
@mesosphere-mergebot bump-ee

branden:
@mesosphere-mergebot bump-ee

packages/rexray/extra/dcos-rexray.service
```diff
@@ -13,4 +13,6 @@ Environment=REXRAY_HOME=/
 ExecStartPre=/bin/rm -f /var/run/libstorage/lsx.lock
 # Allow REX-Ray configuration to be changed during upgrade.
 ExecStartPre=/bin/cp /opt/mesosphere/etc/rexray.conf /etc/rexray/config.yml
+# Prevent startup error: "open /run/docker/plugins/rexray.sock: no such file or directory"
+ExecStartPre=/bin/mkdir -p /run/docker/plugins
```
branden: This `mkdir -p` is apparently what older versions of REX-Ray did implicitly: https://github.com/rexray/rexray/issues/1108#issuecomment-343538897

It looks like I could also accomplish this by setting `CSI_ENDPOINT` to a path that exists (maybe `/run/dcos/rexray.sock`), but I'm not sure if that's a worse hack.

branden:

packages/rexray/buildinfo.json
```diff
@@ -1,9 +1,9 @@
 {
-  "docker": "golang:1.7",
+  "docker": "golang:1.9",
   "single_source" : {
     "kind": "git",
-    "git": "https://github.com/codedellemc/rexray.git",
-    "ref": "2a7458dd90a79c673463e14094377baf9fc8695e",
-    "ref_origin": "v0.9.0"
+    "git": "https://github.com/rexray/rexray.git",
+    "ref": "fc8bfbd2d02c2690fc3a755a9560dd12c88e0852",
```
jieyu: Do we need to bump the version of rexray?

branden: Yes, we need to bump it to pull in https://github.com/rexray/rexray/pull/995. That fix is first available in v0.10, but I bumped to v0.11.2 since it's a more recent release.

packages/rexray/extra/dcos-rexray.service
jdef: Is this because there's a race between docker and RR startup?

packages/rexray/buildinfo.json
jieyu: ah, ic. have you tested the new version, like manually?

branden: I haven't, I was relying on this integration test: https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_rexray.py#L16

Do you have a test plan you run through for REX-Ray changes?

packages/rexray/extra/dcos-rexray.service
branden: What I saw on CF clusters is that this directory was never created, blocking REX-Ray from starting. I'm guessing it's because they're running an older version of Docker. If it were a race between REX-Ray and Docker then I don't think there'd be a problem; REX-Ray would just fail and restart until Docker created `/run/docker/plugins/`.

packages/rexray/buildinfo.json
jdef: Discussed offline w/ @branden and manual testing seems to indicate things are working about as well as can be expected.

PR #2598, opened 2018-03-08
Node Rate Limiter considered HARMFUL

kensipe:
## High-level description

Mesos Core considers this a bug via:

  - [MESOS-7721](https://issues.apache.org/jira/browse/MESOS-7721) Master's agent removal rate limit also applies to agent unreachability.
- [MESOS-5948](https://issues.apache.org/jira/browse/MESOS-5948) Remove rate-limiting for agent removal

A major reason this is not already removed is the PM for the project left the project.   This should have been removed already.   It has likely caused more harm than good.  With companies changing it by our recommendation [under certain situations](https://jira.mesosphere.com/browse/COPS-719)

Mesos is a 2-level scheduler.  It makes NO sense for Mesos to keep and hold information from the "intelligent" scheduler.   `MESOS_SLAVE_REMOVAL_RATE_LIMIT` must die!!  for now `100/1mins` might be ok.

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -448,7 +448,7 @@ package:
       MESOS_REGISTRY_STORE_TIMEOUT=60secs
       MESOS_REGISTRY_FETCH_TIMEOUT=60secs
       MESOS_REGISTRY_STRICT=false
-      MESOS_SLAVE_REMOVAL_RATE_LIMIT=1/20mins
+      MESOS_SLAVE_REMOVAL_RATE_LIMIT=100/1mins
       MESOS_OFFER_TIMEOUT=2mins
       MESOS_WORK_DIR={{ mesos_master_work_dir }}
       MESOS_ZK=zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181/mesos


kensipe:
@mesosphere-mergebot bump-ee

kensipe:

gen/dcos-config.yaml
```diff
@@ -448,7 +448,7 @@ package:
       MESOS_REGISTRY_STORE_TIMEOUT=60secs
       MESOS_REGISTRY_FETCH_TIMEOUT=60secs
       MESOS_REGISTRY_STRICT=false
-      MESOS_SLAVE_REMOVAL_RATE_LIMIT=1/20mins
+      MESOS_SLAVE_REMOVAL_RATE_LIMIT=100/1mins
```
timcharper: Do you think that maybe we should just set the rate to a faster interval?

1/15secs?

kensipe: anything is possible... I don't think it should exist actually... 
Also... I have a strong preference to NOT have what you are suggesting...  any value with a low numerator has the effect of causing seemingly unexplained delays of propagation of information... lets say you lost 100 nodes at once... with a 1/15 secs every 15 secs you would get a new notification of lost node until all 100 are reported... which is 4 each min... it would take 25 mins before you got the last notification.     Mesos is a 2 level scheduler.   The 2nd level is the intelligent layer.  we should not have mesos hide information from the intelligent layer.

orsenthil:
This impacts mesos and mesos team members approval is definitely required.

orsenthil:
@greggomann  / @sambatyon  - Comments on this change affecting `MESOS_SLAVE_REMOVAL_RATE_LIMIT` ?

sambatyon:
I've been looking at this and I don't oppose the change, however I am not fully aware of the whole set of implications here. Perhaps @bmahler could give us his opinion in this?

PR #2685, opened 2018-03-30
Add retries for pkgpanda when downloading packages

ionutbalutoiu:
## High-level description

Deployments on Azure sometimes fail due to pkgpanda not being
able to download some of the packages. This is caused by Azure
transient temporary DNS failures.

This patch adds retries in pkgpanda when downloading packages.

  - [DCOS_OSS-2317](https://jira.mesosphere.com/browse/DCOS_OSS-2317) Pkgpanda lack of retries when downloading packages.

### DIFF
--- a/pkgpanda/util.py
+++ b/pkgpanda/util.py
@@ -18,6 +18,8 @@ from typing import List
 import requests
 import teamcity
 import yaml
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 from teamcity.messages import TeamcityServiceMessages
 
 from pkgpanda.exceptions import FetchError, ValidationError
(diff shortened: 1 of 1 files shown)


pkgpanda/util.py
```diff
@@ -63,6 +65,20 @@ def variant_suffix(variant, delim='.'):
     return delim + variant
 
 
+def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist=[500, 502, 504]):
```
branden: Let's avoid using mutable objects like lists as parameter defaults, since that can lead to surprising behavior. Someone might try to mutate it without realizing it belongs to the module's scope, rather than the function's. We'd want to do something like this instead:

```
def get_requests_retry_session(..., status_forcelist=None):
    status_forcelist = status_forcelist or [500, 502, 504]
    ...
```
This way any mutations of `status_forcelist` won't impact other calls to `get_requests_retry_session()`.

ionutbalutoiu:
@branden, Addressed your review comment.

branden:
@mesosphere-mergebot bump-ee

branden:

PR #2714, opened 2018-04-06
Admin Router: /service endpoint url rewriting is now configurable using the Marathon label

vespian:
## High-level description

This PR makes it possible to configure `/service` endpoint upstream request URL rewriting basing on the `DCOS_SERVICE_REWRITE_REQUEST_URLS` label. See the Jira issue for details.

  - https://jira.mesosphere.com/browse/DCOS-9032 `Adminrouter option to not rewrite URLs`

## Related PRs

EE PR: https://github.com/mesosphere/dcos-enterprise/pull/2570

### DIFF
--- a/packages/adminrouter/extra/src/README.md
+++ b/packages/adminrouter/extra/src/README.md
@@ -479,6 +479,48 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
 be reachable only in their relative location
 `<dcos-cluster>/services/<application ID><link>`.
 
+### Disabling URL path rewriting for selected applications
+Some applications like Jenkins or Artifactory, permit specifying a context path (e.g. `/service/jenkins`) that is used while generating application URLs (both relative and absolute). However, normally Admin Router strips these, and this can lead to breakages, esp. for assets like JavaScript files and images.
+
+It is possible to workaround it by using another another HTTP reverse proxy in the application container which rewrites URLs from / to /service/jenkins. This isn't ideal though, so an extra Marathon label is recognized by Admin Router. If a task has `DCOS_SERVICE_REWRITE_REQUEST_URLS` label set to `false`(string) or `false`(boolean), Admin Router will not strip the context path and the upstream request will be made with the same the URL path as the client request has been made. Please see below for an example of how to enable it in an application definition:
+
+```json
+{
+  "labels": {
+    "DCOS_SERVICE_REWRITE_REQUEST_URLS": "false",
+    "DCOS_SERVICE_NAME": "myapp",
+    "DCOS_SERVICE_SCHEME": "http",
+    "DCOS_SERVICE_PORT_INDEX": "0"
+  },
+  "id": "/myapp",
+  "cmd": "echo \"It works!\" > index.html; /usr/local/bin/python -m http.server $PORT0",
+  "container": {
+    "type": "DOCKER",
+    "volumes": [],
+    "docker": {
+      "image": "python:3.6.5-stretch"
+    }
+  },
+  "cpus": 0.1,
+  "instances": 1,
+  "mem": 128,
+  "networks": [
+    {
+      "mode": "host"
+    }
+  ],
+  "portDefinitions": [
+    {
+      "name": "www",
+      "protocol": "tcp",
+      "port": 10000
+    }
+  ]
+}
+```
+
+This feature is available only for the tasks launched by the root Marathon.
+
 ## Authorization and authentication
 
 DC/OS uses DC/OS authentication token at the core of its authentication and
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -234,8 +234,24 @@ local function fetch_and_store_marathon_apps(auth_token)
           goto continue
        end
 
+       local do_rewrite_requrl = labels["DCOS_SERVICE_REWRITE_REQUEST_URLS"]
+       if do_rewrite_requrl == false or do_rewrite_requrl == 'false' then
+          ngx.log(ngx.INFO, "DCOS_SERVICE_REWRITE_REQUEST_URLS for app '" .. appId .. "' set to 'false'")
+          do_rewrite_requrl = false
+       else
+          -- Treat everything else as true, i.e.:
+          -- * label is absent
+          -- * label is set to "true" (string) or true (bool)
+          -- * label is set to some random string
+          do_rewrite_requrl = true
+       end
+
        local url = scheme .. "://" .. host_or_ip .. ":" .. port
-       svcApps[svcId] = {scheme=scheme, url=url}
+       svcApps[svcId] = {
+         scheme=scheme,
+         url=url,
+         do_rewrite_requrl=do_rewrite_requrl,
+         }
 
        ::continue::
     end
--- a/packages/adminrouter/extra/src/lib/service.lua
+++ b/packages/adminrouter/extra/src/lib/service.lua
@@ -58,7 +58,7 @@ local function upstream_url_from_srv_query(service_name)
     --  A list with:
     --  - upstream_scheme - for now hardcoded to http
     --  - upstream_url
-    --  - err_code, err_text - if an error occured these will be HTTP status
+    --  - err_code, err_text - if an error occurred these will be HTTP status
     --    and error text that should be sent to the client. `nil` otherwise
     local upstream_url = nil
     local upstream_scheme = 'http' -- Hardcoded in case of MesosDNS
@@ -86,12 +86,13 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
     --
     -- Arguments:
     --   service_name (string): service name that should be resolved
+    --   marathon_cache (table): cached root Marathon's apps state
     --
     -- Returns:
     --  A list with:
     --  - true/false depending on whether resolving service name was successful
-    --    or not.
-    --  - err_code, err_text - if an error occured these will be HTTP status
+    --    or not, `nil` if there was an error.
+    --  - err_code, err_text - if an error occurred these will be HTTP status
     --    and error text that should be sent to the client. `nil` otherwise
     if marathon_cache == nil then
         return nil, ngx.HTTP_SERVICE_UNAVAILABLE, "503 Service Unavailable: invalid Marathon svcapps cache"
@@ -107,6 +108,36 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
     return true, nil, nil
 end
 
+local function requrl_rewriting_required(service_name, marathon_cache)
+    -- Determine if the upstream request should contain the
+    -- `/service/<service_name>` part of the URL path or not.
+    --
+    -- Arguments:
+    --   service_name (string): service name that should be resolved
+    --   marathon_cache (table): cached root Marathon's apps state
+    --
+    -- Returns:
+    --  A list with:
+    --  - true/false depending on whether requrl rewriting is enabled or not,
+    --    `nil` if there was an error
+    --  - err_code, err_text - if an error occurred these will be HTTP status
+    --    and error text that should be sent to the client. `nil` otherwise
+    --
+    if service_name == 'marathon' or service_name == 'metronome' then
+      -- These two services should always have the request URL rewriting
+      -- enabled, no matter the marathon_cache, as they are resolved via Mesos
+      -- anyway.
+      return true, nil, nil
+    end
+
+    if marathon_cache[service_name] == nil then
+        -- By default we always rewrite for Mesos Frameworks
+        return true, nil, nil
+    end
+
+    return marathon_cache[service_name]['do_rewrite_requrl'], nil, nil
+end
+
 local function resolve_via_mesos_dns(service_name)
     -- Try to resolve upstream for given service name basing on
     -- MesosDNS SRV entries
@@ -117,8 +148,8 @@ local function resolve_via_mesos_dns(service_name)
     -- Returns:
     --  A list with:
     --  - true/false depending on whether resolving service name was successful
-    --    or not.
-    --  - err_code, err_text - if an error occured these will be HTTP status
+    --    or not, `nil` if there was an error.
+    --  - err_code, err_text - if an error occurred these will be HTTP status
     --    and error text that should be sent to the client. `nil` otherwise
     upstream_scheme, upstream_url, err_code, err_text = upstream_url_from_srv_query(
         service_name)
@@ -146,12 +177,13 @@ local function resolve_via_mesos_state(service_name, mesos_cache)
     --
     -- Arguments:
     --   service_name (string): service name that should be resolved
+    --   mesos_cache (table): cached Mesos's apps state
     --
     -- Returns:
     --  A list with:
     --  - true/false depending on whether resolving service name was successful
-    --    or not.
-    --  - err_code, err_text - if an error occured these will be HTTP status
+    --    or not, `nil` if there was an error.
+    --  - err_code, err_text - if an error occurred these will be HTTP status
     --    and error text that should be sent to the client. `nil` otherwise
     local webui_url = nil
     local service_name_bymesos = nil
@@ -165,7 +197,7 @@ local function resolve_via_mesos_state(service_name, mesos_cache)
     -- services/software relying on that.
     if mesos_cache['f_by_id'][service_name] ~= nil then
         webui_url = mesos_cache['f_by_id'][service_name]['webui_url']
-        -- This effectivelly resolves framework ID into human-friendly service
+        -- This effectively resolves framework ID into human-friendly service
         -- name.
         service_name = mesos_cache['f_by_id'][service_name]['name']
     elseif mesos_cache['f_by_name'][service_name] ~= nil then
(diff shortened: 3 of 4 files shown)


vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/README.md
```diff
@@ -479,6 +479,13 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
 be reachable only in their relative location
 `<dcos-cluster>/services/<application ID><link>`.
 
+### Disabling URL path rewriting for selected applications
+Some applications like Jenkins or Artifactory, permit specifying a context path (e.g. `/service/jenkins`) that is used while generating application URLs (both relative and absolute). However, normally Adminrouter strips these, and this can lead to breakages, esp. for assets like JavaScript files and images.
```
adamtheturtle: `s/Adminrouter/Admin Router`.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
```diff
@@ -842,3 +843,51 @@ def test_if_ar_sets_correct_useragent_while_resolving_via_mesosdns(
 
         assert len(r_reqs) == 1
         verify_header(r_reqs[0]['headers'], 'User-Agent', 'Master Admin Router')
+
+    @pytest.mark.parametrize(
```
adamtheturtle: Not for now but I'd like to point out that this is an *ideal* use case for [Hypothesis](https://hypothesis.readthedocs.io). That is, a test which says "Give me any string which is not `0, false, no` and make the following assumption" could be written very simply.

❓ Ideally for me There would be a docstring something like:

"""
If `DCOS_SERVICE_REWRITE_REQUEST_URLS` is set to `0`, `false` or `no`, Admin Router will not strip the context path and the upstream request will be made with the same the URL path as the client request has been made. Else...
"""

Even better would be two tests, one with `0, false, no` parameters, and one with a bunch of other strings.

Is it the case that `DCOS_SERVICE_REWRITE_REQUEST_URLS` could be not set?

vespian:

packages/adminrouter/extra/src/README.md
```diff
@@ -479,6 +479,13 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
...
+
+It is possible to workaround it by using another Nginx in the application container which rewrites URLs from / to /service/jenkins. This isn't ideal though, so an extra Marathon label is recognized by Admin Router. If a task has `DCOS_SERVICE_REWRITE_REQUEST_URLS` label set to `no` or `0` or `false`, Admin Router will not strip the context path and the upstream request will be made with the same the URL path as the client request has been made.
```
adamtheturtle: > set to `no` or `0` or `false`

Is this a design choice made for this PR or is it something common in DC/OS / AR?

adamtheturtle: > Is this a design choice made for this PR or is it something common in DC/OS / AR?
 
If it is a design choice, I think maybe `true, false` could be the only allowed values, else you get an error. This prevents things like `False` and so on causing confusion.

packages/adminrouter/extra/src/README.md
```diff
@@ -479,6 +479,13 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
...
+
```
jgehrcke: In general I think this documentation should present a Marathon application definition snippet that shows how to properly set this in practice (as done above).

> `no` or `0` or `false`

In general, it's preferable to introduce _one_ definite way (here we offer three) and docs that are precise about that way (`false` as written here could be both, a JSON string or a JSON boolean)

I understand that we should also should "be liberal in what you expect" to a certain degree -- what do you think about only supporting both, the JSON string `"false'` and the JSON boolean `false` (and to document it that way)?

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -234,8 +234,24 @@ local function fetch_and_store_marathon_apps(auth_token)
           goto continue
        end
 
+       local do_rewrite_requrl = labels["DCOS_SERVICE_REWRITE_REQUEST_URLS"]
+       if not do_rewrite_requrl then
+          -- Label is absent, assume 'true' by default:
+          do_rewrite_requrl = true
+       elseif do_rewrite_requrl == 'no' or do_rewrite_requrl == 'false' or do_rewrite_requrl == '0' then
+          ngx.log(ngx.INFO, "DCOS_SERVICE_REWRITE_REQUEST_URLS for app '" .. appId .. "' set to 'false'")
```
jgehrcke: Assuming that this logs upon every service access attempt: don't you think that this might be a little too noisy?

packages/adminrouter/extra/src/README.md
vespian: Let's agree on `false` (string) and `false` (boolean) as @jgehrcke pointed out . 

I was trying to follow Robustness Principle here, but maybe you are right and we should be a bit more strict. Wrt. `true` I am not so sure about it - if the label is absent, then the default is `true`. The label should only be present if there is a need to not to rewrite the upstream URL and it should assume one of two values: `false` (str) or `false` (bool). Having explicit `true` would increase the complexity of the patch without any real gain.

What do you think?

packages/adminrouter/extra/src/lib/cache.lua
vespian: Please note that this is `cache.lua` file not the `service.lua`. The line is printed on every cache refresh, not service access attempt.

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -107,6 +107,43 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
...
+      return true, nil, nil
+    end
+
+    if marathon_cache == nil then
+        -- In theory this should never be reached for non-marathon and
+        -- non-metronome services, as
+        -- `resolve_via_marathon_apps_state` is always called first and once
+        -- the cache value has been set, it is never removed. Still - just in
+        -- case.
+        return nil, ngx.HTTP_SERVICE_UNAVAILABLE, "503 Service Unavailable: invalid Marathon svcapps cache while resolving requrl rewriting"
```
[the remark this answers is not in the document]
vespian: This line causes Nginx to return 503 to the client with the given message. The `return` statement is just a convention/idiom in the openresty world, it is meant to signify that the request processing ends here. The calling function's code is never resumed.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
adamtheturtle: > I get a feeling that introducing Hypothesis here could be a bit of an overkill, plain @pytest.mark.parametrize should just enough. Would you agree?

Strongly "no" - but maybe "too much effort to be worth it right now" which is why I said "Not for now" and wrote the comment just as a brain seed, not as a "do this now" instruction.

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: Discussed via DM, we will pair with Adam this week and see how this could be improved with Hypothesis.

packages/adminrouter/extra/src/lib/service.lua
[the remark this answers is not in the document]
vespian: ouch, yeah you are right - fixing!

I got so used to the `return` idiom that did not notice that docstrings indeed are inaccurate and `ngx.say` is called in the top-level function.

Thanks for pointing it out!

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -291,15 +329,25 @@ local function recursive_resolve(auth, path)
     -- Authorize the request:
     auth.access_service_endpoint(service_realpath)
 
-    -- Trim the URI prefix:
-    prefix = "/service/" .. service_realpath
-    adjusted_prefix = string.sub(ngx.var.uri, string.len(prefix) + 1)
-    if adjusted_prefix == "" then
-        adjusted_prefix = "/"
+    do_requrl_rewriting, err_code = requrl_rewriting_required(service_name, marathon_cache)
```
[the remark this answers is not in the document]
vespian: yep! good catch!

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/lib/service.lua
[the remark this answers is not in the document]
adamtheturtle: I was not clear.

This is a line that we the comment says we should not hit - right? Yet we have it anyways. I'm not 100% sure why - I'm very wary of adding code that we can't hit in a test. That said, if we are to have this code, and we get a report that someone saw this error, what do we do?

packages/adminrouter/extra/src/lib/cache.lua
jgehrcke: Cool, that's convincing.

packages/adminrouter/extra/src/README.md
jgehrcke: > Having explicit true would increase the complexity of the patch without any real gain.

ack!

packages/adminrouter/extra/src/lib/service.lua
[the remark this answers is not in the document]
vespian: This error will be caught earlier by [1]. I have put the same error checking here as well because:
* if in the future the order in which the code is executed changes, lack of error checking here will become a serious bug
* if for some reason, the code starts resetting/invalidating stale cache then we will start having a nasty race condition here which will be resulting in a Lua stacktraces from time to time.
I agree that we should not commit dead code into dc/os, but it seems like a lesser evil.

As for your second question - we investigate why the cache was unavailable. The place when this error is generated does not matter much and the same error message is IIRC emitted in few other places.

[1] https://github.com/dcos/dcos/blob/16ced28a63284de75096e17fa953af08905cd91d/packages/adminrouter/extra/src/lib/service.lua#L96-L98

vespian:
@adamtheturtle  Ready for another pass!

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -234,8 +234,24 @@ local function fetch_and_store_marathon_apps(auth_token)
           goto continue
        end
 
+       local do_rewrite_requrl = labels["DCOS_SERVICE_REWRITE_REQUEST_URLS"]
+       if do_rewrite_requrl == nil then
```
adamtheturtle: I think that this block is not necessary if you change `elseif` to `if` on L241.

vespian: I wanted to be explicit, but yeah you are right. Thanks!

adamtheturtle: Yep, you can put a comment if you like.

packages/adminrouter/extra/src/lib/service.lua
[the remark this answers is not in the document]
vespian: Not at the moment, the error condition that we check for here will always be caught in [1] first.

[1] https://github.com/dcos/dcos/blob/16ced28a63284de75096e17fa953af08905cd91d/packages/adminrouter/extra/src/lib/service.lua#L96-L98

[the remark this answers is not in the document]
adamtheturtle: @jgehrcke tl;dr - We won't hit this code at the time of writing because another function which is called first does a check that the required condition is not true. There is no current infrastructure for unit testing this function. Changing another function, or the order in which functions are called, may cause us to hit a condition for which raising an error here will avoid a bug.

What do you think about keeping this code in?

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/lib/service.lua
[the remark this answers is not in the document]
jgehrcke: I think this PR needs to improve its error handling story/concept.

If we give up and say "we have no idea how our code behaves" (I am not really in favor of that) and want to prepare ourselves for that case then I don't think it is wise to to _not_ log all relevant detail (as far as I read the current code it would simply emit a 503 response, but there would not be a corresponding informative log message).

That is, _iff_ we want to prepare for this rather fatal scenario, then I'd like to see a precise 'error' level log message, stack traces, etc in the Admin Router log.

> which will be resulting in a Lua stacktraces from time to time.

If one handles errors, one needs to handle them very consciously. I think letting things blow up would be better than the current code: it would log a very useful lua stack trace :-) and it would return a '500 internal server error' to the user agent which is about equally useful as the current HTTP error response.

Not adding these 'paranoid' code paths seems like the better thing to do, _unless_ we have better error handling in place than the default error handling (again, which is log a lua stack trace, and return a '500 internal server error').

packages/adminrouter/extra/src/README.md
```diff
@@ -479,6 +479,29 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
 be reachable only in their relative location
 `<dcos-cluster>/services/<application ID><link>`.
 
+### Disabling URL path rewriting for selected applications
+Some applications like Jenkins or Artifactory, permit specifying a context path (e.g. `/service/jenkins`) that is used while generating application URLs (both relative and absolute). However, normally Admin Router strips these, and this can lead to breakages, esp. for assets like JavaScript files and images.
+
+It is possible to workaround it by using another Nginx in the application container which rewrites URLs from / to /service/jenkins. This isn't ideal though, so an extra Marathon label is recognized by Admin Router. If a task has `DCOS_SERVICE_REWRITE_REQUEST_URLS` label set to `no` or `0` or `false`, Admin Router will not strip the context path and the upstream request will be made with the same the URL path as the client request has been made. Please see below for an example of an application definition:
```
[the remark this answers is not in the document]
vespian: Yes, I just forgot to update the documentation. Sorry about that.

Fixed.

packages/adminrouter/extra/src/README.md
```diff
@@ -479,6 +479,29 @@ as `.js` and `.css` files. This is due to the fact that the linked resources wil
...
+    "id": "myapp",
+    "cpus": 0.1,
+    "mem": 16.0,
+    "ports": [
+        0
+    ],
+    "cmd": "/opt/mesosphere/bin/python3 -m http.server $PORT0",
+    "instances": 2,
+    "labels": {
+        "DCOS_SERVICE_REWRITE_REQUEST_URLS": false
```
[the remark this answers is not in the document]
vespian: I am sceptical about it - the purpose of this example to show how to disable request URL rewriting, not to show how to run a Marathon app. 

What do you think about using ellipsis to emphasize that this is an incomplete example and that user needs to make it work on its own accord?

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -107,6 +107,44 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
     return true, nil, nil
 end
 
+local function requrl_rewriting_required(service_name, marathon_cache)
+    -- Determine if the upstream request should contain the
+    -- `/service/<service_name>` part of the URL path or not.
+    --
+    -- Arguments:
+    --   service_name (string): service name that should be resolved
```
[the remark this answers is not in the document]
vespian: > Why is the marathon_cache argument not documented here? I wonder what it...

Fixed, both here and in other places where there was missing argument documentation.

> and why it supposedly can be nil.

Currently, it is only possible as a result of failing to fetch root Marathon apps state during the first cache refresh.

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -107,6 +107,44 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
...
+    if service_name == 'marathon' or service_name == 'metronome' then
+      -- These two services should always have the request URL rewriting
+      -- enabled, no matter the marathon_cache, as they are resolved via Mesos
+      -- anyway.
+      return true, nil, nil
+    end
+
+    if marathon_cache == nil then
+        -- In theory this should never be reached for non-marathon and
+        -- non-metronome services, as
```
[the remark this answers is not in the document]
vespian: Reworded to `reached by services different than 'marathon' (i.e. Root Marathon) and `metronome` (i.e. Root Metronome)`

vespian:
> If we give up and say "we have no idea how our code behaves"

It was meant as a safety net for my future-self, in case he forgets/oversees this edge-case.

> That is, iff we want to prepare for this rather fatal scenario, then I'd like to see a precise 'error' level log message, stack traces, etc in the Admin Router log

The stack trace would tell you something like: `indexing a null variable` plus the place where the code blew up. The error message would tell you `invalid Marathon svcapps cache while resolving requrl rewriting` which IMO is more informative than a pure stack trace which is more probable to get escalated by the field to us.

For the sake of not prolonging this discussion any longer and being pragmatic, I have removed said error handling.

vespian:
@jgehrcke Ready for another pass.

packages/adminrouter/extra/src/test-harness/tests/test_service.py
vespian: We ended up with not implementing it due to https://github.com/pytest-dev/pytest/issues/916. It prevents using pytest fixtures without some extra hacking :(

In overall, I find Hypothesis to be a nice tool. Once the issue is fixed I would like to try to implement test case like these using it.

packages/adminrouter/extra/src/README.md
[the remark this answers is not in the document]
vespian: @jgehrcke I have added a working example - please check it out.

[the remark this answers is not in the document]
jgehrcke: thank you so much

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

vespian:
mergebot/dcos/master/2714

vespian:
mergebot/dcos/master/2714

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

vespian:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -107,6 +108,36 @@ local function resolve_via_marathon_apps_state(service_name, marathon_cache)
     return true, nil, nil
 end
 
+local function requrl_rewriting_required(service_name, marathon_cache)
```
sambatyon: I'm not very sure of the name here. `requrl` doesn't seem to follow the pattern anywhere else.

vespian: Addressed in https://github.com/dcos/dcos/pull/2733

PR #2730, opened 2018-04-11
[DCOS-14199] Write strings to files atomically in pkgpanda.util

adamtheturtle:
## High-level description

This changes `pkgpanda.util.write_string` to avoid a race condition as described in [DCOS-14199](https://jira.mesosphere.com/browse/DCOS-14199).
I have also applied this change to some related functions as a precaution.
I have stopped returning data from the functions as it appears that those return values are never used.

This also adds a test for `write_string` to enshrine the permission requirements and basic functionality of `write_string` (thanks @mhrabovcin for help). That test passes on `master` *and* with the changes here.

  - [DCOS-14199](https://jira.mesosphere.com/browse/DCOS-14199) exhibitor.py try_shortcut() failed due to an empty file

## Checklist for all PRs

  - [ ] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: See https://github.com/dcos/dcos/pull/2730#issuecomment-380506106.
  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This code is already called by the test suite. We could test the race condition with something like the following with a timeout but I'd rather not. I'm happy to discuss this with a reviewer:

```python
# This will trigger the error from the issue if run by two processes in parallel.
# Calling the new function in the loop will not.
filename = 'foo.txt'
while True:
    with open(filename, 'w+') as f:
        f.write('hello')

    with open(filename) as f:
        stripped = f.read().strip()
        assert stripped != ''
```

### DIFF
--- a/pkgpanda/util.py
+++ b/pkgpanda/util.py
@@ -7,7 +7,9 @@ import platform
 import re
 import shutil
 import socketserver
+import stat
 import subprocess
+import tempfile
 from contextlib import contextmanager, ExitStack
 from itertools import chain
 from multiprocessing import Process
@@ -25,12 +27,6 @@ from teamcity.messages import TeamcityServiceMessages
 from pkgpanda.exceptions import FetchError, ValidationError
 
 is_windows = platform.system() == "Windows"
-
-json_prettyprint_args = {
-    "sort_keys": True,
-    "indent": 2,
-    "separators": (',', ':')
-}
 
 
 def is_absolute_path(path):
@@ -254,8 +250,8 @@ def load_yaml(filename):
 
 
 def write_yaml(filename, data, **kwargs):
-    with open(filename, "w+") as f:
-        return yaml.safe_dump(data, f, **kwargs)
+    dumped_yaml = yaml.safe_dump(data, **kwargs)
+    write_string(filename, dumped_yaml)
 
 
 def make_file(name):
@@ -264,13 +260,45 @@ def make_file(name):
 
 
 def write_json(filename, data):
-    with open(filename, "w+") as f:
-        return json.dump(data, f, **json_prettyprint_args)
+    dumped_json = json_prettyprint(data=data)
+    write_string(filename, dumped_json)
 
 
 def write_string(filename, data):
-    with open(filename, "w+") as f:
-        return f.write(data)
+    """
+    Write a string to a file.
+    Overwrite any data in that file.
+
+    We use an atomic write practice of creating a temporary file and then
+    moving that temporary file to the given ``filename``. This prevents race
+    conditions such as the file being read by another process after it is
+    opened here but not yet written to.
+
+    It also prevents us from creating or truncating a file before we fail to
+    write data to it because of low disk space.
+
+    If no file already exists at ``filename``, the new file is created with
+    permissions 0o644.
+    """
+    prefix = os.path.basename(filename)
+    tmp_file_dir = os.path.dirname(os.path.realpath(filename))
+    fd, temporary_filename = tempfile.mkstemp(prefix=prefix, dir=tmp_file_dir)
+
+    try:
+        permissions = os.stat(filename).st_mode
+    except FileNotFoundError:
+        permissions = 0o644
+
+    try:
+        try:
+            os.write(fd, data.encode())
+        finally:
+            os.close(fd)
+        os.chmod(temporary_filename, stat.S_IMODE(permissions))
+        os.replace(temporary_filename, filename)
+    except Exception:
+        os.remove(temporary_filename)
+        raise
 
 
 def load_string(filename):
@@ -279,7 +307,12 @@ def load_string(filename):
 
 
 def json_prettyprint(data):
-    return json.dumps(data, **json_prettyprint_args)
+    return json.dumps(
+        data,
+        sort_keys=True,
+        indent=2,
+        separators=(',', ':'),
+    )
 
 
 def if_exists(fn, *args, **kwargs):
--- a/pkgpanda/test_util.py
+++ b/pkgpanda/test_util.py
@@ -371,3 +371,37 @@ def test_split_by_token():
     assert list(split_by_token('{', '}', 'some text {token} some more text', strip_token_decoration=True)) == [
         ('some text ', False), ('token', True), (' some more text', False)
     ]
+
+
+def test_write_string(tmpdir):
+    """
+    `pkgpanda.util.write_string` writes or overwrites a file with permissions
+    for User to read and write, Group to read and Other to read.
+
+    Permissions of the given filename are preserved, or a new file is created
+    with 0o644 permissions.
+
+    This test was written to make current functionality regression-safe which
+    is why no explanation is given for these particular permission
+    requirements.
+    """
+    filename = os.path.join(str(tmpdir), 'foo_filename')
+    pkgpanda.util.write_string(filename=filename, data='foo_contents')
+    with open(filename) as f:
+        assert f.read() == 'foo_contents'
+
+    pkgpanda.util.write_string(filename=filename, data='foo_contents_2')
+    with open(filename) as f:
+        assert f.read() == 'foo_contents_2'
+
+    st_mode = os.stat(filename).st_mode
+    expected_permission = 0o644
+    assert (st_mode & 0o777) == expected_permission
+
+    os.chmod(filename, 0o777)
+    pkgpanda.util.write_string(filename=filename, data='foo_contents_3')
+    with open(filename) as f:
+        assert f.read() == 'foo_contents_3'
+    st_mode = os.stat(filename).st_mode
+    expected_permission = 0o777
+    assert (st_mode & 0o777) == expected_permission


pkgpanda/util.py
```diff
@@ -264,13 +266,28 @@ def make_file(name):
...
-    with open(filename, "w+") as f:
-        return f.write(data)
+    """
+    Write a string to a file.
+    Overwrite any data in that file.
+
+    We use an atomic write practice of creating a temporary file and then
+    moving that temporary file to the given ``filename``. This prevents race
+    conditions such as the file being read by another process after it is
+    opened here but not yet written to.
```
vespian: Do you think we can also mention here truncating a file in a 'full disk' scenario?

pkgpanda/util.py
```diff
@@ -264,13 +266,28 @@ def make_file(name):
...
+    """
+    suffix = '-{filename}-tmp'.format(filename=filename)
```
vespian: Are we 100% sure that this is always a filename and not file path? :)

adamtheturtle: You're right - this is not safe.

pkgpanda/util.py
adamtheturtle: Done below.

adamtheturtle:
@orsenthil Given your WIP, how do you want a `CHANGES.md` entry to come? Should I just add one and then you resolve conflicts?

orsenthil:
@adamtheturtle  - Given both are in review, let the existing CHANGES.md merge and you could add the entry for this as a follow-up. Would that be okay?

adamtheturtle:
@orsenthil Yes, sure.

adamtheturtle:
Please can an owner override `teamcity/dcos/test/aws/cloudformation/simple` with https://jira.mesosphere.com/browse/DCOS_OSS-2310.

adamtheturtle:
@mesosphere-mergebot bump-ee

adamtheturtle:

adamtheturtle:
cc @vespian @mhrabovcin @jgehrcke @jongiddy @ahairape for review

pkgpanda/util.py
```diff
@@ -264,13 +265,32 @@ def make_file(name):
...
+
+    It also prevents us from creating or truncating a file before we fail to
+    write data to it because of low disk space.
+    """
+    temporary_filename = '{filename}-{random}.tmp'.format(
+        filename=filename,
+        random=uuid.uuid4().hex,
+    )
+
+    with open(temporary_filename, 'w') as f:
```
jongiddy: From creation of the temporary file until the `os.replace` needs exception handling to remove the file if an error occurs, especially since you appear to be worried about disk full conditions. But it's not nice to have random files appearing of unknown significance either.

This can be a bit fiddly with the current code, since you create the file in a context manager. You'd have to start the `try` before the `open`, so the exception handling has to check if the file exists before removing.

With `mkstemp` the file exists when the `mkstemp` call exits, so the code can unconditionally remove the file. Note, it needs a double-try, one to close the file descriptor and one to remove the file on exception.

pkgpanda/util.py
```diff
@@ -264,13 +265,32 @@ def make_file(name):
...
+
+    We use an atomic write practice of creating a temporary file and then
+    moving that temporary file to the given ``filename``. This prevents race
+    conditions such as the file being read by another process after it is
+    opened here but not yet written to.
+
+    It also prevents us from creating or truncating a file before we fail to
+    write data to it because of low disk space.
+    """
+    temporary_filename = '{filename}-{random}.tmp'.format(
```
vespian: Can we please use just the basename of the file in here?

I am afraid some functions may try to call this functions with a full path instead of just a filename.

pkgpanda/util.py
```diff
@@ -264,13 +265,32 @@ def make_file(name):
...
+        filename=filename,
+        random=uuid.uuid4().hex,
+    )
```
jongiddy: Why does this not use the well-tested cross-platform Python library https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp ?

pkgpanda/util.py
jongiddy: The important thing is that the temporary file is in the same directory as the destination file. This ensures that it is on the same disk partition, and hence that `os.replace` will be atomic.

pkgpanda/util.py
jongiddy: And of course, I should point you to the Python 3 version rather than the Python 2 version: https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

pkgpanda/util.py
adamtheturtle: @vespian That was relevant when I was using `mkstemp` but not now, as per @jongiddy 's comment.

pkgpanda/util.py
adamtheturtle: I originally used `mkstemp` and got an error because the created file did not exist on the same partition as the `replace` destination. However, I guess I can use the `prefix` argument.

adamtheturtle:
Builds are blocked on https://jira.mesosphere.com/browse/DCOS_OSS-2372.

pkgpanda/util.py
```diff
@@ -26,12 +27,6 @@
 
 is_windows = platform.system() == "Windows"
 
-json_prettyprint_args = {
```
adamtheturtle: Using this just once and putting the arguments into the `json_prettyprint` function is a purely stylistic change I decided to make.

adamtheturtle: Well - with the new use of `write_string` within `write_json`, it became possible to use `json_prettyprint` and that meant that  `json_prettyprint_args` was used only in one place and could therefore be moved into the `json_prettyprint` function.

adamtheturtle:
Since the last review round I have (with help from @jongiddy and @mhrabovcin ) changed this to use `mkstemp` and added a test for `write_string`.

adamtheturtle:
@mesosphere-mergebot bump-ee

pkgpanda/util.py
```diff
@@ -264,13 +259,36 @@ def make_file(name):
...
+    """
+    prefix = os.path.basename(filename)
+    tmp_file_dir = os.path.dirname(os.path.realpath(filename))
+    fd, temporary_filename = tempfile.mkstemp(prefix=prefix, dir=tmp_file_dir)
+    os.close(fd)
+
+    try:
+        with open(temporary_filename, 'w') as f:
+            f.write(data)
+        os.chmod(temporary_filename, 0o644)
```
jongiddy: You should get the mode from the destination file, using `os.stat`, in case this function is ever used for private files.

adamtheturtle: Good thinking - and keep this if the target file does not exist?

adamtheturtle:
> But, now that it works, would you be willing to go back to using os.write/os.close so that the file doesn't need to be opened and closed twice?

Yes

pkgpanda/util.py
jongiddy: OK, but please add a note to the doc string that if the file does not exist it will be created with permissions 0o644.

adamtheturtle:
@mesosphere-mergebot bump-ee

pkgpanda/util.py
```diff
@@ -264,13 +259,45 @@ def make_file(name):
...
+
+    If no file already exists at ``filename``, the new file is created with
+    permissions 0o644.
+    """
+    prefix = os.path.basename(filename)
+    tmp_file_dir = os.path.dirname(os.path.realpath(filename))
+    fd, temporary_filename = tempfile.mkstemp(prefix=prefix, dir=tmp_file_dir)
+
+    try:
+        permissions = os.stat(filename).st_mode
```
jongiddy: `st_mode` can return bits that aren't valid in a `chmod` call. You need to mask this to ensure you're only getting the permissions, similar to how you do it in the tests. Or better, use `stat.S_IMODE(mode)`

pkgpanda/util.py
```diff
@@ -264,13 +259,45 @@ def make_file(name):
...
+    except FileNotFoundError:
+        permissions = 0o644
```
jongiddy: :grey_question: : You could use the `stat` values from https://docs.python.org/3.5/library/os.html#os.chmod here to avoid the magic number, but 644 is a bit verbose in that form, so it's dubious benefit.

mhrabovcin: I personally prefer `0o644` as it is standard form for linux commands.

adamtheturtle: For this PR I won't make the change, but appreciate the idea.

adamtheturtle:
@mesosphere-mergebot bump-ee

adamtheturtle:
`teamcity/dcos/test/azure/arm` [failed](https://teamcity.mesosphere.io/viewLog.html?buildId=1041202&buildTypeId=DcOs_Open_Test_IntegrationTest_AzureArm&tab=buildLog).

I will re-run and if it is a flake I will report to JIRA.

PR #2817, opened 2018-04-27
Made `make_disk_resources.py` include space used by Mesos' work dir.

chhsia0:
## High-level description

When a Mesos agent is registered as a new one, the data for the old
agent will be garbage collected so the space it takes should be offered
to the frameworks.

This fixes https://jira.mesosphere.com/browse/DCOS_OSS-2418

  - [DCOS_OSS-2418](https://jira.mesosphere.com/browse/DCOS_OSSs2417) `make_disk_space.py` should not exclude disk spaces used by old agents.

## Checklist for all PRs

### DIFF
--- a/packages/mesos/extra/make_disk_resources.py
+++ b/packages/mesos/extra/make_disk_resources.py
@@ -99,14 +99,42 @@ def get_mounts_and_freespace(matching_mounts):
             # Per @cmaloney and @lingmann, we should hard exit here if volume
             # doesn't have sufficient space.
             raise VolumeDiscoveryException(
-                '{} has {} MB net free space, expected > 100M'.format(mount, net_free_space)
+                '{} has {} MB free space, expected > {} MB'.format(mount, free_space, TOLERANCE_MB)
             )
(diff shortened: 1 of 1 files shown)


Gilbert88:
Does anyone know if we had integration tests against this script already?

chhsia0:
I only tested this manually because I'm not sure how we do tests for this script. BTW I need to ensure that `get_path_usage` skips files that are on different devices. Will update this PR shortly.

chhsia0:
Did some experiments with this script on a couple agents in an internal DC/OS cluster. The time is averaged over 10 runs, and the filesystem cache is cleaned up before each run.
Agent 1: 33G used; average real/user/sys time: 5.56s/1.43s/1.48s
Agent 2: 43G used; average real/user/sys time: 14.15s/4.02s/4.44s
Agent 3: 30G used: average real/user/sys time: 10.35s/2.80s/3.07s

Unfortunately, I also observed one test run with 122.05s/4.16s/4.38s on agent 2, which might pass systemd's default start timeout (90s).

PR #2989, opened 2018-06-20
Use ngx.timer.every() for the AR cache update

ahairape:
The lua code has been using ngx.timer.at() for the AR cache update. This can lead to the problem described in the JIRA ticket below. This PR replaces ngx.timer.at() with a more robust alternative of  ngx.timer.every(), which also fixes the problem described in the JIRA ticket below.

https://jira.mesosphere.com/browse/DCOS-38248

## Checklist for all PRs

___

Reviewers should be:

### DIFF
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -602,7 +602,7 @@ local function periodically_refresh_cache(auth_token)
         refresh_cache(true, auth_token)
     end
 
-    -- Trigger the timer, every CACHE_POLL_PERIOD seconds after
+    -- Trigger the initial cache update CACHE_FIRST_POLL_DELAY seconds after
     -- Nginx startup.
     local ok, err = ngx.timer.at(_CONFIG.CACHE_FIRST_POLL_DELAY, timerhandler)
     if not ok then
@@ -612,6 +612,8 @@ local function periodically_refresh_cache(auth_token)
         ngx.log(ngx.INFO, "Created initial timer for cache updating.")
     end
 
+    -- Trigger the timer, every CACHE_POLL_PERIOD seconds after
+    -- Nginx startup.
     local ok, err = ngx.timer.every(_CONFIG.CACHE_POLL_PERIOD, timerhandler)
     if not ok then
         ngx.log(ngx.ERR, "Failed to create timer: " .. err)
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -609,7 +609,7 @@ local function periodically_refresh_cache(auth_token)
         ngx.log(ngx.ERR, "Failed to create timer: " .. err)
         return
     else
-        ngx.log(ngx.INFO, "Created initial recursive timer for cache updating.")
+        ngx.log(ngx.INFO, "Created initial timer for cache updating.")
     end
 
     local ok, err = ngx.timer.every(_CONFIG.CACHE_POLL_PERIOD, timerhandler)
@@ -617,7 +617,7 @@ local function periodically_refresh_cache(auth_token)
         ngx.log(ngx.ERR, "Failed to create timer: " .. err)
         return
     else
-        ngx.log(ngx.INFO, "Created timer for cache updating.")
+        ngx.log(ngx.INFO, "Created periodic timer for cache updating.")
     end
 end
 
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -13,8 +13,27 @@ local util = require "util"
 --
 -- CACHE_BACKEND_REQUEST_TIMEOUT << CACHE_REFRESH_LOCK_TIMEOUT
 --
--- Before changing CACHE_POLL_INTERVAL, please check the comment for resolver
+-- Before changing CACHE_POLL_PERIOD, please check the comment for resolver
 -- statement configuration in includes/http/master.conf
+--
+-- Initial timer-triggered cache update early after nginx startup:
+-- It makes sense to have this initial timer-triggered cache
+-- update _early_ after nginx startup at all, and it makes sense to make it
+-- very early, so that we reduce the likelihood for an HTTP request to be slowed
+-- down when it is incoming _before_ the normally scheduled periodic cache
+-- update (example: the HTTP request comes in 15 seconds after nginx startup,
+-- and the first regular timer-triggered cache update is triggered only 25
+-- seconds after nginx startup).
+--
+-- It makes sense to have this time window not be too narrow, especially not
+-- close to 0 seconds: under a lot of load there *will* be HTTP requests
+-- incoming before the initial timer-triggered update, even if the first
+-- timer callback is scheduled to be executed after 0 seconds.
+-- There is code in place for handling these HTTP requests, and that code path
+-- must be kept explicit, regularly exercised, and well-tested. There is a test
+-- harness test that tests/exercises it, but it overrides the default values
+-- with the ones that allow for testability. So the idea is that we leave
+-- initial update scheduled after 2 seconds, as opposed to 0 seconds.
 --
 -- All are in units of seconds. Below are the defaults:
 local _CONFIG = {}
@@ -587,7 +606,7 @@ local function periodically_refresh_cache(auth_token)
     -- This function is invoked from within init_worker_by_lua code.
     -- ngx.timer.every() is called here, a more robust alternative to
     -- ngx.timer.at() as suggested by the openresty/lua-nginx-module
-    -- documentation: 
+    -- documentation:
     -- https://github.com/openresty/lua-nginx-module/tree/v0.10.9#ngxtimerat
     -- See https://jira.mesosphere.com/browse/DCOS-38248 for details on the
     -- cache update problems caused by the recursive use of ngx.timer.at()
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -585,15 +585,13 @@ end
 
 local function periodically_refresh_cache(auth_token)
     -- This function is invoked from within init_worker_by_lua code.
-    -- ngx.timer.at() can be called here, whereas most of the other ngx.*
-    -- API is not available.
+    -- ngx.timer.every() is called here, a more robust alternative to
+    -- ngx.timer.at() as suggested by the openresty/lua-nginx-module
+    -- documentation: 
+    -- https://github.com/openresty/lua-nginx-module/tree/v0.10.9#ngxtimerat
 
     timerhandler = function(premature)
-        -- Handler for recursive timer invocation.
-        -- Within a timer callback, plenty of the ngx.* API is available,
-        -- with the exception of e.g. subrequests. As ngx.sleep is also not
-        -- available in the current context, the recommended approach of
-        -- implementing periodic tasks is via recursively defined timers.
+        -- Handler for periodic timer invocation.
 
         -- Premature timer execution: worker process tries to shut down.
         if premature then
@@ -602,17 +600,9 @@ local function periodically_refresh_cache(auth_token)
 
         -- Invoke timer business logic.
         refresh_cache(true, auth_token)
-
-        -- Register new timer.
-        local ok, err = ngx.timer.at(_CONFIG.CACHE_POLL_PERIOD, timerhandler)
-        if not ok then
-            ngx.log(ngx.ERR, "Failed to create timer: " .. err)
-        else
-            ngx.log(ngx.INFO, "Created recursive timer for cache updating.")
-        end
     end
 
-    -- Trigger initial timer, about CACHE_FIRST_POLL_DELAY seconds after
+    -- Trigger the timer, every CACHE_POLL_PERIOD seconds after
     -- Nginx startup.
     local ok, err = ngx.timer.at(_CONFIG.CACHE_FIRST_POLL_DELAY, timerhandler)
     if not ok then
@@ -620,6 +610,14 @@ local function periodically_refresh_cache(auth_token)
         return
     else
         ngx.log(ngx.INFO, "Created initial recursive timer for cache updating.")
+    end
+
+    local ok, err = ngx.timer.every(_CONFIG.CACHE_POLL_PERIOD, timerhandler)
+    if not ok then
+        ngx.log(ngx.ERR, "Failed to create timer: " .. err)
+        return
+    else
+        ngx.log(ngx.INFO, "Created timer for cache updating.")
     end
 end
 
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -589,6 +589,8 @@ local function periodically_refresh_cache(auth_token)
     -- ngx.timer.at() as suggested by the openresty/lua-nginx-module
     -- documentation: 
     -- https://github.com/openresty/lua-nginx-module/tree/v0.10.9#ngxtimerat
+    -- See https://jira.mesosphere.com/browse/DCOS-38248 for details on the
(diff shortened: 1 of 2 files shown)


mesosphere-mergebot-dev:
This repo has @mesosphere-mergebot integration. You can interact with the following commands.
```
@mesosphere-mergebot bump-ee  
@mesosphere-mergebot merge-it  
@mesosphere-mergebot label [Request For Comment |Work In Progress |Holding |Ship It |Ready For Review] 
@mesosphere-mergebot override-status  
```

* PR creators can apply one of  `[Ready For Review |Work In Progress]`. [Owners](https://github.com/dcos/dcos/blob/master/owners.json) can apply any label.

ahairape:
@mesosphere-mergebot bump-ee

ahairape:

ahairape:
@vespian @gpaul @jgehrcke please review this PR

mesosphere-mergebot-dev:
:wave: This PR has been inactive for 3 days. Please review the status checks to see what needs to be done to move it forward or change the label to Work in Progress. Thank you!

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -9,7 +9,7 @@ local util = require "util"
 --
 -- Values assigned to these variable need to fufil following condition:
 --
--- CACHE_FIRST_POLL_DELAY << CACHE_EXPIRATION < CACHE_POLL_PERIOD < CACHE_MAX_AGE_SOFT_LIMIT < CACHE_MAX_AGE_HARD_LIMIT
+-- CACHE_EXPIRATION < CACHE_POLL_PERIOD < CACHE_MAX_AGE_SOFT_LIMIT < CACHE_MAX_AGE_HARD_LIMIT
```
gpaul: We're changing the initial startup behaviour by removing `CACHE_FIRST_POLL_DELAY`. I'd like to know why exactly we used to have a different initial delay.

jgehrcke: I remember that I knew exactly. Please get back to me in person if we want to understand that again.

jgehrcke: We must not have race conditions around the startup of Admin Router.

One design idea is that once Admin Router is up and running (in the moment it starts processing HTTP requests) it should _immediately_ be fully functional. That was the case before introducing the cache, and it should be the case with the cache. For all types of HTTP requests.

With the current architecture -- is that true?

When the first HTTP request that needs to have a populated cache is coming in -- do we hold it back, populate the cache, and answer it?

mesosphere-mergebot-dev:
:wave: This PR has been inactive for 3 days. Please review the status checks to see what needs to be done to move it forward or change the label to Work in Progress. Thank you!

packages/adminrouter/extra/src/lib/cache.lua
gpaul: @ahairape I expect the flow to be something like the following if we want to keep the semantics identical:

1. in `init_worker_by_lua_block`
1.1 Schedule a closure to run once with `ngx.timer.at(update_once, CACHE_FIRST_POLL_DELAY)` 
1.2 Schedule a closure to run every 30 seconds with `ngx.timer.every(update_once, CACHE_POLL_DELAY)`

The `update_once` closure calls the existing cache update logic. 
This logic already prevents updates from racing.

Now, why we have any `CACHE_FIRST_POLL_DELAY` at all is something I don't yet understand. Ideally the worker would perform the cache update synchronously in `init_worker_by_lua_block` during startup. 
I believe that isn't possible as the ngx API available during startup is very limited so we background that work instead.
Still, I don't understand why the first poll delay isn't 0.

jgehrcke: > Still, I don't understand why the first poll delay isn't 0.

As far as I remember there are no strong technical reasons.

However, I think there is an important reason in terms of maintenance and making the code understandable: I am pretty sure that I set this to a non-zero number because it should not make a difference whether this is set to 0 s, 2s, or 15 s. I am pretty sure that all these values (including 0) make the system subject to the _same_ type of race condition where "something" happens in the time window between worker process initialization (where the `timer.at()` is being called) and _actual_ execution of the callback; and that something must be explicitly planned for.

Running into the race condition is much more likely with a non-zero value. That is, I think with the current code (which plans for said race condition) we _could_ totally set this to 0 but I think it might raise wrong expectations for new people coming to this code base.

jgehrcke:
In https://github.com/dcos/dcos/pull/2989#discussion_r199490006 I have mentioned a race condition and I also mentioned that we have code in place that takes care of handling it. [This is](https://github.com/dcos/dcos/blob/ae1a69b1bd06e59257a213668af9eadcdd13aa23/packages/adminrouter/extra/src/lib/cache.lua#L631) the code I was referring to:

```
    -- Handle the special case of very early request - before the first
    -- timer-based cache refresh
    local entry_last_refresh = cache:get(name_last_refresh)
    if entry_last_refresh == nil then
        refresh_cache(false, auth_token)

        entry_last_refresh = cache:get(name .. "_last_refresh")
        if entry_last_refresh == nil then
            -- Something is really broken, abort!
            ngx.log(ngx.ERR, "Could not retrieve last refresh time for `" .. name .. "` cache entry")
            return nil
        end
    end
```
Assuming that `timer.every()` only calls the callback _after_ the initial period has passed (and not _immediately_ and then every period) we actually make sure that this piece of code is regularly exercised: it is exercised for all HTTP requests that 
- arrive at Admin Router after worker process initialization and before the first timer-triggered cache refresh
- require a populated cache

For instance, I expect both to be true for our integration test suite.

It is important that we keep exercising this code path in our various test approaches.

When this part is broken / does not work as expected then we will run into issues like the following: https://jira.mesosphere.com/browse/DCOS-5916 (very hard to debug).

Please also see https://github.com/mesosphere/adminrouter/pull/59 for reference how I once stress-tested the startup phase of Admin Router.

ahairape:
Thanks for the useful info, @jgehrcke ! I will also repeat the test https://github.com/mesosphere/adminrouter/pull/59

packages/adminrouter/extra/src/lib/cache.lua
jgehrcke: To conclude the discussion: it looks like we can totally remove the `CACHE_FIRST_POLL_DELAY` (as done by the current version of the patch) if we make sure that we don't "accidentally" set it to 0. That is, the first timer-triggered cache update should come "late" so that the time window during which said race condition exists is large enough for us to regularly exercise the code path which handles the "http request-triggered cache update".

ahairape:
@mesosphere-mergebot bump-ee

jgehrcke:
> sdjlfadjflsajf

After all, let's please have the behavior change and the test change in two different commits, and the test change commit message should briefly explain why it was OK to change the test.

vespian:
> After all, let's please have the behavior change and the test change in two different commits, and the test change commit message should briefly explain why it was OK to change the test.

We paired with @ahairape on this for 1.5 hours. The commit message that you pointed out is just a quick dump of our work and @ahairape is going to polish it soon I think.

ahairape:
yes, I'm going to split the commit above into two and explain the changes

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -621,6 +611,14 @@ local function periodically_refresh_cache(auth_token)
     else
         ngx.log(ngx.INFO, "Created initial recursive timer for cache updating.")
```
vespian: @ahairape Let's please try to adjust the log messages as well, now that there are no recursive timers :) Thanks!

ahairape:
@mesosphere-mergebot bump-ee

ahairape:

ahairape:
After discussing it with @vespian we came to a conclusion that it is simpler to keep close to the the initial business logic of the code and keep the initial delay of 2s (CACHE_FIRST_POLL_DELAY) intact. The first cache update with ngx.timer.every() is then triggered every CACHE_POLL_PERIOD (25s) after the "AR start". This is different from the previous version where the first recursive cache update was triggered CACHE_FIRST_POLL_DELAY + CACHE_POLL_PERIOD = 27s  after the AR start. The modification affected the test test_cache.py, that's why the tests there have been changed as well (the timeouts adjusted) to reflect the modified logic.

ahairape:

packages/adminrouter/extra/src/lib/cache.lua
ahairape: Thanks for the hint, @vespian. I have fixed it.

ahairape:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -585,15 +585,13 @@ end
 
 local function periodically_refresh_cache(auth_token)
     -- This function is invoked from within init_worker_by_lua code.
-    -- ngx.timer.at() can be called here, whereas most of the other ngx.*
-    -- API is not available.
+    -- ngx.timer.every() is called here, a more robust alternative to
+    -- ngx.timer.at() as suggested by the openresty/lua-nginx-module
+    -- documentation:
```
vespian: Can we also link here to Jira issue that made us switch from `.at` to `.every`, please?

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -602,24 +600,24 @@ local function periodically_refresh_cache(auth_token)
...
-            ngx.log(ngx.ERR, "Failed to create timer: " .. err)
-        else
-            ngx.log(ngx.INFO, "Created recursive timer for cache updating.")
-        end
     end
 
-    -- Trigger initial timer, about CACHE_FIRST_POLL_DELAY seconds after
+    -- Trigger the timer, every CACHE_POLL_PERIOD seconds after
     -- Nginx startup.
     local ok, err = ngx.timer.at(_CONFIG.CACHE_FIRST_POLL_DELAY, timerhandler)
```
gpaul: Didn't we decide not to have an initial update? @jgehrcke

ahairape: @gpaul it's triggering the initial update with ngx.timer.at(_CONFIG.CACHE_FIRST_POLL_DELAY,...) but the comment is misleading, I will fix it now. Or do you mean something else with the "initial update"?

vespian: 2cents: Keeping the initial update code increases the chance that the incoming request will not have to wait for cache refresh. I.e. the initial update happens 2s after the initialization is complete, the scheduled update happens 25s after the init. During this 25s  it is much more likely that a request will be blocked while the cache is initialized than during 2s period which is the case when we keep the initial/short update. Also, rewriting the tests to match new logic would take some time extra time.

packages/adminrouter/extra/src/lib/cache.lua
ahairape: ack, done

jgehrcke:
> Didn't we decide not to have an initial update? @jgehrcke

Yes.

> Also, rewriting the tests to match new logic would take some time extra time

This is a bit sad :-). Because otherwise we actually agreed that it's technically meaningful to remove the initial update after 2 seconds. To quote from above:

> To conclude the discussion: it looks like we can totally remove the CACHE_FIRST_POLL_DELAY (as done by the current version of the patch)

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -585,15 +585,15 @@ end
...
 local function periodically_refresh_cache(auth_token)
     -- This function is invoked from within init_worker_by_lua code.
-    -- ngx.timer.at() can be called here, whereas most of the other ngx.*
-    -- API is not available.
+    -- ngx.timer.every() is called here, a more robust alternative to
+    -- ngx.timer.at() as suggested by the openresty/lua-nginx-module
+    -- documentation: 
+    -- https://github.com/openresty/lua-nginx-module/tree/v0.10.9#ngxtimerat
+    -- See https://jira.mesosphere.com/browse/DCOS-38248 for details on the
+    -- cache update problems caused by the recursive use of ngx.timer.at()
```
jgehrcke: Lovely, thanks for this comment. It will be very valuable to your future self, or to another future maintainer. Thanks Pawel for asking for that.

jgehrcke:
Let's also have the report from the manual test from https://github.com/mesosphere/adminrouter/pull/59, please.

jgehrcke:
@gpaul I had a quick discussion with @vespian and now I think we should keep the "Initial timer-triggered cache update *early* after nginx startup:" we had indeed one argument not stressed enough in this current discussion about the purpose of this early update: because the cache update might actually take long we want to minimize the likelihood for an incoming HTTP request to be slowed down (we talk of a slowdown of potentially multiple seconds).

I suggest adding a comment to the code, like this:


```
Initial timer-triggered cache update early after nginx startup:

Triggering a cache update roughly 2 seconds after the
startup of the worker process is a trade-off between the
following two aspects:

- It makes sense to have this initial timer-triggered cache update _early_ after
  nginx startup at all, and it makes sense to make it very early, so that we
  reduce the likelihood for an HTTP request to be slowed down when it is
  incoming _before_ the normally scheduled periodic cache update (example: the
  HTTP request comes in 15 seconds after nginx startup, and the first regular
  timer-triggered cache update is triggered only 25 seconds after nginx
  startup).

- It makes sense to have this time window not be too narrow, especially not
  close to 0 seconds: under a lot of load there *will* be HTTP requests incoming
  before the initial timer-triggered update, even if the first timer callback is
  scheduled to be executed after 0 seconds. There is code in place for handling
  these HTTP requests, and that code path must be kept explicit, regularly
  exercised, and well-tested. That is easier when the initial update is
  scheduled after e.g. 2 seconds, as opposed to 0 seconds.
```

ahairape:
@mesosphere-mergebot bump-ee

jgehrcke:
Thanks for performing the manual test.

jgehrcke:
@mhrabovcin @gpaul can you help pushing this across the finish line? Goal is to land this by EOD.

jgehrcke:
Someone needs to investigate the one failed CI check.

gpaul:
`teamcity/dcos/build/dcos/windows` is broken due to some `dcos-checks-runner on windows` issue:
https://teamcity.mesosphere.io/viewLog.html?buildId=1121238&buildTypeId=DcOs_Windows_Build_BuildDcOs

@branden is this something you can fix? @vespian do you know who is responsible for the "DC/OS on Windows" effort? If building on windows is not `REQUIRED` then it is a matter of time until it breaks.

gpaul:
`test_service_discovery.test_service_discovery_mesos_host` failed in `teamcity/dcos/test/aws/cloudformation/simple` and it appears to be a real error, or at least an extremely rare flake.

gpaul:
The test failure is a known issue https://jira.mesosphere.com/browse/DCOS_OSS-3539 which has already been fixed on master by https://github.com/dcos/dcos/pull/2987.

gpaul:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/cloudformation/simple https://jira.mesosphere.com/browse/DCOS_OSS-3539

vespian:
>  @vespian do you know who is responsible for the "DC/OS on Windows" effort? 

@branden?

gpaul:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/cloudformation/simple 
 https://jira.mesosphere.com/browse/DCOS-22377

gpaul:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/cloudformation/simple
https://jira.mesosphere.com/browse/DCOS-22377

jgehrcke:
@mesosphere-mergebot override-status teamcity/dcos/build/dcos/windows
https://jira.mesosphere.com/browse/DCOS-39083

ahairape:
@mesosphere-mergebot bump-ee

ahairape:
@mesosphere-mergebot bump-ee

ahairape:
Closing this PR in favour of https://github.com/dcos/dcos/pull/3064

PR #3026, opened 2018-07-01
Allocate different port than 8080 on master node (DCOS_OSS-3710)

deric:
## High-level description

Avoids port conflict on DC/OS master nodes, allowing Marathon service to start.

  - [DCOS_OSS-3710](https://jira.mesosphere.com/browse/DCOS_OSS-3710) dcos-log port conflict.

## Checklist for all PRs

  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: I'm not sure how to consistently replicate this behavior.

### DIFF
--- a/docs/admin-ports.md
+++ b/docs/admin-ports.md
@@ -27,6 +27,7 @@ The following is a list of ports used by internal DC/OS services, and their corr
  - 2181: dcos-exhibitor
  - 2888: dcos-exhibitor
  - 3888: dcos-exhibitor
+ - 4040: dcos-log
  - 5050: dcos-mesos-master
  - 7070: dcos-cosmos
  - 8080: dcos-marathon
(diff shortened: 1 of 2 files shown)


kensipe:
any update here?

deric:
@kensipe Any action required from me? I don't have access to CI logs, hard to tell if there is any issue with changes done in the PR. The issue could be OS specific, meaning that TCP port is allocated only on some distributions unless explicitly told to do so. But that's just a hunch.

br-lewis:
I'm surprised you're running into this. From a test cluster I just made, dcos-log is listening on `/run/dcos/dcos-log.sock` which is defined [here](https://github.com/dcos/dcos/blob/master/packages/dcos-log/extra/dcos-log.socket) so shouldn't be using up 8080 even though that's its default.

Can you give some more information about your setup?

I don't have any issue with this change but I'm curious how it came up.

PR #3043, opened 2018-07-09
Add more context to vip test app names.

jeschkies:
## High-level description

Debugging apps with just a random hash is hard. This changed introduces
more context to each app name so that they are unique *and* human
readable.


packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -75,7 +76,7 @@ def marathon_test_app(
     if network == marathon.Network.USER:
         assert host_port != 0, 'Cannot auto-assign a port on USER network!'
 
-    test_uuid = uuid.uuid4().hex
+    test_uuid = uuid.uuid4().hex if app_prefix is None else "{}-{}".format(app_prefix, uuid.uuid4().hex)
```
vespian: :heart:

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -299,13 +301,13 @@ def test_vip(dcos_api_session,
...
             log.error('Exception: {}'.format(e))
-            errors = errors + 1
+            errors.append(e)
         finally:
             log.info('Purging application: {}'.format(origin_app.id))
             origin_app.purge(dcos_api_session)
             log.info('Purging application: {}'.format(proxy_app.id))
             origin_app.purge(dcos_api_session)
-    assert errors == 0
+    assert len(errors) == 0
```
vespian: :heart:

jeschkies:
@mesosphere-mergebot bump-ee

jgehrcke:
```
E     File "/opt/mesosphere/packages/dcos-integration-test--1eef5613e6da82e2bc9955bfe9af8fdb946f3b1b/test_networking.py", line 337
E       def create_app_prefix(vip_net. proxy_net, container, ipv6, named_vip):
E                                    ^
E   SyntaxError: invalid syntax
```

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -332,7 +334,22 @@ def setup_vip_workload_tests(dcos_api_session, container, vip_net, proxy_net, ip
...
+        'ipv6' if ipv6 else 'ipv4'
+        ]
+    if named_vip:
+        prefixes.append('namedvip')
+
+    return '-'.join(prefixes)
+
+
 def vip_workload_test(dcos_api_session, container, vip_net, proxy_net, ipv6, named_vip, same_host):
+    app_prefix = create_app_prefix(vip_net, proxy_net, container, ipv6, named_vip)
```
urbanserj: I like this change very much! However, we would get extremely long names. Even if we use `.value.lower()` it would be 93-char string, e.g. `integration-test-test-vip-user-user-docker-ipv6-namedvip-proxy-9003a0924e8d4807b0ae05dcac8bb13d`. DNS have a strict limitation, it's 64 octets for labels. I would suggest to make a prefix from `named_vip` and `host`, all other information is already logged well by pytest.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -289,7 +291,7 @@ def test_vip(dcos_api_session,
     proxy container that will ping the origin container VIP and then assert
     that the expected origin app UUID was returned
     '''
-    errors = 0
+    errors = list()
```
urbanserj: `errors = []`

packages/dcos-integration-test/extra/test_networking.py
urbanserj: `assert not errors`

jeschkies: I like this one.

packages/dcos-integration-test/extra/test_networking.py
jeschkies: Which `host` parameter do you mean?

urbanserj: I meant `same_host`. if it's false we launch proxy app on second slave, but we can easily get this information from app `ip`.

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -332,7 +337,13 @@ def setup_vip_workload_tests(dcos_api_session, container, vip_net, proxy_net, ip
     return tests
 
 
+def create_app_prefix(vip_net, proxy_net, container, ipv6, named_vip):
```
urbanserj: After your last change, the function uses only 2 out of 5 arguments. I'd recommend to merge it into `vip_workload_test` as a one-liner.

packages/dcos-integration-test/extra/test_helpers.py
urbanserj: This change breaks `test_ipv6` test case. Would you please add `app_prefix` to id and keep test_uuid as is? We use test_uuid as an env variable, see `DCOS_TEST_UUID` [below](https://github.com/dcos/dcos/pull/3043/files#diff-fb347b4b1e1d7ab329071b182c56d964L90).

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -255,7 +260,7 @@ def test_ipv6(dcos_api_session, same_host):
         log.info('Purging application: {}'.format(origin_app.id))
         origin_app.purge(dcos_api_session)
         log.info('Purging application: {}'.format(proxy_app.id))
-        origin_app.purge(dcos_api_session)
+        proxy_app.purge(dcos_api_session)
```
urbanserj: Shame on me 🤦‍♂️

Would you please elaborate on your commit message? Do we call a wrong marathon api for purge?

> Oddly enough `purge` removes all app and pods. I don't know why this
> would be a member method of a Marathon app.

https://github.com/jeschkies/dcos/blob/26c82e1d388529233fe20252f22b643da59acc9a/packages/dcos-integration-test/extra/test_networking.py#L99-L100
https://github.com/jeschkies/dcos/blob/26c82e1d388529233fe20252f22b643da59acc9a/packages/dcos-integration-test/extra/test_networking.py#L155-L156

jeschkies:
@timcharper, @urbanserj and @vespian should we merge this?

jeschkies:
Nevermind, drop in favour of 3282

urbanserj:
@jeschkies we should! I love your changes, to unblock all others asap I had to move a coupe of most obvious fixes to a separate PR. Huge credit to you.

jeschkies:
@urbanserj, be my guest 🙂

PR #3118, opened 2018-07-18
Add a service to automatically drain Mesos node on reboot/halt

surdy:
## High-level description

This PR adds systemd units that are executed when a mesos agent (or public agent) node is rebooted (or shutdown), sends a SIGUSR1 to the agent process to try to gracefully drain the agent of its tasks . Details here : https://jira.mesosphere.com/browse/DCOS_OSS-3672 

  - [DCOS_OSS--3672](https://jira.mesosphere.com/browse/DCOS_OSS-3672) Automatically drain DC/OS agents before host reboot/shutdown

## Checklist for all PRs

___

Reviewers should be:

### DIFF
--- a/packages/mesos/extra/dcos-drain-mesos-agent-public.service
+++ b/packages/mesos/extra/dcos-drain-mesos-agent-public.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Drain Mesos Agent: Stop Mesos tasks on shutdown/reboot
+Before=shutdown.target reboot.target halt.target
+After=dcos-mesos-slave-public.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStop=/usr/bin/systemctl kill -s SIGUSR1 dcos-mesos-slave-public.service
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,8 @@ Format of the entries must be.
 
 * Root Marathon heap size can be customized during installation. (DCOS_OSS-3556)
 
+* Mesos: Drain agent node automatically on reboot or shutdown. (DCOS_OSS-3672) 
+
 ### Security Updates
 
 * Update cURL to 7.59. (DCOS_OSS-2367)
--- a/packages/mesos/extra/dcos-drain-mesos-agent.service
+++ b/packages/mesos/extra/dcos-drain-mesos-agent.service
@@ -1,0 +1,9 @@
+[Unit]
+Description=Drain Mesos Agent: Stop Mesos tasks on shutdown/reboot
+Before=shutdown.target reboot.target halt.target
+After=dcos-mesos-slave.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStop=/usr/bin/systemctl kill -s SIGUSR1 dcos-mesos-slave.service


surdy:

surdy:
@vinodkone @branden It took me a few days but I managed to send the PR. 

I can possibly use some pointers adding tests for this. Please do let me know who I should talk to for that ?

branden:
@surdy I don't think we have any support in our tests for stopping or restarting agents. This is simple enough that I'd be OK with a manual test. Existing integration tests should be enough to make sure that this doesn't hurt deployment or upgrades.

packages/mesos/extra/dcos-drain-mesos-agent-public.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Drain Mesos Agent: Stop Mesos tasks on shutdown/reboot
+Before=shutdown.target reboot.target halt.target
+After=dcos-mesos-slave-public.service
```
branden: nit: Since the only difference between the two units is the name of the Mesos service, we could turn this file into a template with the unit name parameterized. This file could reference the env var, and the build script could set its value and then render the actual unit file with `envsubst`. The Mesos build file is already doing something similar with other units.

packages/mesos/extra/dcos-drain-mesos-agent-public.service
```diff
@@ -0,0 +1,9 @@
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStop=/usr/bin/systemctl kill -s SIGUSR1 dcos-mesos-slave-public.service
```
vinodkone: I don't know systemd enough, but what's the guarantee that the agent *finishes* the draining before the host is rebooted? Does the "After" statement ensure that until the dcos agent service exits, the agent host reboot/shutdown is blocked?

packages/mesos/extra/dcos-drain-mesos-agent-public.service
surdy: Good suggestion. I was just lazy followed what I already saw with dcos-mesos-slave-public.service and dcos-mesos-slave.service. I will template it and update the PR

packages/mesos/extra/dcos-drain-mesos-agent-public.service
surdy: I myself had to spend some time in Systemd man for this. So its understandable to have questions around this. 

I am currently not doing anything to make sure the agent has actually finished draining. Is there a way to do determine programmatically that the agent has finished draining. If yes then I can wrap the SIGUSR1 in a script to make sure this service blocks reboot until the agent is drained.  

After=dcos-mesos-slave.service is there to ensure ensure that this is stopped before dcos-mesos-slave.service (stopping order is reverse of starting)

vinodkone: I think you can wait until the agent process exits to ensure draining is complete. But if for whatever reason the agent doesn't exit by itself (e.g., bug in draining path), then we would be stuck. The script you are proposing or this systemd unit should probably have a timeout as a fail safe.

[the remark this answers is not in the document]
surdy: I did not check if the agent ID persisted after the reboot.  I think it did register as the same agent. I can explicitly test that.

For behaviour of Marathon, I noticed in my test that the drained tasks were attempted to be replaced right away. 

Right now Marathon does have default unreachable strategy to replace unreachable task  immediately. But it does take sometime for a task to be marked as Unreachable after the agent is shutdown (depends on the agent ping timeout if I remember correctly). So the behaviour would be that if the nodes are drained the task would be replace more promptly than if it were not drained. 
Data services would try to wait for the agent to come back online to re-spin the task back on the same node (unless a human intervenes) 

Definitely if Mesos did a TASK_KILLED instead of TASK_LOST it would be awesome.

vinodkone: Also, I'm wondering if this drain should be part of the `service stop` workflow. Right now, we tell users in our docs that they need to drain via SIGUSR1 and then do `service stop`. Would be nice if they just do `stop` and the draining happens. Unless, there are cases where someone would want to stop the agent process but wants to keep the tasks running. Can `stop` take arguments (e.g., --drain)? I'm just spitballing here. I just find it weird that we expose users directly to the SIGUSR1, instead of a `drain` systemd command.

surdy: My very first though when I was playing around with this pre DCOS_OSS-3672 was to just add an ExecStop for the agent service. But then I recalled that we have used systemctl restart dcos-mesos-slave during debugging and have come to expect restarting mesos agent service to be a non-disruptive process and for good reason : Mesos agent service restart have always been non-disruptive long before DC/OS

I did not want to anything very fancy with systemd to I thought doing a separate unit dedicated for draining would be the best approach. I will google a bit to see if there is something else systemd has that we can use. The challenge is going to be the limitation that we only want to drain on system reboot/halt and not on service restart (which is divergent from behaviour a typical systemd service like HTTP server would want to do)

branden: Great point. Another reason not to drain the agent every time we stop it is that we expect DC/OS upgrades to not kill tasks.

surdy:

PR #3148, opened 2018-07-25
Add timestamp for dmesg, distro version, timedatectl and systemd unit status to diag bundle

surdy:
This is only for 1.12 OSS I will backport and do a PR for EE after this gets approved

## High-level description
Get following additional things in a diag bundle 
1) Have dmesg with human readable timestamp
2) `timedatectl` for ruling out time sync issues 
3) Get distro version to rule out untested distro bugs
4) Get DC/OS systemd unit status
5) /v2/pods marathon endpoint
 
What features does this change enable? What bugs does this change fix?
Reduced back and forth while debugging cluster issues since more data is collected in the bundle

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS_OSS-3861](https://jira.mesosphere.com/browse/DCOS_OSS-3861) Add more diagnostics data(e.g. timestamp for dmesg, distro version, systemd unit status etc) into the bundle

## Checklist for all PRs

  - [X] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change:
  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [X] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [X] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and review process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using GitHub's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

Reviewers should be:
* Developers who understand the code being modified.
[...]

PR #3200, opened 2018-08-02
Enable human readable timestamp in dmesg.

janisz:
From dmesg(1):

```
-T, --ctime
          Print human-readable timestamps.

          Be aware that the timestamp could be inaccurate!  The time
          source used for the logs is not updated after system
          SUSPEND/RESUME.
```

  - [DCOS_OSS-1020](https://jira.mesosphere.com/browse/DCOS_OSS-1020) Make dmesg output timestamps in diagnostic bundle human readable.

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -875,7 +875,7 @@ package:
           ],
           "LocalCommands": [
               {
-                  "Command": ["dmesg"]
+                  "Command": ["dmesg", "-T"]
               },
               {
                   "Command": ["ip", "addr"]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ Format of the entries must be.
 
 ### Fixed and improved
 
+* dcos-diagnostics capture `dmesg` with human readable timestamps. (DCOS_OSS-1020)
+
 * Admin Router: Change 'access_log' syslog facility from 'local7' to 'daemon'. (DCOS_OSS-3793)
 
 * Node and cluster checks are executed in parallel. (DCOS_OSS-2239)


gen/dcos-config.yaml
```diff
@@ -875,7 +875,7 @@ package:
           ],
           "LocalCommands": [
               {
-                  "Command": ["dmesg"]
+                  "Command": ["dmesg", "-T"]
```
klueska: @janisz mentioned that this will crash on centos6. @somikbehera @masonse Do we support this version? Do you see a problem with unconditionally adding this even if we don't "officially" support centos6 even if it will crash here?

klueska:
@janisz The summary of the DC/OS JIRA ticket you reference in the description doesn't match the name in JIRA. It should be "Make dmesg output timestamps in diagnostic bundle human readable"

gen/dcos-config.yaml
masonse: @klueska We do not support CentOS 6, I don't see a problem with this small change.

klueska:
It looks like this is failing the integration tests: test_dcos_diagnostics.test_dcos_diagnostics_bundle_download_and_extract:
https://github.com/dcos/dcos/blob/de1cc8abc03eddea81c2795c06cdfcee9c7d235d/packages/dcos-integration-test/extra/test_dcos_diagnostics.py#L481

It may just mean that a change to the test is required (since the output changed, you may need to update the test to account for timestamps now being present in the output)

gen/dcos-config.yaml
klueska: 👍

klueska:
Actually, I triaged this a bit, and it looks like the filename generated has changed from `'172.17.0.2_master/dmesg-0.output.gz'` to `'172.17.0.2_master/dmesg_-T-0.output.gz'` because of the flag you added. Maybe we should chat tomorrow to make sure this makes sense or not and how to adjust for it.

klueska:
Just verified this is expected by looking at other files generated from commands in the config, and I think this name change is OK. You just need to update the expected filename in the integration test.

klueska:
@surdy Are you OK with a name change in the diagnostics bundle from `'172.17.0.2_master/dmesg-0.output.gz'` to `'172.17.0.2_master/dmesg_-T-0.output.gz` in order to support adding timestamps to the dmesg output?

janisz:
Closing as a duplicate of #3148

PR #3220, opened 2018-08-06
Cleanup python versioning

janisz:
It looks like we are inconsistent about Python version used in Windows and Linux. This PR changes all usages of explicit version 3.5 or 3.6 or just 3.

This also fixes problem on Ubuntu 18.04 where there is no `/usr/bin/python` that causes tests to break.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -26,24 +26,15 @@ testpaths =
   ssh
   test_util
 
-[testenv:py35-syntax]
-platform=linux|darwin
+[testenv:py3-syntax]
+platform=linux|darwin|win32
 passenv =
     TEAMCITY_VERSION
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -63,15 +54,14 @@ passenv =
...
-  pytest-catchlog==1.2.2
   PyYAML
   webtest
   webtest-aiohttp==1.1.0
   schema
 commands=
   pytest --basetemp={envtmpdir} {posargs}
 
-[testenv:py36-unittests-win32]
+[testenv:py3-unittests-win32]
```
janisz: @klueska  Do we have a Windows CI to check if it's working?

PR #3246, opened 2018-08-09
Add test validating if bundle contains all data

janisz:
## High-level description

Currently tests only checks if specific files are in the bundle. This test validate if content of the file is correct. This change will not affect end user because it's test only change.

  - [DCOS_OSS-3954](https://jira.mesosphere.com/browse/DCOS_OSS-3954) Add test validating DCOS diagnostic bundle content

## Checklist for all PRs


janisz:
@gpaul @philipnrmn Can you take a look?

gpaul:
Looks good - please fix the link in the PR description, you're missing a ']':

```
[DCOS_OSS-3954(https://jira.mesosphere.com/browse/DCOS_OSS-3954) Add test validating DCOS diagnostic bundle content
```

gen/calc.py
```diff
@@ -1003,6 +1003,7 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
         'dns_search': '',
         'auth_cookie_secure_flag': 'false',
         'marathon_java_args': '',
+        'marathon_port': '8080',
```
branden: This is under the `default` section, which means user config can override it, which isn't what we want. This line needs to go under `must`, which is down below.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -401,51 +396,6 @@ def test_dcos_diagnostics_report(dcos_api_session):
         assert len(report_response['Nodes']) > 0
 
 
-def _get_bundle_list(dcos_api_session):
```
branden: Can you explain the motivation for moving these help functions to `test_helpers.py`? They don't seem to be useful to tests in other files. And we'd want to clean them up a bit before we publish them as general-purpose test utils (they need docstrings, `wait_for_diagnostics_job()` shouldn't be returning info by mutating arguments, `check_json()` shouldn't call `raise_for_status()`, etc.)

packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -23,6 +28,63 @@ def get_exhibitor_admin_password():
...
+
+def check_json(response):
+    response.raise_for_status()
+    try:
+        json_response = response.json()
+        logging.debug('Response: {}'.format(json_response))
+    except ValueError:
+        logging.exception('Could not deserialize response contents:{}'.format(response.content.decode()))
+        raise
+    assert len(json_response) > 0, 'Empty JSON returned from dcos-diagnostics request'
```
branden: This function is presented as a util, but this error messages assumes the response is from dcos-diagnostics.

janisz:
@branden Thanks for review. I will fix it

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
janisz: I agree. The main reason I moved this functions to helper is to solve Python issues with sources path. The test I'm adding here is moved from dcos-enterprise but IMO both OSS and EE version should contain this test. But if I leave this helpers untouched then this tests in EE version fails due to dependencies (`ModuleNotFound`) because OSS tests are imported in sub-directory in EE. 

I agree, this functions are not util functions and probably should not be available in that file. On the other hand fixing it this way is better than something like this:
```python
import sys
sys.path.insert(0,'.')
```

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -669,3 +619,58 @@ def test_diagnostics_bundle_status(dcos_api_session):
...
+            gzipped_contents = z.open(v2_apps_filename)
+            contents = gzip.decompress(gzipped_contents.read())
+            logging.debug("File {} had contents: {}".format(v2_apps_filename, contents))
+            apps = json.loads(contents)
+            # Ensure that the /v2/apps endpoint output captured in the dcos
+            # diagnostics bundle contains details for the sleeper
+            # application that we launched in this test.
+            had_sleeper = False
+            for app in apps['apps']:
+                if app['id'] == app['id']:
```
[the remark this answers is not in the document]
janisz: Good catch!!!

janisz:
It looks like this test need more attention. I'll prepare separate PR with definition of `marathon_port` that we can use later. And then focus on test.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
gpaul: IIRC these functions are imported from dcos-enterprise, hence a move to test_helpers makes sense.

janisz:

janisz:
Close. It need's more time to be fixed

PR #3384, opened 2018-09-04
Allow configuring dcos-diagnostics bundles dir

janisz:
## High-level description

Currently bundles are stored in `/var/lib` which is a root volume.

  - [DCOS_OSS-4040](https://jira.mesosphere.com/browse/DCOS_OSS-4040) dcos-diagnostics bundle files should not be stored in root volume by default

## Checklist for all PRs


janisz:
@mesosphere-mergebot bump-ee

janisz:

janisz:
@gpaul @ArmandGrillet @bamarni Can you take a look? I'm not sure if this is the best way to setup a directory for bundles but I have no idea how to do it better.

janisz:
@mesosphere-mergebot bump-ee

janisz:
@mesosphere-mergebot bump-ee

klueska:
@janisz I'm not sure if I agree with this change. Instead of changing the default location to somehwere non-standard (every other component puts its data in /var/lib), why not just allow the path for it to be cofigurable through `config.yaml`?

Similar to what @branden did here:
https://github.com/dcos/dcos/pull/1992

packages/dcos-diagnostics/extra/dcos-diagnostics-master.service
```diff
@@ -12,6 +12,8 @@ PermissionsStartOnly=True
 User=dcos_diagnostics
 # Allow r/w access to the socket file
 SupplementaryGroups=dcos_adminrouter
+ExecStartPre=/usr/bin/mkdir -p /var/log/dcos/dcos-diagnostics
```
klueska: This will break on Ubuntu. See:
https://github.com/dcos/dcos/pull/3405

gen/calc.py
```diff
@@ -1082,6 +1083,7 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
         'check_search_path': CHECK_SEARCH_PATH,
         'mesos_master_work_dir': '/var/lib/dcos/mesos/master',
         'mesos_agent_work_dir': '/var/lib/mesos/slave',
+        'dcos_diagnostics_bundles_dir': '/var/lib/dcos/dcos-diagnostics/diag-bundles',
```
janisz: @klueska Who is responsible for creating this directory? If I change it to `/var/log` where can I configure permissions and ownership?

janisz:
@klueska Fixed

gen/calc.py
gpaul: I like having the mkdir + chown in the unit. Could you put that back, please?

gen/dcos-config.yaml
```diff
@@ -678,7 +678,7 @@ package:
       {
         "role": "master",
         "pull": true,
-        "diagnostics-bundle-dir": "/var/lib/dcos/dcos-diagnostics/diag-bundles",
+        "diagnostics-bundle-dir": {{ dcos_diagnostics_bundles_dir }},
```
branden: This needs to be in quotes so that it renders as a valid JSON string:
```
        "diagnostics-bundle-dir": "{{ dcos_diagnostics_bundles_dir }}",
```

gen/calc.py
branden: If we assign this param's value to an env var here, the unit can reference that env var and create the directory: https://github.com/dcos/dcos/blob/225b085d6097a29784ee244fd92b1d360d072ab4/gen/dcos-config.yaml#L673-L675

gen/calc.py
```diff
@@ -970,6 +970,7 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
         lambda fault_domain_enabled: validate_true_false(fault_domain_enabled),
         lambda mesos_master_work_dir: validate_absolute_path(mesos_master_work_dir),
         lambda mesos_agent_work_dir: validate_absolute_path(mesos_agent_work_dir),
+        lambda dcos_diagnostics_bundles_dir: validate_absolute_path(dcos_diagnostics_bundles_dir),
```
branden: This is a user-facing config param, so I have to be picky about it. Let's drop the `dcos_` since that's redundant: `diagnostics_bundles_dir`.

gen/calc.py
janisz: Fixed

janisz:
@branden Thanks for comments. I fixed the code. Can you review again?

janisz:
@mesosphere-mergebot bump-ee

packages/dcos-diagnostics/extra/dcos-diagnostics-master.service
```diff
@@ -12,6 +12,8 @@ PermissionsStartOnly=True
 User=dcos_diagnostics
 # Allow r/w access to the socket file
 SupplementaryGroups=dcos_adminrouter
+ExecStartPre=/usr/bin/mkdir -p ${DCOS_DIAGNOSTICS_BUNDLE_DIR}
+ExecStartPre=/usr/bin/chown dcos_diagnostics ${DCOS_DIAGNOSTICS_BUNDLE_DIR}
```
branden: I think this is what's causing `teamcity/dcos/test/azure/arm` to fail. We can't assume `mkdir` and `chown` are at these locations, since they're elsewhere on Ubuntu. You'll need to wrap these commands in a `bash -c` so they get looked up in `$PATH`. See #3405.

janisz: `/bin/bash` might not be available on other systems shouldn't we use `/bin/sh`?

janisz:
@mesosphere-mergebot bump-ee

packages/dcos-diagnostics/extra/dcos-diagnostics-master.service
branden: That's worth considering. It's not a problem right now since every OS we support provides `/bin/bash`, and we already assume it's available on DC/OS nodes in several other places in the codebase. I do like being certain that we're using bash, and not some other shell that the OS decided to place at `/bin/sh` (it's dash on Ubuntu).

janisz:
@mesosphere-mergebot bump-ee

janisz:
@mesosphere-mergebot bump-ee

janisz:
@klueska @branden Can you take a look?

PR #3438, opened 2018-09-13
Merge dependent tests into one big scenario

janisz:
## High-level description

Currently we have 4 tests that depend on creating a bundle tests. This commit merge them into one big integration test simulation basic use case for diagnostics bundles. This mean we will create, check status, download and delete the bundle in single test. This should prevent tests from failing because due to ordering.

This commit also rearrange entries in `tests_dcos_diagnostics.py`. All tests are at the top and helper functions were moved to the bottom. This make code clear and easier to read.

  - [DCOS_OSS-4115](https://jira.mesosphere.com/browse/DCOS_OSS-4115) 

## Checklist for all PRs

### DIFF
--- a/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
+++ b/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
@@ -326,11 +326,11 @@ def test_dcos_diagnostics_bundle_create_download_delete(dcos_api_session):
     """
     test bundle create, read, delete workflow
     """
-    _create_bundle(dcos_api_session)
+    bundle = _create_bundle(dcos_api_session)
     _check_diagnostics_bundle_status(dcos_api_session)
-    _download_and_extract_bundle(dcos_api_session)
-    _download_and_extract_bundle_from_another_master(dcos_api_session)
-    _delete_bundle(dcos_api_session)
+    _download_and_extract_bundle(dcos_api_session, bundle)
+    _download_and_extract_bundle_from_another_master(dcos_api_session, bundle)
+    _delete_bundle(dcos_api_session, bundle)
 
 
 def _check_diagnostics_bundle_status(dcos_api_session):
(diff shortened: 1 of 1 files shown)


packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -640,35 +536,129 @@ def _get_dcos_diagnostics_health(z: zipfile.ZipFile, item: str):
...
+        assert node['host_ip'], 'host_ip cannot be empty'
+        assert node['role'], 'role cannot be empty'
+
+
+def validate_units(units):
+    assert isinstance(units, list), 'input argument must be list'
+    assert len(units) > 0, 'input argument cannot be empty'
+
+    for unit in units:
+        validate_unit(unit)
```
janisz: Here is the only change in code I made. Other changes are only done by moving functions down

janisz:
@mesosphere-mergebot bump-ee

janisz:

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -640,35 +536,139 @@ def _get_dcos_diagnostics_health(z: zipfile.ZipFile, item: str):
...
+        ', '.join(required_fields), unit)
+    for required_field in required_fields:
+        assert required_field in unit, '{} must be in a unit. Actual: {}'.format(required_field, unit)
+
+    # id, name, health, description, help should not be empty
+    assert unit['id'], 'id field cannot be empty'
+    assert unit['name'], 'name field cannot be empty'
+    assert unit['health'] in [0, 1], 'health must be 0 or 1'
+    assert unit['description'], 'description field cannot be empty'
+    assert unit['help'], 'help field cannot be empty'
```
branden: Future improvement for another PR: `validate_unit()` and `validate_units()` check a lot of the same stuff. Either `validate_units()` should call `validate_unit()` inside its for loop, or we should get rid of one and use the other in its place. Looks like `validate_unit()` only has one call site.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -401,55 +321,36 @@ def test_dcos_diagnostics_report(dcos_api_session):
...
+@pytest.mark.supportedwindows
+def test_dcos_diagnostics_bundle_crete_download_delete(dcos_api_session):
+    """
+    test bundle create, read, delete workflow
+    """
+    _create_bundle(dcos_api_session)
+    _check_diagnostics_bundle_status(dcos_api_session)
+    _download_and_extract_bundle(dcos_api_session)
+    _download_and_extract_bundle_from_another_master(dcos_api_session)
+    _delete_bundle(dcos_api_session)
```
branden: Future improvement for another PR: These functions seem to work together by coincidence, with the assumption that they're the only things creating diagnostics bundles as they execute. `_create_bundle()` creates a single bundle, but the other functions operate on all available bundles: all available bundles get downloaded, extracted, and deleted, regardless of what created them. This test could be improved by having `_create_bundle()` return the bundle's name and then passing it to the remaining functions, so that they operate on just that bundle.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -401,55 +321,36 @@ def test_dcos_diagnostics_report(dcos_api_session):
...
-    # if we ran this bit previously compare the current datapoint with the one we saved
-    if last_datapoint['time'] and last_datapoint['value']:
-        if percent_done <= last_datapoint['value']:
-            assert (datetime.datetime.now() - last_datapoint['time']) < datetime.timedelta(seconds=15), (
-                "Job is not progressing"
-            )
-    last_datapoint['value'] = percent_done
-    last_datapoint['time'] = datetime.datetime.now()
+@pytest.mark.supportedwindows
+def test_dcos_diagnostics_bundle_crete_download_delete(dcos_api_session):
```
branden: Typo: `crete` should be `create`.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
janisz: I did it but `validate_units` could not use `validat_unit`. `validate_units` has less fields than `unit`. I'm not sure if we want to loose our assertions just to have clean tests.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
janisz: I love this idea. I'll fix it after weekend.

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
branden: Oh interesting. It's unclear to me why one would test more fields than the other.

janisz: One endpoint returns more data. No idea why

janisz:
@mesosphere-mergebot bump-ee

janisz:
@branden @gpaul Fixed

orsenthil:
@mesosphere-mergebot override-status teamcity/dcos/test/azure/arm 
 https://jira.mesosphere.com/browse/DCOS-41677

orsenthil:

PR #3432, opened 2018-09-12
Ensure that statsd socket is cleaned up before launching telegraf

philipnrmn:
## High-level description

This PR fixes an issue where telegraf could crash-loop on agents. 

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS-41747](https://jira.mesosphere.com/browse/DCOS-41747) Telegraf dcos_statsd plugin cannot listen on its assigned socket file, causing a crash-loop

## Related tickets (optional)

Other tickets related to this change:

  - [SOAK-119](https://jira.mesosphere.com/browse/SOAK-119) 

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: this is a bugfix, and is covered by a previous entry
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a change in the launch script
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #3440, opened 2018-09-14
[1.12.0b1] Ensure that statsd socket is cleaned up before launching telegraf

philipnrmn:
cherry-picks #3432 into beta1


orsenthil:
@mesosphere-mergebot bump-ee

PR #3449, opened 2018-09-17
Add root capabilities to dcos-diagnostics

janisz:
## High-level description

With capabilities we can run root-only commands as a normal user. 

See: 
* https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Capabilities
* http://man7.org/linux/man-pages/man7/capabilities.7.html

  - [DCOS_OSS-3877](https://jira.mesosphere.com/browse/DCOS_OSS-3877) 
  - [DCOS_OSS-4138](https://jira.mesosphere.com/browse/DCOS_OSS-4138)


janisz:
@mesosphere-mergebot bump-ee

janisz:
@mesosphere-mergebot bump-ee

gen/dcos-config.yaml
```diff
@@ -974,6 +974,17 @@ package:
               },
               {
                   "Command": ["systemctl", "list-units", "dcos*"]
+              },
+              {
+                  "Command": ["iptables-save"]
+              },
+              {
+               "Command": ["/bin/sh", "-c", "cat /proc/`pgrep mesos-master`/environ"],
```
fatz: this is IMO a bit optimistic. We should better ask systemctl for the units PID file as this would be the source of truth

e.g.
```
"/bin/sh", "-c", "cat /proc/`systemctl show dcos-mesos-master.service -p MainPID| cut -d'=' -f2`/environ"
```

gen/dcos-config.yaml
```diff
@@ -974,6 +974,17 @@ package:
...
+               "Role": ["master"]
+              },
+              {
+               "Command": ["/bin/sh", "-c", "cat /proc/`pgrep mesos-agent`/environ"],
```
fatz: same here we should use systemd to get the PID

packages/dcos-diagnostics/extra/dcos-diagnostics-master.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=DC/OS Diagnostics Master: aggregates and exposes component health
 [Service]
+AmbientCapabilities=CAP_SYS_PTRACE CAP_DAC_OVERRIDE CAP_NET_RAW CAP_NET_ADMIN
```
fatz: so this might a bit to much. In general this means the process is root as it could overwrite other processes memory and has access to any file. I'm not quite sure if this will be a better solution than running this as root as it would have in fact root like permissions.

Can you maybe describe what each capability is used for in the context of DC/OS Diagnostics?

janisz: * `CAP_NET_RAW CAP_NET_ADMIN` are required by `iptables-save`
* `CAP_SYS_PTRACE CAP_DAC_OVERRIDE` are required to see `/proc/[pid]/environ`

I think this is the minimal set to get this data

janisz:
@mesosphere-mergebot bump-ee

packages/dcos-diagnostics/extra/dcos-diagnostics-master.service
lloesche: What if we had a wrapper utility which gets setuid 4750 perms with same group as dcos-diagnostics and all it's capable of doing is call iptables and read /proc...environ? Sort of like `ping` and `traceroute` for instance used to work.
My thoughts are going in the direction of.. what if future versions of dcos-diagnostics allow running of custom diagnostics scripts, those scripts if not carefully implemented would inherit the elevated rights. And even now it's likely easier than making sure each diagnostics check we're calling can't be used to exploit the system or dropping privileges where applicable.
Just an idea of how we might keep the attack surface somewhat small and manageable.

janisz: I like this idea but it makes it harder to manage. We will need to create `cat` and `iptables-save` command with elevated privileges hidden somewhere in `/opt/mesosphere/bin`

klueska: I'm curious what @gpaul and @jgehrcke think about this. I'm also struggling to see a clean architecture for what @lloesche proposes without a big change to dcos-diagnostics. Is there some middle ground we can arrive at?

gpaul: Let's start with an attack vector. What is the threat model and aren't we already exposed? Should we consider deferring fine-grained permissions to a more comprehensive "selinux policy for dc/os" project?

janisz: @klueska  @gpaul @lloesche @fatz  What is the conclusion?

gpaul:
This PR does not impact the ultimate security of this project:
- `dcos-rexray.service` runs as `root`.
- `dcos-net.service` runs as `root`.
- `dcos-pkgpanda-api.service` runs as `root`.

Using `capabilities` instead of just running dcos-diagnostics as `root` introduces a maintenance burden by haphazardly introducing `capabilitites` into DC/OS. There is no prior art.

What we see here is good folks seeing a bad practice and wanting to do the best we can. That's clearly a good thing. However, deciding to switch from `User=root` to a separate mechanism needs careful consideration. Capabilities are probably _not_ the answer given how quickly (the first PR!) we had to reach for `CAP_SYS_ADMIN`. The security PM and security team should carefully research alternatives like `selinux`, custom binaries (like @lloesche suggested), etc.

Let's not block a very useful PR which makes DC/OS significantly easier to debug with what are essentially feature requests for tighter security.

My recommendation, given the prior art in this project, is:
- Create a COPS ticket assigned to the security team to implement a solution that works _for all components_ instead of introducing a new technique in a lonely corner of the project. I've created this COPS ticket here: https://jira.mesosphere.com/browse/COPS-3847
- Switch this PR to run the service as `root`. This simplifies maintenance by sticking to prior art. Also, capabilities don't work for this use case: the moment you need to reach for CAP_SYS_ADMIN you're doomed.
- Concentrate our well-deserved skepticism and improved designs in COPS-3847 where they can be referenced in future and help form the basis of a project-wide strategy instead. 
- Ship this PR ASAP

janisz:
@klueska @fatz @jgehrcke @lloesche Any opinion about running dcos-diagnostics as a root?

fatz:
@janisz its not ideal. But to me this is a better solution than having an unprivileged user which got high privileges through capabilities.

So I'm fine with running it as root as long as we try to come up with a solution like the one suggested by @lloesche later on. So I would approve using root

janisz:
@mesosphere-mergebot bump-ee

janisz:
@mesosphere-mergebot bump-ee

janisz:
@klueska @gpaul  @jgehrcke @lloesche I altered PR to run dcos-diagnostics as root.

klueska:
That’s fine with me. Interestingly, I think we only decided to go with Capabilites instead of root in the first place because of a suggestion by @gpaul on the ticket this is resolving. Or maybe I suggested it and he +1‘d it. Either way, going with root feels like the right solution fir the time being until we think through how to do it genetically with capabilites for other components as well.

gpaul:
> Interestingly, I think we only decided to go with Capabilites instead of root in the first place because of a suggestion by @gpaul on the ticket this is resolving. Or maybe I suggested it and he +1‘d it.

Hahahahaha this is classic

CHANGES.md
```diff
@@ -15,6 +15,8 @@ Format of the entries must be.
 
 ### What's new
 
+* Add root capabilities to dcos-diagnostics (DCOS_OSS_3877)
```
gpaul: /Add root capabilities to dcos-diagnostics/Run dcos-diagnostics as the `root` user./

janisz:
@gpaul Fixed

@mesosphere-mergebot bump-ee

janisz:
@mesosphere-mergebot bump-ee

gpaul:
@fatz could you approve the bump as well, please? https://github.com/mesosphere/dcos-enterprise/pull/3441

orsenthil:

janisz:
I backported it to 1.12 Can you take a look? https://github.com/dcos/dcos/pull/3578

PR #3473, opened 2018-09-19
Finish porting pkgpanda build process to windows

klueska:
This PR adds a bunch of functionality to the Windows port of the build process for `pkgpanda`.
It is most easily reviewable by going through it commit-by-commit, reading the commit message and verifying that the changes in the commit actually do (and only do!) what the commit message says.

It begins by updating parts of pkgpanda to be more friendly to cross-platform builds, followed by updates to make it fully supported on Windows (some parts of the `pkgpanda` build process had been ported in the past, but weren't fully functional). It also ports pieces of `pkgpanda` that hadn't been ported before, such as the API and the CLI, as well as the release tooling.

All of the relevant `pkgpanda` unit tests are now enabled for Windows as part of this PR!

However, there are still some tests related to the `installer` and `gen` that are being skipped. We need to follow up in a subsequent PR to figure out what's necessary to enable these and get them to pass.

Finally, we have a bunch of TODOs sprinkled throughout the commits in this PR related to improvements that could/should be made in the near future. Most importantly, refactoring `pkgpanda` itself to make unix/windows cross-platform builds more maintainable.

  - [DCOS_OSS-1899](https://jira.mesosphere.com/browse/DCOS_OSS-1899) Build dcos in a windows environment.

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: Not user-facing, so no changes needed here. Once we have *actual* windows support enabled, we will update the changelog.
  - [x] Included a test which will fail if code is reverted but test is not.


klueska:
@mesosphere-mergebot bump-ee

klueska:
The check for `DC/OS/Open/Windows/Build/Build DC/OS` is failing due to a mismatch in the base windows os where the job is run. It looks like there are tow windows agent pools -- one for windows server 1709 machines, and one for windows server 1803 machines.

Prior to this PR, this job required machines from the 1709 pool. This PR (and when this gets merged, all future PRs) will require agents from the 1803 pool.

What is the best way to address this? @dcos/tools-infra

klueska:
Actually, it looks like since the last run, the test mentioned above has somehow been removed. I think we should add it back in either now or after this PR is merged.

pkgpanda/actions.py
```diff
@@ -195,9 +195,19 @@ def _start_dcos_target(block_systemd):
 
 def _get_package_list(package_list_id: str, repository_url: str) -> List[str]:
     package_list_url = repository_url + '/package_lists/{}.package_list.json'.format(package_list_id)
-    with tempfile.NamedTemporaryFile() as f:
-        download(f.name, package_list_url, os.getcwd(), rm_on_error=False)
-        package_list = load_json(f.name)
+
+    # TODO(klueska): On Windows there are issues with following a simple
+    # "delete-on-close" semantic, so we instead pass 'delete=False' when we
```
jongiddy: Is there (a link to) a description of the problem / symptoms here? It would be nice to be able to clean this up one day, but without knowing how this was observed to deviate from the documented behavior, it will become a "safer to leave messy" problem.

pkgpanda/build/__init__.py
```diff
@@ -1114,20 +1114,19 @@ def cache_abs(filename):
...
-                    ["cmd.exe", "/c", "if", "exist", filename, "rmdir", "/s", "/q", filename])
-            filename = PKG_DIR + "\\result"
-            cmd.run("package-cleaner",
-                    ["cmd.exe", "/c", "if", "exist", filename, "rmdir", "/s", "/q", filename])
+            # These native calls handle symlinks better than a simple 'rm -rf'
+            # on windows, as well as perform retries when the operating system
+            # does not release the files quick enough for deletion and renaming.
+            remove_dir = package_store.get_package_cache_folder(name)
+            remove_directory(remove_dir + "\\src")
+            remove_directory(remove_dir + "\\result")
```
jongiddy: Please use `os.path.join` where possible, even when we know which platform we are on.

jongiddy:
@branden Am I right to consider this PR to be Ready for Review rather than Work in Progress?

pkgpanda/util.py
```diff
@@ -62,7 +63,16 @@ def remove_directory(path):
         # so calling out to the cmd prompt to do this fixes that.
         path = path.replace('/', '\\')
         if os.path.exists(path):
-            subprocess.call(['cmd.exe', '/c', 'rmdir', '/s', '/q', path])
+            for retry in range(1, 10):
```
jongiddy: Although the implementation appears to be correct, I would prefer to see this done using one of the Python `retry` or `retrying` modules.

pkgpanda/util.py
```diff
@@ -77,6 +77,23 @@ def remove_directory(path):
         subprocess.check_call(['rm', '-rf', path])
 
 
+def rename_file(src, dst):
+    """Rename a file or directory"""
+    if is_windows:
+        for retry in range(1, 10):
```
jongiddy: Again, there's the Python module `retry` for doing this.

pkgpanda/util.py
```diff
@@ -95,11 +126,82 @@ def copy_directory(src_path, dst_path):
...
+    """Check if a file is a link to another file in an OS independent manner.
+
+    On Linux, this will check for symbolic links for both files and directories.
+    On Windows, this will check for hard links for files, and junctions (a.k.a symbolic links) for directories.
+
+    """
+    if is_windows:
+        if os.path.isdir(path):
+            return _is_junction(path)
+        elif os.stat(path).st_nlink > 1:
```
jongiddy: There are only a few places where `islink` is used. In one, it is only used for directories. In the other, Windows is skipped because this code identifies a real file as a link. Given that we are not using Windows symbolic links (BTW any reason for that?), it seems that it would be OK to return False for all Windows files and simplify the places where `islink` is used.

pkgpanda/util.py
```diff
@@ -160,6 +160,19 @@ def islink(path):
         return os.path.islink(path)
 
 
+def realpath(path):
```
jongiddy: If `islink` is modified to return `False` for all Windows files, then this function can simply be replaced with `os.path.realpath`.

pkgpanda/build/src_fetchers.py
```diff
@@ -220,39 +220,49 @@ def _identify_archive_type(filename):
...
-    dir_contents = os.listdir(path)
+    files = os.listdir(path)
 
-    if len(dir_contents) != 1 or not os.path.isdir(os.path.join(path, dir_contents[0])):
+    if len(files) == 0:
+        raise ValidationError("Extracted archive is empty")
+
+    dirs = [f for f in files if os.path.isdir(os.path.join(path, f))]
+
+    if len(dirs) > 1 or (len(files) > 1 and not allow_extra_files):
```
jongiddy: The original code handled 0 directories by raising this `ValidationError`. The new code returns successfully if the archive contains 0 directories. Change `len(dirs) > 1` to `len(dirs) != 1`.

pkgpanda/build/src_fetchers.py
```diff
@@ -220,39 +220,49 @@ def _identify_archive_type(filename):
...
+    set to True.
 
     Args:
         path: path where extracted archive contents can be found
+        allow_extra_files: if True, allow other files next to the top level directory
     """
-    _check_components_sanity(path)
-
-    top_level_dir = os.path.join(path, os.listdir(path)[0])
+    _check_components_sanity(path, allow_extra_files)
```
jongiddy: Since `_check_components_sanity` gets to the end with a variable `dirs` containing the name of the single directory in the archive, perhaps return it from there and use it to save the scan in the next line.

jongiddy:
I can't get `pytest pkgpanda` to get fewer than 8 failures on Windows. I see these tests have been unskipped, so I expect them to pass. Are there instructions for setting this up so that these tests pass?

My steps were:
```powershell
> virtualenv venv --python="C:\Python36\python.exe"
> Set-ExecutionPolicy -Scope CurrentUser Unrestricted -Force
> . .\venv\Scripts\activate.ps1
> pip install --upgrade pip setuptools wheel
> pip install -e $PWD
> pytest pkgpanda
```

jongiddy:
I tried running `tox` directly, but it gives the same result: 8 failures in the `pkgpanda` tests. 

To be confident that this is working, we need to, at a minimum:
- get 0 failures for a `tox` run.
- ensure that `python setup.py build` works: https://github.com/dcos/dcos/pull/3473#pullrequestreview-190817345
- fix the minor bug for empty archives: https://github.com/dcos/dcos/pull/3473#discussion_r246781390

I'd like to see other issues addressed, but they do not necessarily have to be fixed immediately for this PR to be merged.

klueska:
The goal with this PR was not to try and "fix" things that are broken in `pkgpanda`, but rather to just get the windows port from Paul working with how `pkgpanda` currently operates, and then go back and fix up the problems with it later (with a better architecture for windows support as part of this). 

In terms of testing, are you running in VM on azure that I set up to run all the tests? This mimics the environment that team-city sets up for it when the tests are run there. Last time I was looking at this PR, all tests were passing with a simple `tox` command. @DominikDary should have a document which points out where this dev machine is on the Mesosphere Azure account.

jongiddy:
I figured out that the cause of most of the test failures is due to Windows Git not handling symlinks by default. Reconfiguring Git to support symlinks, and running everything that can create a symlink as Administrator, I'm left with one error. This is because Windows hard links cannot multiple partitions.

Is there a reason not to use symlnks in Windows? I can't believe we plan to support older versions of Windows that don't support symlinks, and requiring filesystems that do support them seems a reasonable limitation to impose.

```
====================================================== FAILURES =======================================================
_______________________________________________ test_activate_packages ________________________________________________

tmpdir = local('C:\\Users\\jongi\\AppData\\Local\\Temp\\pytest-of-jongi\\pytest-9\\test_activate_packages0')

    def test_activate_packages(tmpdir):
        _set_test_config(app)

        install_dir = str(tmpdir.join('install'))
        copytree(resources_test_dir('install'), install_dir, symlinks=True)
        packages_dir = str(tmpdir.join('packages'))
        copytree(resources_test_dir('packages'), packages_dir, symlinks=True)

        # On windows we require 'junctions' (a.k.a directory hardlinks) for any
        # links we create to directories. Unfortunately, git doesn't allow us to
        # commit 'junctions' back to a git repository. Instead, 'junctions' are
        # treated by git as copied directories and committed as such. This is
        # obviously, not what we want for this test, however, because we expect the
        # 'activated' packages to be links, not copies.
        #
        # To address this problem, we keep the committed folders under
        # 'install/active' as symlinks in the git repo, and dynamically change
        # them to 'junctions' when this test is invoked.  Unfortunately, this
        # then causes problems when pytest cleans up the temporary directory
        # where these junctions are created later on.
        #
        # Pytest recursively deletes all files in each 'junction' before
        # attempting to delete the 'junction' itself.  This causes the original
        # files contained in the junction (not copies or symlinks of them).
        # Therefore, whenever pytest removes 'tmpdir' any files underneath any
        # junctions we create within get deleted at their source (which occurs on
        # subsequent invocations of pytest just before running any tests).
        #
        # To address this, we force the 'active_dir' directory (where we create
        # all dynamic 'junctions') to be deleted before the current test run
        # completes. We use our custom 'remove_directory()` function from
        # 'pkgpanda.util', which knows how to safly remove these junctions
        # without deleting the files underneath them.
        #
        # TODO(klueska): Look into using *actual* symlinks on windows so that we
        # can avoid the use of junctions and the mess that come with them.
        #
        # TODO(klueska): Look into using a test fixture instead of `atexit` to do
        # the deletion of the 'active_dir'.
        if is_windows:
            active_dir = install_dir + '\\active'
            _fixup_active_dir_symlinks(active_dir)
            atexit.register(lambda path: remove_directory(path), active_dir)

        app.config['DCOS_ROOT'] = install_dir
        app.config['DCOS_ROOTED_SYSTEMD'] = True
        client = app.test_client()

        # Upgrade from mesos--0.22.0 to mesos--0.23.0.
        old_packages = [
            'mesos--0.22.0',
            'mesos-config--ffddcfb53168d42f92e4771c6f8a8a9a818fd6b8',
        ]
        new_packages = [
            'mesos--0.23.0',
            'mesos-config--ffddcfb53168d42f92e4771c6f8a8a9a818fd6b8',
        ]
        assert_json_response(client.get('/active/'), 200, old_packages)
        assert_response(
            client.put(
                '/active/',
                content_type='application/json',
                data=json.dumps(new_packages),
            ),
            204,
>           b'',
        )

pkgpanda\test_http.py:216:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

response = <Response streamed [500 INTERNAL SERVER ERROR]>, status_code = 204, body = b'', headers = {}
body_cmp = <built-in function eq>

    def assert_response(response, status_code, body, headers=None, body_cmp=operator.eq):
        """Assert response has the expected status_code, body, and headers.

        body_cmp is a callable that takes the response body and expected body and
        returns a boolean stating whether the comparison succeeds.

            body_cmp(response.data, body)

        """
        headers = headers or {}

>       assert response.status_code == status_code, (
            'Expected status code {}, got {}'.format(status_code, response.status_code)
        )
E       AssertionError: Expected status code 204, got 500
E       assert 500 == 204
E         -500
E         +204

pkgpanda\test_http.py:22: AssertionError
-------------------------------------------------- Captured log call --------------------------------------------------
server.py                   28 ERROR    An unexpected error has occurred.
Traceback (most recent call last):
  File "d:\github\dcos\dcos\venv\lib\site-packages\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "d:\github\dcos\dcos\venv\lib\site-packages\flask\app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "D:\github\dcos\dcos\pkgpanda\http\server.py", line 228, in activate_packages
    block_systemd=False)
  File "D:\github\dcos\dcos\pkgpanda\actions.py", line 34, in activate_packages
    install.activate(repository.load_packages(package_ids))
  File "D:\github\dcos\dcos\pkgpanda\__init__.py", line 875, in activate
    symlink_all(pkg_dir, new)
  File "D:\github\dcos\dcos\pkgpanda\__init__.py", line 831, in symlink_all
    symlink_tree(src, dest)
  File "D:\github\dcos\dcos\pkgpanda\__init__.py", line 565, in symlink_tree
    link_file(src_path, dest_path)
  File "D:\github\dcos\dcos\pkgpanda\util.py", line 149, in link_file
    os.link(src, dst)
OSError: [WinError 17] The system cannot move the file to a different disk drive: 'D:\\github\\dcos\\dcos\\pkgpanda\\test_resources\\windows\\packages\\mesos--0.23.0\\bin\\mesos' -> 'C:\\Users\\jongi\\AppData\\Local\\Temp\\pytest-of-jongi\\pytest-9\\test_activate_packages0\\install\\bin.new\\mesos'
================================================== warnings summary ===================================================
pkgpanda\util_test.py:15
  D:\github\dcos\dcos\pkgpanda\util_test.py:15: PytestWarning: cannot collect test class 'TestRepo' because it has a __init__ constructor
    class TestRepo:

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=============================================== short test summary info ===============================================
SKIP [1] pkgpanda\test_util.py:377: Windows does not have a root group
============================= 1 failed, 41 passed, 1 skipped, 1 warnings in 35.79 seconds =============================
```

build_local_windows.ps1
```diff
@@ -29,21 +31,42 @@ $config_yaml =
 "storage: `
    local: `
     kind: local_path `
-    path: $myhome/dcos-artifacts `
+    path: c:/dcos-artifacts `
```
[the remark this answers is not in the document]
branden: Thanks, fixed the comment. I'm not sure whether this will cause any problems. Is there a reason for concern?

pkgpanda/actions.py
branden: The problem is that on Windows you can't call `open()` on a file that's already opened, as it'll raise an exception. Since on Windows we have to close the tempfile after writing before it can be read, `delete=True` renders it useless. Comment updated to clarify.

pkgpanda/docker.windows/dcos-builder/Dockerfile
```diff
@@ -59,5 +66,6 @@ RUN C:\TEMP\vs_buildtools.exe --quiet --wait --norestart --nocache --installPath
     || IF "%ERRORLEVEL%"=="3010" EXIT 0
 
 # Start developer command prompt with any other commands specified.
-
-ENTRYPOINT C:\BuildTools\Common7\Tools\VsDevCmd.bat && powershell.exe -file
\ No newline at end of file
+# note: vsdevcmd.bat builds for x86 by default, use these options for amd64:
```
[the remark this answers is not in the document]
branden: I don't know anything about those two scripts, unfortunately. I think the `&&` is necessary because this is intended to run the script defined in the entrypoint before running the command provided at run-time. (For reference, the image created by this Dockerfile is used in `pkgpanda/build/__init__.py`, in `_build()`.)

pkgpanda/docker.windows/dcos-builder/Dockerfile
```diff
@@ -24,7 +24,8 @@ ADD https://github.com/mesos/3rdparty/raw/master/make-3.81.exe c:\temp\make-3.81
 COPY .\setup-make.ps1 \temp\
 RUN  powershell.exe -file c:\temp\setup-make.ps1
 
-ADD http://erlang.org/download/otp_win64_19.3.exe c:\temp\otp_win64_19.3.exe
+
+ADD https://packages.erlang-solutions.com/erlang/esl-erlang/FLAVOUR_1_general/esl-erlang_20.3~windows_amd64.exe c:\temp\otp_win64_20.3.exe
```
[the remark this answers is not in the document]
branden: Already there :)

pkgpanda/docker.windows/dcos-builder/Dockerfile
```diff
@@ -33,13 +34,19 @@ COPY .\setup-golang.ps1 \temp\
...
 ENV GOPATH 'c:\gopath'
 
+ADD https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-win.zip c:\temp\ninja.zip
+RUN powershell -command "& new-item -itemtype directory ${env:ProgramFiles(x86)}\ninja"
+RUN powershell -command "& Expand-Archive c:\temp\ninja.zip ${env:ProgramFiles(x86)}\ninja"
+
+
 ENV GOLANG_PATH 'C:\gopath\bin;c:\go\bin'
-ENV GIT_PATH 'C:\Program Files\git\cmd'
+ENV GIT_PATH 'C:\Program Files\git\cmd;C:\Program Files\git\bin'
```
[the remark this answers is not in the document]
branden: I'm not sure what command this was intended for.

pkgpanda/docker.windows/dcos-builder/Dockerfile
```diff
@@ -33,13 +34,19 @@ COPY .\setup-golang.ps1 \temp\
 RUN  powershell.exe -file c:\temp\setup-golang.ps1
 ENV GOPATH 'c:\gopath'
 
+ADD https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-win.zip c:\temp\ninja.zip
+RUN powershell -command "& new-item -itemtype directory ${env:ProgramFiles(x86)}\ninja"
+RUN powershell -command "& Expand-Archive c:\temp\ninja.zip ${env:ProgramFiles(x86)}\ninja"
```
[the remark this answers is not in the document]
branden: I'm not sure. It may have been added to support building a package that's not part of this PR. I'll see if I can remove this without breaking anything.

pkgpanda/build/__init__.py
```diff
@@ -1114,20 +1114,19 @@ def cache_abs(filename):
...
         else:
+            # Run a docker container to remove src/ and result/
```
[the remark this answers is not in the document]
branden: I'm honestly not certain why we use a container here. Could also be that it was used to pave over a difference between Linux and OSX. This code has succeeded with the Windows build job back when that was still being run against PRs, so I'm pretty sure it works.

pkgpanda/util.py
branden: I think there's a case in `pkgpanda/__init__.py::symlink_tree()` where we use `islink()` on files:

```python
                if not os.path.isdir(dest_path) and not islink(dest_path):
```

pkgpanda/util.py
branden: See https://github.com/dcos/dcos/pull/3473#discussion_r251722959, I don't think we can get away with always returning `False` on Windows.

pkgpanda/build/src_fetchers.py
branden: This stanza above should already cover that case:

```python
    if len(files) == 0:
        raise ValidationError("Extracted archive is empty")
```

pkgpanda/build/src_fetchers.py
branden: Yeah that is inefficient. However that's the existing behavior, and this PR is already quite large. If you don't mind, I'd rather save that for another PR.

branden:
> The rename of `pkgpanda/http/__init__.py → pkgpanda/http/server.py` breaks the command `python setup build.py`, since Python no longer sees the folder as a Python module.
> 
> Adding an empty `__init__.py` file appears to fix this.
> 
> [EDIT - adding the `__init__.py` appears to prevent `http/server.py` finding the standard module `http.client`. We might have to think a bit more how to solve this properly!]

Yeah, we might. I think being unable to import `http.client` was the problem the rename was intended to solve. But I wasn't able to get an explanation for what was causing that problem, or why the import behaves differently on Windows.

> Is there a reason not to use symlnks in Windows? I can't believe we plan to support older versions of Windows that don't support symlinks, and requiring filesystems that do support them seems a reasonable limitation to impose.

To be honest, I'm not certain. This code seems to work hard to avoid native symlinks, but it's not clear to me why.

I'll follow up with updates that use `retrying` and `os.path` where appropriate.

pkgpanda/build/src_fetchers.py
jongiddy: Consider the case that `allow_extra_files` is `True` and `path` contains one or more (non-directory) files.

PR #3696, opened 2018-10-27
[1.13] Cache executor container metadata in metrics

gracedo:
## High-level description

This fixes a bug in the dcos-metadata plugin in telegraf where executor containers were not being cached. This caused repeated hits on Mesos due to constant cache misses and missing information in the metrics output.

  - [DCOS_OSS-4181](https://jira.mesosphere.com/browse/DCOS_OSS-4181) Telegraf does not retrieve executor metadata from Mesos

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: will be added to 1.12 changelog

### DIFF
--- a/packages/dcos-integration-test/extra/test_metrics.py
+++ b/packages/dcos-integration-test/extra/test_metrics.py
@@ -136,6 +136,81 @@ def test_metrics_agents_statsd(dcos_api_session):
             get_metrics_prom(dcos_api_session, agent, expected_metrics)
 
 
+@retrying.retry(wait_fixed=2000, stop_max_delay=150 * 1000)
+def check_metrics_prom(dcos_api_session, prometheus_port, node, check_func):
+    """Get metrics from prometheus port on node and run check_func function,
+    asserting that it returns True.
+
+    Retries on non-200 status or failed assertions
+    for up to 150 seconds.
+
+    """
+    response = dcos_api_session.session.request(
+        'GET', 'http://{}:{}/metrics'.format(node, prometheus_port))
+    assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
+    assert check_func(response) is True
+
+
+@pytest.mark.parametrize("prometheus_port", [61091])
+def test_metrics_metadata(dcos_api_session, prometheus_port):
+    # install framework
+    install_response = dcos_api_session.cosmos.install_package('kafka')
+    data = install_response.json()
+
+    dcos_api_session.marathon.wait_for_deployments_complete()
+
+    wait = True
+    executor_mesos_id = task_mesos_id = ''
+    while wait:
+        master = dcos_api_session.masters[0]
+        state_response = dcos_api_session.get('/state', host=master, port=5050)
+        assert state_response.status_code == 200
+        state = state_response.json()
+        for framework in state['frameworks']:
+            if framework['name'] == 'marathon':
+                for task in framework['tasks']:
+                    if task['name'] == 'kafka':
+                        executor_mesos_id = task['slave_id']
+            if framework['name'] == 'kafka' and len(framework['tasks']) > 0 and executor_mesos_id:
+                task_mesos_id = framework['tasks'][0]['slave_id']
+                wait = False
+                break
+
+    for agent in state['slaves']:
+        if agent['id'] == executor_mesos_id:
+            executor_node = agent['hostname']
+        if agent['id'] == task_mesos_id:
+            task_node = agent['hostname']
+
+    def executor_checks(response):
+        # check executor metric metadata
+        for line in response.text.splitlines():
+            if '#' in line:
+                continue
+            if 'cpus_nr_periods' in line and 'marathon' in line:
+                assert 'service_name="marathon"' in line
+                assert 'task_name="kafka"' in line
+                return True
+        return False
+    check_metrics_prom(dcos_api_session, prometheus_port, executor_node, executor_checks)
+
+    def task_checks(response):
+        # check kafka task metric metadata
+        for line in response.text.splitlines():
+            if '#' in line:
+                continue
+            if 'cpus_nr_periods' in line and 'marathon' not in line:
+                assert 'service_name="kafka"' in line
+                assert 'task_name=""' in line
+                assert 'executor_name="kafka"' in line
+                return True
+        return False
+    check_metrics_prom(dcos_api_session, prometheus_port, task_node, task_checks)
+
+    # uninstall and cleanup framework
+    dcos_api_session.cosmos.uninstall_package('kafka', app_id=data['appId'])
+
+
 @pytest.mark.supportedwindows
 def test_metrics_node(dcos_api_session):
     """Test that the '/system/v1/metrics/v0/node' endpoint returns the expected
--- a/packages/dcos-integration-test/extra/test_metrics.py
+++ b/packages/dcos-integration-test/extra/test_metrics.py
@@ -3,6 +3,7 @@ import contextlib
 import common
 import pytest
 import retrying
+from test_helpers import expanded_config
 
 
 __maintainer__ = 'mnaboka'
@@ -207,6 +208,8 @@ def get_task_hostname(dcos_api_session, framework_name, task_name):
     return node
 
 
+@pytest.mark.skipif(expanded_config.get('security') == 'strict',
+                    reason="MoM disabled for strict mode")
 def test_task_metrics_metadata(dcos_api_session):
     """Test that task metrics have expected metadata/labels"""
     with deploy_and_cleanup_dcos_package(dcos_api_session, 'marathon', '1.6.535', 'marathon-user'):
@@ -223,6 +226,8 @@ def test_task_metrics_metadata(dcos_api_session):
         check_metrics_metadata()
 
 
+@pytest.mark.skipif(expanded_config.get('security') == 'strict',
+                    reason="Framework disabled for strict mode")
 def test_executor_metrics_metadata(dcos_api_session):
     """Test that executor metrics have expected metadata/labels"""
     with deploy_and_cleanup_dcos_package(dcos_api_session, 'hello-world', '2.2.0-0.42.2', 'hello-world'):
--- a/packages/dcos-integration-test/extra/test_metrics.py
+++ b/packages/dcos-integration-test/extra/test_metrics.py
@@ -46,36 +46,49 @@ def test_metrics_masters_prom(dcos_api_session):
         assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
 
 
-@retrying.retry(wait_fixed=2000, stop_max_delay=150 * 1000)
-def get_metrics_prom(dcos_api_session, node, expected_metrics):
-    """Assert that expected metrics are present on prometheus port on node.
+@retrying.retry(wait_fixed=2000, stop_max_delay=300 * 1000)
+def get_metrics_prom(dcos_api_session, node):
+    """Gets metrics from prometheus port on node and returns the response.
 
-    Retries on non-200 status or missing expected metrics
-    for up to 150 seconds.
+    Retries on non-200 status for up to 300 seconds.
 
     """
     response = dcos_api_session.session.request('GET', 'http://{}:61091/metrics'.format(node))
     assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
-    for metric_name in expected_metrics:
-        assert metric_name in response.text
+    return response
 
 
 def test_metrics_agents_mesos(dcos_api_session):
     """Assert that mesos metrics on agents are present."""
     for agent in dcos_api_session.slaves:
-        get_metrics_prom(dcos_api_session, agent, ['mesos_slave_uptime_secs'])
+        @retrying.retry(wait_fixed=2000, stop_max_delay=300 * 1000)
+        def check_mesos_metrics():
+            response = get_metrics_prom(dcos_api_session, agent)
+            assert 'mesos_slave_uptime_secs' in response.text
+        check_mesos_metrics()
 
 
 def test_metrics_masters_mesos(dcos_api_session):
     """Assert that mesos metrics on masters are present."""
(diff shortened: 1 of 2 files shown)


gracedo:

gracedo:
@mesosphere-mergebot bump-ee

gracedo:
@philipnrmn @branden @janisz this is ready for review! :)

gracedo:
teamcity/dcos/test/dcos-docker/stati, teamcity/dcos/test/aws/onprem/static and teamcity/dcos/test/aws/cloudformation/simple failing on https://jira.mesosphere.com/browse/DCOS_OSS-4342

gracedo:
@janisz I added a line to the changelog in 1.12 denoting the bug fix - I don't think I need to add it to 1.13 as well?

branden:
Yeah, users won't encounter this bug in a 1.13 release, so I don't think we need to mention the fix in the 1.13 changelog.

branden:
This Telegraf bump appears to break something. The integration test jobs are all failing due to errors in existing metrics tests.

gracedo:
I'm guessing installing kafka in the new integration test is making something unhappy.. taking a closer look

gracedo:
@mesosphere-mergebot bump-ee

gracedo:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+    asserting that it returns True.
+
+    Retries on non-200 status or failed assertions
+    for up to 300 seconds.
+
+    """
+    response = dcos_api_session.session.request(
+        'GET', 'http://{}:61091/metrics'.format(node))
+    assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
+    assert check_func(response) is True
```
branden: The only difference between `check_metrics_prom()` and `get_metrics_prom()` is the assertion after retrieving metrics from the Prometheus endpoint -- the request and timeout logic is duplicated. Let's refactor `get_metrics_prom()` so it only gets and returns the metrics response (while still asserting it's a 200), reuse that where `check_metrics_prom()` is used, and leave it up to test functions to assert whatever they need to in the response that gets returned.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+
+    def check_task_metrics(response):
+        for line in response.text.splitlines():
+            if '#' in line:
+                continue
+            if 'marathon-user' in line:
+                return ('service_name="marathon"' in line and
+                        'task_name="marathon-user"' in line)
+        return False
+    check_metrics_prom(dcos_api_session, node, check_task_metrics)
```
branden: Instead of passing in a callable for `check_metrics_prom()` to execute in order to make assertions about the response, I think it'd be more straightforward to just get the response and then make assertions about it.

```
response = get_metrics_prom(dcos_api_ession, node)
for line in response.text.splitlines():
    # etc.
```

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+
+    # uninstall and cleanup framework
+    dcos_api_session.cosmos.uninstall_package('marathon', app_id=data['appId'])
+
+
+def test_executor_metrics_metadata(dcos_api_session):
```
branden: Docstring needed here as well.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+    # get count of all running containers to check against later for complete teardown
+    num_containers = 0
+    for agent_ip in dcos_api_session.slaves:
+        response = dcos_api_session.metrics.get('/containers', node=agent_ip)
+        assert response.status_code == 200
+        num_containers += len(response.json())
```
branden: This assumes that all currently running containers are intended to still be running when this test completes, and that no containers are intended to start during the test. I think those assumptions may make for a brittle test. If this test is run concurrently with other tests, or if another test fails to clean up its containers before executing, this could fail. We should look for a more direct way of asserting that the framework's containers are finished executing during teardown.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+            assert state_response.status_code == 200
+            state = state_response.json()
+            for framework in state['frameworks']:
+                if framework['name'] == 'marathon':
+                    for task in framework['tasks']:
+                        if task['name'] == 'hello-world':
+                            mesos_id = task['slave_id']
+                            wait = False
+                            break
+                    break
```
branden: This stanza that blocks until the package is deployed could be factored out into a separate function. It's separable from the test and is duplicated in `test_task_metrics_metadata()`.

This would work nicely as a context manager. It could install a package and block until it's deployed on entry, then uninstall it and block until all its containers are finished on exit. Quick sketch:

```python
import contextlib


@contextlib.contextmanager
def dcos_package(dcos_api_session, package_name, package_version)
    app_id = dcos_api_session.cosmos.install_package(package_name, package_version=package_version).json()['appId']
    dcos_api_session.marathon.wait_for_deployments_complete()
    wait_for_package_deployment(dcos_api_session, package_name)
    yield
    dcos_api_session.cosmos.uninstall_package(package_name, app_id=app_id)
    wait_for_package_teardown(dcos_api_session, package_name, ...)  # Not sure what else this might need


def test_executor_metrics_metadata(dcos_api_session):
    with dcos_package(dcos_api_session, 'hello-world', '2.2.0-0.42.2'):
        executor_node = get_task_hostname(framework_name='marathon', task_name='hello-world')
        metrics_response = get_metrics_prom(dcos_api_session, executor_node)
        for line in metrics_response.text.splitlines():
            # assert stuff
    # done!
```

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+                    continue
+                # ignore metrics from hello-world task started by marathon by checking
+                # for absence of 'marathon' string.
+                if 'cpus_nr_periods' in line and 'marathon' not in line:
+                    return ('service_name="hello-world"' in line and
+                            'task_name=""' in line and  # this is an executor, not a task
+                            # hello-world executors can be named "hello" or "world"
+                            ('executor_name="hello"' in line or 'executor_name="world"' in line))
+            return False
+        check_metrics_prom(dcos_api_session, executor_node, check_executor_metrics)
```
branden: See above comment about getting the metrics response and asserting it's correct inline with the rest of the test.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+    for up to 300 seconds.
+
+    """
+    response = dcos_api_session.session.request(
+        'GET', 'http://{}:61091/metrics'.format(node))
+    assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
+    assert check_func(response) is True
+
+
+def test_task_metrics_metadata(dcos_api_session):
```
branden: This could use a docstring explaining what the test asserts.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+        mesos_id = ''
+        while wait:
+            master = dcos_api_session.masters[0]
+            state_response = dcos_api_session.get('/state', host=master, port=5050)
+            assert state_response.status_code == 200
+            state = state_response.json()
+            for framework in state['frameworks']:
+                if framework['name'] == 'marathon':
+                    for task in framework['tasks']:
+                        if task['name'] == 'hello-world':
```
branden: Wouldn't this get us the task that's running the hello-world framework, rather than the hello-world framework's own tasks? The tasks launched by the framework may not be scheduled on the same node as where the framework is running. Is the intention to look for the Marathon task running the framework, rather than one of the framework's own tasks? If that's the case, why do we need to install the hello-world package, instead of e.g. running a single-task Marathon app that only sleeps?

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+                            mesos_id = task['slave_id']
+                            wait = False
+                            break
+                    break
+
+        for agent in state['slaves']:
+            if agent['id'] == mesos_id:
+                executor_node = agent['hostname']
```
branden: This bit is also duplicated in the other test. Sounds like we need a helper function for getting the hostname for an agent on one of a framework's tasks is running. `def get_task_hostname(framework_name, task_name)`?

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -134,6 +134,127 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+            for line in response.text.splitlines():
+                if '#' in line:
+                    continue
+                # ignore metrics from hello-world task started by marathon by checking
+                # for absence of 'marathon' string.
+                if 'cpus_nr_periods' in line and 'marathon' not in line:
+                    return ('service_name="hello-world"' in line and
+                            'task_name=""' in line and  # this is an executor, not a task
+                            # hello-world executors can be named "hello" or "world"
+                            ('executor_name="hello"' in line or 'executor_name="world"' in line))
```
branden: We should really be parsing these Prometheus responses into a data structure, especially now that we have multiple tests duplicating the logic of e.g. skipping comments. https://github.com/prometheus/client_python#parser

packages/dcos-integration-test/extra/test_metrics.py
gracedo: I pulled these checks out into a separate method primarily to use the retry decorator since it sometimes takes some time for the metrics/labels to show up

branden: Yeah the retry on getting the metrics and asserting their content is necessary. We can still get that while keeping `get_metrics_prom()` focused on retrying the metrics endpoint until it returns 200, without it taking on the responsibility of making assertions about its content.

```python
@retrying.retry(wait_fixed=2000, stop_max_delay=300 * 1000)
def get_metrics_prom(dcos_api_session, node)
    response = dcos_api_session.session.request(
        'GET', 'http://{}:61091/metrics'.format(node))
    assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
    return response


def test_task_metrics_metadata(dcos_api_session)
    # <snip>
    @retrying.retry(...)
    def check_task_metrics():
        response = get_metrics_prom(dcos_api_ession, node)
        for line in response.text.splitlines():
            # etc.
    check_task_metrics()
    # <snip>
```

gracedo: ahh I see, didn't think to add the decorator to the inner func, will fix 👍

packages/dcos-integration-test/extra/test_metrics.py
gracedo: I think this gets the same node that the hello-world executor is running on (I believe it is indicated by the node that the Marathon task running the framework is on)

packages/dcos-integration-test/extra/test_metrics.py
gracedo: I can't seem to find that jira ticket you created a while back, but maybe we can wait for that ticket to do the work to adopt the parser for all these tests all at once?

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -131,7 +141,109 @@ def test_metrics_agents_statsd(dcos_api_session):
...
+
+
+@retrying.retry(wait_fixed=2000, stop_max_delay=300 * 1000)
+def get_task_hostname(dcos_api_session, framework_name, task_name):
+    mesos_id = node = ''
+    state_response = dcos_api_session.get('/state', host=dcos_api_session.masters[0], port=5050)
+    assert state_response.status_code == 200
+    state = state_response.json()
+
+    for framework in state['frameworks']:
```
janisz: I think I've seen it before [here](https://github.com/dcos/dcos/pull/3764/files#diff-a99db445cccf205fe11239e4be6de2aaR157) can you extract it to function?

gracedo: sorry, all 4 of these dcos prs I have open are all basically adding the same code - let me try to consolidate them better and get as similar versions as I can in all of these

packages/dcos-integration-test/extra/test_metrics.py
gracedo: I think it works to pull `/state` and check the response for the status of frameworks/tasks

packages/dcos-integration-test/extra/test_metrics.py
branden: Yeah that SGTM.

packages/dcos-integration-test/extra/test_metrics.py
gracedo: @philipnrmn to fact-check me :)

gracedo:
@mesosphere-mergebot bump-ee

gracedo:
FYI: teamcity/dcos/build/dcos/windows failing due to https://jira.mesosphere.com/browse/DCOS-45547

gracedo:
@mesosphere-mergebot bump-ee

philipnrmn:

PR #3703, opened 2018-10-30
Bump dcos-net

urbanserj:
## High-level description

Bump dcos-net.

  - [DCOS_OSS-4395](https://jira.mesosphere.com/browse/DCOS_OSS-4395) dcos-net ignores some tcp/udp discovery ports for tasks on host network
  - [DCOS_OSS-1954](https://jira.mesosphere.com/browse/DCOS_OSS-1954) dcos-l4lb routes traffic until the first failed health check
  - [DCOS_OSS-4398](https://jira.mesosphere.com/browse/DCOS_OSS-4398) dcos-net continously restarting systemd-networkd on a bare-metal server with bond interfaces

## Related tickets (optional)

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: none

  - [x] Test Results: [link to CI job test results for component]
  - [x] Code Coverage (if available): [link to code coverage report]

https://github.com/dcos/dcos-net/compare/b02814510c1d4686b7055c4f70f44410bc3892c9...03f249b1c5f1d8a12b571a35ffa34efa07c52375
https://circleci.com/gh/dcos/dcos-net/749
https://codecov.io/github/dcos/dcos-net/commit/03f249b1c5f1d8a12b571a35ffa34efa07c52375

___

Reviewers should be:

### DIFF
--- a/packages/dcos-net/buildinfo.json
+++ b/packages/dcos-net/buildinfo.json
@@ -4,7 +4,7 @@
     "dcos-net": {
       "kind": "git",
       "git": "https://github.com/dcos/dcos-net.git",
-      "ref": "b02814510c1d4686b7055c4f70f44410bc3892c9",
+      "ref": "03f249b1c5f1d8a12b571a35ffa34efa07c52375",
       "ref_origin": "master"
     }
   },
--- a/packages/dcos-net/extra/dcos-net-setup.py
+++ b/packages/dcos-net/extra/dcos-net-setup.py
@@ -15,6 +15,9 @@ Also the script prevents from duplicating iptables rules [3]
 The script allows to add configuration for networkd
 """
 
+import datetime
+import filecmp
+import logging
 import os
 import shutil
 import subprocess
@@ -50,7 +53,7 @@ def main():
 def add_networkd_config(src):
     networkd = b'systemd-networkd.service'
 
-    # check if there is networkd
+    # Check if there is networkd
     result = subprocess.run(['systemctl', 'list-unit-files', networkd],
                             stdout=subprocess.PIPE)
     if result.returncode != 0:
@@ -58,10 +61,12 @@ def add_networkd_config(src):
     if networkd not in result.stdout:
         return result
 
-    # copy the configuration
+    # Copy the configuration
     bname = os.path.basename(src)
     dst = os.path.join('/etc/systemd/network', bname)
-    shutil.copyfile(src, dst)
+
+    if not safe_filecmp(src, dst):
+        shutil.copyfile(src, dst)
 
     # Restart networkd only if it's active
     result = subprocess.run(['systemctl', 'is-active', networkd],
@@ -70,7 +75,32 @@ def add_networkd_config(src):
         result.returncode = 0
         return result
 
+    # Restart networkd only if the configuration is updated
+    mtime = os.path.getmtime(dst)
+    result = subprocess.run(['systemctl', 'show', '--value',
+                             '--property', 'ActiveEnterTimestamp',
+                             networkd], stdout=subprocess.PIPE)
+    if result.returncode == 0:
+        active_enter_timestamp = result.stdout.strip().decode()
+        try:
+            started = datetime.datetime.strptime(
+                active_enter_timestamp,
+                '%a %Y-%m-%d %H:%M:%S %Z')
+            if started.timestamp() > mtime:
+                return result
+        except ValueError:
+            logging.warning('Unexpected ActiveEnterTimestamp value: "%s"',
+                            active_enter_timestamp)
+
+    # Restart networkd
     return subprocess.run(['systemctl', 'restart', networkd])
+
+
+def safe_filecmp(src, dst):
+    try:
+        return filecmp.cmp(src, dst)
+    except FileNotFoundError:
+        return False
 
 
 if __name__ == "__main__":


urbanserj:
@mesosphere-mergebot bump-ee

urbanserj:

packages/dcos-net/buildinfo.json
```diff
@@ -4,7 +4,7 @@
     "dcos-net": {
       "kind": "git",
       "git": "https://github.com/dcos/dcos-net.git",
-      "ref": "b02814510c1d4686b7055c4f70f44410bc3892c9",
+      "ref": "03f249b1c5f1d8a12b571a35ffa34efa07c52375",
```
Fabs: Have you considered tagging the releases so it makes easier to track down what version of dcos-net is shipped with every dcos bump?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -70,8 +74,28 @@ def add_networkd_config(src):
         result.returncode = 0
         return result
 
+    # Restart networkd only if the configuration is updated
+    mtime = os.path.getmtime(dst)
+    result = subprocess.run(['systemctl', 'show', '--value',
```
Fabs: In which cases you expect this command to have `returncode != 1`?

I ran this on Fedora 29, and all those commands result in statuscode 0:
```sh
> systemctl show --value --property ActiveEnterTimestamp docker
Mon 2018-11-05 04:35:54 CET
> echo $?
0
> systemctl show --value --property ActiveEnterTimestamp bananas

> echo $?
0
> systemctl show --value --property NO_PROPERTY docker
> echo $?
0
>
```

Does that means that you can end up passing `empty string` to strptime, and run into an exception?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,6 +15,8 @@
 The script allows to add configuration for networkd
 """
 
+import datetime
```
Fabs: Should we consider this code critical and cover it with some `integration-tests`? One idea would be to spin up containers with all the linux versions we support, and tests that this can drive a dummy systemd unit.

packages/dcos-net/buildinfo.json
vespian: @Fabs I am not sure if pkgpanda supports it. Besides that we have here the `master` branch of DC/OS reffering to the tip of the `master` branch of DC/OS net. The way it is handled e.g. for bouncer is that e.g. `1.12` branch of DC/OS reffers to the tip fo the `1.12` branch of Bouncer. I do not know if tags would be of any help in our release model.

packages/dcos-net/extra/dcos-net-setup.py
vespian: Valid point, but I do not think that this fits the scope of this PR :)

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -70,8 +74,28 @@ def add_networkd_config(src):
...
+            return result
+
+    # Restart networkd
     return subprocess.run(['systemctl', 'restart', networkd])
 
 
+def safe_filecmp(src, dst):
+    try:
+        return filecmp.cmp(src, dst)
+    except:
```
vespian: Can we please try to avoid catching all the exceptions and explicitly list the ones that we expect to ignore? This shold make the code more robust.

packages/dcos-net/extra/dcos-net-setup.py
vespian: +1

packages/dcos-net/extra/dcos-net-setup.py
urbanserj: This script already has 90%+ code coverage by integration tests, you run it every time when you deploy dc/os. We may consider adding some unit tests for failing path, but, as @vespian said, it's out of scope for this PR.

packages/dcos-net/extra/dcos-net-setup.py
urbanserj: Exception is exactly that behavior that I would expect in that case. It's the best that we can do here. Let me explain why.

`ActiveEnterTimestamp` is a well-known documented property, all active systemd services have it. Right before `systemctl show` the script checks that `systemd-networkd.service` is active.

Non-zero exit code of `systemctl show` is an expected case, i.e. `--value` argument is not supported in very old systemd versions, it can fail if there is an issue with dbus, etc. In that case we will just try to restart networkd.

If `systemd-networkd.service` is active and `systemctl show --value --property ActiveEnterTimestamp systemd-networkd.service` exists with successful exit code and returns empty string, there is nothing much we can do and we have to panic. It's not expected behavior. We can't ignore the result, we can catch the exception and print a human-readable error message, but I don't see what kind of problem we are solving by that.

packages/dcos-net/extra/dcos-net-setup.py
urbanserj: That's a good recommendation in general, in this case I'd really like to copy `src` file to `dst` if any error with `filecmp.cmp` happens. It could be posix error (`enoent`, `eaccess`, etc) or logical error in `filecmp` module, that doesn't matter. `filecmd` must always return boolean value with no exceptions. It catches all errors for one small function call, there is no other logic here. How would it make the code more robust if it catches all possible cases explicitly in this particular case?

packages/dcos-net/extra/dcos-net-setup.py
Fabs: Thanks for the clarification @urbanserj.

vespian: >  but I don't see what kind of problem we are solving by that.

This helps debugging.

Instead of rediscovering what you just described, a clear error message would allow the operator/support/us to quickly narrow down the problem. You also provided a really good description - can we please record it in the source code?  We can just copy-paste it as a comment.

These two things will definatelly help people working on this code base in the future.

packages/dcos-net/extra/dcos-net-setup.py
vespian: If you get the permission error, then it does not really make sense to try to copy the file I believe. What is more - the result of the comparison is invalid (you  were not able to compare these files in the end). 

I would rather catch the following errors:
* file does not exist

urbanserj:
@mesosphere-mergebot bump-ee

PR #2819, opened 2018-04-27
Check that ifconfig is available on cluster nodes

branden:
## High-level description

DC/OS network components depend on `ifconfig`, and will fail if it's not installed. This PR adds a check to the advanced install script that causes it to fail if `ifconfig` is not available on the host. It also adds a poststart node check that will fail if `ifconfig` is not available. This should shorten the debug loop by preventing DC/OS from deploying to hosts without `ifconfig`.

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS_OSS-2395](https://jira.mesosphere.com/browse/DCOS_OSS-2395) Fail install if ifconfig isn't available

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: Doesn't seem significant enough for a changelog entry.
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Tested manually on a CentOS host with ifconfig made unavailable. Install fails as expected.
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-checks/compare/c5a667cd20713af21c618df48b0ab2b3fc5ed0a6...c34c5a76bd49bf5df224e0c8abca51a71dece475)
  - [x] Test Results: https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-cluster-ops/job/dcos-checks/job/dcos-checks-pulls-pipeline/14/
  - [x] Code Coverage (if available): https://jenkins.mesosphere.com/service/jenkins/job/public-dcos-cluster-ops/job/dcos-checks/job/dcos-checks-pulls-pipeline/14/

PR #3734, opened 2018-11-07
Remove deprecated ifconfig dependency

lloesche:
## High-level description

ifconfig is deprecated. This PR removes it from DC/OS.

After talking with @urbanserj yesterday we no longer need the ifconfig dependency in DC/OS. It was introduced via https://github.com/dcos/dcos/pull/2819 and used by the network/cni isolator in Mesos. That's no longer the case so we can remove it.

I also found it used in the `azure.sh` ip-detect script so removed its use in there. Also updated the `vagrant.sh` ip-detect script (the two are now identical) to properly detect the source IP in the `ip route` output. Before it was hard coded to the 7th place which depending on the network configuration could produce wrong output.

E.g.
```
lukas@bigmac:~$ ip r g 8.8.8.8 | awk '{print $7}'
proto
```

  - [DCOS-44380](https://jira.mesosphere.com/browse/DCOS-44380) GCP Install of DC/OS 1.12.0 is missing prereq ifconfig for current Centos.


lloesche:
@mesosphere-mergebot bump-ee

lloesche:

MrMarvin:
@mesosphere-mergebot bump-ee

PR #3744, opened 2018-11-08
Collect ZooKeeper Metrics using DC/OS Telegraf

timaa2k:
## High-level description

This PR adds metrics collection from ZooKeeper instance on master nodes by the `dcos-telegraf` service. The collected logs correspond to the output from the `mntr` command:
https://zookeeper.apache.org/doc/r3.4.8/zookeeperAdmin.html#sc_zkCommands

  - [DCOS_OSS-4477](https://jira.mesosphere.com/browse/DCOS_OSS-4477) Instrument and Transmit ZooKeeper Metrics to dcos-telegraf service.

## Checklist for all PRs

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -1381,6 +1381,22 @@ package:
           "registrar",
           "allocator",
        ]
+      # Reads 'mntr' health checks from one or many zookeeper servers
+      [[inputs.zookeeper]]
+        ## An array of addresses to gather stats about. Specify an ip or hostname
+        ## with port. ie localhost:2181, 10.0.0.1:2181, etc.
+        ## If no hosts are specified, then localhost is used as the host.
+        ## If no port is specified, 2181 is used
+        servers = [":2181"]
+        ## Timeout for metric collections from all servers.  Minimum timeout is "1s".
+        timeout = "10s"
+        ## Optional TLS Config
+        # enable_tls = true
+        # tls_ca = "/etc/telegraf/ca.pem"
+        # tls_cert = "/etc/telegraf/cert.pem"
+        # tls_key = "/etc/telegraf/key.pem"
+        ## If true while TLS is enabled skip chain & host verification
+        # insecure_skip_verify = true
       # Expose metrics via the dcos-metrics v0 API.
       [[outputs.dcos_metrics]]
         dcos_node_role = "master"
--- a/packages/dcos-integration-test/extra/test_metrics.py
+++ b/packages/dcos-integration-test/extra/test_metrics.py
@@ -70,6 +70,12 @@ def test_metrics_masters_mesos(dcos_api_session):
         get_metrics_prom(dcos_api_session, master, ['mesos_master_uptime_secs'])
 
 
+def test_metrics_masters_zookeeper(dcos_api_session):
+    """Assert that ZooKeeper metrics on masters are present."""
+    for master in dcos_api_session.masters:
+        get_metrics_prom(dcos_api_session, master, ['zookeeper_'])
+
+
 def test_metrics_agents_statsd(dcos_api_session):
     """Assert that statsd metrics on agent are present."""
     if len(dcos_api_session.slaves) > 0:


timaa2k:

gen/dcos-config.yaml
```diff
@@ -1350,6 +1350,22 @@ package:
           "registrar",
           "allocator",
        ]
+      # Reads 'mntr' stats from one or many zookeeper servers
+      [[inputs.zookeeper]]
+        ## An array of address to gather stats about. Specify an ip or hostname
+        ## with port. ie localhost:2181, 10.0.0.1:2181, etc.
+        ## If no servers are specified, then localhost is used as the host.
```
jongiddy: Maybe use "If no host is specified," here, to match port line below and also to make it clear that the value used `":2181"` represents `"localhost:2181"`.

gen/dcos-config.yaml
```diff
@@ -1350,6 +1350,22 @@ package:
           "registrar",
           "allocator",
        ]
+      # Reads 'mntr' stats from one or many zookeeper servers
```
jongiddy: "Reads Zookeeper 'mntr' health checks from..." would help for those who don't know what `mntr` is without looking.

gen/dcos-config.yaml
```diff
@@ -1350,6 +1350,22 @@ package:
...
+      [[inputs.zookeeper]]
+        ## An array of address to gather stats about. Specify an ip or hostname
```
jongiddy: address -> addresses

gen/dcos-config.yaml
```diff
@@ -1350,6 +1350,22 @@ package:
...
+        servers = [":2181"]
+        ## Timeout for metric collections from all servers.  Minimum timeout is "1s".
+        timeout = "10s"
+        ## Optional TLS Config
+        # enable_ssl = true
+        # tls_ca = "/etc/telegraf/ca.pem"
+        # tls_cert = "/etc/telegraf/cert.pem"
+        # tls_key = "/etc/telegraf/key.pem"
+        ## If false, skip chain & host verification
+        # insecure_skip_verify = true
```
jongiddy: Is this correct? My intuitive interpretation is that setting this to `true` would cause verification to be skipped and `false` would cause verification to be done.  Either way, the default setting should be the one that cause verification to happen. All this, despite it being commented out! :smiling_imp:

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -336,6 +336,14 @@ def test_metrics_containers_nan(dcos_api_session):
         assert metric_value == '', 'unexpected metric value: {}'.format(metric_value)
 
 
+@retrying.retry(wait_fixed=(2 * 1000), stop_max_delay=(150 * 1000))
+def test_zookeeper_metrics_collected(dcos_api_session):
+    for master in dcos_api_session.masters:
+        response = dcos_api_session.session.request('GET', 'http://' + master + ':{}/metrics'.format(61091))
```
jongiddy: Maybe use
`dcos_api_session.session.get('http://' + master + ':{}/metrics'.format(61091))`
or even
`dcos_api_session.get('/metrics', scheme='http', host=master, port=61091)`?

timaa2k:
Some other owner than me please run:
`@mesosphere-mergebot override-status teamcity/dcos/test/aws/onprem/static https://jira.mesosphere.com/browse/DCOS_OSS-2115`

jongiddy:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/onprem/static https://jira.mesosphere.com/browse/DCOS_OSS-2115

jongiddy:
Failed test matches https://jira.mesosphere.com/browse/DCOS_OSS-3747 which is marked as a duplicate of https://jira.mesosphere.com/browse/DCOS_OSS-2115

gen/dcos-config.yaml
timaa2k: It took me some time to get what you meant here, but yes this should not be the default. I've made it explicit now to skip verification until we make ZooKeeper TLS-enabled :)

jongiddy: Sorry, I think it's the multiple negatives (false, skip) that make it confusing, along with the fact that this line is inoperative until the we setup TLS.
1. In the first line, where the comment says "If false, skip...verification", I would expect the value `true` to skip verification, so I think this comment may be incorrect.
2. In the second line, to ensure we default to secure, the value should be the value that does not skip verification. If the comment is indeed correct, then `insecure_skip_verify = true` is correct. If the comment is incorrect, then this line should be `insecure_skip_verify = false`.
3. There is no need to uncomment this line until the `tls_*` lines are also uncommented.

timaa2k: Thanks, I've looked up the code and updated the config description accordingly.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -336,6 +336,14 @@ def test_metrics_containers_nan(dcos_api_session):
         assert metric_value == '', 'unexpected metric value: {}'.format(metric_value)
 
 
+@retrying.retry(wait_fixed=(2 * 1000), stop_max_delay=(150 * 1000))
```
adamtheturtle: Why retry this test?

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -336,6 +336,14 @@ def test_metrics_containers_nan(dcos_api_session):
...
+def test_zookeeper_metrics_collected(dcos_api_session):
+    for master in dcos_api_session.masters:
+        response = dcos_api_session.get('/metrics', scheme='http', host=master, port=61091)
+        assert response.status_code == 200, 'Status code: {}'.format(response.status_code)
```
adamtheturtle: ❓ I think `pytest`'s assertion will be clear enough without the explanation

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -336,6 +336,14 @@ def test_metrics_containers_nan(dcos_api_session):
...
+        assert 'zookeeper_' in response.text
```
adamtheturtle: Is this enough to tell us what we want to know?

jgehrcke:
@mesosphere-mergebot override-status teamcity/dcos/test/dcos-docker/static DCOS_OSS-2115

jgehrcke:
In `teamcity/dcos/test/aws/cloudformation/simple` we see `test_metrics.test_metrics_containers` fail with


```
                assert 'dimensions' in container_metrics, 'got {}'.format(container_metrics)
>               assert 'task_name' in container_metrics['dimensions'], 'got {}'.format(
                    container_metrics['dimensions'])
E               AssertionError: got {'mesos_id': '', 'cluster_id': '80efb312-752d-4d44-979d-c4d7061c999f', 'container_id': 'b8d224bc-b202-42b9-840a-6cad6632b7ad', 'hostname': '10.0.1.122', 'labels': {'dcos_cluster_id': '80efb312-752d-4d44-979d-c4d7061c999f', 'dcos_cluster_name': 'teamcity-dcos-test-aws-cloudformation-simple-rpdbjtp72e', 'host': 'ip-10-0-1-122.us-west-2.compute.internal'}}
E               assert 'task_name' in {'cluster_id': '80efb312-752d-4d44-979d-c4d7061c999f', 'container_id': 'b8d224bc-b202-42b9-840a-6cad6632b7ad', 'hostna...': 'teamcity-dcos-test-aws-cloudformation-simple-rpdbjtp72e', 'host': 'ip-10-0-1-122.us-west-2.compute.internal'}, ...}

test_metrics.py:193: AssertionError
```

jgehrcke:
> In teamcity/dcos/test/aws/cloudformation/simple we see test_metrics.test_metrics_containers fail with

This looks like https://jira.mesosphere.com/browse/DCOS_OSS-4342 which seems to have been addressed in master yesterday, i.e. a rebase would probaby fix this. Overriding now.

jgehrcke:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/cloudformation/simple https://jira.mesosphere.com/browse/DCOS_OSS-4342

packages/dcos-integration-test/extra/test_metrics.py
jgehrcke: The output of `/metrics` must be massive, wow. We didn't specify any filter criterion and simply get everything? :) I should have a look at that API to better understand this.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -336,6 +336,14 @@ def test_metrics_containers_nan(dcos_api_session):
         assert metric_value == '', 'unexpected metric value: {}'.format(metric_value)
 
 
+@retrying.retry(wait_fixed=(2 * 1000), stop_max_delay=(150 * 1000))
+def test_zookeeper_metrics_collected(dcos_api_session):
+    for master in dcos_api_session.masters:
+        response = dcos_api_session.get('/metrics', scheme='http', host=master, port=61091)
```
jgehrcke: Why `port=61091`? Is this an internal interface? Does this assume that the test runner is executed on a master node? It's been a while since I worked with the api session interface, but `host=master` is confusing to me. `master` is not a valid DNS name.

jongiddy: Good point! We could use the `/system/v1/metrics` route on AR, which means we don't need the scheme or port.

However, `master` is a variable here. What makes you think it is a DNS name?

jongiddy:
OK, that doesn't work! Feel free to dismiss my review then.

packages/dcos-integration-test/extra/test_metrics.py
timaa2k: Thanks @adamtheturtle! I did not intend to retry the test. The `GET` request should be retried because it may take up to 20 seconds for the metrics to appear in Telegraf.

packages/dcos-integration-test/extra/test_metrics.py
timaa2k: It probably is, I just copied it from the other tests in the same file. Removing that.

packages/dcos-integration-test/extra/test_metrics.py
timaa2k: @adamtheturtle All ZooKeeper metrics start with this string. I could specify a particular one that we expect but I don't want the test to break if different metrics are outputted or the old ones are renamed.
@jgehrcke It's pretty neat how this works:
https://github.com/dcos/dcos/blob/53a31071349cd0cf6560ef48e5bbd36489e97022/gen/dcos-config.yaml#L1254
The reponse will no grow beyond a specified size, but that's why we need to retry the request like a Prometheus client to not miss the metrics being outputted.

packages/dcos-integration-test/extra/test_metrics.py
timaa2k: dcos-telegraf is outputting to `0.0.0.0:61091` which means the `/metrics` endpoint can be read from anywhere. I guess that's not ideal but currently I think we do not care about shielding metrics from the outside. `dcos_api_sessions.masters` returns the list of IP addresses specified the environment variable `MASTER_HOSTS` that we set before starting our CI. Since the test intends to assert that the metrics are reported on every master node we cannot utilize `/system/v1/metrics` through Admin Router. TBH I don't even know whether this is the same metrics stack that this endpoint exposes. I did not manage to use it in my test cluster via Admin Router. So I believe apart from the `retry` the test is exactly how it should be.

timaa2k:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_metrics.py
jgehrcke: > What makes you think it is a DNS name?

My brain made me think this, and clearly there was something wrong with my brain :)

jgehrcke: > dcos-telegraf is outputting to 0.0.0.0:61091 which means the /metrics endpoint can be read from anywhere. I guess that's not ideal but currently I think we do not care about shielding metrics from the outside.

Even more important than security is clear, rational thinking about what is a public interface and what not. We should always focus on testing a public interface first. It's what must "never" break. 

> TBH I don't even know whether this is the same metrics stack that this endpoint exposes.

That worries me a little bit. Can we invest a little more time, discussion, research in order to understand what's meant to be consumed by whom here? Ideally we know exactly what _external_ tooling expects, and test things from the point of view of this external tooling.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -496,3 +502,15 @@ def get_app_metrics(dcos_api_session, node: str, container_id: str):
     assert 'datapoints' in app_metrics, 'got {}'.format(app_metrics)
     assert 'dimensions' in app_metrics, 'got {}'.format(app_metrics)
     return app_metrics
+
+
+@retrying.retry(stop_max_delay=(30 * 1000))
+def assert_metric_reported(dcos_api_session, master: str, metric_substring: str):
```
gracedo: A [similar helper defined here](https://github.com/dcos/dcos/pull/3744/files#diff-a99db445cccf205fe11239e4be6de2aaR48) can be used instead

timaa2k: Too many trees; I could not see the forest :+1:

timaa2k:
@mesosphere-mergebot bump-ee

packages/dcos-integration-test/extra/test_metrics.py
timaa2k: Agreed, however I think that this is a larger topic that should be part of discussion with @leehambley

timaa2k:
Please run `@mesosphere-mergebot override-status teamcity/dcos/test/aws/onprem/static https://jira.mesosphere.com/browse/DCOS_OSS-4487`

jongiddy:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/onprem/static https://jira.mesosphere.com/browse/DCOS_OSS-4487

jongiddy:
@mesosphere-mergebot override-status teamcity/dcos/test/azure/arm https://jira.mesosphere.com/browse/DCOS-41675

packages/dcos-integration-test/extra/test_metrics.py
leehambley: > That worries me a little bit. Can we invest a little more time, discussion, research in order to understand what's meant to be consumed by whom here?

This is the Prometheus output which is enabled by default in Telegraf which exposes all metrics known to Telegraf in Promethus standard format.

- https://github.com/dcos/dcos/blob/279685203d9159526408f759cf6a5a700b25f7bc/gen/dcos-config.yaml#L1357-L1359

@jgehrcke if you have a suggestion to change the listener from `:61091` to something more concrete (`::1:61091` or whatever) we're all ears. (/cc @philipnrmn) 

This is the same endpoint you should be able to reach "from anywhere" and is our recommended public interface:

- https://docs.mesosphere.com/1.12/metrics/prometheus/

PR #3823, opened 2018-11-27
Seccomp

abudnik:
### High-level description

This PR introduces two new flags **mesos_seccomp_enabled** and **mesos_seccomp_profile_name**. The first flag is used to enable Seccomp isolator. The second flag specifies a name of a default Seccomp profile, e.g. `default.json`. Also, this PR includes a default Seccomp profile which is installed into `/etc/dcos/mesos/seccomp/default.json`. For more details, see [design doc](https://docs.google.com/document/d/1BxCa_MiUpItAzVS5Gy_BmxQrH2IvlsszYyfLvmk8XsU).

### Corresponding DC/OS tickets (obligatory)

* [DCOS_OSS-3830](https://jira.mesosphere.com/browse/DCOS_OSS-3830) Enable `linux/seccomp` isolator by default for mesos in gen/calc.py
* [DCOS_OSS-3829](https://jira.mesosphere.com/browse/DCOS_OSS-3829) Add `--enable-seccomp-isolator` flags to mesos build script
* [DCOS_OSS-3828](https://jira.mesosphere.com/browse/DCOS_OSS-3828) Add default seccomp config in DC/OS


abudnik:

orsenthil:
Hey @abudnik  / @Gilbert88  - What kind of integration test will be suitable for this in DC/OS?

I can imagine DC/OS being very high level abstraction for this feature, but still interested in knowing thoughts on having any integration test addition that will be suitable for this feature.

Tests that live at one of these places:

* https://github.com/dcos/dcos/blob/master/packages/dcos-integration-test/extra/test_mesos.py 
* https://github.com/dcos/dcos/tree/master/test-e2e

orsenthil:
There is one test that can exercise that the default profile is placed at the correct location.

Gilbert88:
If this feature is enabled on dc/os, any UCR based integration tests would exercise the code change.

@orsenthil are you saying that we should verify that the default.json is at the correct location?

orsenthil:
Hi @Gilbert88 , (cc: @abudnik )

Think about these scenarios. 

1. When `seccomp isolator` is enabled in a DC/OS cluster, and an job that is run on UCR tries to invoke a restricted system call, feature should give desired error response or failure vs the differing response when `isolator` is disabled in the DC/OS cluster.

2. Also, a test that makes sure that seccomp isolation feature is available only on DC/OS 1.13 and above and not in earlier versions of DC/OS.

I was suggesting that these kind of tests will be valuable addition along with this feature.

gen/dcos-config.yaml
```diff
@@ -581,6 +1365,12 @@ package:
       MESOS_NETWORK_CNI_PLUGINS_DIR=/opt/mesosphere/active/cni/:/opt/mesosphere/active/dcos-cni/:/opt/mesosphere/active/mesos/libexec/mesos
       MESOS_RECONFIGURATION_POLICY=additive
       MESOS_RECOVERY_TIMEOUT={{ mesos_recovery_timeout }}
+      MESOS_SECCOMP_CONFIG_DIR=/opt/mesosphere/etc/dcos/mesos/seccomp
```
qianzhangxa: I see you put the default seccomp profile `default.json` under `/etc/dcos/mesos/seccomp/`, should it be put under `/opt/mesosphere/etc/dcos/mesos/seccomp` instead? Or actually they are the same directory?

gen/calc.py
```diff
@@ -1105,6 +1108,8 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
         'mesos_container_log_sink': 'logrotate',
         'mesos_max_completed_tasks_per_framework': '',
         'mesos_recovery_timeout': '24hrs',
+        'mesos_seccomp_enabled': 'false',
```
qianzhangxa: So by default we will NOT enable this feature even a default seccomp profile is included in DC/OS?

gen/calc.py
```diff
@@ -1105,6 +1108,8 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
...
+        'mesos_seccomp_profile_name': '',
```
qianzhangxa: Since a default seccomp profile `default.json` is included, should we set the default value of this parameter to `default.json`?

qianzhangxa:
@orsenthil @Gilbert88 @abudnik 
I agree that we should add an integration test for this feature. However, to add an integration test for a feature, I think that feature should be enabled in DC/OS by default, right? In this PR, I see this seccomp feature is disabled by default.

gen/calc.py
abudnik: Yes, correct. By default, `linux/seccomp` isolator is disabled.

gen/dcos-config.yaml
abudnik: Yeah, they are the same.

gen/calc.py
abudnik: We could do that.
@Gilbert88 WDYT?

Gilbert88: No, we cannot do that. Do you forget the following case:

If a framework wants to enable seccomp but no default config. User could prefer a seccomp profile name if they need but by default all containers run without seccomp. How should the operator configure these two flags?

Gilbert88:
@qianzhangxa probably we could enable it by default in the future. But today only a small group of customers request seccomp and leveraging libseccomp may have some perf impact which we need more testing against. I would suggest do not turn it on for 1.13.0

orsenthil:
Hi @Gilbert88  - Yeah, I did notice that the feature is disabled by default, but it definitely gives an option to enable the feature, that's the reason I requested for a test.

You could make a note and JIRA, and please plan to bring in the test separately.

gen/calc.py
qianzhangxa: @Gilbert88 In that case, I think the operator can set `mesos_seccomp_enabled` to true and set `mesos_seccomp_profile_name` to empty string, and framework can specify the specific seccomp profile name in ContainerInfo.

Anyway, I am OK with the current changes in this PR (i.e., set `mesos_seccomp_enabled` to false and set `mesos_seccomp_profile_name` to empty string), but I think we need to document these two parameters somewhere, otherwise how can operator know there is a default seccomp profile included in DC/OS installer package?

qianzhangxa:
@Gilbert88 I agree to disable this feature for now. And another comment is, will we document the two new parameters (`mesos_seccomp_enabled` & `mesos_seccomp_profile_name`) and also the default seccomp profile included in DC/OS installer package?

gen/calc.py
Gilbert88: @qianzhangxa , both ways sgtm. It is just inclusive VS exclusive.

@abudnik Qian's suggestion makes more sense, but we might have to add more logic to allow DC/OS to unset mesos_seccomp_profile_name as empty string `""`, and not passing the empty string to Mesos. Could you make that change?

Gilbert88:
> @Gilbert88 I agree to disable this feature for now. And another comment is, will we document the two new parameters (`mesos_seccomp_enabled` & `mesos_seccomp_profile_name`) and also the default seccomp profile included in DC/OS installer package?

@qianzhangxa I think you were expecting all documentation would be included in this PR. Actually we have a JIRA to follow up docs https://jira.mesosphere.com/browse/DCOS_OSS-3870

gen/calc.py
abudnik: @qianzhangxa We'll document these flags and will give a short instruction on how to enable a default profile.
I'm leaning towards empty (disable) Seccomp profile by default - as it described in the design doc.

Gilbert88: If we step back, seems like in DC/OS, currently we don't support default config `unset` by passing empty string `""`.

Also, having the operator to explicitly set the profile name would allow the operator knows what exactly the default.json is.

Gilbert88: cc @qianzhangxa

qianzhangxa: @Gilbert88 Yeah, that makes sense to me.

Gilbert88:
To follow up with the integration test:
https://jira.mesosphere.com/browse/DCOS_OSS-4808

PR #3887, opened 2018-12-05
Remove deprecated ifconfig dependency (copy)

MrMarvin:
**This is a PR copy of https://github.com/dcos/dcos/pull/3734#issue-229039518, as I was unable to push my additional commits to @lloesche fork for the existing PR. If this one lands, we can safely close #3734.**

## High-level description
ifconfig is deprecated. This PR removes it from DC/OS.

After talking with @urbanserj yesterday we no longer need the ifconfig dependency in DC/OS. It was introduced via #2819 and used by the network/cni isolator in Mesos. That's no longer the case so we can remove it.

I also found it used in the azure.sh ip-detect script so removed its use in there. Also updated the vagrant.sh ip-detect script (the two are now identical) to properly detect the source IP in the ip route output. Before it was hard coded to the 7th place which depending on the network configuration could produce wrong output.

- [DCOS-44380](https://jira.mesosphere.com/browse/DCOS-44380) GCP Install of DC/OS 1.12.0 is missing prereq ifconfig for current Centos.


MrMarvin:
@mesosphere-mergebot bump-ee

MrMarvin:

MrMarvin:
@mesosphere-mergebot bump-ee

PR #3899, opened 2018-12-07
Change test_checks to improve its output and reduce the number of irrelevant failures

br-lewis:
## High-level description
This sort of addresses test flakiness for `test_checks.test_checks_api`. It now no longer checks the result code from the checks API because `dcos_api_session` will essentially wait for and require a healthy cluster in order to continue and run this test. Since we're assuming a healthy cluster I don't think it makes sense to check the status because we thus always expect it to be 0 and if something unexpected goes wrong, it's very likely to be some transient failure in a different component as seen by the issue in the associated ticket.

  - [DCOS-42722](https://jira.mesosphere.com/browse/DCOS-42722) test_checks.test_checks_api is flaky

## Checklist for all PRs

  - [X] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: This is a change to a test
  - [X] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a change to a test

___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/test_checks.py
+++ b/packages/dcos-integration-test/extra/test_checks.py
@@ -96,7 +96,10 @@ def test_checks_api(dcos_api_session):
             results = r.json()
             assert isinstance(results, dict)
 
-            logging.info("returned checks: {}".format(checks))
-            logging.info("run checks     : {}".format(results['checks'].keys()))
-            # Make sure we ran all the listed checks.
-            assert set(checks.keys()) == set(results['checks'].keys())
+            # check that the returned statuses of each check is 0
+            expected_status = {c: 0 for c in checks.keys()}
+            response_status = {c: v['status'] for c, v in results['checks'].items()}
+            assert expected_status == response_status
+
+            # check that overall status is also 0
+            assert results['status'] == 0
--- a/packages/dcos-integration-test/extra/test_checks.py
+++ b/packages/dcos-integration-test/extra/test_checks.py
@@ -63,7 +63,14 @@ def test_checks_cli(dcos_api_session):
 
 
 def test_checks_api(dcos_api_session):
-    """Test the checks API at /system/checks/"""
+    """
+    Test the checks API at /system/checks/
+    This will only test the checks API can be successfully reached on a real
+    cluster and that the checks it returns when running GET will match those it
+    reports when running the checks with POST. It does not check the results
+    status as `dcos_api_session` will wait while being built for the cluster to
+    be running so we won't expect any checks here to fail anyway.
+    """
     checks_uri = '/system/checks/v1/'
     # Test that we can list and run node and cluster checks on a master, agent, and public agent.
     check_nodes = []
@@ -88,7 +95,8 @@ def test_checks_api(dcos_api_session):
             assert r.status_code == 200
             results = r.json()
             assert isinstance(results, dict)
-            assert results['status'] == 0
 
+            logging.info("returned checks: {}".format(checks))
+            logging.info("run checks     : {}".format(results['checks'].keys()))
             # Make sure we ran all the listed checks.
             assert set(checks.keys()) == set(results['checks'].keys())
--- a/packages/dcos-integration-test/extra/test_checks.py
+++ b/packages/dcos-integration-test/extra/test_checks.py
@@ -65,11 +65,9 @@ def test_checks_cli(dcos_api_session):
 def test_checks_api(dcos_api_session):
     """
     Test the checks API at /system/checks/
-    This will only test the checks API can be successfully reached on a real
-    cluster and that the checks it returns when running GET will match those it
-    reports when running the checks with POST. It does not check the results
-    status as `dcos_api_session` will wait while being built for the cluster to
-    be running so we won't expect any checks here to fail anyway.
+    This will test that all checks run on all agents return a normal status. A
+    failure in this test may be an indicator that some unrelated component
+    failed and dcos-checks functioned properly.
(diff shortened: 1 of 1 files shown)


br-lewis:
@orsenthil @branden can you review?

branden:
I'm +1 better error messages, but I don't think we should remove the status check from the test.

> `dcos_api_session` will essentially wait for and require a healthy cluster in order to continue and run this test.

It doesn't run checks, however. Maybe it should? https://github.com/dcos/dcos-test-utils/blob/2cca7625217952a6d7ee78b13f5f8d6a03f81a09/dcos_test_utils/dcos_api.py#L428

> Since we're assuming a healthy cluster

I don't believe any of our tests should assume a healthy cluster!

>  I don't think it makes sense to check the status because we thus always expect it to be 0 and if something unexpected goes wrong, it's very likely to be some transient failure in a different component as seen by the issue in the associated ticket.

That failure really ought to be captured, and it ought to result in a failing test run. We can do a better job of making clear where that failure lies, but we shouldn't suppress it entirely.

br-lewis:
Here's what this test actually checks (prior to this PR):
for each node in the cluster this tests:
    - that the dcos-checks endpoint serves GET requests
    - that the response to the GET is a dictionary
    - that the dcos-checks endpoint serves POST requests
    - that the response to the POST is a dictionary
    - that the `status` key in the response is 0 which indicates that all checks that are run on POST returned a normal status
    - that the response to the GET has the same keys as the `checks` dictionary in the response to the POST which indicates that the dcos-checks API ran all the checks it said it would.

So what is this really supposed to test?

Is it testing if the dcos-checks API is working properly? If so, then a failing component on the cluster should result in a non-zero status which would cause a failing test but that's an indication that dcos-checks is working properly so should this specific test fail? I'd think it shouldn't because it says that dcos-checks is doing its job. But what we're actually testing currently doesn't cover this because it assumes that the cluster is healthy enough to return a 0 status. Changing it to adequately test this would require us to verify everything dcos-checks checks out-of-band to ensure that we assert against the expected status.

Is it testing that the cluster is working well enough to fulfill whatever checks are being run? If so, 1) we're assuming that dcos-checks works consistently enough for this purpose and 2) that would mean this test should never be considered flaky because if it fails, it could very well mean something unexpected happened in the cluster which resulted in a failure and that this "test" didn't really flake but failed properly in pointing this out. In this case, the only change necessary would be improved reporting for what specifically caused the failure but that doesn't solve the problem that people are overriding this because it's failing when supposedly everything is fine or at least everything relevant to the change being made.

If we look at the specific issue that was cited in the JIRA ticket (Metronome would OOM which naturally caused dcos-checks to return a non-zero status), we can see exactly how this causes problems. Yes, dcos-checks did catch that Metronome wasn't working but it caused a failure in the `dcos-checks` tests. Should `dcos-checks`'s tests fail because some other component wasn't working? Is that helpful information when Metronome's tests also failed? 

I took the stance that all this test is really trying to do is make sure that dcos-checks is running, reachable, and returns something resembling what it's supposed to return. In that sense, removing an assert on `status` being a specific value makes sense because we clearly aren't really interested in the value it returns as much as we are in it returning a value.

Thinking about it a bit more, I could certainly make some changes to better test that dcos-checks returns what we expect like making sure that `status` is some int between 0 and 3 and that the checks also report valid statuses but I really don't think it makes sense to assert against any specific status value or to have this test fail because a component in the cluster failed and dcos-checks was able to pick up on that properly.

br-lewis:
After talking with Lee, we decided to continue asserting on status but instead assert on the individual check statuses first and after those all show to be 0, we assert on the overall status. This makes it so the error reported is more descriptive of what went wrong.

```>               assert expected_status == response_status
E               AssertionError: assert {'clock_sync'...nfig': 1, ...} == {'clock_sync':...nfig': 0, ...}
E                 Differing items:
E                 {'mesos_master_replog_synchronized': 1} != {'mesos_master_replog_synchronized': 0}
E                 {'clock_sync': 1} != {'clock_sync': 0}
E                 {'xz': 1} != {'xz': 0}
E                 {'components_master': 1} != {'components_master': 0}
E                 {'tar': 1} != {'tar': 0}
E                 {'journald_dir_permissions': 1} != {'journald_dir_permissions': 0}...
E
E                 ...Full output truncated (6 lines hidden), use '-vv' to show```

File: packages/dcos-integration-test/extra/test_checks.py [hunk 1]
```diff
@@ -59,7 +59,14 @@ def test_checks_cli(dcos_api_session):
...
 
 def test_checks_api(dcos_api_session):
-    """Test the checks API at /system/checks/"""
+    """
+    Test the checks API at /system/checks/
+    This will only test the checks API can be successfully reached on a real
+    cluster and that the checks it returns when running GET will match those it
+    reports when running the checks with POST. It does not check the results
+    status as `dcos_api_session` will wait while being built for the cluster to
+    be running so we won't expect any checks here to fail anyway.
```
branden: This docstring is out of date after your latest update, which restores the status check.

br-lewis:

br-lewis:
@mesosphere-mergebot bump-ee

br-lewis:
@mesosphere-mergebot bump-ee

br-lewis:
@branden I think you need to reapprove. I think I invalidated your review when I rebased to update to master.

janisz:
@mesosphere-mergebot sync

janisz:
@mesosphere-mergebot sync

PR #3920, opened 2018-12-11
Set DC/OS CLI used in fetch_cluster_logs.bash to 1.11.

ArmandGrillet:
## High-level description

Set DC/OS CLI used in fetch_cluster_logs.bash to https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.11/dcos which is the version we expect to be used for 1.11 clusters.

  - [DCOS-46149](https://jira.mesosphere.com/browse/DCOS-46149) TeamCity log fetcher broken: started using new CLI version, breaks on setup.

## Related tickets (optional)

## Checklist for all PRs

___

Reviewers should be:

### DIFF
--- a/fetch_cluster_logs.bash
+++ b/fetch_cluster_logs.bash
@@ -108,7 +108,7 @@ fi
 
 # download the DC/OS CLI
 if [ ! -f dcos-cli ]; then
-  wget https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.12/dcos --output-document=dcos-cli
+  wget https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.11/dcos --output-document=dcos-cli
   chmod +x dcos-cli
 fi
 


ArmandGrillet:
@mesosphere-mergebot bump-ee

fetch_cluster_logs.bash
```diff
@@ -108,7 +108,7 @@ fi
 
 # download the DC/OS CLI
 if [ ! -f dcos-cli ]; then
-  wget https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.12/dcos --output-document=dcos-cli
+  wget https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.11/dcos --output-document=dcos-cli
```
jgehrcke: Thanks for the patch.

Why did this ever work?

jgehrcke: But should we not learn from that; and *not consider the DC/OS CLI retro-compatible*? Because here we learned that it is not?

jgehrcke:
Does this PR have any confirmation that log collection will now work upon failure?

cprovencher:
@ArmandGrillet @jgehrcke 
The problems I see with this change is:
1) https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.11/dcos is still a moving dependency
2) It doesn't solve log collection for 1.12 and master branches

This issue was not caused by using the same dcos-cli version for all dcos versions, it was caused by a moving dependency. Currently dcos-cli looks broken to me because I cannot call "dcos cluster setup" with a token. Please see https://jira.mesosphere.com/browse/DCOS-46149 for details. As soon as this is fixed, 0.7.5 could be released and used across all dcos versions.
I would pin to a stable version, but the earliest I can pin to is https://downloads.dcos.io/cli/releases/binaries/dcos/linux/x86-64/0.7.0/dcos, which still has the issue.

fetch_cluster_logs.bash
ArmandGrillet: The DC/OS CLI is not considered/promoted as retro-compatible, something that has been made clearer with our new warning when there is a mismatch between the CLI and the DC/OS version of the cluster. However, having a mismatch does not return an error thus inconsistencies such as the one fixed here exist in our codebase.

ArmandGrillet:
> https://downloads.dcos.io/binaries/cli/linux/x86-64/dcos-1.11/dcos is still a moving dependency
Is this a problem? The endpoints `https://downloads.dcos.io/binaries/cli/linux/x86-64/<dcos-version>/dcos` return the latest associated CLI (in this branch, 0.6.x) and we only backport bugfixes on such branches. I understand the need for fixed dependencies when using libraries developed by 3rd parties but not in this case.

> It doesn't solve log collection for 1.12 and master branches.
It doesn't, fixes for that are/will need to be done on https://github.com/dcos/dcos-cli and https://github.com/dcos/dcos-core-cli. But this PR and https://github.com/dcos/dcos/pull/3953 make the scripts more correct and allow us to update the CLI repos later on and see them impacting the correct branches of DC/OS and DC/OS EE.

rgo3:
@cprovencher Why would a different CLI version for the respective branch certainly cause problems in the future? The reasoning for @ArmandGrillet `s PRs was that we do not want to constantly update the CLI version here if we make new releases. Now for instance we already have a 0.7.6 release but 0.7.5 will be used everywhere. The CLI releases for 1.12, 1.11 and 1.10 clusters will only include bug fixes. IMO we should trust our own latest stable release. And even though 0.7.5 works everywhere, 0.7.x releases bundle the `dcos-core-cli` plugin specifically for 1.12 clusters. So for 1.12, 1.11 and 1.10, the respective CLI version should be used.
When it comes to the master branch I would vote for using the `testing` release because this is also the fastest way to catch errors, that previous testing didn't reveal. But in this case I would agree that using a moving target, which is `master`, should be considered carefully.

PR #3935, opened 2018-12-12
[1.13] Allow user-editable Telegraf configs

gracedo:
## High-level description

Previously, the only way for a user to supply additional telegraf configuration was to modify/add conf files to the telegraf config directory, which is within `/opt/mesosphere`. Now, users can add additional conf files to the `/var/lib/dcos/telegraf/telegraf.d/` directory, which gets symlinked into the Telegraf config directory.

Will add documentation on this in a docs pr.

  - [DCOS-42214](https://jira.mesosphere.com/browse/DCOS-42214) support methods to configure additional settings on dcos-telegraf

## Checklist for all PRs

  - [ ] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: changelog entry added in 1.12
  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Mainly a configuration change, not easily testable

### DIFF
--- a/packages/telegraf/extra/start_telegraf.sh
+++ b/packages/telegraf/extra/start_telegraf.sh
@@ -24,6 +24,14 @@ mkdir -p "${TELEGRAF_CONTAINERS_DIR}"
 # Migrate old containers dir to new location in case the cluster was upgraded.
 /opt/mesosphere/active/telegraf/tools/migrate_containers_dir.sh "${LEGACY_CONTAINERS_DIR}" "${TELEGRAF_CONTAINERS_DIR}"
 
+# Add symlinks to any additional configuration files in the user-editable Telegraf config directory
+shopt -s nullglob
+for file in ${TELEGRAF_EXTRA_CONFIG_DIR}*; do
+	ln -sfn $file ${TELEGRAF_CONFIG_DIR}$(basename "$file")
+done
+# Remove any broken symlinks
+find ${TELEGRAF_CONFIG_DIR} -type l ! -exec test -e {} \; -exec rm {} \;
+
 # Ensure that old socket file is removed, if present
 # TODO(philipnrmn): investigate whether moving to a systemd-managed socket
 # would be a better solution than manually creating and removing this file.
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -1275,6 +1275,7 @@ package:
     content: |
       TELEGRAF_CONFIG_FILE=/opt/mesosphere/etc/telegraf/telegraf.conf
       TELEGRAF_CONFIG_DIR=/opt/mesosphere/etc/telegraf/telegraf.d/
+      TELEGRAF_EXTRA_CONFIG_DIR=/var/lib/dcos/telegraf/telegraf.d/
       TELEGRAF_CONTAINERS_DIR=/run/dcos/telegraf/dcos_statsd/containers
       LEGACY_CONTAINERS_DIR=/run/dcos/mesos/isolators/com_mesosphere_MetricsIsolatorModule/containers
   - path: /etc/telegraf/telegraf.conf


gracedo:
@mesosphere-mergebot bump-ee

packages/telegraf/extra/start_telegraf.sh
```diff
@@ -24,6 +24,14 @@ mkdir -p "${TELEGRAF_CONTAINERS_DIR}"
 # Migrate old containers dir to new location in case the cluster was upgraded.
 /opt/mesosphere/active/telegraf/tools/migrate_containers_dir.sh "${LEGACY_CONTAINERS_DIR}" "${TELEGRAF_CONTAINERS_DIR}"
 
+# Add symlinks to any additional configuration files in the user-editable Telegraf config directory
+shopt -s nullglob
+for file in ${TELEGRAF_EXTRA_CONFIG_DIR}*; do
+	ln -sfn $file ${TELEGRAF_CONFIG_DIR}$(basename "$file")
```
branden: What happens if there's a collision with a user-created config file?

I wonder if it would be a good idea to add something to the filenames of these symlinks that suggests they come from builtin config:
```bash
  ln -sfn "$file" "${TELEGRAF_EXTRA_CONFIG_DIR}${TELEGRAF_BUILTIN_CONFIG_PREFIX}$(basename "$file")"
```
where `TELEGRAF_BUILTIN_CONFIG_PREFIX` is defined in `gen/dcos-config.yaml` to be `dcos.` or something.

This may help prevent collisions with user config, and we can filter by that prefix when cleaning up symlinks, to make sure we don't touch something we don't own.

```bash
find ${TELEGRAF_CONFIG_DIR} -type l -maxdepth 1 -name "${TELEGRAF_BUILTIN_CONFIG_PREFIX}*" ! -exec test -e {} \; -exec rm {} \;
```

Even with this change, collisions are possible, so we should know what that looks like and how a user could detect and work around it.

philipnrmn: If a user adds a configuration file and then removes it, what happens? Does it just leave behind a broken symlink, or does it cause problems?

gracedo: Broken symlinks are removed from the telegraf config dir [below](/var/lib/dcos/telegraf/telegraf.d)!

philipnrmn: how did I miss this 🤦‍♂️

packages/telegraf/extra/start_telegraf.sh
```diff
@@ -24,6 +24,14 @@ mkdir -p "${TELEGRAF_CONTAINERS_DIR}"
...
+done
```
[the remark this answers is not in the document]
gracedo: The intention was to create symlinks within `/opt/mesosphere/etc/telegraf/telegraf.d/` to the `/var/lib/dcos/telegraf/telegraf.d/` user-added config files, but I wasn't aware that we shouldn't be modifying `/opt/mesosphere`, even in a dcos startup script. Flipping all this logic and making the telegraf config dir point to the `/var/lib` path instead works, but still feels weird to me, and brings us back to [the original (similar) solution](https://github.com/dcos/dcos/pull/3572/files) which was to use the `/var/lib` dir and copy the `/opt/mesosphere` configs into it. Let me look more closely at modifying Telegraf itself to accept multiple dirs, that does seem like the cleanest solution.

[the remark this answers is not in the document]
branden: Yeah, nothing should be editing anything in `/opt/mesosphere/`, including DC/OS components aside from Pkgpanda. It's meant to be totally static apart from installs and upgrades.

FWIW we have at least one other package that copies config out of `/opt/mesosphere/` into a different location where it ultimately gets read from, so that users may edit it: https://github.com/dcos/dcos/blob/abaeb5cceedd5661b8d96ff47f8bb5ef212afbdc/packages/rexray/extra/dcos-rexray.service#L16

But I agree that it's not ideal.

PR #3941, opened 2018-12-13
cockroach: use exhibitor_wait as ExecStartPre

jgehrcke:
From the commit message:
```
This fixes DCOS_OSS-4601. Without this step
register.py is highly likely to find the
ZooKeeper cluster in a state where individual
ZooKeeper instances are still in standalone
mode. This on the other hand leads to individual
CockroachDB instances being launched without
being aware of each other.
```

https://jira.mesosphere.com/browse/DCOS_OSS-4601

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: bug was not yet released
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: there is no multi-master test in our CI

### DIFF
--- a/packages/bootstrap/extra/dcos_internal_utils/cli.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/cli.py
@@ -109,6 +109,7 @@ bootstrappers = {
     'dcos-mesos-slave': noop,
     'dcos-mesos-slave-public': noop,
     'dcos-cosmos': noop,
+    'dcos-cockroach': noop,
     'dcos-metronome': noop,
     'dcos-history': noop,
     'dcos-mesos-dns': noop,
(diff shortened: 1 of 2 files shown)


adamtheturtle:
> there is no multi-master test in our CI

We have a multi-master test in Enterprise at https://github.com/mesosphere/dcos-enterprise/blob/82a6c18c35327691223a8868297fdc8d71bb16cc/test-e2e/test_master_node_replacement.py#L43.

Let's bring this over to OSS, either with this change or immediately after to prevent regressions.

For sure at least please comment on this PR exactly how you know that this PR fixes the issue.

jgehrcke:
> Something seems strange with this PR - the onprem tests are all failing with semi-identical failures. I'll take a look in a bit.

The exhibitor_wait script does not exist anymore (since https://github.com/dcos/dcos/commit/6d61b889a2ec98ffd600edbf8101fc41cba7aa7d) :-) We actually have to use a noop bootstrapper. Pushed a corresponding commit.

> Why do we do this rather than adding a dcos_cockroach bootstrap? Using bootstrap would make the systemd file more consistent between Open and Enterprise.

Now we indeed use a noop bootstrapper for achieving the exhibitor_wait effect.

jgehrcke:
@mesosphere-mergebot bump-ee

gpaul:
Lovely!

jgehrcke:
> a quick manual test verified this fixes the issue.

Thank you @DominikDary !!

kensipe:
@jgehrcke should we consider back porting this?  or is this just new to 1.13?

PR #3952, opened 2018-12-14
Attempt to reduce flakes and gather data on bundle generation.

br-lewis:
## High-level description

This ups the timeout for bundle generation to 20 minutes instead of 2 to
ensure that the cluster has enough time to generate the bundle which
is the cause of the vast majority of the flakes for this test.

It also adds logging the time taken to generate the bundle to the output
to get a better idea of how much time it really takes to generate the
bundle.

  - [DCOS-41819]https://jira.mesosphere.com/browse/DCOS-41819) dcos-e2e/docker/static/strict tests (open_source_tests.test_units.test_socket_files_teardown) failed with "ResourceWarning: unclosed SSL socket"

## Checklist for all PRs

  - [X] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change:
  - [X] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a test change

___

Reviewers should be:

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -153,9 +153,9 @@ def _dump_diagnostics(request, dcos_api_session):
 
     make_diagnostics_report = os.environ.get('DIAGNOSTICS_DIRECTORY') is not None
     if make_diagnostics_report:
+        creation_start = datetime.datetime.now()
         last_datapoint = {
             'time': None,
-            'start': None,
             'value': 0
         }
 
@@ -177,7 +177,7 @@ def _dump_diagnostics(request, dcos_api_session):
         log.info('\nWait for diagnostics job to complete')
         diagnostics.wait_for_diagnostics_job(last_datapoint=last_datapoint)
 
-        duration = last_datapoint['time'] - last_datapoint['start']
+        duration = last_datapoint['time'] - creation_start
         log.info('\nDiagnostis bundle took {} to generate'.format(duration))
 
         log.info('\nWait for diagnostics report to become available')
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -155,6 +155,7 @@ def _dump_diagnostics(request, dcos_api_session):
     if make_diagnostics_report:
         last_datapoint = {
             'time': None,
+            'start': None,
             'value': 0
         }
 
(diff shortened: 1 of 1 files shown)


br-lewis:
@jgehrcke @janisz could you please review this?

br-lewis:

br-lewis:
@orsenthil @gpaul could you add a ship-it label or review please?

janisz:
@mesosphere-mergebot bump-ee

adamtheturtle:
In particular, what I'm asking for is that we merge https://github.com/dcos/dcos/pull/4148 and then this change is made in DC/OS Test Utils and then the version of DC/OS Test Utils is bumped.

br-lewis:
Sure that seems fine. I didn't feel it made much sense to keep non-test code in a file for test code anyway.

br-lewis:
Made a PR for the timeout bump and start time tracking in dcos-test-utils: https://github.com/dcos/dcos-test-utils/pull/76

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -162,12 +162,16 @@ def _dump_diagnostics(request, dcos_api_session):
 
         last_datapoint = {
             'time': None,
+            'start': None,
```
adamtheturtle: Mmm... this pattern of passing a `last_datapoint` dictionary is a mess but it is inherited and it makes sense.

It tells us "what was the time and value of the last datapoint observed" - at the start there is no time.

What does it mean to have `last_datapoint['start']`?

Can we instead have a variable here - `time_before_waiting = datetime.datetime.now()`, and not change `wait_for_diagnostics_job`?

adamtheturtle: Separately and not for now - should be be using monotonic time rather than `datetime.datetime.now()`?

br-lewis: Start just tracks the start time of the bundle gathering. I considered giving it an initial value of now but decided against it, I think because it would keep all datetime grabbing in one place, but I'm not strongly attached to that.

The idea was that we could then use it to see how long a bundle takes to generate since I don't think we have any idea how long it takes normally especially because this test has a 0% success rate for all the runs it's still tracking so we know that it's more than 2 minutes.

From a practical standpoint though, I'm not sure if anyone will really look at that number because they'll have no reason to unless the bundle fails to generate and if that happens nothing is logged and even if we did it would just log the roughly the length of the timeout.

I wouldn't be against making this PR only the timeout bump.

adamtheturtle: To be clear - I think that this variable should exist but not in the `last_datapoint` dictionary - it can be its own variable and it can be set in this function.

adamtheturtle:
@br-lewis - so will you create another PR to bump the DC/OS Test Utils version in DC/OS? I'm approving assuming that.

PR #3996, opened 2018-12-17
Fine-grained Admin Router metrics including Bouncer and MesosDNS upstreams

timaa2k:
## High-level description

This PR annotates the Admin Router config for requests going to the IAM and MesosDNS endpoints respectively. It allows for collecting statistics and outputting metrics about Admin Router upstreams as well as requests not going to any upstream. The latter ones are then classified as server level requests.
The metrics are gathered through the utilization of the nginx_vts_filter_by_set_key directive of the Nginx VTS module plugin. This directive is used in creative ways to transmit request properties like status, URI and client to the `/nginx/metrics` Nginx Prometheus endpoint.

This leads to metrics that look to something similar to this:

> nginx_vts_filter_requests_total{
>     filter = ",upstream=Bouncer,backend=,status=401,",
>     filter_name = "client=Mesos/1.8.0 authorizer (master)",
> } 4

The intended way to use these metrics is to parse them in the Prometheus config to make them usable in Grafana, like we do in the following PR that is part of this initiative:
https://github.com/mesosphere/dcos-monitoring/pull/99/files

This allows for fine-grained Admin Router Nginx metrics to be generated from the annotated Admin Router config. So far we only annotated the IAM and MesosDNS, but other upstreams can be annotated in exactly the same way.

Generating Admin Router metrics by the aforementioned approach with the VTS module comes with a number of trade-offs:
1. The size of the VTS module shared memory must be large enough to accommodate all possibly generated metrics for at least the time of the Telegraf scraping interval to not miss any data.
2. The size of the VTS module metrics output from the Admin Router Prometheus endpoint must be small enough to be downloaded within the scrape interval and timeout.
3. Without a dedicated Telegraf plugin for Admin Router the vast amount of useless metrics data must be able to be scraped by the `dcos-monitoring` Prometheus server within its scraping interval and timeout.

These changes were tested manually with the `dcos-monitoring` package with DC/OS OSS and Enterprise master branches and trade-off descisions were made based on extrapolation of MWT #11 data. These are mentioned in the code comments where appropriate.

  - [DCOS_OSS-4606](https://jira.mesosphere.com/browse/DCOS_OSS-4606) Tailor Nginx VTS metrics to yield fine-grained HTTP time-series.
  - [DCOS_OSS-4654](https://jira.mesosphere.com/browse/DCOS_OSS-4654) Limit number of time-series stored in Nginx VTS module shared memory.

## Related tickets (optional)

  - [DCOS_OSS-4596](https://jira.mesosphere.com/browse/DCOS_OSS-4596) Create Dashboards for Admin Router Nginx metrics.

## Checklist for all PRs

We don't want to test for specific metrics, a test for generic nginx metrics already exists.


timaa2k:
@mesosphere-mergebot bump-ee

timaa2k:
@mesosphere-mergebot bump-ee

timaa2k:
@mesosphere-mergebot bump-ee

timaa2k:
@mesosphere-mergebot bump-ee

timaa2k:
@mesosphere-mergebot bump-ee

timaa2k:
@mesosphere-mergebot bump-ee

packages/adminrouter/extra/src/includes/http/open/master.conf
```diff
@@ -1,3 +1,10 @@
+# Enable VTS module for metrics
+# Set the shared memory zone size to 256MB to be sure to
```
adamtheturtle: Let's be more precise with why 256 was chosen, and how someone wanting to change this number can know if their choices are sensible.

Does this mean that we have increased the memory requirement for DC/OS by 256 MB / master + 32 MB / agent?

packages/adminrouter/extra/src/includes/http/open/master.conf
```diff
@@ -1,3 +1,10 @@
...
+# contain all metrics in the VTS module configuration.
+# If the VTS module configuration enables dynamic metric
```
adamtheturtle: ❓ Maybe point to where this dynamic metric is set.

packages/adminrouter/extra/src/includes/metrics-http-level.conf
```diff
@@ -0,0 +1,10 @@
+# Output fine-grained http directive level metrics using the VTS module.
+# 
+# Caution:
+# A request which matches a location that has a VTS module filter defined will be
+# counted on the Nginx server and location directive level filter as well.
+#
+# This only exists in order for the Nginx VTS module to calculate the average latency
+# per second over all requests for us. This could potentially be removed once we
+# calculate the average latency ourselves in Grafana.
+vhost_traffic_status_filter_by_set_key http= ,all=,;
```
adamtheturtle: Please explain what it means to have `http=` and `all=` with no values.

packages/adminrouter/extra/src/includes/server/open/master.conf
```diff
@@ -109,6 +113,10 @@ location /acs/api/v1/auth/ {
 # Group: Metadata
 # Description: DC/OS GUI configuration (unauthenticated)
 location /dcos-metadata/ui-config.json {
+
+    set $upstream_tag Bouncer;
```
adamtheturtle: This means that the name "Bouncer" will be exposed to users who use Grafana and the monitoring package, right? Perhaps "IAM" or similar is more appropriate, as Bouncer, I think is meant to be an internal name (though it does appear in our docs).

packages/adminrouter/extra/src/includes/http/open/master.conf
timaa2k: Yes, that's what it means (I believe).

packages/adminrouter/extra/src/includes/server/open/master.conf
timaa2k: Renamed to IAM.

packages/adminrouter/extra/src/includes/metrics-location-level.conf
```diff
@@ -0,0 +1,40 @@
...
+# identify the upstream that the metrics belong to. This constitutes a mechanism to
+# annotate the Admin Router config, allowing one to collect metrics of dynamically
+# resolved upstreams on the fly. It is also the name that will show up in Grafana when
+# referring to a particular upstream. Grouping metrics based on upstream would normally
+# only be possible with Nginx Plus because only then they can be defined in a upstream
+# directive and continue to be dynamically resolved.
+# https://tenzer.dk/nginx-with-dynamic-upstreams/
+
+# Create a counter for every backend that requests go to when passing this line in a location directive.
+# Tag the counter with the manually set upstream.
```
adamtheturtle: ❓ Add: This means that before including this file, a line like `set $upstream_tag <CUSTOM_UPSTREAM_TAG>;`.

packages/adminrouter/extra/src/includes/metrics-location-level.conf
```diff
@@ -0,0 +1,40 @@
...
+# Tag the counter with the manually set upstream, the backend that the request was passed on to and the resulting status code.
+vhost_traffic_status_filter_by_set_key uri=$uri ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},uri;
+
+# Create a counter for every client that requests where sent from that pass this line in a location directive.
+# Tag the counter with the manually set upstream, the backend that the request was passed on to and the resulting status code.
+vhost_traffic_status_filter_by_set_key client=$http_user_agent ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},client;
+
+# Another thing to keep in mind is that if this file is imported before a LUA auth check
+# block and authorization for the request is denied, the reported $upstream_addr value
+# would be empty. We chose to import it afterwards so that an unauthorized request is
```
adamtheturtle: `s/would/will`.

timaa2k:

timaa2k:
@jongiddy @mhrabovcin @adamtheturtle @ahairape 
Please have a look on the changes, if you'd like anything changed.

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,16 @@
...
-# Set the shared memory zone size to 32MB to be sure to
-# contain all metrics in the VTS module configuration.
-# If the VTS module configuration enables dynamic metric
-# keys this needs to be increased to accommodate for that.
-vhost_traffic_status_zone shared:vhost_traffic_status:32m;
+# Enable VTS module for metrics.
+# The shared memory zone size must be sufficiently large to ensure that all
+# metrics that could potentially be generated by the VTS module's
+# vhost_traffic_status_filter_by_set_key directive can be stored. If the
+# shared memory zone limit is hit Admin Router will crash. The metrics data
```
jongiddy: Do we know what this crash would look like? Are we likely to get an error message, or will the logs just show Admin Router restarting? Is there anything we can do to make it obvious if this failure occurs? Can we log how close we are to the limit?

packages/adminrouter/extra/src/includes/metrics-http-level.conf
```diff
@@ -0,0 +1,15 @@
...
+#
+# This only exists in order for the Nginx VTS module to calculate the average latency
+# per second over all requests for us. This could potentially be removed once we
+# calculate the average latency ourselves in Grafana.
+# 
+# By supplying no value to http= and all= the keys will only be used by our custom
+# Prometheus config in the dcos-monitoring package to relabel the metric __name__
+# but will result in empty labels http and all. Empty labels are dropped by
+# Prometheus and thus they will not show up in Grafana.
+vhost_traffic_status_filter_by_set_key http= ,all=,;
```
jongiddy: Is it significant that there is a space after `http=` but not after `all=`? If there is, can we document the reason. If not, can we make them the same, so I don't ask this question another time?

packages/adminrouter/extra/src/includes/metrics-location-level.conf
```diff
@@ -0,0 +1,43 @@
...
+
+# Create a counter for every status of requests which pass this line in a location directive.
+# Tag the counter with the manually set upstream and the backend that the request was passed on to.
+vhost_traffic_status_filter_by_set_key status=$status ,upstream=${upstream_tag},backend=${upstream_addr},;
+
+# Create a counter for every URI that requests go to when passing this line in a location directive.
+# Tag the counter with the manually set upstream, the backend that the request was passed on to and the resulting status code.
+vhost_traffic_status_filter_by_set_key uri=$uri ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},uri;
+
+# Create a counter for every client that requests where sent from that pass this line in a location directive.
```
jongiddy: where -> were, or "for every client that sent requests"

Also, I'm not clear what "pass this line in a location directive" means, here and in previous stanza.

packages/adminrouter/extra/src/includes/metrics-location-level.conf
```diff
@@ -0,0 +1,43 @@
...
+vhost_traffic_status_filter_by_set_key client=$http_user_agent ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},client;
+
+# Another thing to keep in mind is that if this file is imported before a LUA auth check
+# block and authorization for the request is denied, the reported $upstream_addr value
+# would be empty. We chose to import it afterwards so that an unauthorized request is
+# counted at the server directive level filter. This choice implies losing the information
+# that the denied request was going to go to e.g. the IAM, which might be an unacceptable
+# trade-off in favor cleanliness and easy of building Grafana graphs. The nice property
+# of this solution is that only requests which really went to the upstream are counted
+# in the filters introduced through upstream metric annotation.
```
jongiddy: This whole paragraph is confusing because it is easy to lose track of which choice has which effect. Maybe:
```
If this file goes before any LUA auth check block, then:
1. if the request is denied, the reported $upstream_addr value will be empty;
2. we count requests that were not forwarded.

If this file goes after any LUA auth check block, then:
1.
2.

We choose the second option.
```

packages/adminrouter/extra/src/includes/metrics-http-level.conf
timaa2k: Yes, the reason is that `http=` is the value and `,all=,` is the key. They are separated in every filter directive by a space :+1:

packages/adminrouter/extra/src/includes/http/common.conf
timaa2k: We have a metric in place that is displayed in the Grafana dashboards for the used shared memory size. Ideally we set up an alarm for this later on. I have not actually tried to crash Admin Router, but due to how DC/OS handles Admin Router crashes I suspect that Admin Router is restarted by the systemd unit and the in-memory metric state is lost.
If you want I can try it out I suppose.

packages/adminrouter/extra/src/includes/metrics-location-level.conf
timaa2k: Nginx works by matching request properties first to `server` and then `location` directives with the corresponding server name and URI regex.
The term `passing this line` refers to having matched the regex of a particular a directive so that the filter code defined under it is executed. I imagined this as the code flow for handling the request literally passes by this line of the config.
Do you want me to note that in a comment?

packages/adminrouter/extra/src/includes/metrics-http-level.conf
jongiddy: OK, I did notice that it occurred elsewhere, but I guess it looks most confusing here. And the comma at the start of the key doesn't help. I think this is just one of those cases where I should lookup the directive to understand its format before deciding what it means. So, no change required.

packages/adminrouter/extra/src/includes/metrics-location-level.conf
jongiddy: Maybe just change `pass` to `match`?

packages/adminrouter/extra/src/includes/http/common.conf
timaa2k: Thanks for suggesting this @jongiddy. I ran a couple of tests and it appears that the module does not crash Admin Router. When reaching about 90% of the shared memory max size the VTS module stops to create new metrics nodes for new unique URIs, however the requests themselves are still counted. This means we definitely need to incorporate a LRU mechanism but it also means that we don't need to worry about the crash case. I think we should still set up an alert later at maybe around 80%.

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,17 @@
...
+# Enable VTS module for metrics.
+# The shared memory zone size must be sufficiently large to ensure that all
+# metrics that could potentially be generated by the VTS module's
+# vhost_traffic_status_filter_by_set_key directive can be stored. The metrics
+# data scales with the number of metric nodes. A metric node in our case is a
+# set of data about one detail, such as the number of requests made with a
+# particular client, or the number of responses with a particular status code,
+# or the number of URIs requested from Admin Router. In theory, the number of
+# URIs can be infinitely large. If the shared memory zone is near 86% of the
+# declared maxuimum size the VTS module will stop creating new metric nodes.
```
adamtheturtle: `s/maxuimum/maximum`

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,17 @@
...
-vhost_traffic_status_zone shared:vhost_traffic_status:32m;
+# Enable VTS module for metrics.
+# The shared memory zone size must be sufficiently large to ensure that all
+# metrics that could potentially be generated by the VTS module's
+# vhost_traffic_status_filter_by_set_key directive can be stored. The metrics
+# data scales with the number of metric nodes. A metric node in our case is a
+# set of data about one detail, such as the number of requests made with a
+# particular client, or the number of responses with a particular status code,
+# or the number of URIs requested from Admin Router. In theory, the number of
+# URIs can be infinitely large. If the shared memory zone is near 86% of the
```
adamtheturtle: Ideally let's get a reference to 86% - how this was determined and ideally a link to code in another tool which chooses it.

packages/adminrouter/extra/src/includes/metrics-http-level.conf
```diff
@@ -0,0 +1,15 @@
+# Output fine-grained http directive level metrics using the VTS module.
+# 
+# Caution:
+# A request which matches a location that has a VTS module filter defined will be
+# counted on the Nginx server and location directive level filter as well.
+#
+# This only exists in order for the Nginx VTS module to calculate the average latency
+# per second over all requests for us. This could potentially be removed once we
+# calculate the average latency ourselves in Grafana.
```
adamtheturtle: Is there a JIRA for this?

packages/adminrouter/extra/src/includes/metrics-http-level.conf
```diff
@@ -0,0 +1,15 @@
...
+# 
+# By supplying no value to http= and all= the keys will only be used by our custom
+# Prometheus config in the dcos-monitoring package to relabel the metric __name__
```
adamtheturtle: "our custom Prometheus config in the dcos-monitoring package" - is the relabelling always going to be there?

packages/adminrouter/extra/src/includes/metrics-location-level.conf
```diff
@@ -0,0 +1,50 @@
...
+
+# Create a counter for every URI that requests go to when matching the location directive that this filter appears in.
+# Tag the counter with the manually set upstream, the backend that the request was passed on to and the resulting status code.
+vhost_traffic_status_filter_by_set_key uri=$uri ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},uri;
+
+# Create a counter for every client that requests were sent from which match the location directive that this filter appears in.
+# Tag the counter with the manually set upstream, the backend that the request was passed on to and the resulting status code.
+vhost_traffic_status_filter_by_set_key client=$http_user_agent ,upstream=${upstream_tag},backend=${upstream_addr},status=${status},client;
+
+# Another thing to keep in mind is the consideration of whether to import this file before
```
adamtheturtle: Why is `LUA` capitalised?

packages/adminrouter/extra/src/includes/metrics-http-level.conf
adamtheturtle: https://jira.mesosphere.com/browse/DCOS_OSS-4665

packages/adminrouter/extra/src/includes/metrics-location-level.conf
timaa2k: Shouldn't be, fixed.

packages/adminrouter/extra/src/includes/metrics-http-level.conf
timaa2k: Nope, fixed.

packages/adminrouter/extra/src/includes/http/common.conf
adamtheturtle: If we can't, let's at least put a comment how we determined this.
Why only 86%!

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,17 @@
...
+# declared maximum size the VTS module will stop creating new metric nodes.
+# Therefore, we risk missing out on data if the URIs requested grows large.
+# See DCOS_OSS-4677 for limiting the number of nodes and employing a LRU
+# mechanism.
+vhost_traffic_status_zone shared:vhost_traffic_status:8m;
```
jongiddy: As discussed with @timaa2k, we should ensure that this is large enough to handle MWT sized clusters in a reasonable way. Tim suggests that 64MB is a reasonable value. If this causes problems with more constrained environments (e.g. minidcos on Docker/Vagrant) then we should look later at possibly adding a configuration variable for this, but also at other processes in DC/OS that use significantly more memory than Admin Router.

timaa2k: Further testing revealed that 16MB is more appropriate than 64MB because until there is a dedicated Telegraf plugin for Admin Router that drops useless metrics generated in the process.
Any Prometheus server that scrapes Telegraf must download these within its scrape interval and timeout. At 9.2MB Admin Router metrics nodes this amounts to 53MB of time-series data for Telegraf. That's close to a factor of 5 of useless metrics. Going from 16MB to 64MB would mean approximately 50MB metric node data, thus 250MB to download from Telegraf. That would require an optimistic network connection of like 5MB/s. Once the Telegraf plugin is in place we could still tune the shared memory size.

packages/adminrouter/extra/src/includes/http/common.conf
timaa2k: This is no longer the case, we employ an LRU mechanism now.

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,87 @@
...
-# If the VTS module configuration enables dynamic metric
-# keys this needs to be increased to accommodate for that.
-vhost_traffic_status_zone shared:vhost_traffic_status:32m;
+# Enable VTS module for metrics.
+#
+# The VTS module uses shared memory to store metrics data.
+# In particular, we store metrics data generated by the
+# ``vhost_traffic_status_filter_by_set_key`` directive.
+#
+# If this memory is used, Admin Router does not crash, but errors are shown
```
jongiddy: Given the more specific information two paragraphs down, this sentence could be removed.

packages/adminrouter/extra/src/includes/http/common.conf
```diff
@@ -1,9 +1,87 @@
...
+# See https://jira.mesosphere.com/browse/DCOS_OSS-4695 for an issue to create
+# a test which shows that Admin Router survives when the shared memory is full.
+#
+# Set shared memory size to 16MB. This accommodates ~4000 metric nodes. Reserve
+# 1300 metric nodes for Admin Router agent forwarding where each agent is
+# counted as distinct upstream in the order of at maximum 1000 agents.
+# Upstreams are seemingly not accounted for by the VTS module LRU mechanism so
+# we must always keep enough room for them.
+vhost_traffic_status_zone shared:vhost_traffic_status:16m;
+#
```
jongiddy: It would be nice to have a blank line after each directive (and possibly before) to highlight the actual directives amongst the copious comments.

packages/adminrouter/extra/src/includes/metrics-server-level.conf
```diff
@@ -0,0 +1,18 @@
+# Output fine-grained server directive level metrics using the VTS module.
+#
+# Caution:
+# A request which matches a location that has a VTS module filter defined will be
+# counted on the Nginx location directive level filter but it will not be counted
+# on a filter defined on the server directive level that it matches as well.
```
jongiddy: It would be nice to explain how this works in an overall comment somewhere - i.e. what is the relationship and priority for http/location/server filters. Not sure where is best though.

timaa2k: I'm writing up a google doc where I can add a section for that.

PR #4036, opened 2018-12-20
[master] Allow the collection of tests without a cluster

adamtheturtle:
## High-level description

This allows one to run `pytest --collect-only --confcutdir .` in the integration tests directory of a DC/OS OSS checkout without a cluster running, if the right requirements are installed.

This is part of https://jira.mesosphere.com/browse/DCOS-46438.

This will allow a dashboard which knows about test flakes.
This is also progress towards a test suite which can be run from any host against any cluster.

* Put "skip" conditions which require a cluster into functions, rather than as decorators
* Put import-level opening of files which exist on a cluster into functions

### DIFF
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -24,7 +24,7 @@ def deploy_test_app_and_check(dcos_api_session, app: dict, test_uuid: str):
     is the default, root
     """
     expanded_config = test_helpers.get_expanded_config()
-    default_os_user = 'nobody' if test_helpers.get('security') == 'strict' else 'root'
+    default_os_user = 'nobody' if expanded_config.get('security') == 'strict' else 'root'
 
     if 'container' in app and app['container']['type'] == 'DOCKER':
         marathon_user = 'root'
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -189,7 +189,7 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
 def test_octarine(dcos_api_session, timeout=30):
     expanded_config = test_helpers.get_expanded_config()
     if expanded_config.get('security') == 'strict':
-        pytest.skip(reason='See: https://jira.mesosphere.com/browse/DCOS-14760')
+        pytest.skip('See: https://jira.mesosphere.com/browse/DCOS-14760')
     # This app binds to port 80. This is only required by the http (not srv)
     # transparent mode test. In transparent mode, we use ".mydcos.directory"
     # to go to localhost, the port attached there is only used to
--- a/packages/dcos-integration-test/extra/test_applications.py
+++ b/packages/dcos-integration-test/extra/test_applications.py
@@ -23,7 +23,8 @@ def deploy_test_app_and_check(dcos_api_session, app: dict, test_uuid: str):
     from the app's Dockerfile, which, for the test application
     is the default, root
     """
-    default_os_user = 'nobody' if test_helpers.expanded_config.get('security') == 'strict' else 'root'
+    expanded_config = test_helpers.get_expanded_config()
+    default_os_user = 'nobody' if test_helpers.get('security') == 'strict' else 'root'
 
     if 'container' in app and app['container']['type'] == 'DOCKER':
         marathon_user = 'root'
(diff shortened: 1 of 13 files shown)


packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -185,17 +186,16 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
         pass
 
 
-@pytest.mark.skipif(
-    test_helpers.expanded_config.get('security') == 'strict',
-    reason='See: https://jira.mesosphere.com/browse/DCOS-14760')
 def test_octarine(dcos_api_session, timeout=30):
+    expanded_config = test_helpers.get_expanded_config()
+    if expanded_config.get('security') == 'strict':
```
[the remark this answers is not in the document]
adamtheturtle: Two reasons I'm happy with this:

* We do have a precedent in the code (e.g. in `test_marathon_authn_authz.py`)
* We have to mark this as skipped as run-time and not collect time, if we are going to collect tests without a cluster, as getting the configuration needs a cluster

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -508,9 +511,6 @@ def thread_request():
     assert set(expected_uuids) == set(received_uuids)
 
 
-@pytest.mark.skipif(not lb_enabled(), reason='Load Balancer disabled')
-@pytest.mark.xfail(test_helpers.expanded_config.get('security') == 'strict',
```
adamtheturtle: This is a change from `xfail` to `skip` because it is complex to achieve `xfail` within a function, and I see little value.

adamtheturtle:
@rukletsov The reason that tests could not be imported without a cluster was that in multiple places we had `open('/some/path/on/a/cluster)` on import or collect.

For those which were run on import, I put the `open` calls in a function.
For those which were run on `collect` (e.g. in `skipif` decorators), I put the `skip` checks inside the test functions so that they are not run on `collect` but only when the test is run.

packages/dcos-integration-test/extra/test_applications.py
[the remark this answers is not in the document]
janisz: We need to document it somewhere but I do not know where :(

packages/dcos-integration-test/extra/test_helpers.py
```diff
@@ -23,14 +23,16 @@ def get_exhibitor_admin_password():
...
 
-# make the expanded config available at import time to allow determining
-# which tests should run before the test suite kicks off
-with open('/opt/mesosphere/etc/expanded.config.json', 'r') as f:
-    expanded_config = json.load(f)
-    # expanded.config.json doesn't contain secret values, so we need to read the Exhibitor admin password from
-    # Exhibitor's config.
-    # TODO: Remove this hack. https://jira.mesosphere.com/browse/QUALITY-1611
-    expanded_config['exhibitor_admin_password'] = get_exhibitor_admin_password()
+def get_expanded_config():
```
janisz: Why we need this change?

janisz: Can you add a commit description for https://github.com/dcos/dcos/pull/4036/commits/9408bc2646dc04ffbfd69a15bdd3989d90eee92b

adamtheturtle: The file being opened exists only on a cluster. This change means that the file is opened only when the function is called, not on import.

adamtheturtle:

adamtheturtle:
I want to add a manual Ship It because there is a Mergebot issue preventing the EE PR from getting a Ship It.

packages/dcos-integration-test/extra/test_applications.py
```diff
@@ -189,7 +189,7 @@ def test_if_marathon_pods_can_be_deployed_with_mesos_containerizer(dcos_api_sess
 def test_octarine(dcos_api_session, timeout=30):
     expanded_config = test_helpers.get_expanded_config()
     if expanded_config.get('security') == 'strict':
-        pytest.skip(reason='See: https://jira.mesosphere.com/browse/DCOS-14760')
+        pytest.skip('See: https://jira.mesosphere.com/browse/DCOS-14760')
```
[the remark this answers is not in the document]
mhrabovcin: This is not a new test skip. The PR is only changing when the decision about skipping a test is evaluated from decorator to a test.

orsenthil:
I think, it might be a good idea to land this directly as it touches of number of test files and I noticed conflict when the mergebot was trying to create a train with this PR. ( The conflict rejection feature is not yet in place).

I am going ahead with direct merge after updating, so that the rest of the PRs can be go through. Thank you!

orsenthil:
@mesosphere-mergebot bump-ee

adamtheturtle:
Thank you @orsenthil for handling this!

orsenthil:
Np, @adamtheturtle, and this is a great change.

PR #4038, opened 2018-12-20
[1.12] Allow the collection of tests without a cluster

adamtheturtle:
## High-level description

This allows one to run `pytest --collect-only --confcutdir .` in the integration tests directory of a DC/OS OSS checkout without a cluster running, if the right requirements are installed.

This will allow a dashboard which knows about test flakes.
This is also progress towards a test suite which can be run from any host against any cluster.

* Put "skip" conditions which require a cluster into functions, rather than as decorators
* Put import-level opening of files which exist on a cluster into functions


mhrabovcin:

adamtheturtle:
I am manually applying Ship It because the EE PR is failing because overrides are not working.

PR #4074, opened 2019-01-02
Add 'show' action for pkgpanda.

lava:
## High-level description

This adds a 'show' action which can be used to view the
definition of a package, analogous to e.g. 'aptitude show'
or 'guix package --show'.

  - [DCOS_OSS-4623](https://jira.mesosphere.com/browse/DCOS_OSS-4623) Add a method to view a package description in pkgpanda.

## Checklist for all PRs

Only touching pkgpanda internals.
Only triggered manually by users, existing tests are sufficient to verify existing usage is not broken.

___

Reviewers should be:

### DIFF
--- a/pkgpanda/cli.py
+++ b/pkgpanda/cli.py
@@ -1,6 +1,7 @@
 """Panda package management
 
 Usage:
+  pkgpanda show <package-id>
   pkgpanda activate <id>... [options]
   pkgpanda swap <package-id> [options]
   pkgpanda active [options]
@@ -26,6 +27,7 @@ Options:
                                 rather than /etc/systemd/system/dcos.target.wants
 """
 
+import json
 import os
 import sys
 from itertools import groupby
@@ -165,6 +167,18 @@ def main():
             for pkg in sorted(install.get_active()):
                 print(pkg)
             sys.exit(0)
+
+        if arguments['show']:
+            package_id = arguments['<package-id>']
+            try:
+                pkginfo = repository.load(package_id).pkginfo
+                print(json.dumps(pkginfo, indent=4))
+                sys.exit(0)
+            except PackageNotFound:
+                print(
+                    "Package not found: {}".format(package_id),
+                    file=sys.stderr)
+                sys.exit(1)
 
         if arguments['add']:
             actions.add_package_file(repository, arguments['<package-tarball>'])
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -244,6 +244,10 @@ class Package:
         return self.__path
 
     @property
+    def pkginfo(self):
+        return self.__pkginfo
+
+    @property
     def variant(self):
         return self.__pkginfo.get('variant', None)
 
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -149,7 +149,7 @@ class Systemd:
 
         for unit_name in self.unit_names(self.__unit_directory):
             systemd_file_path = os.path.join(self.__base_systemd, unit_name)
-            os.rename(systemd_file_path + self.new_unit_suffix, systemd_file_path)
+            shutil.move(systemd_file_path + self.new_unit_suffix, systemd_file_path)
 
     @staticmethod
     def unit_names(unit_dir):
@@ -492,7 +492,7 @@ class Repository:
         remove_directory(tmp_path)
 
         fetcher(id, tmp_path)
-        os.rename(tmp_path, pkg_path)
+        shutil.move(tmp_path, pkg_path)
         return True
 
     def remove(self, id):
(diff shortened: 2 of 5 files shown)


pkgpanda/actions.py
```diff
@@ -67,6 +68,18 @@ def swap_active_package(install, repository, package_id, systemd, block_systemd)
     activate_packages(install, repository, new_active, systemd, block_systemd)
 
 
+def show_package(repository, package_id):
```
branden: The functions in `actions` are intended to not be coupled to the application's interface, but `show_package()` prints its result, which means it's only useful in the context of a terminal. This function should return the `pkginfo` object and let the caller decide how to display it, but then it would do so little that it probably isn't worth defining as its own function.

pkgpanda/cli.py
```diff
@@ -166,6 +167,10 @@ def main():
                 print(pkg)
             sys.exit(0)
 
+        if arguments['show']:
+            actions.show_package(repository, arguments['<package-id>'])
```
branden: Per my other inlined comment, we can get away with a simple one-liner here:

```suggestion
            print(json.dumps(repository.load(arguments['<package-id>']).pkginfo, indent=4))
```

However I'm not sure if JSON is the best output format, it's not very human readable compared to e.g . YAML. If you need the JSON, maybe we can default to something more human readable and then support an optional `--json` flag to print JSON instead?

pkgpanda/actions.py
lava: Agreed, inlined this into cli.py.

pkgpanda/cli.py
lava: I think JSON is fine as output format here, since the underlying `pkginfo.json` was originally json-formatted anyways.

lava:
Note that I appended to smaller bugfix commits to this review that I had sitting in my worktree - admittedly not the best style, but I feel that the DC/OS PR process is too heavy to realistically turn all of these into separate PR's.

pkgpanda/cli.py
branden: I don't think the storage format is relevant to how we display it in the terminal. Pkgpanda CLI subcommands produce human-readable output, and I'd like that to stay consistent. A `--json` flag or similar should be suitable for use cases that need machine-readable output.

pkgpanda/cli.py
```diff
@@ -166,6 +168,17 @@ def main():
...
 
+        if arguments['show']:
+            package_id = arguments['<package-id>']
+            try:
+                pkginfo = repository.load(package_id).pkginfo
+                print(json.dumps(pkginfo, indent=4))
+                sys.exit(0)
+            except PackageNotFound:
+                print("Package not found: {}".format(package_id),
+                      file=sys.stderr)
```
branden: To be consistent with the style in the rest of this file, the newline that breaks this up should go after `print(`. But it looks like this file has a line length limit of 120 columns (there are a few existing lines that go beyond 80), so you can go ahead and put this on one line.

branden:
@lava The `shutil.move()` and doc fix commits LGTM. Thanks for writing descriptive commit messages!

pkgpanda/cli.py
lava: I think the idea behind having mandatory `pep8` linting on every PR is to be able to avoid these kind of style discussions, so given that this is valid according to https://www.python.org/dev/peps/pep-0008/#id17 I'm just going to leave it as is.

branden: From the doc you linked to:

> A style guide is about consistency. Consistency with this style guide is important. Consistency within a project is more important. *Consistency within one module or function is the most important.*

Emphasis mine. If you insist that pep8 catch all style inconsistencies before you make a change, then feel free to add the corresponding pep8 config to this PR. Otherwise, please make the simple style change I requested in order to keep this module consistent. Or find someone else to review this PR.

pkgpanda/cli.py
lava: I'm not sure that I follow your reasoning, can you explain why you think that JSON is not human-readable, and that YAML would be human-readable?

pkgpanda/cli.py
lava: To be clear, the portion you're quoting is defining exceptions for when you should *not* follow pep8, i.e. local consistency trumps the rules in this guide. Since we are enforcing pep8 in DC/OS, it is not really relevant here since all code is consistently pep8 anyways. Also, there's not a single instance of a `print(` followed by a newline in the current DC/OS codebase.

However, since you seem to care strongly about this, I went ahead and applied the style fix.

pkgpanda/cli.py
branden: > I'm not sure that I follow your reasoning, can you explain why you think that JSON is not human-readable, and that YAML would be human-readable?

YAML's purpose is to be a "human-friendly" serialization format, with syntax that's intended to be friendly to the eye. https://en.wikipedia.org/wiki/YAML I tossed it out as an example since it better resembles the output of the other pkgpanda subcommands, but I'm open to others, or even just a `for` loop of `print` statements. I'm concerned with keeping this CLI consistent as new subcommands are added.

PR #4148, opened 2019-01-09
[master/1.13] Use DC/OS Test Utils helpers for diagnostics tests

adamtheturtle:
## High-level description

There are multiple benefits to merging this pair of PRs (this PR + the DC/OS Enterprise change).

The driver for this is to progress towards making it possible to use `pytest` to collect and later run integration tests from a computer which is not a DC/OS cluster, such as a developer's laptop.
That will allow us to complete [DCOS-46439](https://jira.mesosphere.com/browse/DCOS-46439) but will also allow for greater developer productivity.
This PR alone does not achieve that, but it helps, as someone with a checkout of the DC/OS Enterprise repository will not necessarily be able to import open source tests.

I also consider this to be a clean up of some technical debt - this removes:

* Imports from test files - all common helpers should go in a common helpers file, or in DC/OS Test Utils
* Imports of private functions
* Duplication of code between DC/OS and DC/OS Test Utils - this is about to be a problem as these two code bases diverge in https://github.com/dcos/dcos/pull/3952/files

  - [DCOS-46438](https://jira.mesosphere.com/browse/DCOS-46438) Make it possible to collect DC/OS integration tests without a running cluster.

## Checklist for all PRs

  - [X] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: This is a test-only change
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This is a test-only change

### DIFF
--- a/packages/dcos-integration-test/extra/util/delete_ec2_volume.py
+++ b/packages/dcos-integration-test/extra/util/delete_ec2_volume.py
@@ -88,7 +88,7 @@ def delete_ec2_volume(name, timeout=600):
         try:
             log.info("Issuing volume.delete()")
             volume.delete()  # Raises ClientError (VolumeInUse) if the volume is still attached.
-        except exceptions.ClientError as exc:
+        except exceptions.ClientError:
             log.exception("volume.delete() failed.")
             raise
 
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -4,12 +4,7 @@ import os
 
 import pytest
 import requests
-from test_dcos_diagnostics import (
-    _get_bundle_list,
-    check_json,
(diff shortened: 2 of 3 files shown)


packages/dcos-integration-test/extra/util/delete_ec2_volume.py
```diff
@@ -88,7 +88,7 @@ def _delete_volume(volume):
         try:
             log.info("Issuing volume.delete()")
             volume.delete()  # Raises ClientError (VolumeInUse) if the volume is still attached.
-        except exceptions.ClientError as exc:
```
adamtheturtle: My local linter picked this unused variable up so I removed it.

rukletsov: Please as a separate commit next time.

PR #4165, opened 2019-01-10
[1.12] Fix hello-world flaky teardown

gracedo:
## High-level description

Tried wrapping the `cosmos.uninstall_package` in a 15min retry, but the retry complained that the package was already uninstalled. Switched to looking at `completed_frameworks` mesos /state response, but also ran into flake. I added some logs of the frameworks response, and found that in the flake, the framework remained under `'frameworks'` object but the `tasks` list within eventually emptied out to 0 (moved to `completed_tasks`). I think this could be a good way of assessing the framework as been torn down sufficiently if there are no longer any running tasks.

- [DCOS_OSS-4568](https://jira.mesosphere.com/browse/DCOS_OSS-4568) test_metrics.deploy_and_cleanup_dcos_package Framework hello-world still running

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: integration test change

### DIFF
--- a/packages/dcos-integration-test/extra/test_metrics.py
+++ b/packages/dcos-integration-test/extra/test_metrics.py
@@ -75,9 +75,13 @@ def deploy_and_cleanup_dcos_package(dcos_api_session, package_name, package_vers
             assert state_response.status_code == 200
             state = state_response.json()
 
-            for framework in state['frameworks']:
-                if framework['name'] == framework_name:
-                    raise Exception('Framework {} still running'.format(framework_name))
+            # Rarely, the framework will continue to show up in 'frameworks' instead of
+            # 'completed_frameworks', even after teardown. To avoid this causing a test
+            # failure, if the framework continues to show up in 'frameworks', we instead
+            # check if there are any running tasks.
+            frameworks = {f['name']: f for f in state['frameworks']}
+            assert framework_name not in frameworks or len(
+                frameworks[framework_name]['tasks']) == 0, 'Framework {} still running'.format(framework_name)
         wait_for_package_teardown()
 
 


gracedo:
I was finally able to hit a test failure!! https://teamcity.mesosphere.io/viewLog.html?buildId=1506697& as I suspected, wrapping the uninstall call wasn't the solution - it errored when it was called again
```
[2019-01-23 22:34:59,972|urllib3.connectionpool|DEBUG]: http://10.10.0.145:80 "POST /package/uninstall HTTP/1.1" 400 178
[2019-01-23 22:34:59,973|dcos_test_utils.package|INFO]: Response from cosmos: '{"type":"UninstallNonExistentAppForPackage","message":"Package [hello-world] with id [/hello-world] is not installed","data":{"packageName":"hello-world","appId":"/hello-world"}}'
```
So - it's claiming that the package was actually uninstalled, but the mesos `/state` checks were still showing the `hello-world` framework running under the `frameworks` object. Not sure what the best method is for checking whether a framework is still running or not, but maybe there is an rare upstream bug somewhere that continues to report a framework in /state even after it's been uninstalled.. I made a change to revert the uninstall retry loop and to look at `completed_frameworks` to check for uninstallation completion instead, we'll see how this goes.

packages/dcos-integration-test/extra/test_metrics.py
```diff
@@ -75,9 +75,19 @@ def wait_for_package_teardown():
...
+            completed = True
             for framework in state['frameworks']:
                 if framework['name'] == framework_name:
-                    raise Exception('Framework {} still running'.format(framework_name))
+                    completed = False
+                    if len(framework['tasks']) == 0:
+                        return
+            if completed:
+                return
+            raise Exception('Framework {} still running'.format(framework_name))
```
branden: Adding the `completed` variable complicates this `for` loop to the point where I had a hard time following it. Let's clarify this by building a `dict` of frameworks from the response, and then making assertions on that data structure:

```python
frameworks = {f['name']: f for f in state['frameworks']}
assert framework_name not in frameworks or len(frameworks[framework_name]['tasks']) == 0
```

gracedo: whoaaa dict comprehensions 😱

PR #3855, opened 2018-11-30
AR: supporting endpoints for services running in CNI

ahairape:
## High-level description
Due to Marathon app description syntax change https://github.com/mesosphere/marathon/blob/master/docs/docs/upgrade/network-api-migration.md#example-definitions AR is not able to expose apps defined with the new syntax. This PR fixes this issue.

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:
https://jira.mesosphere.com/browse/DCOS-45468

## Checklist for all PRs

  - [ ] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change:
  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [ ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

PR #4204, opened 2019-01-14
Add AR routing logic support for ip-per-container feature

ahairape:
## High-level description
This PR is a particular case of https://github.com/dcos/dcos/pull/3855 for DC/OS v1.11.

Due to Marathon app description syntax change https://github.com/mesosphere/marathon/blob/master/docs/docs/upgrade/network-api-migration.md#example-definitions AR is not able to expose apps requiring ip-per-container feature at /service/<app-id> endpoint. This PR fixes this issue.

https://jira.mesosphere.com/browse/DCOS-45468

## Related tickets (optional)

https://jira.mesosphere.com/browse/DCOS-45468

## Checklist for all PRs

  - [ x ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here:
  - [ x ] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [ x ] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

___

### DIFF
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -137,6 +137,8 @@ local function request(url, accept_404_reply, auth_token)
 end
 
 local function is_container_network(app)
+  -- Networking mode for a Marathon application is defined in
+  -- app["container"]["networks"][1]["mode"].
     local container = app["container"]
     local network = app["networks"][1]
     return container and network and (network["mode"] == "container")
@@ -217,6 +219,16 @@ local function fetch_and_store_marathon_apps(auth_token)
           )
 
        local host_or_ip = task["host"] --take host  by default
+       -- In "container" networking mode the task/container will get its own IP
+       -- (ip-per-container).
+       -- The task will be reachable:
+       -- 1) through the container port defined in portMappings.
+       -- If the app is using DC/OS overlay network
+       -- it will be also reachable on
+       -- 2) task["host"]:task["ports"][portIdx] (<private agent IP>:<hostPort>).
+       -- However, in case of CNI networks (e.g. Calico), the task might not be
+       -- reachable on task["host"]:task["ports"][portIdx], so we choose option 2)
+       -- for routing.
        if is_container_network(app) then
           ngx.log(ngx.NOTICE, "app '" .. appId .. "' is in a container network")
           -- override with the ip of the task
@@ -234,10 +246,6 @@ local function fetch_and_store_marathon_apps(auth_token)
           goto continue
        end
 
-       -- task["ports"] will contain the host ports made available to the task:
-       -- see https://mesosphere.github.io/marathon/docs/ports.html#definitions,
-       -- note on the hostPort definitions.
-       -- AR, however, needs the container port to route the request to.
        -- In "container" mode we find the container port out from portMappings array
        -- for the case when container port is fixed (non-zero value specified).
        -- When container port is specified as 0 it will be set the same as the host port:
@@ -282,8 +290,6 @@ local function fetch_and_store_marathon_apps(auth_token)
        end
 
        local url = scheme .. "://" .. host_or_ip .. ":" .. port
-       ngx.log(ngx.INFO, "Routing app " .. appId .. " to " .. url)
-
        svcApps[svcId] = {
          scheme=scheme,
          url=url,
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -136,13 +136,10 @@ local function request(url, accept_404_reply, auth_token)
     return res, nil
 end
 
-local function is_ip_per_task(app)
-    return app["ipAddress"] ~= nil and app["ipAddress"] ~= cjson_safe.null
-end
-
-local function is_user_network(app)
+local function is_container_network(app)
     local container = app["container"]
-    return container and container["type"] == "DOCKER" and container["docker"]["network"] == "USER"
+    local network = app["networks"][1]
+    return container and network and (network["mode"] == "container")
 end
 
 local function fetch_and_store_marathon_apps(auth_token)
@@ -220,8 +217,8 @@ local function fetch_and_store_marathon_apps(auth_token)
           )
 
        local host_or_ip = task["host"] --take host  by default
-       if is_ip_per_task(app) then
-          ngx.log(ngx.NOTICE, "app '" .. appId .. "' is using ip-per-task")
+       if is_container_network(app) then
+          ngx.log(ngx.NOTICE, "app '" .. appId .. "' is in a container network")
           -- override with the ip of the task
           local task_ip_addresses = task["ipAddresses"]
           if task_ip_addresses then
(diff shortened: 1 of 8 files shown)


packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -431,6 +432,65 @@ def test_ip_per_container(dcos_api_session):
...
+        'DCOS_SERVICE_SCHEME': 'http',
+    }
+
+    #  Arbitrary buffer time, accounting for propagation/processing delay.
+    buffer_time = 5
+
+    #  Cache refresh in Adminrouter takes 30 seconds at most.
+    #  CACHE_POLL_PERIOD=25s + valid=5s Nginx resolver DNS entry TTL
+    #  https://github.com/dcos/dcos/blob/cb9105ee537cc44cbe63cc7c53b3b01b764703a0/
+    #  packages/adminrouter/extra/src/includes/http/master.conf#L21
```
adamtheturtle: Thanks @timaa2k - we borrowed this from you!

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -1033,18 +1024,30 @@ def test_ip_per_task_app_without_user_networking(
...
-        specify 'ipAddress' at all, is successfully cached.
-        """
+    @pytest.mark.parametrize(
+        'networking_mode, container_port',
+        [['container/bridge', '80'], ['container/bridge', '0'], ['host', '80'], ['host', '0']]
+        )
+    def test_app_with_bridge_and_host_networking(
+            self, nginx_class, mocker, valid_user_header, container_port, networking_mode):
+        # Testing the cases when networking mode is either 'container' or 'host'.
+        # The host port can non-zero or 0. In the latter case Marathon will
```
timaa2k: s/can/can\ be/

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -238,23 +235,26 @@ local function fetch_and_store_marathon_apps(auth_token)
...
-         else
-            --override with the discovery ports
-            local discovery_ports = app["ipAddress"]["discovery"]["ports"]
-            for _, discovery_port in ipairs(discovery_ports) do
-                table.insert(ports, discovery_port["number"])
-            end
-         end
-       end
+         local port_mappings = app["container"]["portMappings"] or {}
+          for _, port_mapping in ipairs(port_mappings) do
```
jongiddy: This seems to be a lot of effort to get a single port. At this point we know that `app["container"]["portMappings"][portIdx]["containerPort"]` contains the value we want for `port`. Do we need to build a whole other array?

Could we just have something like:
```
if is_container_network(app) and app["container"]["portMappings"][portIdx]["containerPort"] ~= 0 then
    port = app["container"]["portMappings"][portIdx]["containerPort"]
else
    port = task["ports"][portIdx]
end
```

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -948,45 +948,24 @@ def test_if_backend_requests_have_useragent_set_correctly(
...
+        # randomly assigned by Marathon (0) or is not present at all:
+        # Admin Router must route the request to the specified container port.
         app = self._scheduler_alwaysthere_app()
-        app['ipAddress'] = {'networkName': 'samplenet'}
+        app['networks'] = [{
+            'mode': 'container',
+            'name': 'samplenet'
+        }]
+        if host_port:
+            app['container']['portMappings'] = [{'containerPort': 80, 'hostPort': host_port}]
```
jongiddy: This creates a dictionary with one integer value and one string value. Should both ports be of the same type? If that type is integer then you'll also need to change `if host_port:` to `if host_port is not None:` as `0` would trigger the other branch.

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -1005,18 +984,30 @@ def test_ip_per_task_app_with_user_networking_and_portdefinitions(
...
+        # Testing the case when container port is specified as 0
+        # in Marathon app definition with networking mode 'container'.
+        # This means that the Marathon app container port is randomly assigned
+        # by Marathon. The chosen container port number is same as hostPort
+        # and is available through task["ports"] array. We are reusing port
+        # 16000 on 127.0.0.1 exposed by the mock server, as the one randomly
+        # chosen by Marathon.
+        # It does not matter if the host port is fixed (non-zero),
+        # randomly assigned by Marathon (0) or is not present at all:
+        # Admin Router must route the request to the specified container port.
```
jongiddy: Possibly I misunderstand the meaning of this, but can you please clarify why it is necessary for AR to route to the container port? This port is internal to the container, so how can AR route to the container port?

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -948,45 +948,24 @@ def test_if_backend_requests_have_useragent_set_correctly(
...
-                                headers=valid_user_header)
-
-            assert resp.status_code == 200
-            req_data = resp.json()
-            assert req_data['endpoint_id'] == 'http://127.0.0.2:80'
-
-    def test_ip_per_task_app_with_user_networking_and_portdefinitions(
-            self, nginx_class, mocker, valid_user_header):
+    @pytest.mark.parametrize('host_port', ['12345', '0', None])
+    def test_app_with_container_networking_and_fixed_container_port(
```
jongiddy: Should we have a test for multiple `container` networks? (See the last point of [Marathon Port Mappings](https://mesosphere.github.io/marathon/docs/networking.html#port-mappings))

ahairape: We have got no request for supporting that case, so I think that would be out of the scope of this PR. We can add it as a feature work, if needed.

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
ahairape: Every task/container  gets its own IP in `container` networking mode (exposed via `task[“ipAddresses”][1][“ipAddress”]`). On the host with that IP, the containerPort is used to reach the task. 

For an application using container networking mode (app[“networks”][1][“mode”] == “container”) with dcos overlay network (app[“networks”][1][“name”] == “dcos”) it is possible to route to the application with both <host>:<port> == task[“ipAddresses”][1][“ipAddress”] : app[“container”][“portMappings”][1][“containerPort”] as well as <host>:<port> == tasks[1][“host”] : tasks[1][“ports”][portIdx].

For a Calico-networked application, however, it is possible to route to the application only using <host>:<port> == task[“ipAddresses”][1][“ipAddress”] : app[“container”][“portMappings”][1][“containerPort”] combination. 

That combination, therefore, has been chosen for routing by AR to support both cases.

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -237,38 +246,56 @@ local function fetch_and_store_marathon_apps(auth_token)
...
        end
 
-       local port = ports[portIdx]
        if not port then
           ngx.log(ngx.NOTICE, "Cannot find port at Marathon port index '" .. (portIdx - 1) .. "' for app '" .. appId .. "'")
           goto continue
        end
 
+       local do_rewrite_req_url = labels["DCOS_SERVICE_REWRITE_REQUEST_URLS"]
+       if do_rewrite_req_url == false or do_rewrite_req_url == 'false' then
```
jongiddy: :grey_question: Maybe adding a link to https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/README.md#disabling-url-path-rewriting-for-selected-applications just before this line would provide info about the effect of values in this variable, allowing you to remove the comment in the `else` clause.

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -237,38 +246,56 @@ local function fetch_and_store_marathon_apps(auth_token)
...
+       else
+          -- Treat everything else as true, i.e.:
+          -- * label is absent
+          -- * label is set to "true" (string) or true (bool)
+          -- * label is set to some random string
+          do_rewrite_req_url = true
+       end
+
+       local do_request_buffering = labels["DCOS_SERVICE_REQUEST_BUFFERING"]
+       if do_request_buffering == false or do_request_buffering == 'false' then
```
jongiddy: Ditto https://github.com/dcos/dcos/blob/master/packages/adminrouter/extra/src/README.md#disabling-request-buffering-for-selected-applications

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
jongiddy: Thanks for fixing this.

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -1005,18 +984,30 @@ def test_ip_per_task_app_with_user_networking_and_portdefinitions(
...
-    def test_ip_per_task_app_without_user_networking(
-            self, nginx_class, mocker, valid_user_header):
+    @pytest.mark.parametrize('host_port', [12345, 0, None])
+    def test_app_with_container_networking_and_random_container_port(
+            self, nginx_class, mocker, valid_user_header, host_port):
+        # Testing the case when container port is specified as 0
+        # in Marathon app definition with networking mode 'container'.
+        # This means that the Marathon app container port is randomly assigned
+        # by Marathon. The chosen container port number is same as hostPort
+        # and is available through task["ports"] array. We are reusing port
```
jongiddy: The container port is "randomly assigned" but also "is same as hostPort". Should the sentence "The chosen container port number is same as hostPort and is available through task["ports"] array" be in this comment?

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
jdef: A container port *may* be internal to the container, but AR (and Marathon) have no way of knowing that. I'm pretty sure that Marathon defaults to using *host* ip/port when in host or bridged networking mode, and container-ip:container-port when in *container* networking mode. I forget what's reported by `task["ipAddresses"][1]["ipAddress"]` - that might actually depend on the version of Marathon you're testing against, since networking changes/fixes are made from release to release.

All that said, I don't know how many of our customers use host-private container networks that aren't cluster-routeable. They don't seem very useful on a cluster like DC/OS, but I suppose for some specialized, highly secure environment there's probably an argument to be made for them. So the assumptions made here are probably fine. It might be worth documenting somewhere higher up in the docs/stack for people trying to debug service connectivity.

jgehrcke:
@ahairape this is lacking a `CHANGES.md` entry, right? If so I will add one.

PR #4232, opened 2019-01-16
DCOS-46438 - Support collecting integration tests locally without a cluster.

jbarrick-mesosphere:
## High-level description

This enables us to collect the integration tests locally without running inside of a cluster. While it doesn't allow us to actually run the tests (due to missing DCOS files), it allows us to import the tests and gather some statistics (for example, to unblock DCOS-46439).

  - [DCOS-46438](https://jira.mesosphere.com/browse/DCOS-46438) Make it possible to collect DC/OS integration tests without a running cluster

## Related tickets (optional)

- [DCOS-46439](https://jira.mesosphere.com/browse/DCOS-46439) xfailflake dashboard - collect marked issues without a cluster and without regex
- [DCOS-47044](https://jira.mesosphere.com/browse/DCOS-47044) Hand off the xfailflake dashboard script.
- [DCOS-46438](https://jira.mesosphere.com/browse/DCOS-46438) Make it possible to collect DC/OS integration tests without a running cluster

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: N/A, change is for CI

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -154,3 +154,34 @@ changedir=packages/bootstrap/extra
 commands=
   pip install .
   pytest --basetemp={envtmpdir} {posargs}
+
+
+[testenv:py35-collect-tests]
+platform=linux|darwin
+# See the following link for AWS_* environment variables:
+# http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials
+passenv =
+  TEAMCITY_VERSION
+  SSH_AUTH_SOCK
+  AZURE_PROD_STORAGE_ACCOUNT
+  AZURE_PROD_STORAGE_ACCESS_KEY
+  AZURE_DEV_STORAGE_ACCOUNT
+  AZURE_DEV_STORAGE_ACCESS_KEY
+  AWS_ACCESS_KEY_ID
+  AWS_SECRET_ACCESS_KEY
+  AWS_SESSION_TOKEN
+  AWS_PROFILE
+  AWS_CONFIG_FILE
+  AWS_DEFAULT_REGION
+deps =
+  git+https://github.com/dcos/dcos-test-utils.git
+  dnspython
+  pytest==3.8.2
+  PyYAML
+  webtest
+  webtest-aiohttp==1.1.0
+  schema
+  kazoo
+
+commands=
+  pytest --collect-only --basetemp={envtmpdir} packages/dcos-integration-test/extra/
--- a/tox.ini
+++ b/tox.ini
@@ -60,14 +60,7 @@ passenv =
   AWS_PROFILE
   AWS_CONFIG_FILE
   AWS_DEFAULT_REGION
-deps =
-  dnspython
-  pytest==3.3.2
-  pytest-catchlog==1.2.2
-  PyYAML
-  webtest
-  webtest-aiohttp==1.1.0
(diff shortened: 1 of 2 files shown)


jbarrick-mesosphere:

tox.ini
```diff
@@ -154,3 +154,34 @@ changedir=packages/bootstrap/extra
...
+  AZURE_PROD_STORAGE_ACCESS_KEY
+  AZURE_DEV_STORAGE_ACCOUNT
+  AZURE_DEV_STORAGE_ACCESS_KEY
+  AWS_ACCESS_KEY_ID
+  AWS_SECRET_ACCESS_KEY
+  AWS_SESSION_TOKEN
+  AWS_PROFILE
+  AWS_CONFIG_FILE
+  AWS_DEFAULT_REGION
+deps =
```
adamtheturtle: Let's pin dependencies so that our CI is predictable.

tox.ini
```diff
@@ -154,3 +154,34 @@ changedir=packages/bootstrap/extra
 commands=
   pip install .
   pytest --basetemp={envtmpdir} {posargs}
+
+
+[testenv:py35-collect-tests]
```
adamtheturtle: There are multiple test suites so let's be clear - this is integration tests.

tox.ini
adamtheturtle: Let's be careful - we are duplicating dependencies - anyone adding some in one place should know to add them in another.

Perhaps this should be a `requirements.txt` file?

jbarrick-mesosphere: Possibly, yeah.. but I ran into this today: https://github.com/tox-dev/tox/issues/149

adamtheturtle: For sure let's pin it.

adamtheturtle: Maybe https://github.com/signalpillar/tox-battery would help?

tox.ini
```diff
@@ -154,3 +154,34 @@ changedir=packages/bootstrap/extra
...
+platform=linux|darwin
+# See the following link for AWS_* environment variables:
+# http://boto3.readthedocs.io/en/latest/guide/configuration.html#configuring-credentials
+passenv =
```
adamtheturtle: Why do these have to be set? Locally I can collect without these.

tox.ini
jbarrick-mesosphere: It seems like you have to install that dependency before you run tox :/

adamtheturtle: Is that a big problem? Is that worse than not having a `requirements.txt` (and therefore not really helping folks who are not using `tox`)?

tox.ini
jbarrick-mesosphere: I'm not sure if they have to, but I copied them from the other tox environments. I wasn't sure if them being set would make a difference.

tox.ini
jbarrick-mesosphere: It might not be a problem, but we've already got a ton of dependencies to install before you can get started: https://github.com/dcos/dcos/tree/dfb9df89f0863d9423bf814578c7c9ceb05afb30#development-environment

I'm not keen on adding more, but I don't feel super strongly about it. Maybe we could patch upstream tox.

PR #4362, opened 2019-01-29
Open /package/describe and /package/resource to any authenticated user

bamarni:
These endpoints are used to install CLI extensions through the `dcos
package install` subcommand (eg. `dcos package install
dcos-enterprise-cli`) and currently require the full
dcos:adminrouter:package permission even though they do not install
anything on the cluster side.

This change opens these endpoints to all authenticated users, making it
easier for unprivileged CLI users to install package subcommands.

https://jira.mesosphere.com/browse/DCOS_OSS-4651

### DIFF
--- a/packages/adminrouter/extra/src/lib/auth/open.lua
+++ b/packages/adminrouter/extra/src/lib/auth/open.lua
@@ -159,7 +159,7 @@ function _M.init(use_auth)
     end
 
     -- /package/
-    res.access_package_endpoint = function()
+    res.access_package_endpoint = function(package_path)
         return res.do_authn_and_authz_or_exit()
     end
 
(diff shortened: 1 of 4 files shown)


bamarni:

packages/adminrouter/extra/src/lib/auth/open.lua
```diff
@@ -159,7 +159,7 @@ function _M.init(use_auth)
     end
 
     -- /package/
-    res.access_package_endpoint = function()
+    res.access_package_endpoint = function(package_path)
```
ahairape: Do we need to pass `package_path` argument if we're not using it later?

bamarni: I'm not sure, but this is consistent with https://github.com/dcos/dcos/blob/82abd9676231e24c8a888e3e8c2b65df82d800e0/packages/adminrouter/extra/src/lib/auth/open.lua#L101

bamarni:

timaa2k:

PR #4388, opened 2019-01-31
Use Python's tarfile in make_tar to avoid Linux/macOS difference in tar distributions

timaa2k:
## High-level description

This PR removes operating system specific tar commands in favor of Python's built-in `tarfile` module. It adds a new DC/OS package `liblzma` which is used by `tarfile` to do `lzma` compression for `xz` archives.

This was a collaborative effort together with @adamtheturtle. It enables running DC/OS OSS unit tests natively on macOS.

  - [DCOS_OSS-4764](https://jira.mesosphere.com/browse/DCOS_OSS-4764) Make it possible/documented to run unit tests on macOS.

## Checklist for all PRs

### DIFF
--- a/packages/liblzma/build
+++ b/packages/liblzma/build
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -o errexit -o nounset
+
+mkdir -p build
+pushd build
+
+export CFLAGS=-I/opt/mesosphere/include
+export LDFLAGS="-L/opt/mesosphere/lib -Wl,-rpath=/opt/mesosphere/lib"
+export CPPFLAGS=-I/opt/mesosphere/include
+
+cd /pkg/src/liblzma
+./autogen.sh
+./configure --prefix="$PKG_PATH"
+make
+make install
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -9,20 +9,11 @@ Each package contains a pkginfo.json. That contains a list of requires as well a
 environment variables from the package.
 
 """
-try:
-    import grp
-except ImportError:
-    pass
 import json
 import os
 import os.path
-try:
-    import pwd
-except ImportError:
-    pass
 import re
 import shutil
-import sys
 import tempfile
 from collections import Iterable
 from itertools import chain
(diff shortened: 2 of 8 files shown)


adamtheturtle:
@mesosphere-mergebot bump-ee

pkgpanda/__init__.py
```diff
@@ -12,6 +12,7 @@
 try:
     import grp
 except ImportError:
+    # It is not possible to import grp on Windows.
```
jongiddy: Thanks for adding an explanation. Even better, I think we should move this and `import pwd` below the import of `is_windows` and then replace the `try` with an `if not is_windows:`. This documents the case where it is not supported, but also errors if, for some reason, another platform does not have these modules available.

We can also get rid of the later `assert` checks that `pwd` and `grp` exist.

packages/liblzma/build
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
jongiddy: Should we add `set -o errexit -o nounset` to make the build fail on error?

I notice a lot of `build` scripts don't do this, but some do.

pkgpanda/__init__.py
timaa2k: Beautiful, I like it!

packages/liblzma/build
timaa2k: Adding this one as well.

timaa2k:
cc @mrnugget @branden

timaa2k:
@mesosphere-mergebot sync

timaa2k:
@mesosphere-mergebot label Holding

timaa2k:
As discussed with @DominikDary we do not intend to revert this merge.

orsenthil:
@timaa2k  / @DominikDary With mergebot mistakenly applying ship-it label in open, and thus merging both open and EE PR, our existing tests were successful with this when exercised in train PR.

* Something to note and keep in mind, Windows build has been broken  prior to this merge,  https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcOs_Windows_Build_BuildDcOs&page=2&tab=buildTypeHistoryList&branch_DcOs_Windows_Build=%3Cdefault%3E  and we have not focused to bring it back to healthy state yet.

* If this PR helps with building DC/OS on mac, and running the tests on Mac, that feature could be verified and shared in the description of this PR.

PR #4473, opened 2019-02-11
Revert back to platform-dependent tar decompression

timaa2k:
## High-level description

This PR reverts the decompression part of: [#4388](https://github.com/dcos/dcos/pull/4388/files)
After the aforementioned PR was merged we noticed that the CI runs into this Python bug:
https://bugs.python.org/issue21872

Due to complications to reproduce the issue locally this should be reverted to the old functionality for now.

  - [DCOS-48220](https://jira.mesosphere.com/browse/DCOS-48220) pkgpanda: Compressed file ended before the end-of-stream marker was reached.

## Checklist for all PRs

### DIFF
--- a/pkgpanda/util.py
+++ b/pkgpanda/util.py
@@ -15,6 +15,7 @@ from contextlib import contextmanager, ExitStack
 from itertools import chain
 from multiprocessing import Process
 from shutil import rmtree
+from subprocess import check_call
 from typing import List
 
 import requests
@@ -218,8 +219,13 @@ def extract_tarball(path, target):
         assert os.path.exists(path), "Path doesn't exist but should: {}".format(path)
         make_directory(target)
 
-        with tarfile.open(name=str(path), mode='r') as tar:
-            tar.extractall(path=str(target), numeric_owner=True)
+        # TODO(tweidner): https://jira.mesosphere.com/browse/DCOS-48220
+        # Make this cross-platform via Python's tarfile module once
+        # https://bugs.python.org/issue21872 is fixed.
+        if is_windows:
+            check_call(['bsdtar', '-xf', path, '-C', target])
+        else:
+            check_call(['tar', '-xf', path, '-C', target])
 
     except:
         # If there are errors, we can't really cope since we are already in an error state.


timaa2k:

adamtheturtle:
I can confirm that all unit tests pass on macOS! Thank you @timaa2k !

pkgpanda/util.py
```diff
@@ -218,8 +219,13 @@ def extract_tarball(path, target):
         assert os.path.exists(path), "Path doesn't exist but should: {}".format(path)
         make_directory(target)
 
-        with tarfile.open(name=str(path), mode='r') as tar:
-            tar.extractall(path=str(target), numeric_owner=True)
```
adamtheturtle: Do we lose anything by no longer specifying numeric owner anywhere?

timaa2k: We might, I guess only CI can answer that question at the moment.

adamtheturtle: And if we find some edge case later we find missing tests - we must be willing to take those risks else we will be paralyzed by the past.

timaa2k: The numeric owner option is applied when creating the archive in our case via the `tarfile.TarInfo`.
It is in the archive and the decompression is the same as before so I guess we do not lose anything.

adamtheturtle: Is it? Before we touched anything there was the UID and GID set but there was also an option which forced numeric owner:

```
--numeric-owner
always use numbers for user/group names
```

timaa2k: I've checked it by looking into a tar'd DC/OS package (using `dcos-shell` tar) before and after the change.
Before the change there is no non-numeric owner and group information stored:
```
$ tar -tvJf bootstrap--8704a6496c6dfe56980c4926f40e94b31da7f067.tar.xz
-rwxr-xr-x 0/0            2568 2018-03-22 19:45 ./bin/bootstrap
```
Afterwards there is also only numeric information about the owner:
```
$ tar -tvJf bootstrap--8704a6496c6dfe56980c4926f40e94b31da7f067.tar.xz
-rwxr-xr-x 0/0            2568 2019-01-28 17:27 ./bin/bootstrap
```
On extraction if the superuser extracts a tar archive the information in the archive will be used.
Look for `--same-owner` in the manual:
https://www.gnu.org/software/tar/manual/html_node/Attributes.html#SEC138
For non-root users this applies (from the same manual):
```
When restoring, [tar] tries to look the name (if one was written) up in `/etc/passwd'. If it fails, then it uses the user ID stored in the archive instead.
```

timaa2k: `bsdtar` has the same semantics if running as `root`:
https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&manpath=freebsd-release-ports

timaa2k:
@mesosphere-mergebot sync

timaa2k:

timaa2k:
@mesosphere-mergebot merge-it

PR #4549, opened 2019-02-19
DCOS-46615 - Retry downloading pkgpanda pkg if file size incorrect

mrnugget:
## High-level description

AWS S3, from which we download pre-built pkgpanda packages, might sometimes end the connection before the package has been fully downloaded. (See https://github.com/aws/aws-sdk-js/issues/312 and https://github.com/boto/boto3/issues/540 for more context on this)

In our case that leads to invalid `tar.xz` archives which cause `unexpected end of input` errors and result in faulty builds of DC/OS (`dcos_generate_config.sh`) which contain these invalid archives and then result in install-time errors for users. See this comment for an investigation into how this can affect installation errors: https://jira.mesosphere.com/browse/DCOS_OSS-4097?focusedCommentId=219259&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-219259

This change here adds a sanity check to the `download` function used by pkgpanda. It checks whether the final file size (after flushing the file to disk) matches the `Content-Length` we get from S3. If it doesn't match, it sleep for 2, 4, 8, ... seconds and then retries until it either downloads the file completely or the retries are exhausted. In that case, it raises an exception.

  - [[DCOS-46615] Pkgpanda - package dcos-image - xz: (stdin): Unexpected end of input - JIRA](https://jira.mesosphere.com/browse/DCOS-46615)

## Related tickets (optional)

  - [[DCOS_OSS-4097] pkgpanda: Explicitly create user groups for all users - JIRA](https://jira.mesosphere.com/browse/DCOS_OSS-4097)

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: it's internal behaviour that's not exposed to the user
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: I didn't include a test yet, since that would need to involve setting up a mock http server etc. and I wanted to get input on this first

___

Reviewers should be:

### DIFF
--- a/pkgpanda/test_util.py
+++ b/pkgpanda/test_util.py
@@ -1,8 +1,11 @@
 import os
 import tempfile
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from subprocess import CalledProcessError
+from threading import Thread
 
 import pytest
+import requests
 
 import pkgpanda.util
 from pkgpanda import UserManagement
@@ -412,3 +415,49 @@ def test_write_string(tmpdir):
     st_mode = os.stat(filename).st_mode
     expected_permission = 0o777
     assert (st_mode & 0o777) == expected_permission
+
+
+class MockDownloadServerRequestHandler(BaseHTTPRequestHandler):
+    def do_GET(self):  # noqa: N802
+        body = b'foobar'
+
+        self.send_response(requests.codes.ok)
+        self.send_header('Content-Length', '6')
+        self.send_header('Content-Type', 'text/plain')
+        self.end_headers()
+
+        if self.server.requests_received == 0:
+            # Don't send the last byte of the response body.
+            self.wfile.write(body[:len(body) - 1])
+        else:
+            self.wfile.write(body)
+        self.server.requests_received += 1
+
+        return
+
+
+class MockHTTPDownloadServer(HTTPServer):
+    requests_received = 0
+
+
+def test_stream_remote_file_with_retries(tmpdir):
+    mock_server = MockHTTPDownloadServer(('localhost', 0), MockDownloadServerRequestHandler)
+    mock_server_port = mock_server.server_port
+
+    mock_server_thread = Thread(
+        target=mock_server.serve_forever,
+        daemon=True)
+    mock_server_thread.start()
+
+    url = 'http://localhost:{port}/foobar.txt'.format(port=mock_server_port)
+
+    out_file = os.path.join(str(tmpdir), 'foobar.txt')
+    response = pkgpanda.util._download_remote_file(out_file, url)
+
+    response_is_ok = response.ok
+    assert response_is_ok
+
+    assert mock_server.requests_received == 2
+
+    with open(out_file, 'rb') as f:
+        assert f.read() == b'foobar'
--- a/pkgpanda/util.py
+++ b/pkgpanda/util.py
@@ -19,13 +19,14 @@ from subprocess import check_call
 from typing import List
 
 import requests
+import retrying
 import teamcity
 import yaml
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 from teamcity.messages import TeamcityServiceMessages
 
-from pkgpanda.exceptions import FetchError, ValidationError
+from pkgpanda.exceptions import FetchError, IncompleteDownloadError, ValidationError
 
 is_windows = platform.system() == "Windows"
 
@@ -150,6 +151,33 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
     return session
 
 
+def _is_incomplete_download_error(exception):
+    return isinstance(exception, IncompleteDownloadError)
+
+
+@retrying.retry(
+    stop_max_attempt_number=3,
+    wait_random_min=1000,
+    wait_random_max=2000,
+    retry_on_exception=_is_incomplete_download_error)
+def _download_remote_file(out_filename, url, retries=4):
+    with open(out_filename, "wb") as f:
+        r = get_requests_retry_session().get(url, stream=True)
+        r.raise_for_status()
+
+        content_length = int(r.headers['content-length'])
+
+        total_bytes_read = 0
+        for chunk in r.iter_content(chunk_size=4096):
+            f.write(chunk)
+            total_bytes_read += len(chunk)
+
+        if total_bytes_read != content_length:
+            raise IncompleteDownloadError(url, total_bytes_read, content_length)
+
+        return r
+
+
 def download(out_filename, url, work_dir, rm_on_error=True):
     assert os.path.isabs(out_filename)
     assert os.path.isabs(work_dir)
@@ -167,14 +195,7 @@ def download(out_filename, url, work_dir, rm_on_error=True):
                 src_filename = work_dir + '/' + src_filename
             shutil.copyfile(src_filename, out_filename)
         else:
-            # Download the file.
-            with open(out_filename, "w+b") as f:
-                r = get_requests_retry_session().get(url, stream=True)
-                if r.status_code == 301:
-                    raise Exception("got a 301")
-                r.raise_for_status()
-                for chunk in r.iter_content(chunk_size=4096):
-                    f.write(chunk)
+            _download_remote_file(out_filename, url)
     except Exception as fetch_exception:
         if rm_on_error:
             rm_passed = False
--- a/pkgpanda/exceptions.py
+++ b/pkgpanda/exceptions.py
@@ -16,6 +16,22 @@ class FetchError(Exception):
         return msg
 
 
+class IncompleteDownloadError(Exception):
+
+    def __init__(self, url, total_bytes_read, content_length):
+        self.url = url
+        self.total_bytes_read = total_bytes_read
+        self.content_length = content_length
+
+    def __str__(self):
+        msg = "Problem fetching {} - bytes read {} does not match content-length {}".format(
+            self.url,
+            self.total_bytes_read,
+            self.content_length)
+
+        return msg
+
+
 class InstallError(Exception):
     pass
 


pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
 
+def stream_remote_file_with_retries(out_filename, url, retries=4):
+    def cleanup():
+        try:
+            os.remove(out_filename)
+        except Exception:
+            pass
+
+    while True:
+        with open(out_filename, "w+b") as f:
```
gpaul: I think you should drop the `+` here since you're not reading, only writing.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+            r = get_requests_retry_session().get(url, stream=True)
```
gpaul: By default, `get` will follow redirects (http://docs.python-requests.org/en/v3.0.0/_modules/requests/api/#get). If you really want to ensure that the request did not get redirected you should pass `allow_redirects=False` here.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+            if r.status_code == 301:
```
gpaul: Why do you care about redirects? It seems reasonable that the CDN / S3 would redirect you to a specific download URL.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+                raise Exception("got a 301")
+            r.raise_for_status()
+
+            for chunk in r.iter_content(chunk_size=4096):
+                f.write(chunk)
+            f.flush()
```
gpaul: Consider maintaining "total bytes read" by accumulating `len(chunk)` instead of `stat`ing the file and explicitly flushing it.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+
+            content_length = int(r.headers['content-length'])
+            final_size = os.fstat(f.fileno()).st_size
+            if final_size == content_length:
+                return r
+            else:
+                f.close()
```
gpaul: Calling `f.close()` is often a sign that you could restructure your code, as the context manager `with open(...) as f:` will do that automatically when you leave its scope.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+                cleanup()
+                retries -= 1
```
gpaul: There's a very lovely little python library called `retrying` (https://pypi.org/project/retrying/) that let's you annotate a function (even an inline function) with retry logic of this sort.) We should use that instead of rolling our own.

For example:
https://github.com/dcos/dcos/blob/5391993fa0ca8457dae4d2126d2c0c515dd8264a/packages/dcos-integration-test/extra/util/delete_ec2_volume.py#L46
https://github.com/dcos/dcos/blob/57350f12eb7c1ed82eeb7e8944c72776b936160d/packages/dcos-integration-test/extra/test_networking.py#L60-L62

I like declaring little inner functions in my larger function and annotating the inner ones with retrying logic.

pkgpanda/util.py
```diff
@@ -167,14 +200,7 @@ def download(out_filename, url, work_dir, rm_on_error=True):
...
         else:
-            # Download the file.
-            with open(out_filename, "w+b") as f:
-                r = get_requests_retry_session().get(url, stream=True)
-                if r.status_code == 301:
-                    raise Exception("got a 301")
-                r.raise_for_status()
-                for chunk in r.iter_content(chunk_size=4096):
-                    f.write(chunk)
+            stream_remote_file_with_retries(out_filename, url)
```
gpaul: I think `_with_retries` is an implementation detail, let's just call the function "_download_file" ?

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
     return session
 
 
+def stream_remote_file_with_retries(out_filename, url, retries=4):
```
adamtheturtle: Can we use a known-good retrying library. We already have one on clusters (see [here](https://github.com/dcos/dcos/blob/86a6934657c3d1c22f77d3cd21e112f940988043/ssh/test_integration.py#L13) for an example use).

adamtheturtle: `retries=4`! Let's explain all of these choices, even if the explanation is "this is my best guess for a compromise and it seems to work". That means that the next maintainer will know whether it is safe to change.

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+    def cleanup():
+        try:
+            os.remove(out_filename)
+        except Exception:
```
adamtheturtle: Can you perhaps be more particular with the exceptions that you wish to ignore?

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+            pass
+
+    while True:
+        with open(out_filename, "w+b") as f:
+            r = get_requests_retry_session().get(url, stream=True)
+            if r.status_code == 301:
+                raise Exception("got a 301")
```
adamtheturtle: Where do you expect this error message to be seen - what do you expect someone who sees it to do?

pkgpanda/util.py
```diff
@@ -150,6 +151,38 @@ def get_requests_retry_session(max_retries=4, backoff_factor=1, status_forcelist
...
+
+            for chunk in r.iter_content(chunk_size=4096):
+                f.write(chunk)
+            f.flush()
+
+            content_length = int(r.headers['content-length'])
+            final_size = os.fstat(f.fileno()).st_size
+            if final_size == content_length:
+                return r
+            else:
```
adamtheturtle: There is no need for `else` here.

pkgpanda/util.py
adamtheturtle: I see we had the same thought! https://github.com/dcos/dcos/pull/4549/files#r258438916

pkgpanda/test_util.py
```diff
@@ -412,3 +416,53 @@ def test_write_string(tmpdir):
     st_mode = os.stat(filename).st_mode
     expected_permission = 0o777
     assert (st_mode & 0o777) == expected_permission
+
+
+class MockDownloadServerRequestHandler(BaseHTTPRequestHandler):
+    def do_GET(self):  # noqa: N802
+        self.send_response(requests.codes.ok)
```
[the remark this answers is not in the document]
mrnugget: Will do that, yes!

And, yes, `BaseHTTPRequestHandler` calls `flush`: https://github.com/python/cpython/blob/ca5e6aaea5168bff274f3e98d7970bfa1bc366a2/Lib/http/server.py#L415

pkgpanda/test_util.py
```diff
@@ -412,3 +416,53 @@ def test_write_string(tmpdir):
...
+
+
+class MockHTTPDownloadServer(HTTPServer):
+    requests_received = 0
+
+
+def test_stream_remote_file_with_retries(tmpdir):
+    def _get_free_port():
+        s = socket.socket(socket.AF_INET, type=socket.SOCK_STREAM)
+        s.bind(('localhost', 0))
```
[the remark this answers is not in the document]
mrnugget: Thanks! The `_get_free_port` was a straight prototyping-let's-get-something-to-run-copy from an example. Now I learned something and will use your suggestion :)

pkgpanda/test_util.py
```diff
@@ -412,3 +416,53 @@ def test_write_string(tmpdir):
...
+    out_file = os.path.join(str(tmpdir), 'foobar.txt')
+    response = pkgpanda.util.stream_remote_file_with_retries(out_file, url)
+
+    response_is_ok = response.ok
+    assert response_is_ok
+
+    assert mock_server.requests_received == 2
+
+    with open(out_file) as f:
+        assert f.read() == 'foobar'
```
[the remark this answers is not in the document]
mrnugget: Your attention to consistency is balm for this nitpicking programmer's soul! :)

pkgpanda/util.py
mrnugget: As I wrote in Slack: I — personally — don't care 😄 I've taken this from the existing (untested) code that was already there, presumably for years. And when working with legacy code that doesn't have tests to confirm something. I tread lightly and only really change what needs to be changed. Or, in other words: if I don't know why code is there and don't know why it _shouldn't_ be there I'll leave it alone — for now, because I don't want to break something that I didn't know existed.

pkgpanda/util.py
mrnugget: I wanted to differentiate it from the existing `download` function and make its possibly slow behaviour explicit, but I get your point.

I'll rename it to `_download_remote_file`, since `download` has two branches: "download" local file or download remote file. The latter is what's encapsulated in this function.

pkgpanda/util.py
mrnugget: Also taken from the existing code, but it's probably safe to drop it.

pkgpanda/util.py
mrnugget: Ah, yes, I wanted to clean this up after confirming general approach. Let's see if we can get rid of a lot of it when using `retrying`...

pkgpanda/util.py
mrnugget: This is taken from here:

https://github.com/dcos/dcos/blob/2897e2459e460033fb2b72099d4aac29e2b58937/pkgpanda/util.py#L138

No idea why it's `4` but I thought I'd keep the behaviour the same.

pkgpanda/util.py
mrnugget: This is also taken from here: https://github.com/dcos/dcos/blob/2897e2459e460033fb2b72099d4aac29e2b58937/pkgpanda/util.py#L185-L189

pkgpanda/util.py
mrnugget: See here: https://github.com/dcos/dcos/pull/4549#discussion_r259736459

pkgpanda/util.py
mrnugget: Fixed that :)

pkgpanda/util.py
gpaul: SGTM

pkgpanda/util.py
mrnugget: That's a good idea!

mrnugget:

pkgpanda/util.py
mrnugget: Switched to `retrying` — it's lovely!

mrnugget:
@gpaul @adamtheturtle Can you re-review? I removed the useless legacy code, worked in all of your suggestions and switched to `retrying` – the only thing left is the long-running thread for the mock http server, which will run until the end of the test suite, I guess. I don't think that's a huge issue, but I'll look into it.

mrnugget:
@mesosphere-mergebot bump-ee

gpaul:
@mesosphere-mergebot sync

gpaul:
@mesosphere-mergebot sync

PR #4642, opened 2019-02-28
Update the cloud-formation templates to use specific Availability Zone

orsenthil:
DCOS-18825 - Pin the Availability Zone per region for the AWS CF Template

**NOTE** - Cloudformation tests are successful with this change:

1. https://teamcity.mesosphere.io/viewLog.html?buildId=1642996&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsCloudFormationSimple 

2. The list of regions map to the supported ones listed here:
https://github.com/dcos/dcos/blob/master/gen/build_deploy/aws.py#L129

3. If we have to update the instance support from `m4.xlarge` (https://github.com/dcos/dcos/blob/master/gen/aws/templates/cloudformation.json#L35) to something else like m5, we will have to update the list of amis per region that is built for the instance in step 2. That's a larger change, and this change makes sure that our default cloudformation templates will keep working.

### DIFF
--- a/gen/aws/templates/cloudformation.json
+++ b/gen/aws/templates/cloudformation.json
@@ -73,15 +73,8 @@
           "default": "PT45M"
       }},
     "RegionMap" : {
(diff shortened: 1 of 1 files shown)


orsenthil:

PR #4712, opened 2019-03-07
dcos-net: ensure the desination directory exists

jkoelker:
## High-level description

Makes the systemd network configuration directory (`/etc/systemd/network`) if it does not already exists.

  - [DCOS-49711](https://jira.mesosphere.com/browse/DCOS-49711) COPS-4621: dcos-net-setup.py fails if systemd-networkd is installed but not active.

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: test would just end up testing `os.makedirs`

### DIFF
--- a/packages/dcos-net/extra/dcos-net-setup.py
+++ b/packages/dcos-net/extra/dcos-net-setup.py
@@ -52,6 +52,7 @@ def main():
 
 def add_networkd_config(src):
     networkd = b'systemd-networkd.service'
+    networkd_path = '/etc/systemd/network'
 
     # Check if there is networkd
     result = subprocess.run(['systemctl', 'list-unit-files', networkd],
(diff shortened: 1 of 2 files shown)


jkoelker:

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -52,6 +52,7 @@ def main():
 
 def add_networkd_config(src):
     networkd = b'systemd-networkd.service'
+    networkd_path = '/etc/systemd/network'
```
Fabs: is there a reason the `networkd` string is annotated with the `b` and the `networkd_path` is not?

urbanserj: `subprocess.run()` returns binary data and `b''` is here just for networkd to avoid type casting.

jkoelker: The type of `result.stdout` is a`bytes` so it needs to be a byte string to do `if networkd not in result.stdout:`. The alternative would be to do `if networkd not in result.stdout.decode():` to compare it as utf-8 strings instead of opaque bytes.

PR #4730, opened 2019-03-12
DCOS-18825- Update CORE OS AMI used in CloudFormation templates.

orsenthil:
* Use m5.large as the default instance type.
* Update CoreOS AMI used from the AMI in dcos-images.


orsenthil:

PR #4744, opened 2019-03-13
DCOS-18825 - Update CORE OS AMI used CloudFormation templates.

orsenthil:
* Remove the Availability Zone Pinning.
* Update EL7 AMIs.
* We are not including gov cloud region now.


orsenthil:

PR #4885, opened 2019-03-25
Fix dcos_add_user.py script for OSS

ahairape:
## High-level description

This PR fixes the bug in dcos_add_user.py script and adds integration tests to check the user addition in Open DC/OS.

https://jira.mesosphere.com/browse/DCOS_OSS-4930

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: This fixes a bug which we have never released.

### DIFF
--- a/packages/dcos-integration-test/extra/test_legacy_user_management.py
+++ b/packages/dcos-integration-test/extra/test_legacy_user_management.py
@@ -200,21 +200,21 @@ def test_dcos_add_user(dcos_api_session):
     command = ['python', '/opt/mesosphere/bin/dcos_add_user.py', email_address]
     cli.exec_command(command)
 
-    r = dcos_api_session.get('/acs/api/v1/users')
-    r.raise_for_status()
-    expected_user_data = {
-        "uid": email_address,
-        "description": "",
-        "url": "/acs/api/v1/users/" + email_address,
-        "is_remote": True,
-        "is_service": False,
-        "provider_type": "oidc",
-        "provider_id": "https://dcos.auth0.com/"
-    }
-
-    delete_user(dcos_api_session, email_address)
-
-    assert expected_user_data in r.json()['array']
+    try:
+        r = dcos_api_session.get('/acs/api/v1/users')
+        r.raise_for_status()
+        expected_user_data = {
+            "uid": email_address,
+            "description": "",
+            "url": "/acs/api/v1/users/" + email_address,
+            "is_remote": True,
+            "is_service": False,
+            "provider_type": "oidc",
+            "provider_id": "https://dcos.auth0.com/"
+        }
+        assert expected_user_data in r.json()['array']
+    finally:
+        delete_user(dcos_api_session, email_address)
 
 
 def test_check_message_on_adding_user_twice(dcos_api_session):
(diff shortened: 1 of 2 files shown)


packages/dcos-integration-test/extra/test_legacy_user_management.py
```diff
@@ -188,8 +189,55 @@ def test_dynamic_ui_config(dcos_api_session):
     assert 'uiConfiguration' in data
 
 
-def test_dcos_add_user_script_exists_oss():
+def test_dcos_add_user(dcos_api_session):
     """
-    dcos_add_user.py script exists in /opt/mesosphere/bin directory.
+    dcos_add_user.py script adds a user to IAM using the
```
adamtheturtle: There is an unnecessary repetition of "dcos_add_user.py" in this docstring.

adamtheturtle:
We ran these tests locally on an OSS cluster and there is a helper in the new test file which skips the test on Enterprise.

We also ran the tests without the change and saw the 400 error.

packages/dcos-integration-test/extra/test_legacy_user_management.py
```diff
@@ -188,8 +189,55 @@ def test_dynamic_ui_config(dcos_api_session):
...
+        "url": "/acs/api/v1/users/" + email_address,
+        "is_remote": True,
+        "is_service": False,
+        "provider_type": "oidc",
+        "provider_id": "https://dcos.auth0.com/"
+    }
+
+    delete_user(dcos_api_session, email_address)
+
+    assert expected_user_data in r.json()['array']
```
mhrabovcin: I know that this is technically correct, but it is odd to assert for user existence after you've deleted the user. I'd swap both lines here and in the following test.

timaa2k:
LGTM mod CI

packages/dcos-integration-test/extra/test_legacy_user_management.py
adamtheturtle: I would rather keep it as is. This is so that the test does not leave behind a user if the assert fails. We could do `try...finally` but I thought that this was the neatest option.

jongiddy: The problem is that the user doesn't get deleted if anything else fails (e.g. the `raise_for_status`). This is more obvious in the following test, where some `assert`s exist before the `delete_user` and some after.

Could we use a fixture to do the cleanup?

packages/dcos-integration-test/extra/test_legacy_user_management.py
```diff
@@ -188,8 +189,55 @@ def test_dynamic_ui_config(dcos_api_session):
...
+
+    # Adding the same user second time
+    stdout, stderr = cli.exec_command(command)
+
+    expected_error = '[INFO] User `' + email_address + '` already exists\n'
+
+    delete_user(dcos_api_session, email_address)
+
+    assert expected_error == stderr
+    assert '' == stdout
```
jongiddy: This result implies an exit status of `0`. Should it be `1` if the user exists?

jongiddy: This result implies an exit status of `0`. Should it be `1` if the user exists?

packages/dcos-integration-test/extra/test_legacy_user_management.py
adamtheturtle: Yes. We have https://github.com/mesosphere/dcos-enterprise/blob/master/packages/dcos-integration-test/extra/conftest.py#L164 in DC/OS Enterprise. I considered adding similar to OSS but chose not to for a few reasons:

* It seemed like a bit too much for this one issue - I considered this balance pragmatic, but I am happy to do that work if you like
* The Enterprise helper already has a TODO, stating that it isn't done as someone would like
* There is currently no `dcos_api_session.iam` module, so ideally that would come first - I already plan to make a JIRA for that.

How do you feel if I create a JIRA to add this as follow-up?

packages/dcos-integration-test/extra/test_legacy_user_management.py
adamtheturtle: It does have an exit status of `0`.
That isn't just implied, it is tested for - else `exec_command` would raise a `subprocess.CalledProcessError`.

I would like to not change this here.

Would you like me to make an issue to change this? Or just leave it?

jongiddy: Leave it. It is kind of a success in an idempotent way. Although it is also something an admin might want to flag up.

packages/dcos-integration-test/extra/test_legacy_user_management.py
jongiddy: I certainly don't want you to put much effort in right now. How about going with the `try...finally` option for now?

adamtheturtle: Will do.

adamtheturtle:
We rebased and I believe that all tests will pass apart from flakes so I am putting this into review.

ahairape:

PR #4887, opened 2019-03-25
Annotate all upstreams including original URI

timaa2k:
## High-level description

This PR annotates all Admin Router location directives that include a `proxy_pass` other than
agent proxy locations or locations under `/service`. We refer to the currently annotated set of locations as Upstreams (apart from Mesos all of the annotated requests are handled locally on a master node).
Each Upstream is given an `$upstream_tag` that identifies it in terms of DC/OS terminology. Each location annotation also makes sure that Upstream metrics are tagged with the `$original_uri` to Admin Router rather than the rewritten `$uri` that was actually used in the `proxy_pass`.

  - [DCOS_OSS-4945](https://jira.mesosphere.com/browse/DCOS_OSS-4945) Annotate all upstreams including original URI for Admin Router metrics.

## Checklist for all PRs

  - [ ] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: We have had an entry for collecting Admin Router metrics.

___

### DIFF
--- a/packages/adminrouter/extra/src/includes/server/master.conf
+++ b/packages/adminrouter/extra/src/includes/server/master.conf
@@ -76,6 +76,11 @@ location ~ ^/system/v1/agent/(?<agentid>[0-9a-zA-Z-]+)(?<url>/logs.*|/metrics/v0
 # Group: System
 # Description: Node and cluster checks
 location /system/checks/v1 {
+    # Annotation for DCOS Checks upstream metrics.
+    set $upstream_tag DCOSChecks;
+    set $original_uri $uri;
+    include includes/metrics-upstream-location-level.conf;
+
     access_by_lua_block {
         auth.access_system_checks_endpoint();
         util.clear_dcos_cookies();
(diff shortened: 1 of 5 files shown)


timaa2k:

packages/adminrouter/extra/src/includes/server/master.conf
```diff
@@ -434,6 +460,11 @@ location /capabilities {
 # Group: Authentication
 # Description: Access Control Service
 location /acs/api/v1 {
+    # Annotation for IAM upstream metrics.
+    set $upstream_tag IAM;
```
jgehrcke: Don't we want to distinguish `/internal/acs/api/...` from `/acs/api/...`?

jgehrcke: Same for MesosDNS.

packages/adminrouter/extra/src/includes/server/open/master.conf
```diff
@@ -1,6 +1,11 @@
-# Group: Pkgpanda
+#IAtrea Group: Pkgpanda
```
jongiddy: What does this change do?

packages/adminrouter/extra/src/includes/server/master.conf
jongiddy: The intent is to allocate requests to upstream services. In that sense, a call to `/internal/X` is the same as a call to `/X`. There will be a breakdown by URI, which, with this change, will separate the `internal` calls from the external calls.

packages/adminrouter/extra/src/includes/server/open/master.conf
jgehrcke: Seems like a typo, should be [iatreia](http://aprendeenlinea.udea.edu.co/revistas/index.php/iatreia)

timaa2k: Thanks for spotting this @jongiddy . Lol @jgehrcke

PR #4890, opened 2019-03-25
Update NAT AMIS to address CloudFormation Resource Errors

orsenthil:
#### StackStatus changed unexpectedly to: ROLLBACK_IN_PROGRESS - Resource unavailability issue

was observed due to resource unavailability of the NAT AMI. The error message in the console had this errror

```
Your requested instance type (m3.medium) is not supported in your requested Availability Zone (us-west-2d). Please retry your request by not specifying an Availability Zone or choosing us-west-2c, us-west-2b, us-west-2a. (Service: AmazonEC2; Status Code: 400; Error Code: Unsupported; Request ID: 11bba17d-87a4-4e7c-a0b1-3d3c09257a4a)
```
Attempting to resolve this issue by upgrading the NAT AMI provided by Amazon to the latest version (https://aws.amazon.com/amazon-linux-ami/2018.03-release-notes/) 

* https://jira.mesosphere.com/browse/DCOS-18825

PR #4891, opened 2019-03-25
Revert to m4 and pin the region map

orsenthil:
Another attempt to solve - https://jira.mesosphere.com/browse/DCOS-18825

* Revert to m4 instance and pin the region map.


orsenthil:
closing in favor of https://github.com/dcos/dcos/pull/4890

PR #4939, opened 2019-03-29
Telegraf scale tuning

branden:
## High-level description

This PR edits Telegraf config to help prevent dropped metrics on busy agents. This is accomplished by increasing the number of metrics output pluggins can buffer before they start dropping metrics (`metric_buffer_limit`), and by increasing the collection `interval` for the `dcos_statsd` input plugin, which reduces the rate at which it collects metrics, thus reducing pressure on output buffers.

This PR also reduces the default `interval` for all input plugins to 10s. Most inputs don't need to be throttled on a busy agent, since they emit a consistent number of metrics that doesn't increase as workloads are executed on an agent.

Last, this PR sets `expiration_interval` for the `prometheus_client` output to its default value, instead of using it implicitly.

I'll backport this fix to 1.12, so no changelog update is necessary.

  - [DCOS-50994](https://jira.mesosphere.com/browse/DCOS-50994) Telegraf started to drop metrics on some agents.

## Related tickets (optional)

N/A

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: This will be backported to 1.12, so it won't be a change for 1.13.
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: This solves a problem that only appears at scale, which our integration tests don't cover.

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -2130,7 +2130,7 @@ package:
 {% endswitch %}
       [agent]
         ## Default data collection interval for all inputs
-        interval = "20s"
+        interval = "10s"
         ## Rounds collection interval to 'interval'
         ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
         round_interval = true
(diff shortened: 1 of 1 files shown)


branden:

gen/dcos-config.yaml
```diff
@@ -2130,7 +2130,7 @@ package:
 {% endswitch %}
       [agent]
         ## Default data collection interval for all inputs
-        interval = "20s"
+        interval = "10s"
```
bbannier: This value is crucial knowledge to reliably report metrics from short-lived jobs (they need to live at least `interval` to make sure that their metrics enter the system). https://jira.mesosphere.com/browse/DCOS-49035

branden: 👍 We probably ought to fix the dcos_statsd input so that it doesn't discard a container's metrics before they've been gathered.

PR #4943, opened 2019-03-29
Migrate telegraf containers dir as root

philipnrmn:
The mesos module's containers directory is owned by root. We therefore migrate
its contents in an ExecPre step, instead of in Exec using the dcos_telegraf
user.

## High-level description

  - [DCOS-50697](https://jira.mesosphere.com/browse/DCOS-50697) Evidence of Telegraf failures after upgrading from 1.11 to 1.13.0-alpha

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: this stops the 1.11->1.13 upgrade breaking, and changes nothing else
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: this is covered by upgrade tests (which discovered the issue in the first place)

### DIFF
--- a/packages/telegraf/extra/dcos-telegraf.service
+++ b/packages/telegraf/extra/dcos-telegraf.service
@@ -20,6 +20,9 @@ ExecStartPre=/bin/bash -c "mkdir -p /run/dcos/telegraf"
 ExecStartPre=/bin/bash -c "chmod 775 /run/dcos/telegraf"
 ExecStartPre=/bin/bash -c "chown root:dcos_telegraf /run/dcos/telegraf"
 
+# Migrate old containers dir to new location in case the cluster was upgraded.
+ExecStartPre=/opt/mesosphere/active/telegraf/tools/migrate_containers_dir.sh "${LEGACY_CONTAINERS_DIR}" "${TELEGRAF_CONTAINERS_DIR}"
+
 ExecStartPre=/bin/bash -c "mkdir -p ${TELEGRAF_USER_CONFIG_DIR}"
 
 ExecStartPre=/opt/mesosphere/bin/bootstrap ${SERVICE}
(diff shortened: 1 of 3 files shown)


philipnrmn:

philipnrmn:
@branden and @gracedo, would you kindly review this (one-line!) PR when you have a minute?

packages/telegraf/extra/dcos-telegraf.service
```diff
@@ -19,6 +19,8 @@ LimitNOFILE=16384
 ExecStartPre=/bin/bash -c "mkdir -p /run/dcos/telegraf"
 ExecStartPre=/bin/bash -c "chmod 775 /run/dcos/telegraf"
 ExecStartPre=/bin/bash -c "chown root:dcos_telegraf /run/dcos/telegraf"
+# The metrics isolator module dir may not exist; if so, chown it to dcos_telegraf; if not, no-op.
+ExecStartPre=/bin/bash -c "chown -R dcos_telegraf /run/dcos/mesos/isolators/com_mesosphere_MetricsIsolatorModule || :"
```
branden: Can we get away with doing the migration itself as root in this step? That seems less risky than changing the ownership of a runtime directory that belongs to Mesos.

philipnrmn: Yes, that should work. I feel like there is a reason we didn't do it that way to begin with, but it's lost in the mists of time and the original PR doesn't help.

branden: Yeah I don't recall either. Could just be that the startup script was an easy place to put a bit of bash.

philipnrmn: Entirely possible :grimacing: 

I've made the change you suggested.

PR #4981, opened 2019-04-03
Pass Accept-Encoding header to service backends

branden:
## High-level description

This PR causes adminrouter to stop stripping the `Accept-Encoding` header from requests to `/service/<service-name>` before it proxies them to service backends. This allows service backends to compress responses.

  - [DCOS_OSS-4906](https://jira.mesosphere.com/browse/DCOS_OSS-4906) Admin Router service endpoint: forward Accept-Encoding header to upstream

## Related tickets (optional)

N/A

## Checklist for all PRs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -90,6 +90,8 @@ This change also aligned the authentication architectures between DC/OS Enterpri
 
 ### Fixed and improved
 
+* Backends for requests to `/service/<service-name>` may compress responses (DCOS_OSS-4906)
+
 * Fixed issue where Metronome did not handle restart policy is ON_FAILURE correctly, not restarting the task. (DCOS_OSS-4636 )
 
 * Prefix illegal prometheus metric names with an underscore (DCOS_OSS-4899)
--- a/packages/adminrouter/extra/src/includes/service-location.common.conf
+++ b/packages/adminrouter/extra/src/includes/service-location.common.conf
@@ -1,5 +1,3 @@
-    more_clear_input_headers Accept-Encoding;
-
     include includes/proxy-headers.conf;
     include includes/websockets.conf;
 
--- a/packages/adminrouter/extra/src/test-harness/tests/test_master.py
+++ b/packages/adminrouter/extra/src/test-harness/tests/test_master.py
@@ -23,7 +23,7 @@ log = logging.getLogger(__name__)
 
 class TestServiceEndpoint:
     # Majority of /service endpoint tests are done with generic tests framework
-    def test_if_accept_encoding_header_is_removed_from_upstream_request(
+    def test_if_accept_encoding_header_is_in_upstream_request(
             self, master_ar_process_perclass, mocker, valid_user_header):
         headers = copy.deepcopy(valid_user_header)
         headers['Accept-Encoding'] = 'gzip'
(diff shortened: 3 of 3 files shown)


jgehrcke:
@ahairape @DominikDary I think we should work towards marking `continuous-integration/jenkins/pr-head` as a `required` status check? So far we always treated it as such. CC @jongiddy @orsenthil @adamtheturtle 

Thanks for taking initiative here @branden!

@vespian I would appreciate to hear your thoughts. Can you imagine a scenario that would break as of this change?

jongiddy:
Yes, this test is broken by the change, so should be fixed. It should just require a change of the `assert_headers_absent` line to `assert_headers={'Accept-Encoding': 'gzip'},` and changing the test name to something more appropriate (e.g. `test_if_accept_encoding_header_is_kept_in_upstream_request`).

jongiddy:
(and of course you need a CHANGELOG entry for this).

branden:
Thanks for pointing out the test I need to fix, @jongiddy. I pushed a test fix and a changelog update. Once this passes CI, I'll mark it ready for review.

branden:
@mesosophere-mergebot label ready for review

branden:

ahairape:
> @ahairape @DominikDary I think we should work towards marking `continuous-integration/jenkins/pr-head` as a `required` status check. Do you agree? So far we always treated it as such. CC @jongiddy @orsenthil @adamtheturtle

I agree @jgehrcke. 
It would be helpful to know what were the reasons back at the time the check was written not to mark it as `required`. Are those reasons still valid?

vespian:
Hey,

The test was added in order to freeze existing behaviour during the AdminRouter refactoring that happened 1.5 years ago. This was one of many legacy settings that were simply migrated over without modifications as researching and double-checking each and every one of them would have take some extra time and the refactoring effort took almost 5 months on its own. 

The clearing of the headers can be traced as far as the 2nd commit of the *original* adminrouter repo (i.e. even before the split to `ee` and `open` repositories):

https://github.com/mesosphere/adminrouter/commit/466127f607e7447e9b9abd612959bbaedca1b432#diff-daaf2fceca49ed034eace91c4e875653R62

It was almost 4 years ago on 13 May 2015, done by Seb while he was still with us. I wish I could be more helpfull here, but unfortunatelly I do not know what was the original problem that this config was solving. I can only guess that *maybe* there were some services running on DC/OS that did not do well with these headers?

> @vespian I would appreciate to hear your thoughts. Can you imagine a scenario that would break as of this change?

Nothing comes to my mind - AFAIK nginx does not decompress the responses it gets and just forwards them, so there should be no negative impact on performance.

branden:
> I can only guess that _maybe_ there were some services running on DC/OS that did not do well with these headers?

I wondered about that too. My other guess was that maybe adminrouter was modifying response bodies, which would require that it receives them uncompressed.  But I couldn't find any evidence of either of these situations in JIRA or in git history.

CHANGES.md
```diff
@@ -90,6 +90,8 @@ This change also aligned the authentication architectures between DC/OS Enterpri
 
 ### Fixed and improved
 
+* Backends for requests to `/service/<service-name>` may compress responses (DCOS_OSS-4906)
```
jgehrcke: Note to self, maybe for an editorial change later on:

Admin Router's service endpoint now forwards the `Accept-Encoding` header if present in the request. This for example allows for gzip-compressed responses.

PR #4988, opened 2019-04-04
Changed order of containerizers on Mesos agents.

kaysoky:
## High-level description

This is a subtle change that changes the processing order for
operations associated with launching, monitoring, or stopping
containers.  Effectively, the Mesos agent will try the Mesos
containerizer before the Docker containerizer.

This change improves performance of launching Mesos containers when
there are a large number of Docker containers running on the same
agent.  The Docker containerizer tends to spend most of its time
waiting on the Docker daemon to respond, sometimes blocking further
progress if there are too many containers.  Sometimes, this waiting
will cause Mesos containers to take excessively long times to launch,
as we get blocked while executing the launch command on each
containerizer, *in order*.  If the Mesos containerizer comes first,
then we skip the wait entirely.

This is unlikely to cause the opposite scenario (too many Mesos
containers blocking a Docker container from launching) because the
Mesos containerizer does not have long blocking sections.

  - [DCOS-42757](https://jira.mesosphere.com/browse/DCOS-42757) Switch composing containerizer order in agent flag.

## Related tickets (optional)

  - [COPS-3794](https://jira.mesosphere.com/browse/COPS-3794) Tasks stuck in STAGING.

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: This should be transparent to the user, as we are flipping the order of some operations.  Functionality does not change.
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: There is no test because this is a performance improvement for an uncommon case (launching ~50 docker containers, then a single UCR container on the same agent).  We can independently check how much time is saved though.

  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review): N/A

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -1330,7 +1330,10 @@ package:
       GLOG_drop_log_memory=false
       MESOS_CGROUPS_ENABLE_CFS=true
       MESOS_CGROUPS_LIMIT_SWAP=false
-      MESOS_CONTAINERIZERS=docker,mesos
+      # The order of containerizers affects the order of certain operations
+      # on the agent.  See https://github.com/dcos/dcos/pull/4988 before
+      # making any changes to this variable.
+      MESOS_CONTAINERIZERS=mesos,docker
       MESOS_CONTAINER_LOGGER=com_mesosphere_mesos_JournaldLogger
       MESOS_DEFAULT_CONTAINER_DNS=file:///opt/mesosphere/etc/mesos-slave-dns.json
       MESOS_DISALLOW_SHARING_AGENT_PID_NAMESPACE=true


gen/dcos-config.yaml
```diff
@@ -1330,7 +1330,7 @@ package:
       GLOG_drop_log_memory=false
       MESOS_CGROUPS_ENABLE_CFS=true
       MESOS_CGROUPS_LIMIT_SWAP=false
-      MESOS_CONTAINERIZERS=docker,mesos
+      MESOS_CONTAINERIZERS=mesos,docker
```
lava: Since this is very subtle, I'd recommend a comment like
```
# See https://github.com/dcos/dcos/pull/4988 before changing this variable
```

kaysoky: I was hoping the commit history would provide enough context in case anyone wants to change the line back.  (The PR description is taken from the commit description.)
But it certainly won't hurt to add.

lava: It would certainly provide enough context for anyone caring to look into the `git blame` for that line, but I can imagine e.g. someone re-ordering them alphabetically without looking at the history first because it's an 'obviously' harmless change.

kaysoky:

PR #4990, opened 2019-04-04
[1.12] Fix lua variable scope bug in service.lua

ahairape:
## High-level description

This PR fixes a bug in service.lua library. See corresponding JIRA ticket for details. This is a backport of a change implemented as part of DCOS_OSS-4950.

https://jira.mesosphere.com/browse/DCOS_OSS-4999

## Checklist for all PRs

No test is included since the behaviour of AR resulting from the bug is non-deterministic.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,7 +17,7 @@ Format of the entries must be.
 
 ### Notable changes
 
-* Fix a bug in service.lua which was causing some wrong metrics labels for Admin Router and affecting request buffering by Admin Router. (DCOS_OSS-4950, DCOS_OSS-4999)
+* Fix a bug in Admin Router's service endpoint as of which the DCOS_SERVICE_REQUEST_BUFFERING setting was not adhered to in all cases. (DCOS_OSS-4999)
 
 * Update ZooKeeper to release [3.4.14](https://zookeeper.apache.org/doc/r3.4.14/releasenotes.html). (DCOS_OSS-4988)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,12 +17,12 @@ Format of the entries must be.
 
 ### Notable changes
 
-* Fix a bug in Admin Router's service endpoint as of which the DCOS_SERVICE_REQUEST_BUFFERING setting was not adhered to in all cases. (DCOS_OSS-4999)
-
 * Update ZooKeeper to release [3.4.14](https://zookeeper.apache.org/doc/r3.4.14/releasenotes.html). (DCOS_OSS-4988)
 
 
 ### Fixed and improved
+
+* Fix a bug in Admin Router's service endpoint as of which the DCOS_SERVICE_REQUEST_BUFFERING setting was not adhered to in all cases. (DCOS_OSS-4999)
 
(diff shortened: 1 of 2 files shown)


ahairape:

packages/adminrouter/extra/src/lib/service.lua
```diff
@@ -312,6 +312,7 @@ local function recursive_resolve(auth, path, marathon_cache, mesos_cache)
     local do_request_buffering = true
     local err_code = nil
     local err_text = nil
+    local service_name = ""
```
jgehrcke: Next time it would be nice to add a code comment here stating the importance of this declaration.

(the non-obvious part here is that things "seem to work" even without this declaration)

ahairape: @jgehrcke There are backport foreseen for 1.11 and 1.10: https://jira.mesosphere.com/browse/DCOS_OSS-4999
Does it make sense to add it to 1.10's CHANGES.md?

ahairape:
@jgehrcke There are backport foreseen for 1.11 and 1.10: https://jira.mesosphere.com/browse/DCOS_OSS-4999
Does it make sense to add the note to 1.10's CHANGES.md?

packages/adminrouter/extra/src/lib/service.lua
jongiddy: I asked @ahairape to remove the comment that he initially added. Declaring variables as local should be the default expectation. All these local variables could cause problems if they are not declared local. We should add an explicit comment when something is _not_ local. (or Lua should default to local unless a variable is global - viva Python!)

jgehrcke:
> Does it make sense to add the note to 1.10's CHANGES.md?

Discussed this offline. If we backport the patch to 1.10.x and 1.11.x, then yes, the changelog entry needs to go into all those branches.

ahairape:
@jgehrcke I've added the an entry to CHANGES.md, PTAL.

CHANGES.md
```diff
@@ -17,6 +17,8 @@ Format of the entries must be.
 
 ### Notable changes
 
+* Fix a bug in service.lua which was causing some wrong metrics labels for Admin Router and affecting request buffering by Admin Router. (DCOS_OSS-4950, DCOS_OSS-4999)
+
```
jgehrcke: Let's put this into the `Fixed and improved` section. I believe from 1.12's point of view the reference to metrics labels does not apply.

jgehrcke:
What about

```
Fix a bug in Admin Router's service endpoint as of which the `DCOS_SERVICE_REQUEST_BUFFERING` setting was not adhered to in all cases.
```

Would that be an accurate technical description of the bug from a user's point of view?

ahairape:
@jgehrcke I have changed the `CHANGES.md` entry as you suggested, PTAL.
@jongiddy Could you PTAL?

Thank you!

jongiddy:
@mesosphere-mergebot override-status teamcity/dcos/test/aws/cloudformation/simple https://jira.mesosphere.com/browse/DCOS-51332

PR #5132, opened 2019-04-16
[1.13] Tune fluent bit's configuration

philipnrmn:
## High-level description

Backport of #5131

  - [DCOS-51758](https://jira.mesosphere.com/browse/DCOS-51758) Masters and many agents got unreachable after upgrading soak113s from 1.13.0-alpha to 1.13.0-beta

## Related tickets (optional)

## Checklist for all PRs

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: this is a configuration change, no dc/os code is involved here. 
  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: this simply ensures that fluentbit will not cause performance issues on startup, there is no user-facing change

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -2785,6 +2785,9 @@ package:
           # are further forwarded into elasticsearch, where fields with leading
           # underscore are reserved for internal use.
           Strip_Underscores On
+          # Tail the journal (i.e., only send new logs). Without this option set
+          # fluent bit will attempt to send the entire contents of the journal
+          # on startup, which can cause performance issues.
           Read_From_Tail    On
       # Log forwarding is disabled by default.
       [OUTPUT]
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -2785,6 +2785,7 @@ package:
           # are further forwarded into elasticsearch, where fields with leading
           # underscore are reserved for internal use.
           Strip_Underscores On
+          Read_From_Tail    On
       # Log forwarding is disabled by default.
       [OUTPUT]
           Name  null


gen/dcos-config.yaml
```diff
@@ -2785,6 +2785,7 @@ package:
           # are further forwarded into elasticsearch, where fields with leading
           # underscore are reserved for internal use.
           Strip_Underscores On
+          Read_From_Tail    On
```
jdef: Why this change instead of setting `DB` to some value in order to configure a persistent cursor?

philipnrmn: We should enable the DB (this will require setting some build flags, as [FLB_SQLDB is not enabled by default](https://github.com/fluent/fluent-bit/blob/master/CMakeLists.txt#L62)), but we would still need the `Read_From_Tail` flag - when a user upgrades to 1.13, they will have logs on disk which are unread, and it's this read that caused issues in soak.

jdef: Understood. It sounds like we should have a JIRA (if we don't already) to track the tasks associated with enabling the `DB` flag here. Can you link such to this comment thread?

philipnrmn:
@jdef, the reason this isn't marked ready for review is that those things are not done.

gen/dcos-config.yaml
philipnrmn: https://jira.mesosphere.com/browse/DCOS-51855

philipnrmn:

philipnrmn:
@gracedo and @branden, would you kindly review this PR?

PR #5167, opened 2019-04-17
Add information about deprecation DC/OS

janisz:
## High-level description

Add information about deprecation DC/OS

  - [DCOS-51011](https://jira.mesosphere.com/browse/DCOS-51011)

## Related tickets (optional)

## Checklist for all PRs

___

Reviewers should be:


janisz:
:wave: @drewvanstone @br-lewis @makkes  Can you review, please?

jgehrcke:
As I argued before (elsewhere) I think we can just remove this service, applying some structured thinking around private and public interface: this service can be looked at as an implementation detail of a UI feature (connected via a private interface). That UI feature (which is the public interface to users) was removed, and so that very implementation detail can also be removed.

What do I miss; why do we think we need to keep the service around?

High level point of view: what we do here seems like a measure of precaution, which I honor, but I'd like us to approach this more systematically, so that we can iterate faster and focus our energy on deprecation cycles that we actually need :-).

makkes:
> As I argued before (elsewhere) I think we can just remove this service, applying some structured thinking around private and public interfaces: this service can be looked at as an implementation detail of a UI feature (connected via a private interface). That UI feature (which is the public interface to users) was removed, and so that very implementation detail can also be removed.
> 
> What do I miss; why do we think we need to keep the service around? Did we identify unexpected consumers that we cannot break yet?
> 
> High level point of view: what we do here seems like a measure of precaution, which I honor, but I'd like us to approach this more systematically, so that we can iterate faster and focus our energy on deprecation cycles that we actually need :-).

Thanks @jgehrcke for your valuable input. I think the reason for deprecating it before removal is that we just applied our standard deprecation policy to it and honestly no one on the team is 100% sure that this service is not used by any customer. So we're riding the safe road here. If anyone could assure us that the standard deprecation policy doesn't have to be applied here we'd be more than happy to shortcut our way to getting rid of dcos-history.

janisz:

janisz:
@jgehrcke At least one user is dcos-diagnostic https://github.com/dcos/dcos-diagnostics/blob/1389e1499b0790a9bdf041514fe173a395a4b7b8/dcos/finder.go#L28

jgehrcke:
> If anyone could assure us that the standard deprecation policy doesn't have to be applied here we'd be more than happy to shortcut our way to getting rid of dcos-history.

I'd be happy to take the responsibility for that; this is what I was trying to say.

The major conceptual argument is that this is an implementation detail where a deprecation policy built for a public interface does not apply.

The most important practical argument is that it is _extremely_ unlikely for external tooling to consume the history service API. The information it provides is pretty much useless for consumers except for our UI.

"Not knowing with 100 % certainty" of course always implies a risk, and I am rarely willing to take it. Here, however, I am convinced that it's the right thing to do.

Let's talk about the very unlikely event of an external dependency:

1) In the unlikely event that anyone out there depends on the history service API: once we remove it we will most likely surprise them -- the deprecation notice will _most likely_ not prevent that surprise from happening.
2) Based on (1) the major reason to do the deprecation is to, in the event of the surprise, be able to say "but we properly deprecated it!", preventing embarrassment from happening. 
3) A similar diplomatic advantage can be achieved by saying "but you consumed an implementation detail, probably accidentally, sorry for the hassle", also preventing embarrassment from happening. 

So, what arguments do we have for removing the history service? Getting rid of maintenance burden (see e.g. COPS-4652). The earlier we simplify the less work we have in the future.

makkes:
Ok, if we want to got for just removing dcos-history service, I'd be happy to do so. Let's do it! First, though, we should at least talk once to Ernest and/or Seth from Product about this because there's a whole bunch of tickets in Jira referring dcos-history service decommisioning, removing dependencies and so forth. If they agree and we get rid of all the Caps/Features in JIRA we can go for it.

PR #5226, opened 2019-04-19
DCOS_OSS-5083 Use PATH for coreutils and procps bins

wolf31o2:
Rather than supply an explicit path to the binaries provided by the
coreutils or procps packages, fetch them from the PATH. These are
binaries such as `cp` or `kill` and tend to be assumed to be in the
default PATH on any system. However, these are different prefixes on
RHEL-based systems and Debian-based systems. Rather than be explicit on
the prefix, let's assume the PATH includes these items. This will always
be true on Redhat and Ubuntu installations.

Signed-off-by: Chris Gianelloni <cgianelloni@applause.com>

## High-level description

Removes explicit PATHs on coreutils or procps binaries so we can install/run on Ubuntu.

  - [DCOS_OSS-5083](https://jira.mesosphere.com/browse/DCOS_OSS-5083) Use common tools from PATH versus explicit paths

## Checklist for all PRs

___

Reviewers should be:

### DIFF
--- a/gen/azure/cloud-config.yaml
+++ b/gen/azure/cloud-config.yaml
@@ -11,7 +11,7 @@ root:
       StandardOutput=journal+console
       StandardError=journal+console
       ExecStartPre=/usr/bin/curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/tmp/d.deb https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/docker-ce_17.09.0~ce-0~ubuntu_amd64.deb
-      ExecStart=/usr/bin/bash -c "try=1;until dpkg -D3 -i /var/tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
+      ExecStart=/usr/bin/env bash -c "try=1;until dpkg -D3 -i /var/tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
   - path: /etc/systemd/system/docker.service.d/execstart.conf
     permissions: "0644"
     content: |
@@ -46,17 +46,8 @@ root:
       Type=oneshot
       EnvironmentFile=/etc/environment
       EnvironmentFile=/opt/mesosphere/environment
-      ExecStart=/usr/bin/bash -c "echo $(detect_ip) $(hostname) > /etc/hosts"
+      ExecStart=/usr/bin/env bash -c "echo $(detect_ip) $(hostname) > /etc/hosts"
 runcmd:
-    - [ ln, -s, /bin/rm, /usr/bin/rm ]
-    - [ ln, -s, /bin/mkdir, /usr/bin/mkdir ]
-    - [ ln, -s, /bin/tar, /usr/bin/tar ]
-    - [ ln, -s, /bin/ln, /usr/bin/ln ]
-    - [ ln, -s, /bin/cp, /usr/bin/cp ]
-    - [ ln, -s, /bin/systemctl, /usr/bin/systemctl ]
-    - [ ln, -s, /bin/mount, /usr/bin/mount ]
-    - [ ln, -s, /bin/bash, /usr/bin/bash ]
-    - [ ln, -s, /usr/sbin/useradd, /usr/bin/useradd ]
     - [ systemctl, disable, --now, resolvconf.service ]
     - [ systemctl, mask, --now, lxc-net.service ]
     - [ systemctl, disable, --now, unscd.service ]
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -1425,7 +1425,7 @@ package:
     content: |
       {
         "PATH": "/usr/bin:/bin",
-        "SHELL": "/usr/bin/bash",
+        "SHELL": "/bin/bash",
         "LIBPROCESS_NUM_WORKER_THREADS": "8"
       }
   - path: /etc/dcos-log.env
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -612,8 +612,12 @@ class UserManagement:
                                   "automatic user addition is disabled".format(username))
 
         # Add the user:
+        useradd_cmd = 'useradd'
+        if os.path.exists('/usr/sbin/useradd'):
+            useradd_cmd = '/usr/sbin/useradd'
+
         add_user_cmd = [
-            'useradd',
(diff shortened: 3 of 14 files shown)


gen/dcos-config.yaml
```diff
@@ -1425,7 +1425,7 @@ package:
     content: |
       {
         "PATH": "/usr/bin:/bin",
-        "SHELL": "/usr/bin/bash",
+        "SHELL": "/bin/bash",
```
jongiddy: Why are you confident `bash` will be available in `/bin` here, when in `gen/azure/cloud-config.yaml` you have used `/usr/bin/env bash`?

gen/azure/cloud-config.yaml
```diff
@@ -11,7 +11,7 @@ root:
       StandardOutput=journal+console
       StandardError=journal+console
       ExecStartPre=/usr/bin/curl -fLsSv --retry 20 -Y 100000 -y 60 -o /var/tmp/d.deb https://download.docker.com/linux/ubuntu/dists/xenial/pool/stable/amd64/docker-ce_17.09.0~ce-0~ubuntu_amd64.deb
-      ExecStart=/usr/bin/bash -c "try=1;until dpkg -D3 -i /var/tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
+      ExecStart=/usr/bin/env bash -c "try=1;until dpkg -D3 -i /var/tmp/d.deb || ((try>9));do echo retry $((try++));sleep $((try*try));done;systemctl --now start docker;systemctl restart docker.socket"
```
jongiddy: Since you use relative paths for the other `ExecStart` lines, why use a specific path here for `env`?

jgehrcke:
Tanks @wolf31o2! I have asked around internally towards getting this some reviews. Sorry for the delay so far.

gen/azure/cloud-config.yaml
wolf31o2: `env` isn't necessary, at all... I'll give this another pass later today, as I realize there's no reason to touch the provider-specific configs such as this one

wolf31o2:
I'm going to give this another pass. The intent was to generalize the Ubuntu changes in the Azure cloud config so DC/OS can run on stock Ubuntu, without symlinking everything.

orsenthil:
Hi @wolf31o2  - Thanks for your contribution.

I noticed failures like this:

> 2019-04-30 14:39:18 ERROR    dcos_e2e._common | Failed to start dcos-link-env.service: Unit is not loaded properly: Invalid argument.

It means that the changes being introduced are adversely affecting the start of systemd units.

We also have note that, our default testing environment is Cent OS 7.4, and we also support CoreOS. So, changes will have to be compatible with CentOs and CoreOS. 

Thanks, again!

orsenthil:
Please ping us for reviews when you are ready!

surdy:
My understanding of Systemd is that it need absolute paths for the binaries. Even if it is for binaries that would be standard linux binaries found in PATH of the user shell. 
Perhaps my understanding is outdated.  @wolf31o2 did you test the changes you made? Last time I tiried to do this systemd failed to recognize the unit as  valid unless I used absolute path in ExecStart[*] lines

surdy:
Ohh looks like SystemD 239 removed that requirement : https://github.com/systemd/systemd/blob/master/NEWS#L1085
But RHEL/Centos 7.x are stuck on systemd 219 and Container Linux has had exactly one stable release with systemd > 239. We have to support them for the near future based on https://docs.mesosphere.com/version-policy/#dcos-platform-version-compatibility-matrix

Based on this I'm going to decline the change. Hopefully in a brighter future when EL (and all supported distros for DC/OS) has moved to newer Systemd. I would be happy to approve a PR that simplifies systemd units.

pkgpanda/__init__.py
```diff
@@ -612,8 +612,12 @@ def add_user(self, username, groupname):
                                   "automatic user addition is disabled".format(username))
 
         # Add the user:
+        useradd_cmd = 'useradd'
+        if os.path.exists('/usr/sbin/useradd'):
+            useradd_cmd = '/usr/sbin/useradd'
+
         add_user_cmd = [
-            'useradd',
+            useradd_cmd,
```
branden: Why is this change necessary? Is there a competing `useradd` earlier in the `PATH` that we need to avoid?

surdy:
Great point @branden `/bin/bash -c '<cmd>'` might be a great workaround for what we are doing with those binaries.

gen/dcos-config.yaml
wolf31o2: I prefer to use `/usr/bin/env bash` in cross-distribution scripts. However, you can't use that for a shell.

```
# CentOS defaults to /usr/bin/bash by PATH
cgianelloni@chris:~/Repos/mine/dcos$ docker run --rm -ti centos:7 bash -c 'type -P bash'
/usr/bin/bash
# However, it's also available as /bin/bash
cgianelloni@chris:~/Repos/mine/dcos$ docker run --rm -ti centos:7 bash -c 'ls -l /bin/bash'
-rwxr-xr-x 1 root root 964608 Oct 30  2018 /bin/bash
# And it's /bin/bash on Ubuntu
cgianelloni@chris:~/Repos/mine/dcos$ docker run --rm -ti ubuntu:18.04 bash -c 'type -P bash'
/bin/bash
```

pkgpanda/__init__.py
wolf31o2: I'll add a comment. This is because the PATH used by pkgpanda doesn't include `/usr/sbin` which is the location on Ubuntu. On Azure, it's symlinked into `/usr/bin`, but I was attempting to get this working on stock Ubuntu without customization.

wolf31o2:
I'm going to use `/bin/bash -c` where appropriate, as `/bin/bash` exists on CentOS, CoreOS, and Ubuntu.

wolf31o2:
It looks like either `/bin/bash -c <command>` or `/usr/bin/env <command>` would be appropriate. Any preference?

```
$ docker run --rm -ti centos:7 bash -c 'ls -l /usr/bin/env'
-rwxr-xr-x 1 root root 29048 Oct 30  2018 /usr/bin/env
$ docker run --rm -ti ubuntu:18.04 bash -c 'ls -l /usr/bin/env'
-rwxr-xr-x 1 root root 35000 Jan 18  2018 /usr/bin/env
```

surdy:
@wolf31o2 I have not used either of them in a real world scenario, so I am not aware of limitation/advantages of one over other. However since these are all simple commands (mkdir, ping, sleep etc) I could not think of any corner cases.
@branden do you have any preferences ?

PR #5334, opened 2019-04-26
ZooKeeper backup/restore script E2E tests

timaa2k:
## High-level description

This PR adds two E2E tests for our ZooKeeper backup/restore procedures.
The actual backup/restore test is split into two because we want to test are two different scenarios:
1. Restore from a ZooKeeper transaction log 
2. Restore from a ZooKeeper snapshot

Normally one would like to test 1. with a snapshot+transaction log but success is difficult to assert in this case. Splitting it into to different test cases gives us confidence that a transaction log is replayed from our backup (1.) and that we could restore from snapshot directly, which limits the fallback of a backup to the `snapCount`. That's not a perfect test but it's reasonably close.

The scripts have been tested manually to work on CCM (AWS) in a non-root environment as well as on an Enterprise strict mode cluster.

  - [DCOS-53076](https://jira.mesosphere.com/browse/DCOS-53076) Automate ZooKeeper backup and restoration testing with an E2E test.

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: It's just a test.

### DIFF
--- a/test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
+++ b/test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+SSH_USER='root'
+REMOTE_TMP_DIR=/tmp
+
+usage() {
+  echo "Usage: $0 [-h] [-l SSH_USER] [-i SSH_KEY_PATH] [-t REMOTE_TMP_DIR] ZK_BACKUP_DEST_DIR MASTER_HOSTNAME" 1>&2
+}
+exit_abnormal() {
+  usage
+  exit 1
+}
+while getopts ":hl:i:t:" opt; do
+  case $opt in
+    h)
+      usage
+      exit 0
+    ;;
+    l) SSH_USER=$OPTARG
+    ;;
+    i) SSH_KEY_PATH=$OPTARG
+    ;;
+    t) REMOTE_TMP_DIR=$OPTARG
+    ;;
+    \?) echo "Error: Invalid option -$OPTARG" 1>&2
+        exit_abnormal
+    ;;
+    :) echo "Error: -$OPTARG requires an argument." 1>&2
+       exit_abnormal
+    ;;
+    *) exit_abnormal
+    ;;
+  esac
+done
+
+shift $((OPTIND - 1))
+if (($# != 2)); then
+  exit_abnormal
+fi
+
+DESTINATION_DIR=$1
+if [ ! -d "$DESTINATION_DIR" ]; then
+  echo "Error: $DESTINATION_DIR does not exist." 1>&2
+  exit 1
+fi
+
+MASTER_HOST=$2
+REMOTE_BACKUP_DIR=${REMOTE_TMP_DIR}/zk_backup.$$
+KEY=${SSH_KEY_PATH:+"-i $SSH_KEY_PATH"}
+OPTS='-o ConnectTimeout=3 -o ConnectionAttempts=1
+-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
+
+ssh $OPTS $KEY -tt ${SSH_USER}@${MASTER_HOST} << EOF
+  set -euxo pipefail
+  if [ -d "$REMOTE_BACKUP_DIR" ]; then
+    exit 1
+  fi
+  sudo mkdir -p $REMOTE_BACKUP_DIR
+  sudo systemctl stop dcos-exhibitor
+  sudo cp -pr /var/lib/dcos/exhibitor/zookeeper ${REMOTE_BACKUP_DIR}/zookeeper
+  sudo systemctl start dcos-exhibitor
+  sudo tar --exclude 'myid' --exclude 'zookeeper.out' -pcvzf \\
+  ${REMOTE_BACKUP_DIR}/zk_backup-'$(date +%Y-%m-%d_%H-%M-%S)'.tar.gz -C $REMOTE_BACKUP_DIR ./zookeeper
+  sudo rm -rf ${REMOTE_BACKUP_DIR}/zookeeper
+  exit 0
+EOF
+
+scp $OPTS $KEY -r ${SSH_USER}@${MASTER_HOST}:${REMOTE_BACKUP_DIR}/zk_backup-'*'.tar.gz ${DESTINATION_DIR}/ 2> /dev/null
+
+ssh $OPTS $KEY -tt ${SSH_USER}@${MASTER_HOST} << EOF
+  set -euxo pipefail
+  sudo rm -rf $REMOTE_BACKUP_DIR
+  exit 0
+EOF
--- a/test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
+++ b/test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
@@ -52,11 +52,11 @@ OPTS='-o ConnectTimeout=3 -o ConnectionAttempts=1
 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
 
 ssh $OPTS $KEY -tt ${SSH_USER}@${MASTER_HOST} << EOF
-  set -euxo pipefail
+  set -exo pipefail
   if [ -d "$REMOTE_BACKUP_DIR" ]; then
     exit 1
   fi
-  sudo mkdir -p $REMOTE_BACKUP_DIR
+  mkdir -p $REMOTE_BACKUP_DIR
   sudo systemctl stop dcos-exhibitor
   sudo cp -pr /var/lib/dcos/exhibitor/zookeeper ${REMOTE_BACKUP_DIR}/zookeeper
   sudo systemctl start dcos-exhibitor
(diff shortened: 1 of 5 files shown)


timaa2k:
@meosphere-mergebot label Ready For Review

timaa2k:
PTAL @jongiddy @jgehrcke @mhrabovcin @adamtheturtle @ahairape

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,394 @@
...
+  exit 0
+EOF
+scp $OPTS $KEY $BACKUP_PATH $SSH_USER@${i}:${REMOTE_TMP_DIR}/zk_backup.tar.gz 2> /dev/null
+done
+
+for i in "$@"
+do
+ssh $OPTS -tt $KEY ${SSH_USER}@${i} << EOF
+  set -e
+  sudo rm -rf /var/lib/dcos/exhibitor/zookeeper
```
timaa2k: Yes, but that undo is typically worth 18GB of data :)

timaa2k: Also I haven't found a simple way to make the script idempotent then.
`cp` does a `stat` which fails on repeat.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,394 @@
...
+        LOGGER.error(exc.output.decode())
+        raise
+
+
+BACKUP_SCRIPT = textwrap.dedent(
+    """\
+#!/bin/bash -e
+
+DESTINATION_DIR=$PWD
+REMOTE_TMP_DIR=/tmp/zk_backup
```
timaa2k: :+1: I guess you are aware that all shell expansion is done on the local host, so that pid will be used then.

jongiddy: Yes. You could add the `$$` on the remote scripts, but that would make it difficult to `scp` the archive. So using the local pid is reasonable.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,407 @@
...
+            stderr=subprocess.STDOUT,
+            shell=True,
+        )
+        LOGGER.debug(result.decode())
+    except subprocess.CalledProcessError as exc:
+        LOGGER.error(exc.output.decode())
+        raise
+
+
+BACKUP_SCRIPT = textwrap.dedent(
```
adamtheturtle: ❓ (Optional) - if this is in a separate file that you read here then that means that:

* We have an easier transition for when this is on a remote server
* Folks who edit it can use their own tools (shellcheck!) while editing this.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,407 @@
...
+    The `FLAG` value is found in ZooKeeper at the `znode` path.
+    """
+    zk_client = KazooClient(hosts=hosts, read_only=True)
+    zk_client.start()
+    value = zk_client.get(znode)
+    zk_client.stop()
+    assert value[0] == FLAG
+
+
+def _bash(cmd: List[str]) -> None:
```
adamtheturtle: @mhrabovcin @jongiddy - This would be a great case for the `run_subprocess` helper we want to bring out of E2E!

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,407 @@
...
+        )
+        LOGGER.debug(result.decode())
+    except subprocess.CalledProcessError as exc:
+        LOGGER.error(exc.output.decode())
+        raise
+
+
+BACKUP_SCRIPT = textwrap.dedent(
+    """\
+#!/bin/bash -e
```
adamtheturtle: So this must be run on a user's laptop? Can users be expected to have `bash`? I guess so but just want to make sure!

If so - let's document this, or confirm that it is clear - Windows users don't necessarily have bash.

adamtheturtle:
Is this meant to go in conjunction with the "The following backup procedure is currently considered"  docs in https://jira.mesosphere.com/browse/DCOS-51647?

It'd be good to see what the user side of this is documentation-wise, so we can assess that this is a good representation of that. Bullet point docs-structure is plenty, and if that is already in the JIRA, great.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,407 @@
...
+    znode = '/{}'.format(uuid.uuid4())
+    zk_client = KazooClient(hosts=hosts)
+    zk_client.start()
+    zk_client.create(znode, makepath=True, ephemeral=ephemeral)
+    zk_client.set(znode, FLAG)
+    zk_client.stop()
+    return znode
+
+
+def _zk_check_flag(hosts: str, znode: str) -> None:
```
adamtheturtle: A key part of this is that it raises `NoNodeError` if the flag is not there.
Instead, what do you think of:

```
def _zk_flag_exists(hosts: str, znode: str, flag: bytes) -> bool:
    ...
```

Then, a test doesn't need the `with` but instead `assert not _zk_flag_exists(...)` and `assert _zk_flag_exists(...)`.

test-e2e/test_zookeeper_backup.py
adamtheturtle: Why bash and not Python?

adamtheturtle: How about `#!/usr/bin/env bash`?

timaa2k: Because I thought it's gonna be a 3-liner ...

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def _bash(cmd: List[str]) -> None:
+    """
+    Execute `cmd` in a subprocess, log potential errors.
+    """
+    LOGGER.debug('BASH ARGS: {}'.format(' '.join(cmd)))
+    try:
+        result = subprocess.check_output(
+            args=' '.join(cmd),
```
jongiddy: For safety, you should `shlex.quote` each argument added to the command. e.g. `' '.join(shlex.quote(arg) for arg in cmd)`.

test-e2e/test_zookeeper_backup.py
jongiddy: There's some risk of a collision here. Either two backups running concurrently could share this tmp directory, or a failed command could leave a tmp directory for another run to find. 

Redefine `REMOTE_TMP_DIR` to be a shared temp directory (e.g. `/tmp` but configurable by user), and then create a `REMOTE_BACKUP_DIR=${REMOTE_TMP_DIR}/zkbackup.$$`. This incorporates the local pid and will have a better chance to be unique. The ssh script then creates the `REMOTE_BACKUP_DIR` and puts everything inside it.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,394 @@
...
+fi
+
+CONN=$1
+KEY=${SSH_KEY_PATH:+"-i $SSH_KEY_PATH"}
+OPTS='-o ConnectTimeout=3 -o ConnectionAttempts=1
+-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
+
+ssh $OPTS $KEY -tt $CONN << EOF
+  set -e
+  sudo mkdir -p $REMOTE_TMP_DIR
```
jongiddy: If this already exists, then fail. (Existing will be less likely if the name has been changed to include a pid).

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,394 @@
...
+exit 0
+    """
+)
+
+
+RESTORE_SCRIPT = textwrap.dedent(
+    """\
+#!/bin/bash -e
+
+SSH_USER='root'
```
jongiddy: Why are the user@host ssh details configured differently to backup? Can we make them more symmetric?

test-e2e/test_zookeeper_backup.py
jongiddy: Maybe move this directory to `/var/lib/dcos/exhibitor/zookeeper.old` to provide some form of undo?

test-e2e/test_zookeeper_backup.py
timaa2k: :+1:

test-e2e/test_zookeeper_backup.py
timaa2k: :+1:

test-e2e/test_zookeeper_backup.py
adamtheturtle: Ok - I'm ok with either but let's consider what we should do if we were starting knowing that this is not three lines.

test-e2e/test_zookeeper_backup.py
timaa2k: done now.

test-e2e/test_zookeeper_backup.py
timaa2k: You mean documenting it in the test?

test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
adamtheturtle: Let's discuss tomorrow converting this to Python. That might be Python which does a bunch of work in `subprocess.run`.

test-e2e/fixtures/zk_scripts/dcos-zk-backup.sh
```diff
@@ -0,0 +1,75 @@
...
+
+DESTINATION_DIR=$1
+if [ ! -d "$DESTINATION_DIR" ]; then
+  echo "Error: $DESTINATION_DIR does not exist." 1>&2
+  exit 1
+fi
+
+MASTER_HOST=$2
+REMOTE_BACKUP_DIR=${REMOTE_TMP_DIR}/zk_backup.$$
+KEY=${SSH_KEY_PATH:+"-i $SSH_KEY_PATH"}
```
adamtheturtle: As per our discussion (but note for others): We cannot guarantee that every master will use the same SSH key.

timaa2k:
Superseded by #5575

PR #5426, opened 2019-05-06
DCOS-53467 - test_mom_installation fails on Permissive Mode Cluster

orsenthil:
## High-level description

* DCOS-53467 - test_mom_installation fails on Permissive Mode Cluster 

Skip the test_mom_installation as the test hangs on the permissive mode cluster.


orsenthil:

PR #5496, opened 2019-05-13
Set the 1.11 branch UI version to 1.11.11-dev.

orsenthil:
## High-level description

* Set the 1.11 branch UI version to 1.11.11-dev

* DCOS-53938 - Set UI versions consistently across all in-development branches

### DIFF
--- a/dcos_installer/test_backend.py
+++ b/dcos_installer/test_backend.py
@@ -83,7 +83,7 @@ def test_version(monkeypatch):
     monkeypatch.setenv('BOOTSTRAP_VARIANT', 'some-variant')
     version_data = subprocess.check_output(['dcos_installer', '--version']).decode()
     assert json.loads(version_data) == {
-        'version': '1.11-dev',
+        'version': '1.11.11-dev',
         'variant': 'some-variant'
     }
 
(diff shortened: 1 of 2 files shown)


orsenthil:

orsenthil:
> But 1.11-dev and the likes do not really fit that picture.

I get this. I iterated "unofficially" with various formats, before deciding on making the `major.minor` branches as `major.minor-dev` version so that we will have minimal periodic work and sometimes don't have to do at-the-wish, "version updates" in major branches like `major.minor`, and will have same consistent version for "in-development" branches like 1.11, and 1.12 and 1.13 and master. 

When it becomes a release branch of the format `major.minor.release`, it becomes the only isolated change to do, and doing it only once without being limited by other factors like full CI cycle, or getting blocked by other changes, which will be the situation, if you want to do it in major branches like `x.y`

> In the UI I absolutely would like to see whether I am on 1.13.0-dev vs. 1.13.2-dev. It's a quite relevant difference.

Agree again. I wasn't seeing people concerned about UI versions for the clusters they tried with in-development version of the DC/OS.

> I think while we are here we should do "the right thing". We don't really use the two-digit notation elsewhere so it does not seem to be right.

I agree. I will change it. I will have think about another solution (/automation) for quick periodic updates in the major branch now.

orsenthil:
@jgehrcke , please take a look again.

PR #5561, opened 2019-05-21
[master] Ensure dcos-checks-poststart waits for spartan to be available

drewvanstone:
## High-level description

This PR updates the dcos-checks-poststart.service to check to make sure the spartan service is running. Spartan takes care of setting up DNS, so if it hasn't started yet, other services depending on DNS can fail. 

I've made the change to conform with how this check on spartan is done in other parts of the system, but I do think there is opportunity for improvement by using the `After=` directive. If our systemd units properly use the `After=`, `Before=` and `Wants=` directives, we can generate a proper dependency graph.

I think our current approach is an attempt at spinning up systems more quickly, since all services start at the same time, and continuously reboot until they are successful or exceed the threshold and ultimately fail. 

But if we properly configure our systemd dependency graph, I think startup be faster, and we should see an improvement in stability.

  - [DCOS-47608](https://jira.mesosphere.com/browse/DCOS-47608): Error executing DC/OS components health check: unable to create HTTP client: dial tcp: lookup master.mesos on [::1]:53: dial udp [::1]:53: connect: network is unreachable

### DIFF
--- a/packages/dcos-check-runner/extra/dcos-checks-poststart.service
+++ b/packages/dcos-check-runner/extra/dcos-checks-poststart.service
@@ -1,9 +1,15 @@
 [Unit]
 Description=DC/OS Poststart Checks: Run node-poststart checks
+
 [Service]
+Restart=always
+RestartSec=5
+StartLimitInterval=0
 EnvironmentFile=/opt/mesosphere/environment
 EnvironmentFile=/opt/mesosphere/etc/dcos-check-runner.env
 StandardOutput=journal
 User=root
 Group=root
+# Ensure dns is ready
+ExecStartPre=/bin/ping -c1 ready.spartan
 ExecStart=/opt/mesosphere/bin/dcos-check-runner check node-poststart --config ${DCOS_CHECK_RUNNER_CONFIG_PATH}


drewvanstone:

packages/dcos-check-runner/extra/dcos-checks-poststart.service
```diff
@@ -1,9 +1,15 @@
 [Unit]
 Description=DC/OS Poststart Checks: Run node-poststart checks
+
 [Service]
+Restart=always
+RestartSec=5
+StartLimitInterval=0
```
jgehrcke: Do we know how these settings interact with the corresponding timer settings?

```
[Unit]
Description=DC/OS Checks Timer: timer for DC/OS Checks service
[Timer]
OnCalendar=minutely
AccuracySec=10
```

jgehrcke:
@branden can you also have a look, please?

branden:
his is effectively running a check before we run checks, which I don't understand. Shouldn't this be implemented as another check that the check runner executes, rather than an `ExecStartPre`? As a bonus, checks would assert that Spartan is available in the other contexts in which they run.

packages/dcos-check-runner/extra/dcos-checks-poststart.service
branden: We don't want any of these settings. This unit doesn't provide a long-running service, so there's never a need to restart.

makkes:
> This is effectively running a check before we run checks, which I don't understand. Shouldn't this be implemented as another check that the check runner executes, rather than an `ExecStartPre`? As a bonus, checks would assert that Spartan is available in the other contexts in which they run, outside of just this systemd unit.

@branden Thanks for your feedback. Please have a look at the ticket https://jira.mesosphere.com/browse/DCOS-47608 for context. We are running into problems running a check (`components_agent`) at all because the (invariant!) prerequisites ("network is available" in this case) aren't met. `dcos-check-runner` doesn't support specifying an order of checks execution so that we could make sure the prerequisite is met before running any other checks nor does `dcos-check-runner` support specifying retries in case a check fails temporarily. So making sure spartan is up BEFORE running `dcos-check-runner` is the simplest and IMHO a viable solution.

packages/dcos-check-runner/extra/dcos-checks-poststart.service
```diff
@@ -1,9 +1,15 @@
 [Unit]
 Description=DC/OS Poststart Checks: Run node-poststart checks
+
 [Service]
+Restart=always
```
makkes: `Restart=on-failure` in conjunction with `RestartSec=1` and `StartLimitBurst=5` will do what we need here in that it runs the poststart checks and when they fail restarts them 5 times at one-second intervals.

jgehrcke:
> I do think there is opportunity for improvement by using the After= directive. If our systemd units properly use the After=, Before= and Wants= directives, we can generate a proper dependency graph. [...] But if we properly configure our systemd dependency graph, I think startup be faster, and we should see an improvement in stability.

After experimenting with this one finds that systemd does not provide strong guarantees around the "dependency graph", and the reasoning about the state of the system would become really difficult. Stability would after all suffer, probably rather heavily, if we were to deviate from the **simple** crashloop approach where dependencies are tested for directly and explicitly.

We should definitely value fault tolerance much higher than "faster bootstrap".

We have little documentation around DC/OS internals, but this part is covered here, at least the gist of it: https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md, quote:

> Before=, After= are discouraged. They are not strong guarantees, software needs to check that pre-requisites are up and working correctly

drewvanstone:

packages/dcos-check-runner/extra/dcos-checks-poststart.service
branden: It's unclear to me why we want that behavior. If the intention is to quickly re-run checks in the event of a flaky check, our effort is probably better spent on fixing the flake. If we want to run checks more frequently, we should just reduce the interval between check runs. This unit is often an early stop when debugging a cluster, so it's valuable that it can be described simply: it runs checks once every minute.

PR #5575, opened 2019-05-23
Automated DC/OS ZooKeeper backup and restore 

timaa2k:
## High-level description

The backup strategy was shifted to a backup/restore script shipped with DC/OS that will do a node local backup/restore on the particular master node that it is executed on.

The backup/restore script is now ported to Python3 and comprises a single file + CLI interface with commands backup and restore.

This PR adds two E2E tests for our ZooKeeper backup/restore procedures.
The actual backup/restore test is split into two because we want to test are two different scenarios:
* Restore from a ZooKeeper transaction log
* Restore from a ZooKeeper snapshot
Normally one would like to test `1.` with a snapshot+transaction log but success is difficult to assert in this case. Splitting it into to different test cases gives us confidence that a transaction log is replayed from our backup (`1.`) and that we could restore from snapshot directly, which limits the fallback of a backup to the `snapCount`. It's not a perfect test but it's reasonably close.

  - [DCOS_OSS-5186](https://jira.mesosphere.com/browse/DCOS_OSS-5186) Automate ZooKeeper backup and restoration testing with an E2E test.


timaa2k:

timaa2k:
Ready for Review now, all the remaining failures are due to infra issues.
PTAL @jongiddy @mhrabovcin @jr0d

CHANGES.md
```diff
@@ -6,6 +6,7 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 ### What's new
 
+* ZooKeeper instances on master nodes can now be backed up and restored via a dedicated command line script `dcos-zk` that is shipped with DC/OS.
```
mhrabovcin: Nit - add a new line

https://github.com/dcos/dcos/wiki/CHANGES.md-guidelines#format

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+    See https://jira.mesosphere.com/browse/DCOS_OSS-5185
+    for changing this procedure to allow a backup without downtime.
+
+    If backing up ZooKeeper via a custom temporary directory, a failing backup procedure does not
+    clean up files created in the temporary directory in the process.
+    """
+    zookeeper_dir = EXHIBITOR_DIR / 'zookeeper'
+    tmp_zookeeper_dir = tmp_dir / 'zookeeper'
+
+    print('Backing up ZooKeeper into {backup} via {tmp_zookeeper_dir}'.format(
```
mhrabovcin: I would replace `print` statements with `logging` module use.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+    tmp_dir: Path,
+    verbose: bool,
+) -> None:
+    """
+    DC/OS ZooKeeper instance backup procedure.
+
+    ZooKeeper changes files while it is running.
+    In order to have a consistent backup stop ZooKeeper, run the backup procedure, then start ZooKeeper.
+
+    Running this procedure will temporarily bring down the ZooKeeper instance on this master node.
```
mhrabovcin: This comment seems outdated. Running this function will not temporarily bring down the ZK instance.

I would prefer if we had a check (short circuit) that would check if the ZK instance is running and if is it would exit with non-0 exit code.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+
+    print('Deleting temporary files in {tmp_zookeeper_dir}'.format(
+        tmp_zookeeper_dir=tmp_zookeeper_dir,
+    ))
+    shutil.rmtree(path=str(tmp_zookeeper_dir), ignore_errors=True)
+
+    print('Local ZooKeeper instance restored successfully')
+
+
+def _non_existing_file_path_existing_parent_dir(value) -> Path:
```
mhrabovcin: Missing type hint for `value` argument.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+            'command',
+            type=str,
+            choices=[
+                'backup',
+                'restore',
+            ],
+            help="CLI commands available",
+        )
+        args = parser.parse_args(sys.argv[1:2])
+        getattr(self, args.command)()
```
mhrabovcin: Argument parser library supports subcommands - https://docs.python.org/3/library/argparse.html#sub-commands

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,278 @@
...
+                request=request,
+                log_dir=log_dir,
+            )
+            master = next(iter(cluster.masters))
+            master_ips = [str(m.public_ip_address) for m in cluster.masters]
+            zk_hostports = ','.join([m + ':2181' for m in master_ips])
+
+            # Write to ZooKeeper before backup
+            persistent_flag = _zk_set_flag(zk_hostports)
+            ephemeral_flag = _zk_set_flag(zk_hostports, ephemeral=True)
```
mhrabovcin: You can remove `ephemeral_flag` test.

> ZooKeeper also has the notion of ephemeral nodes. These znodes exists as long as the session that created the znode is active. When the session ends the znode is deleted. Ephemeral nodes are useful when you want to implement [tbd].

With the current implementation of `_zk_set_flag` the ephemeral node is removed from ZK at the moment when you stop zk client.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,278 @@
...
+            # Assert that DC/OS is intact.
+            wait_for_dcos_oss(
+                cluster=cluster,
+                request=request,
+                log_dir=log_dir,
+            )
+
+            # Read from ZooKeeper after restore
+            assert _zk_flag_exists(zk_hostports, persistent_flag)
+            assert not _zk_flag_exists(zk_hostports, ephemeral_flag)
```
mhrabovcin: Remove this assertion.

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,278 @@
...
+                log_dir=log_dir,
+            )
+
+            # Begin the backup/restore test procedure
+            master_ips = [str(m.public_ip_address) for m in masters]
+            zk_hostports = ','.join([m + ':2181' for m in master_ips])
+
+            # Write to ZooKeeper multiple times before backup
+            persistent_flag = _zk_set_flag(zk_hostports)
+            ephemeral_flag = _zk_set_flag(zk_hostports, ephemeral=True)
```
mhrabovcin: Remove ephemeral flag test.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def _non_existing_file_path_existing_parent_dir(value) -> Path:
+    """
+    Validate that the value is a path to a file which does not exist
+    but the parent directory tree exists.
+    """
+    path = Path(value)
+    if path.exists():
+        raise ArgumentTypeError('Error: {} already exists'.format(path))
```
mhrabovcin: Do these messages need `Error:` prefix? How does the final error message presented to a user look like?

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+    if not Path(path.parent).exists():
+        raise ArgumentTypeError(
+            'Error: {} parent directory does not exist'.format(path),
+        )
+    return path.absolute()
+
+
+def _existing_file_path(value) -> Path:
```
mhrabovcin: Missing type hint for `value` argument.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+    """
+    path = Path(value)
+    if not path.exists():
+        raise ArgumentTypeError('Error: {} does not exist'.format(path))
+    if not path.is_file():
+        raise ArgumentTypeError('Error: {} is not a file'.format(path))
+    return path.absolute()
+
+
+def _existing_dir_path(value) -> Path:
```
mhrabovcin: Missing type hint for `value` argument.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,322 @@
...
+
+if __name__ == "__main__":
+    try:
+        DCOSZooKeeperCli()
+    except subprocess.CalledProcessError as exc:
+        if exc.output:
+            sys.stdout.buffer.write(exc.output)
+        sys.exit(exc.returncode)
+
+# vi: ft=python
```
mhrabovcin: Is this line necessary?

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: Good point. I consulted the Python3 documentation and then decided to use `print` instead of logging: https://docs.python.org/3/howto/logging.html#when-to-use-logging
It seems reasonable to me to distinguish the two, `print` as a CLI user interface, logging as a debugging help for developers.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: Thanks, missed that.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: I left this out on purpose because it's tricky to figure out.
I like mypy while it does not get in the way but this one is just pain.

test-e2e/test_zookeeper_backup.py
timaa2k: Unfortunately the ephemeral node is in the backup but cannot be restored because of course we need to restart all ZooKeeper instances. It was not clear to me what would happen to a client in that case. The test was mere a confirmation that ephemeral nodes are not preserved and I would like to keep it to make the behavior clear.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: Good catch, it is no longer necessary because `dcos_zk_backup.py` is renamed in the `build` script so that `tox` recognizes it as a Python script and runs linting on it. Before it was `dcos_zk_backup` without `.py`, then this line was essential to me :)

CHANGES.md
timaa2k: Haha, I removed the new-line in a separate commit because I noticed the over entries didn't have one :D

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: I've read up and tried `argparse` `subcommands` and I think the current way is a superior solution, especially if we ever want add more commands.
After manually digging into a transaction log I can totally see a `dcos-zk open logfile` command that we should implement.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: `error: argument backup_path: Error: sessions already exists`
Removing `Error` then :+1:

tox.ini
```diff
@@ -166,6 +166,7 @@ deps=
   mypy-mypyc==0.660
 commands=
   mypy ./test-e2e
+  mypy ./packages/exhibitor/extra/dcos_zk_backup.py
```
adamtheturtle: ❤️ Thank you @timaa2k

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,350 @@
...
+            'command',
+            type=str,
+            choices=[
+                'backup',
+                'restore',
+            ],
+            help='CLI commands available',
+        )
+        args = parser.parse_args(sys.argv[1:2])
+        getattr(self, args.command)()
```
adamtheturtle: @timaa2k We discussed using `click` but concluded that you would rather not add a dependency for this script. However, I see we already have `click` as an [indirect dependency](https://github.com/dcos/dcos/blob/1c8088873cb8cbe5abb514deea35375f2a08b313/packages/flask/buildinfo.json#L9) and we import it directly:

https://github.com/dcos/dcos/blob/c71cf838783ee7f69dfb8d51cda01f5f03c6b025/packages/dcos-integration-test/extra/get_test_group.py#L16

Can we use it here?

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,350 @@
...
+    zookeeper_dir = EXHIBITOR_DIR / 'zookeeper'
+    tmp_zookeeper_dir = tmp_dir / 'zookeeper'
+
+    print('Backing up ZooKeeper into {backup} via {tmp_zookeeper_dir}'.format(
+        backup=backup,
+        tmp_zookeeper_dir=tmp_zookeeper_dir,
+    ))
+
+    print('Validate that ZooKeeper is not running')
+    assert not _is_zookeeper_running(verbose)
```
adamtheturtle: `AssertionError` is not a good error for a user to see. Instead, let's do:

```
if not _is_zookeeper_running(verbose):
   print('...') # or write to stderr?
   sys.exit(1)
```

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,350 @@
...
+
+    _copy_dir_and_preserve_ownership(
+        src=zookeeper_dir,
+        dst=tmp_zookeeper_dir,
+        verbose=verbose,
+    )
+
+    print('Creating ZooKeeper backup tar archive at {backup}'.format(backup=backup))
+
+    def _tar_filter(tar_info: tarfile.TarInfo) -> Optional[tarfile.TarInfo]:
```
adamtheturtle: Could you add a comment explaining why the files which match these names are not welcome in the backup?

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,350 @@
...
+    # We copy files rather than move them so that if this script is interrupted
+    # while running, no data is lost inplace.
+    # However, this has a downside that we may use a lot of disk space -
+    # we have seen ZooKeeper files up to 20 GB in size.
+    _copy_dir_and_preserve_ownership(
+        src=zookeeper_dir,
+        dst=tmp_zookeeper_dir,
+        verbose=verbose,
+    )
+    shutil.rmtree(path=str(zookeeper_dir), ignore_errors=True)
```
jongiddy: Why do we want to ignore errors here? Wouldn't a failure to remove the directory tree be a problem?

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,283 @@
...
+        hosts=zk_hostports,
+        # Avoid failure due to client session timeout.
+        timeout=60,
+    )
+    def zk_listener(state: KazooState) -> None:
+        # Open session after temporary connection loss.
+        if state == KazooState.CONNECTED:
+            zk_client.start()
+
+    zk_client.add_listener(zk_listener)
```
jongiddy: Why do we need this? The docs at https://kazoo.readthedocs.io/en/latest/basic_usage.html#connection-handling state that:
```
Once connected, the client will attempt to stay connected regardless of intermittent connection loss or Zookeeper session expiration.
```

test-e2e/test_zookeeper_backup.py
```diff
@@ -0,0 +1,283 @@
...
+        # Assert that DC/OS is intact.
+        wait_for_dcos_oss(
+            cluster=three_master_cluster,
+            request=request,
+            log_dir=log_dir,
+        )
+
+        # Read from ZooKeeper after restore
+        assert _zk_flag_exists(zk_client, persistent_flag)
+        assert _zk_flag_exists(zk_client, ephemeral_flag)
```
jongiddy: How do we know that the restore and `wait_for_dcos_oss` took less than the ephemeral flag timeout?

test-e2e/test_zookeeper_backup.py
timaa2k: `Connected` doesn't seem to imply able to do `get()` operations. I'm still in the progress of figuring this out.

timaa2k: Actually, with a long enough connection timeout we don't seem to need this.

test-e2e/test_zookeeper_backup.py
timaa2k: What do you mean by the `ephemeral flag timeout`?
In my understanding there are two timeouts active here:
* Client connection timeout (60 seconds, set above). This needs to be longer than the restore procedure (we can set it arbitrarily)
* ZooKeeper session timeout -> resets after ZooKeeper comes back alive. Default is 10 seconds.
The client would need to reconnect within 10 seconds after ZooKeeper comes up again.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: Good catch, that was left over from the `rmtree` below.

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: I should merge as is once the tests pass! Let's do it once we add another command.

test-e2e/test_zookeeper_backup.py
timaa2k: As it appears this was the culprit! Thanks Jon for raising this in the first place :100:

timaa2k:
I've added excessive retrying to get rid of flakiness. The snapshot test now passed `4/4` times locally for me now.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,358 @@
...
+        check=True,
+    )
+
+
+def _is_zookeeper_running(verbose: bool) -> bool:
+    """
+    Exhibitor and therefore ZooKeeper is in active state (running).
+    """
+    try:
+        run_command('systemctl status dcos-exhibitor', verbose)
```
mhrabovcin: I think this is OK, but there is a dedicated way to tell if service is running

https://unix.stackexchange.com/questions/396630/the-proper-way-to-test-if-a-service-is-running-in-a-script
https://www.freedesktop.org/software/systemd/man/systemctl.html#is-active%20PATTERN%E2%80%A6

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,358 @@
...
+    tmp_zookeeper_dir = tmp_dir / 'zookeeper'
+
+    print('Backing up ZooKeeper into {backup} via {tmp_zookeeper_dir}'.format(
+        backup=backup,
+        tmp_zookeeper_dir=tmp_zookeeper_dir,
+    ))
+
+    print('Validate that ZooKeeper is not running')
+    if _is_zookeeper_running(verbose):
+        print('ZooKeeper must not be running. Aborting.')
```
mhrabovcin: As @adamtheturtle pointed out earlier, should this be printed to `stderr`?

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,358 @@
...
+
+    print('Local ZooKeeper instance restored successfully')
+
+
+def _non_existing_file_path_existing_parent_dir(value: str) -> Path:
+    """
+    Validate that the value is a path to a file which does not exist
+    but the parent directory tree exists.
+    """
+    print(type(value))
```
mhrabovcin: Debug leftover?

timaa2k: Yes, /facepalm

packages/exhibitor/extra/dcos_zk_backup.py
timaa2k: Done

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,357 @@
...
+    tmp_zookeeper_dir = tmp_dir / 'zookeeper'
+
+    print('Backing up ZooKeeper into {backup} via {tmp_zookeeper_dir}'.format(
+        backup=backup,
+        tmp_zookeeper_dir=tmp_zookeeper_dir,
+    ))
+
+    print('Validate that ZooKeeper is not running')
+    if _is_zookeeper_running(verbose):
+        sys.stderr.write('ZooKeeper must not be running. Aborting.\n')
```
jongiddy: :question: Should this refer to the `dcos-exhibitor` unit, to ensure that the customer shuts-down ZK correctly?

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,357 @@
...
+    Minimal CLI to backup/restore the local DC/OS ZooKeeper instance.
+    """
+
+    def __init__(self) -> None:
+        """
+        Present CLI command choices.
+        """
+        parser = ArgumentParser(
+            description=(
+                'Command line utility to more safely interact with the local '
```
jongiddy: I don't think we need the "more safely".

timaa2k:
@jongiddy thanks for the comments.
The `usage: dcos-zk [-h] [-t TMP_DIR] [-v] backup_path` (options before args) pattern is used by basic unix commands like `mv` and `cp` `--help` descriptions. I really like it and would like to keep it as is.
The superfluous messages are not necessary but help when something went wrong :)

CHANGES.md
```diff
@@ -6,6 +6,8 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 ### What's new
 
+* ZooKeeper instances on master nodes can now be backed up and restored via a dedicated command line script `dcos-zk` that is shipped with DC/OS. (DCOS_OSS-5186)
```
jgehrcke: We will add this to 1.13, correct? Then that's not news anymore. Please remove in that case :-).

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,357 @@
...
+                'ZooKeeper instance on DC/OS master nodes'
```
jgehrcke: Also this description should be rewritten to summarize the actual purpose as concisely as possible. It should probably contain the words "backup" and "restore" :-).

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,357 @@
...
+                'File path that the gzipped ZooKeeper backup tar archive will '
+                'be written to'
+            ),
+        )
+        parser.add_argument(
+            '-t', '--tmp-dir',
+            type=_existing_dir_path,
+            help=(
+                'Location of an existing directory to be used as temporary '
+                'directory in the process'
```
jgehrcke: What's temporary about this? Will this script work if I do not provide this? What's the default?

jongiddy:
> The usage: dcos-zk [-h] [-t TMP_DIR] [-v] backup_path (options before args)

To clarify, I didn't mean that the parameter `backup_path` should move to be after `dcos-zk`. I mean that the literal word `backup` should appear after `dcos-zk`.

The usage should say:
`dcos-zk backup [-h] [-t TMP_DIR] [-v] backup_path`

packages/exhibitor/extra/dcos_zk_backup.py
jgehrcke: maybe say something like "use this if the default does not work".

timaa2k:
@jongiddy I cheated a little but that does the trick.

packages/exhibitor/extra/dcos_zk_backup.py
```diff
@@ -0,0 +1,365 @@
...
+                'be written to.'
+            ),
+        )
+        parser.add_argument(
+            '-t', '--tmp-dir',
+            type=_existing_dir_path,
+            help=(
+                'Location of an existing directory to be used as temporary '
+                'directory. A temporary directory will be created if not '
+                'specified.'
```
jgehrcke: thanks for the amendment!

jgehrcke:
@mesosphere-mergebot sync

timaa2k:
@mesosphere-mergebot sync

timaa2k:
@mesosphere-mergebot sync

adamtheturtle:
@mesosphere-mergebot sync

adamtheturtle:
@mesosphere-mergebot sync

PR #5632, opened 2019-05-31
Collect procstat metrics only for components

branden:
## High-level description

This PR removes config that directs Telegraf to monitor resource consumption for all processes running on all cluster nodes. This is wasteful on agents with many Mesos tasks, because task metrics are already collected by other inputs. Instead we provide a whitelist of component systemd units to monitor. This causes Telegraf to monitor the main PID for each component.

I tested this manually by installing dcos-monitoring and querying Prometheus to confirm that procstat metrics are present for all components.

This will be backported to 1.13, so no changelog entry is necessary.

  - [DCOS-53589](https://jira.mesosphere.com/browse/DCOS-53589) Whitelist the processes to monitor when using the procstat input plugin

## Related tickets (optional)

N/A

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -2275,16 +2275,29 @@ package:
       [[inputs.net]]
       # Get the number of processes and group them by status
       [[inputs.processes]]
-      # Gather resource usage of processes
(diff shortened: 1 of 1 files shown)


branden:

PR #5669, opened 2019-06-06
Upgrade CockroachDB to version 2.1.7

timaa2k:
## High-level description

This PR upgrades CockroachDB to version [2.1.7](
https://www.cockroachlabs.com/docs/releases/v2.1.7.html) in DC/OS.

The PR removes `set_cluster_version` which was previously used to `finalize` an upgrade of CockroachDB. With version 2.1.x CockroachDB [auto-finalizes](https://www.cockroachlabs.com/docs/v2.1/upgrade-cockroach-version.html#step-3-decide-how-the-upgrade-will-be-finalized) upgraded nodes by default.

The PR fixes a race-condition in our CockroachDB init procedure where we first start one CockroachDB instance that initializes a new cluster before it get killed to then restart and `join` the cluster it previously created. The problem there appeared to be that the instance got terminated to quickly which led to the cluster not being properly initialized in rare cases.

  - [DCOS_OSS-5262](https://jira.mesosphere.com/browse/DCOS_OSS-5262) [CRDB update] Upgrade to the latest CockroachDB 2.1.x in dcos/dcos.

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain.
  - [x] Change log from the last version integrated (this should be a link to commits for easy verification and review):
    - [2.1.7](https://www.cockroachlabs.com/docs/releases/v2.1.7.html) (Now)
    - [2.1.0](https://www.cockroachlabs.com/docs/releases/v2.1.0.html)
    - [2.0.7](https://www.cockroachlabs.com/docs/releases/v2.0.7.html) (Previous)

### DIFF
--- a/packages/cockroach/buildinfo.json
+++ b/packages/cockroach/buildinfo.json
@@ -1,9 +1,8 @@
 {
-  "requires": ["ncurses"],
   "single_source" : {
       "kind": "url_extract",
-      "url": "https://binaries.cockroachdb.com/cockroach-v2.0.7.src.tgz",
-      "sha1": "8b3940ef9ab46646b4178bc93c9904b57d9f3c50"
+      "url": "https://binaries.cockroachdb.com/cockroach-v2.1.7.linux-amd64.tgz",
+      "sha1": "68021a4779a5df27dfd8a8e72dbd20599018623d"
   },
   "username": "dcos_cockroach",
   "state_directory": true
--- a/packages/cockroach/extra/cockroachdb-change-config.py
+++ b/packages/cockroach/extra/cockroachdb-change-config.py
@@ -20,63 +20,69 @@ log = logging.getLogger(__name__)
 logging.basicConfig(format='[%(levelname)s] %(message)s', level='INFO')
 
 
-def set_cluster_version(my_internal_ip: str, version: str) -> None:
-    """
-    Use `cockroach sql` to set the cluster-wide configuration setting
-    version to `version`. This requires that all nodes are running the
-    given version of CockroachDB and are operating successfully.
-
-    Relevant JIRA ticket: https://jira.mesosphere.com/browse/DCOS-19427
-
-    Args:
-        my_internal_ip: The internal IP of the current host.
-        version: The current 'major.minor' version of CockroachDB.
-    """
-    command = [
-        '/opt/mesosphere/active/cockroach/bin/cockroach',
-        'sql',
-        '--insecure',
-        '--host={}'.format(my_internal_ip),
-        '-e',
-        "SET CLUSTER SETTING version = '{}';".format(version),
-        ]
-    config_text = 'version: {}'.format(version)
-    log.info('Set `%s` via command `%s`', config_text, ' '.join(command))
-    subprocess.run(command, input=config_text.encode('ascii'))
-    log.info('Command returned')
-
-
 def set_num_replicas(my_internal_ip: str, num_replicas: int) -> None:
     """
-    Use `cockroach zone set` to set the cluster-wide configuration setting
-    num_replicas to `num_replicas`. This does not matter on a 3-master
-    DC/OS cluster because the CockroachDB default for num_replicas is 3.
-    This however ensures that num_replicas is set to 5 on a 5-master DC/OS
-    cluster. Feed the configuration key/value pair to the `cockroach` program
-    via stdin.
+    CockroachDB version 2.1.7 sets the internal system range replication
+    factor to `5` by default:
+    https://www.cockroachlabs.com/docs/stable/configure-replication-zones.html#view-all-replication-zones
 
+    In order to adjust this to the maximum replication factor possible for a
+    given DC/OS cluster and to provide maximum fault-tolerance we set the
+    cluster-wide configuration setting `num_replicas` for any database entity
+    to the given `num_replicas` count which equals the number of DC/OS master
+    nodes.
     Relevant JIRA ticket: https://jira.mesosphere.com/browse/DCOS-20352
     """
+    zone_config = 'num_replicas = {}'.format(num_replicas)
 
-    def _set_replicas_for_zone(zone: str) -> None:
-        command = [
-            '/opt/mesosphere/active/cockroach/bin/cockroach',
-            'zone',
-            'set',
-            zone,
-            '--insecure',
-            '--host={}'.format(my_internal_ip),
-            '-f',
-            '-'
-            ]
-        config_text = 'num_replicas: %s' % (num_replicas, )
-        log.info('Set `%s` via command `%s`', config_text, ' '.join(command))
-        subprocess.run(command, input=config_text.encode('ascii'))
+    # If more entities that must have their `num_replicas` setting adjusted are
+    # added to CockroachDB, a DC/OS check will fail on all clusters (DC/OS OSS
+    # 1.13+, DC/OS Enterprise 1.12+).
+    # describing that there are "underreplicated ranges".
+    # The DC/OS check is named "cockroachdb_replication".
+    #
+    # To resolve this, add more items to the following list.
+    # To display CockroachDB entities that must have their `num_replicas` setting
+    # adjusted, issue the following SQL command: `SHOW ALL ZONE CONFIGURATIONS;`
+    #
+    # One option is to parse the output of the above SQL command.
+    # However, there is a plan for CockroachDB that all replication will be
+    # derived from ``.default``.
+    # See https://forum.cockroachlabs.com/t/change-replication-factor/2052/3.
+    # Should this happen, we can replace the following loop initialisation with:
+    # zone = '.default'
+    # db_entity = 'RANGE default'
+    for zone, db_entity in [
+        ('.default', 'RANGE default'),
+        ('system', 'DATABASE system'),
+        ('system.jobs', 'TABLE system.public.jobs'),
+        ('.meta', 'RANGE meta'),
+        ('.system', 'RANGE system'),
+        ('.liveness', 'RANGE liveness'),
+    ]:
+        sql_command = (
+            'ALTER {db_entity} CONFIGURE ZONE USING {zone_config};'
+        ).format(
+            db_entity=db_entity,
+            zone_config=zone_config,
+        )
+        command = (
+            '/opt/mesosphere/active/cockroach/bin/cockroach '
+            'sql -e "{sql_command}" --insecure --host={host}'
+        ).format(
+            sql_command=sql_command,
+            host=my_internal_ip,
+        )
+        message = (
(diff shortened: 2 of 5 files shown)


timaa2k:

packages/cockroach/extra/cockroachdb-change-config.py
```diff
@@ -20,64 +20,70 @@
...
+        )
+        command = (
+            '/opt/mesosphere/active/cockroach/bin/cockroach '
+            'sql -e "{sql_command}" --insecure --host={host}'
+        ).format(
+            sql_command=sql_command,
+            host=my_internal_ip,
+        )
+        message = (
+            'Set {zone_config} for {zone} via command {command}'
```
adamtheturtle: Backticks or quotes around the variables - particularly "command" please.

packages/cockroach/extra/cockroachdb-change-config.py
```diff
@@ -20,64 +20,70 @@
...
+        ).format(
+            zone_config=zone_config,
+            zone=zone,
+            command=command,
+        )
+        log.info(message)
```
adamtheturtle: ❤️

packages/cockroach/extra/cockroachdb-change-config.py
```diff
@@ -20,64 +20,70 @@
...
+        subprocess.run(command, shell=True)
```
adamtheturtle: Let's keep in mind that if we are going to do a lot of work on this we should use Python rather than Bash - https://www.cockroachlabs.com/docs/stable/build-a-python-app-with-cockroachdb.html#basic-statements.

packages/cockroach/extra/cockroachdb-change-config.py
jongiddy: Maybe wrap the format arguments with `repr`?

packages/cockroach/extra/cockroachdb-change-config.py
```diff
@@ -107,9 +113,6 @@ def main() -> None:
 
     set_num_replicas(my_internal_ip, master_node_count)
 
-    # We are running CockroachDB v2.0.x so pass '2.0'.
-    set_cluster_version(my_internal_ip, '2.0')
```
jongiddy: If we no longer need this, should we be explictly unsetting this for 2.1?

timaa2k: No, it will automatically be set to `2.1`

packages/cockroach/buildinfo.json
```diff
@@ -1,9 +1,8 @@
 {
-  "requires": ["ncurses"],
   "single_source" : {
       "kind": "url_extract",
-      "url": "https://binaries.cockroachdb.com/cockroach-v2.0.7.src.tgz",
-      "sha1": "8b3940ef9ab46646b4178bc93c9904b57d9f3c50"
+      "url": "https://binaries.cockroachdb.com/cockroach-v2.1.7.linux-amd64.tgz",
```
mhrabovcin: Is this going to be statically linked against any library that isn't shipped with DC/OS in `/opt/mesosphere`?

timaa2k: Maybe I don't understand the question, but why would it matter?

mhrabovcin: We have to understand what exact dependencies are we expecting to be installed on the host system and document them - https://docs.mesosphere.com/1.13/installing/production/system-requirements/#software-prerequisites

The reason why we ship almost every dependency with DC/OS is that we can control software versions that we use to run DC/OS and test such configuration. If you're deciding to rely on OS provided library dependency it should be clear which versions we're going to test with.

timaa2k:
```
ldd dependencies of CockroachDB 2.0.7 (previous, self-built binary)
	linux-vdso.so.1 =>  (0x00007ffe0f5f1000)
	librt.so.1 => /lib64/librt.so.1 (0x00007fe526549000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fe52632d000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fe52602b000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fe525c5e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe526751000)
```
```
ldd dependencies of CockroachDB 2.1.7 (downloaded binary)
	linux-vdso.so.1 (0x00007ffe39de7000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f30b286a000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f30b264b000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f30b22b7000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f30b1ef8000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f30b2a72000)
```

mhrabovcin: Nice, this can be closed.

packages/cockroach/extra/register.py
```diff
@@ -262,14 +264,12 @@ def _zk_lock(zk: KazooClient, lock_path: str, contender_id: str, timeout: int) -
...
+def _init_cockroachdb_cluster(ip: str) -> None:
     """
-    Starts CockroachDB listening on `ip`. It waits until the PID file
-    appears, signalling that the instance has started successfully
-    and is serving requests. Thereafter it shuts down the bootstrap
-    instance again.
+    Starts CockroachDB listening on `ip`. It waits until the cluster ID is
+    published via the local gossip endpoint, signalling that the instance has
+    successfully initialized the cluster. Thereafter it shuts down the
+    bootstrap CockroachDB instance again.
```
jgehrcke: Both, the old docstring and the new docstring are lifesaving.

jgehrcke: (thanks!)

packages/cockroach/extra/register.py
```diff
@@ -278,21 +278,69 @@ def _init_cockroachdb_cluster(
     """
     @retrying.retry(
         wait_fixed=1000,
-        stop_max_attempt_number=10,
+        stop_max_attempt_number=60,
```
jgehrcke: Why did we change this to 60?

packages/cockroach/extra/register.py
```diff
@@ -278,21 +278,69 @@ def _init_cockroachdb_cluster(
...
         retry_on_result=lambda x: x is False,
-        )
-    def _wait_for_pidfile(pid: int) -> bool:
+    )
+    def _wait_for_cluster_init() -> bool:
         """
-        Try and read the PID from `PID_FILE_PATH` and match it to the given `pid`.
-        If, after 10 seconds, the PIDs don't match, raise an exception.
+        CockroachDB Cluster initialization takes a certain amount of time
+        while the cluster ID and and node ID are written to the storage.
```
jgehrcke: and and

packages/cockroach/extra/register.py
```diff
@@ -278,21 +278,69 @@ def _init_cockroachdb_cluster(
...
+        a big deal and we have never seen this take longer than that.
+
+        If after 60 attempts the cluster ID is not available raise an exception.
         """
+        gossip_url = 'http://localhost:8090/_status/gossip/local'
+        try:
+            response = requests.get(
+                gossip_url,
+                # We expect this request to get a response very quickly
+                # since it is a request to localhost.
```
jgehrcke: The network connectivity is not a strong argument here.

More important is the question whether the code, the HTTP handler, is expected to serve _quickly_ or if it may sometimes be slow (but operational).

Is we ever run into a situation where this request would be reproducibly and accurately served after e.g. 2 seconds then this tight timeout control would result in an unnecessary crash loop. Let's please explain why we know that this cannot happen, otherwise let's set this to something more conservative like 30 seconds where I think it's much more obvious that we want to stop waiting.

packages/cockroach/extra/register.py
```diff
@@ -278,21 +278,69 @@ def _init_cockroachdb_cluster(
...
-    def _wait_for_pidfile(pid: int) -> bool:
+    )
+    def _wait_for_cluster_init() -> bool:
         """
-        Try and read the PID from `PID_FILE_PATH` and match it to the given `pid`.
-        If, after 10 seconds, the PIDs don't match, raise an exception.
+        CockroachDB Cluster initialization takes a certain amount of time
+        while the cluster ID and and node ID are written to the storage.
+
+        We wait 1 second to avoid having a huge amount of logs in the journal.
```
jgehrcke: I don't understand this argument. Maybe write "We wait 1 s instead of ..... to avoid" so that it's clearer what the train of thought was. Notably, I have commented on the 1 second timeout constant below.

packages/cockroach/extra/register.py
```diff
@@ -278,21 +278,69 @@ def _init_cockroachdb_cluster(
...
+            response.raise_for_status()
+        except HTTPError as exc:
+            # 150 bytes was chosen arbitrarily as it might not be so long as to
+            # cause annoyance in a console, but it might be long enough to show
+            # some useful data.
+            first_150_bytes = response.content[:150]
+            decoded_first_150_bytes = first_150_bytes.decode(
+                encoding='ascii',
+                errors='backslashreplace',
+            )
```
jgehrcke: :rocket:  @adamtheturtle  :D

packages/cockroach/extra/register.py
jgehrcke: My thinking...: when this program crashes we hand back control to systemd. This change delays this. Why do we want this? It probably makes sense, but I'd love to read about a reason :-)

timaa2k: If the initialization takes longer than 10 seconds DC/OS would never be able to start up.

packages/cockroach/extra/cockroachdb-change-config.py
```diff
@@ -20,64 +20,70 @@
...
 def set_num_replicas(my_internal_ip: str, num_replicas: int) -> None:
     """
-    Use `cockroach zone set` to set the cluster-wide configuration setting
-    num_replicas to `num_replicas`. This does not matter on a 3-master
-    DC/OS cluster because the CockroachDB default for num_replicas is 3.
-    This however ensures that num_replicas is set to 5 on a 5-master DC/OS
-    cluster. Feed the configuration key/value pair to the `cockroach` program
-    via stdin.
-
+    CockroachDB version 2.1.7 sets the internal system range replication
```
jongiddy: As for EE, this can just be 2.1 - but it's only worth changing if re-run is necessary

adamtheturtle: Oops!

PR #5741, opened 2019-06-19
[1.14 / master] Bump DC/OS E2E to 2019.06.19.0

adamtheturtle:
## High-level description

This change bumps the version of DC/OS E2E used in tests.
The main goal of this is to keep more logs in the event of a `DCOSTimeoutError` - a long running test which times out after an hour of waiting for DC/OS previously kept only ~20 minutes of logs.

In particular, [DC/OS E2E PR #1857](https://github.com/dcos/dcos-e2e/pull/1857) provides a change we want to bring in.

This PR is submitted without evidence that more logs are kept by DC/OS.
One test we could do is add a loop to bootstrap which would mean that the cluster start up time is > 1h, and then we could check the amount of logs kept with or without this change.
We prefer not to do this and instead reason that this change will work, and then come back if we get evidence that it has not worked.
Reviewer feedback on that strategy is welcome!

  - [DCOS-55311](https://jira.mesosphere.com/browse/DCOS-55311) DC/OS tests - E2E test logs are rotated so we are missing logs.

## Related tickets (optional)

  - [DCOS-55257](https://jira.mesosphere.com/browse/DCOS-55257) test-e2e.test_master_node_replacement.test_dynamic_cluster - DCOSTimeoutError.

### DIFF
--- a/test-e2e/test_e2e_module.py
+++ b/test-e2e/test_e2e_module.py
@@ -13,12 +13,12 @@ from dcos_e2e.backends import Docker
 @pytest.fixture(scope='session', autouse=True)
 def configure_logging() -> None:
     """
-    Surpress DEBUG log messages from libraries that log excessive amount of
-    debug output that isn't useful for debugging e2e tests.
+    Surpress INFO, DEBUG and NOTSET log messages from libraries that log
+    excessive amount of debug output that isn't useful for debugging e2e tests.
     """
-    # Disble debug output from `docker` and `urllib3` libraries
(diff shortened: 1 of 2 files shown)


test-e2e/test_e2e_module.py
```diff
@@ -13,12 +13,12 @@
 @pytest.fixture(scope='session', autouse=True)
```
adamtheturtle: While here I also decided to update the comment (review comment from @jgehrcke a while back) and reduce the amount of logs in the build log, as was previously done in DC/OS Enterprise by suppressing some logs from `sarge`.

PR #5743, opened 2019-06-19
Use extra requirements for tox to catch more mypy errors

adamtheturtle:
## High-level description

I spotted while bumping DC/OS E2E in another branch that `tox -e mypy` did not catch as much as it should - in particular an API change in DC/OS E2E (a typed library) itself.
This is because `tox` was not using the requirements of the `test-e2e` tests.
This change makes `tox` more strict by installing those requirements, and it fixes the issues which came up.

  - [DCOS-55311](https://jira.mesosphere.com/browse/DCOS-55311) DC/OS tests - E2E test logs are rotated so we are missing logs

### DIFF
--- a/test-e2e/test_zookeeper_backup.py
+++ b/test-e2e/test_zookeeper_backup.py
@@ -5,7 +5,7 @@ Tests for the ZooKeeper backup/restore via `dcos-zk`.
 import logging
 import uuid
 from pathlib import Path
-from typing import Set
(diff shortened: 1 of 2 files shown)


jgehrcke:
`teamcity/dcos/build/tox` failed so this needs more work I suppose

jgehrcke:

adamtheturtle:
Oops, shouldn't be in review!

PR #5792, opened 2019-07-03
Fix incorrect licensing socket location

drewvanstone:
## High-level description

DCOS-Signal is not getting the license key from deployed clusters due to having the wrong dcos-licensing.sock path. This PR corrects this issue.

  - [DCOS-55882](https://jira.mesosphere.com/browse/DCOS-55882)

  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain.
  - [x] Test Results: [link to CI job test results for component]
  - [x] Code Coverage (if available): [link to code coverage report]

### DIFF
--- a/packages/dcos-signal/extra/dcos-signal.service
+++ b/packages/dcos-signal/extra/dcos-signal.service
@@ -14,4 +14,4 @@ ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-signal
 
 # Signal depends on the dcos-licensing service for the license_id
-ExecStart=/opt/mesosphere/bin/dcos-signal -licensing-socket /run/dcos/dcos-license.sock -segment-key 51ybGTeFEFU1xo6u10XMDrr6kATFyRyh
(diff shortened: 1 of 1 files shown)


packages/dcos-signal/extra/dcos-signal.service
```diff
@@ -14,4 +14,4 @@ ExecStartPre=/bin/ping -c1 leader.mesos
 ExecStartPre=/opt/mesosphere/bin/bootstrap dcos-signal
 
 # Signal depends on the dcos-licensing service for the license_id
-ExecStart=/opt/mesosphere/bin/dcos-signal -licensing-socket /run/dcos/dcos-license.sock -segment-key 51ybGTeFEFU1xo6u10XMDrr6kATFyRyh
+ExecStart=/opt/mesosphere/bin/dcos-signal -licensing-socket /run/dcos/dcos-licensing.sock -segment-key 51ybGTeFEFU1xo6u10XMDrr6kATFyRyh
```
makkes: I'm a bit confused why we would need the licensing-socket at all in dcos/dcos since dcos-licensing isn't even running there but only in EE, is it?

drewvanstone: Ya, that's a fair point. The reasoning was that since dcos-signal is in open source, this change should be there as well. But I think you're right, we should just move this to enterprise. I'll close this PR and make the change in enterprise instead.

PR #5810, opened 2019-07-05
Assert ZooKeeper is not running on backup

timaa2k:
## High-level description

This PR changes the ZooKeeper backup/restore script so that it checks that the ZooKeeper process is not running. Before it only asserted that `dcos-exhibitor` was not running which we suspect to not coincide with the former 100% of the cases.

The change enables getting a better understanding of the failure mode next time the flaky test in the JIRA fails. This is not a fix but a mere improvement of the debug output.

## Corresponding DC/OS tickets (required)

  - [DCOS-55827](https://jira.mesosphere.com/browse/DCOS-55827) TestZooKeeperBackup.test_transaction_log_backup_and_restore if flaky.

**edit by @adamtheturtle : We now have an issue just for this, https://jira.mesosphere.com/browse/DCOS-56097 **

PR #5817, opened 2019-07-08
[1.13] Backport tweidner/assert-zk-not-running-during-backup

adamtheturtle:
Backport of @timaa2k 's #5810, with `CHANGES.md` change added and a comment added as suggested by @jgehrcke on the original PR at https://github.com/dcos/dcos/pull/5810/files#r300683262.

Original PR's description:

```
## High-level description

This PR changes the ZooKeeper backup/restore script so that it checks that the ZooKeeper process is not running. Before it only asserted that `dcos-exhibitor` was not running which we suspect to not coincide with the former 100% of the cases.

The change enables getting a better understanding of the failure mode next time the flaky test in the JIRA fails. This is not a fix but a mere improvement of the debug output.

  - [DCOS-55827](https://jira.mesosphere.com/browse/DCOS-55827) TestZooKeeperBackup.test_transaction_log_backup_and_restore if flaky.

**edit by @adamtheturtle : We now have an issue just for this, https://jira.mesosphere.com/browse/DCOS-56097 **
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,7 +13,7 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Changed `iam-database-restore` to work when no database exists. (DCOS_OSS-5317)
 
-* Changed `iam-database-backup` to exit early if ZooKeeper is not running. (DCOS_OSS-5353)
+* Changed `dcos-zk backup` and `dcos-zk restore` to exit early if ZooKeeper is running. (DCOS_OSS-5353)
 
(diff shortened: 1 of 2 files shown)


adamtheturtle:
> need to iterate on the changelog entry. goal is to have ZooKeeper not running before backup, and we don't change iam-database-backup here, the program is called differently

Thanks for the thorough review - updated now.

adamtheturtle:
`teamcity/dcos/test/test-e2e/group1` is not reported and it is also not in the build queue. I'll make sure to make sure that this is represented in JIRA.

adamtheturtle:
JIRA issue made [here](https://jira.mesosphere.com/browse/DCOS_OSS-5355) for the above.

PR #5823, opened 2019-07-09
DCOS-56234 - Do not collect logs from successful jobs.

jbarrick-mesosphere:
## High-level description

Do not collect logs from successful jobs.

We clean up these lines by just checking for the `all_tests_passed` marker file. I am not sure what the purpose of the removed lines was, but they made the logic more complicated and brittle.

  - [DCOS-56234](https://jira.mesosphere.com/browse/DCOS-56234) Do not collect logs from successful jobs.

### DIFF
--- a/test_util/terraform_logs.sh
+++ b/test_util/terraform_logs.sh
@@ -7,8 +7,8 @@ master_ip=$(./terraform output --json masters-ips |jq -r '.value[0]')
 master_fqdn=$(./terraform output --json cluster-address |jq -r '.value')
 ssh_user=$(./terraform output --json -module dcos.dcos-infrastructure masters.os_user |jq -r '.value')
 
-ssh -i id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null ${ssh_user}@$master_ip -- "ls pytest_output || exit 0; cat pytest_output | tail -n 10 | grep -q xfailed"
-if [[ ( $? != 0 ) && ( -f all_tests_passed ) ]]; then
+# Exit early if all tests passed.
+if [[ -f all_tests_passed ]]; then
   exit 0
 fi
 


jbarrick-mesosphere:

jbarrick-mesosphere:
@mesosphere-mergebot changelog-not-required test only changes

jbarrick-mesosphere:
@mesosphere-mergebot request-review @cprovencher @orsenthil @pcrews

jbarrick-mesosphere:
Running my changes: https://teamcity.mesosphere.io/viewQueued.html?itemId=2054361&buildTypeId=DcOs_Open_Test_IntegrationTest_AwsOnpremWStaticBackendGroup1TerraformSpotBlock&tab=buildLog

jgehrcke:
Why do we do this? To save money? We rarely use these logs, but when we use them they are very valuable to us.

jbarrick-mesosphere:
> Why do we do this? To save money? We rarely use these logs, but when we use them they are very valuable to us (usually in a comparative scenario, good case vs. bad case).
> 
> I'd love to be pointed to some argument here.
> 
> I trust that you've already discussed and properly decided, and I approve. But still, this makes me sad.

This is something we've been doing forever and the main reason for it is to save on disk usage. I just broke it during the Terraform migration.

I think if we'd like to *add* collection of logs on successful jobs, we should open a ticket for it and make a case for why it's worth the disk usage.

jgehrcke:
Thanks for the feedback Justin.

> This is something we've been doing forever and the main reason for it is to save on disk usage.

Oh. Well, not the "forever" from my perspective, but probably for a long time w/o me personally noticing.

> I think if we'd like to add collection of logs on successful jobs, we should open a ticket for it and make a case for why it's worth the disk usage.

@makkes once advocated for it and I think this is a bigger topic in his team.

PR #5903, opened 2019-07-22
Added an integration test for sharing /dev/shm in a pod.

qianzhangxa:
## High-level description

Add an integration test for sharing /dev/shm in a pod.

  - [DCOS-54620](https://jira.mesosphere.com/browse/DCOS-54620) Add an integration test for sharing /dev/shm in a pod.

### DIFF
--- a/packages/dcos-integration-test/extra/test_ucr.py
+++ b/packages/dcos-integration-test/extra/test_ucr.py
@@ -348,3 +348,44 @@ def test_if_ucr_pods_can_be_deployed_with_non_root_user_ephemeral_volume(dcos_ap
     with dcos_api_session.marathon.deploy_pod_and_cleanup(pod_definition):
         # Trivial app if it deploys, there is nothing else to check
         pass
+
+
+def test_if_ucr_pods_can_share_shm_with_childs(dcos_api_session):
+    """Marathon pods inside ucr deployment integration test.
+
+    This test launches a marathon ucr pod with a specified size (1234MB)
+    of private /dev/shm and share it with one of its child containers but
+    not the other one, and then verifies the size of one child container's
+    /dev/shm is 1234MB and the other's is not.
+    """
+    test_uuid = uuid.uuid4().hex
+    pod_definition = {
+        'id': '/integration-test-pods-{}'.format(test_uuid),
+        'scaling': {'kind': 'fixed', 'instances': 1},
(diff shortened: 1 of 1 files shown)


packages/dcos-integration-test/extra/test_ucr.py
```diff
@@ -348,3 +348,38 @@ def test_if_ucr_pods_can_be_deployed_with_non_root_user_ephemeral_volume(dcos_ap
...
+                'resources': {'cpus': 0.1, 'mem': 32},
+                'image': {'kind': 'DOCKER', 'id': 'library/alpine'},
+                'exec': {'command': {'shell': 'df -m /dev/shm | grep -w 1234'}},
+                'linuxInfo': {'ipcInfo': {'mode': 'SHARE_PARENT'}}
+            },
+            {
+                'name': 'container2',
+                'resources': {'cpus': 0.1, 'mem': 32},
+                'exec': {'command': {'shell': 'echo $PING > foo; while true; do sleep 1; done'}},
+                'healthcheck': {'command': {'shell': 'test $PING = `cat foo`'}}
```
vinodkone: do you also want to check that this container doesnt get access to the shm?

qianzhangxa:

qianzhangxa:
@mesosphere-mergebot changelog-not-required Added an integration test

packages/dcos-integration-test/extra/test_ucr.py
abudnik: how about
```
'exec': {'command': {'shell': 'test -z "$(df -m /dev/shm | grep -w 1234)" && echo $PING > foo; while true; do sleep 1; done'}},
```

Gilbert88: > do you also want to check that this container doesnt get access to the shm?

We don't have this case. Containers may:
1. have their own shm
2. share from parent
3. share from the agent

https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L3301

Gilbert88: @qianzhangxa As Andrei commented above, we probably need to verify the size is actually 1234 by test -z 'df -m /dev/shm | grep -w 1234'

qianzhangxa: Done.

Gilbert88:

PR #5921, opened 2019-07-23
Add integration test for node draining and deactivation.

greggomann:
## High-level description

This PR adds two minimal integration tests which verify that node draining and deactivation work correctly.

  - [DCOS-56345](https://jira.mesosphere.com/browse/DCOS-56345) DC/OS integration test for node draining

### DIFF
--- a/packages/dcos-integration-test/extra/test_mesos.py
+++ b/packages/dcos-integration-test/extra/test_mesos.py
@@ -457,3 +457,116 @@ def reserved_disk(dcos_api_session):
                 'unreserve_resources': resources}
             r = dcos_api_session.post('/mesos/api/v1', json=request)
             assert r.status_code == 202, r.text
+
+
+def reactivate_agent(dcos_api_session, agent_id):
+    request = {
+        'type': 'REACTIVATE_AGENT',
+        'reactivate_agent': {
+            'agent_id': {
+                'value': agent_id
+            }
+        }
+    }
+
+    r = dcos_api_session.post('/mesos/api/v1', json=request)
+    assert r.status_code == 200, r.text
+
+
+def test_node_drain_and_reactivate(dcos_api_session):
+    app, test_uuid = test_helpers.marathon_test_app()
+    app_id = 'integration-test-{}'.format(test_uuid)
+
+    with dcos_api_session.marathon.deploy_and_cleanup(app):
+        # Fetch the mesos master state once the task is running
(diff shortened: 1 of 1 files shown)


greggomann:

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -457,3 +457,116 @@ def principal():
...
+
+
+def test_node_deactivate_and_reactivate(dcos_api_session):
+    # Fetch the mesos master state to find an agent ID
+    master_ip = dcos_api_session.masters[0]
+    r = dcos_api_session.get('/state', host=master_ip, port=5050)
+    assert r.status_code == 200
+    state = r.json()
+
+    assert 'slaves' in state and len(state['slaves']) > 0, 'No agents found in master state'
```
bbannier: nit: In Python a non-empty array is falsy and the more idiomatic way to check would be

    assert 'slaves' in state and state['slaves'], ...

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -457,3 +457,116 @@ def principal():
...
+                for agent in state['slaves']:
+                    if agent['id'] == agent_id:
+                        agent_found = True
+                        assert 'drain_info' in agent, 'Agent {} does not include drain info'.format(agent_id)
+                        assert agent['drain_info']['state'] == 'DRAINED'
+
+                assert agent_found
+
+            get_state_and_verify_drained()
+        except Exception as ex:
```
bbannier: Let's add a comment that this is part of the test teardown and needs to be invoked unconditionally. This would be clearer if the `try` block would contain all our mutations to the cluster (starting agent draining, confirm agent is draining, reactivating agent).

A more scalable alternative would be to use a proper test fixutre with a tear down function.

Here and below.

packages/dcos-integration-test/extra/test_mesos.py
```diff
@@ -457,3 +457,116 @@ def principal():
                 'unreserve_resources': resources}
             r = dcos_api_session.post('/mesos/api/v1', json=request)
             assert r.status_code == 202, r.text
+
+
+def reactivate_agent(dcos_api_session, agent_id):
```
bbannier: This is invoked from test teardowns and we should make sure that it succeeds. We should probably add retry logic to it either globally or on a per-invocation basis (for all cleanups).

PR #5993, opened 2019-07-30
Replace artisanal docer version detect script

janisz:
## High-level description

Fix problems with docker version > 1.19 which added new entries to `docker version` output.

Latest docker hase more entries in `docker version` output. See folowing examples

```yaml
Client: Docker Engine - Community
 Version:           19.03.1
 API version:       1.35 (downgraded from 1.40)
 Go version:        go1.12.5
 Git commit:        74b1e89
 Built:             Thu Jul 25 21:21:07 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          17.12.1-ce
  API version:      1.35 (minimum version 1.12)
  Go version:       go1.9.4
  Git commit:       7390fc6
  Built:            Tue Feb 27 22:20:43 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

```yaml
Client: Docker Engine - Community
 Version:           19.03.1
 API version:       1.40
 Go version:        go1.12.5
 Git commit:        74b1e89
 Built:             Thu Jul 25 21:21:07 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.1
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.5
  Git commit:       74b1e89
  Built:            Thu Jul 25 21:19:36 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

The previous script instead of `Engine: Version:` took `docker-init: Version:` and then took lowest version from client and server. This PR removes AWK script and replace it with `sort -V` which can correctly sort versions and `docker version -f` which allow us to use go template to get version in desired output`. Here we are interested only in client and server versions and choosing the lowest.

  - [DCOS-56831](https://jira.mesosphere.com/browse/DCOS-56831)
  - [DCOS-57154](https://jira.mesosphere.com/browse/DCOS-57154)

## Related tickets (optional)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -77,3 +77,7 @@ The following parameters have been removed from the DC/OS installer:
 * --install-prereqs
 * --deploy
 * --postflight
+
+### Fixed and improved
+
+* Fix preflight docker version check failing for docker 1.19. (DCOS-56831)
--- a/gen/build_deploy/bash.py
+++ b/gen/build_deploy/bash.py
@@ -465,52 +465,11 @@ function check_all() {
 
     check_docker_running
 
-    local docker_version=$(command -v docker >/dev/null 2>&1 && docker version 2>/dev/null | awk '
-        BEGIN {
-            version = 0
-            client_version = 0
-            server_version = 0
-        }
-        {
-            if($1 == "Server:") {
-                server = 1
-                client = 0
-            } else if($1 == "Client:") {
-                server = 0
-                client = 1
-            } else if ($1 == "Server" && $2 == "version:") {
-                server_version = $3
-            } else if ($1 == "Client" && $2 == "version:") {
-                client_version = $3
-            }
-            if(server && $1 == "Version:") {
-                server_version = $2
-            } else if(client && $1 == "Version:") {
-                client_version = $2
-            }
-        }
-        END {
-            if(client_version == server_version) {
-                version = client_version
-            } else {
-                cv_length = split(client_version, cv, ".")
-                sv_length = split(server_version, sv, ".")
+    local docker_version=$(command -v docker >/dev/null 2>&1 \
+        && docker version --format '{{{{printf "%s\\n%s" .Server.Version .Client.Version}}' \
+        | sort -V \
+        | head -n 1)
 
-                y = cv_length > sv_length ? cv_length : sv_length
-
-                for(i = 1; i <= y; i++) {
-                    if(cv[i] < sv[i]) {
-                        version = client_version
-                        break
-                    } else if(sv[i] < cv[i]) {
-                        version = server_version
-                        break
-                    }
-                }
-            }
-            print version
-        }
-    ')
     # CoreOS stable as of Aug 2015 has 1.6.2
     check docker 1.6 "$docker_version"
 


janisz:
@drewvanstone Updated description

gen/build_deploy/bash.py
```diff
@@ -465,52 +465,7 @@ def validate_custom_check_bins_dir(custom_check_bins_dir):
...
-                    } else if(sv[i] < cv[i]) {
-                        version = server_version
-                        break
-                    }
-                }
-            }
-            print version
-        }
-    ')
+    local docker_version=$(command -v docker >/dev/null 2>&1 && docker version --format '\{\{printf "%s\n%s" .Server.Version .Client.Version\}\}' | sort -V | head -n 1)
```
makkes:
```suggestion
    local docker_version=$(command -v docker >/dev/null 2>&1 && docker version --format '\{\{printf "%s\n%s" .Server.Version .Client.Version\}\}' | sort -rV | head -n 1)
```
As far as I understand the old AWK script the lesser of both versions should be returned here.

janisz: The code it good
```[root@dcos-e2e-default-4cc8d-master-0 /]# echo -e '18.06.3-ce\n17.06.3' | sort -V | head -n 1
17.06.3
[root@dcos-e2e-default-4cc8d-master-0 /]# echo -e '17.06.1\n18.06.3-ce\n17.06.3' | sort -V | head -n 1
17.06.1
```

makkes: Aaah, sorry, you're using `head`, not `tail`. All good.

janisz:

janisz:
@mesosphere-mergebot request-review @drewvanstone @br-lewis @makkes @fatz @vinodkone

janisz:
@mesosphere-mergebot sync

janisz:
@mesosphere-mergebot sync

CHANGES.md
```diff
@@ -77,3 +77,7 @@ The following parameters have been removed from the DC/OS installer:
 * --install-prereqs
 * --deploy
 * --postflight
+
+### Fixed and improved
```
jgehrcke: Please don't add new section headings.

jgehrcke:
Is the plan to "fix" this in 1.13.x?

CHANGES.md
```diff
@@ -77,3 +77,7 @@ The following parameters have been removed from the DC/OS installer:
...
+
+* Fix preflight docker version check failing for docker 1.19. (DCOS-56831)
```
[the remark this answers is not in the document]
jgehrcke: Is the plan to "fix" this in 1.13.x?

Then this entry should not be here at all.

janisz:
@jgehrcke Here are backports:

- https://github.com/mesosphere/dcos-enterprise/pull/6309
- https://github.com/mesosphere/dcos-enterprise/pull/6308
- https://github.com/dcos/dcos/pull/5995
- https://github.com/dcos/dcos/pull/5994

CHANGES.md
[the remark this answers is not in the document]
janisz: > Is the plan to "fix" this in 1.13.x?
> Then this entry should not be here at all.

Why? if something is fixed in 1.13.x it does not imply it's fixed in 1.14. Bugfix release for 1.13 may be months after 1.14 so IMO we should include all fixes in changelog. Look at [Mesos Changelog](https://github.com/apache/mesos/blob/master/CHANGELOG) 1.8.0 release contains bugfixes that were also listed in [1.7.3 CHANGELOT](https://github.com/apache/mesos/blob/1.7.x/CHANGELOG)

Do we have a different policy for DCOS changelog? Were can I find information about it?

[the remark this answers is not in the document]
jgehrcke: > Why? if something is fixed in 1.13.x it does not imply it's fixed in 1.14.

That's right, but here we know already more, and we want to use that information: we _know_ that the next 1.13.x release will be much earlier than the GA 1.14.0 release. 

You are right that once 1.14.0 is already out the door and if we are then going to release 1.13.9 that this is a different situation. But that's not where we are yet.

> Do we have a different policy for DCOS changelog? Were can I find information about it?

https://github.com/dcos/dcos/wiki/CHANGES.md-guidelines#major-releases

Quote:

> In case of a bug fix forward port into the development version of the next major release there is no need to mention this in the changelog of the major release. Example: a bug is fixed for the upcoming 1.12.3 release while 1.13.0 is still in development. As of common sense the same bug fix gets forward-ported into 1.13.0-dev. That very forward-port does not need to be accompanied by a changelog entry. It's expected behavior for this bug fix to be included in 1.13.0. If it's not, it's a bug in itself and will be fixed with with a future 1.13.x release, accompanied with a relevant changelog entry.

I hope you appreciate that this
- saves you work (forward ports don't need to be noted during *all* of 1.14.0 dev, this saves a bunch of work).
- keeps the information in the 1.14.0 changelog more interesting (all the goodies from previous major / patch releases are contained, everyone assume that anyway, we don't need to mention them again).

I want to say again: in a couple of months from now you will be right. But not yet :)

Hope that helps @janisz

janisz:
@deric How did you test it? We need double double curly bracket (`{{{{`) to escape it so in final script it will be `{{`

> [# To escape the template signature, put four `{`, so `{{{{` would result in the final result getting a `{{`.](https://github.com/dcos/dcos/blob/2163801370fb195cb09acf0b88ee945c257f3c4a/gen/template.py#L8)

deric:
Ahh, sorry. I was directly copying the PR's code into bash script as there's no release. I'll try to recompile DC/OS and test it the whole script.

janisz:
@deric no worries. I've made the same mistake. It took me hours to debug that escaping is necessary. Thankfully @makkes helped with it.

PR #6006, opened 2019-08-02
Revert to default journald rate limit

jongiddy:
## High-level description

Remove the increased journald rate limit. Scale testing demonstrates that the raised limit causes problems for other components that get delayed or lost logs. Logging that exceeds the default indicates that a component is logging excessively, and should be remedied in that component.

  - [DCOS-53763](https://jira.mesosphere.com/browse/DCOS-53763) "too many" Admin Router access log lines went missing in MWST 15

## Related tickets (optional)

Change to logging only.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,6 +64,8 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Prune VIPs with no backends in order to avoid unbounded growth of state and messages exchanged among `dcos-net` processes. (DCOS_OSS-5356)
 
+* DC/OS no longer increases the rate limit for journald logging. (DCOS-53763)
+
 
 ### Breaking changes
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -64,7 +64,7 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Prune VIPs with no backends in order to avoid unbounded growth of state and messages exchanged among `dcos-net` processes. (DCOS_OSS-5356)
 
-* DC/OS no longer increases the rate limit for journald logging. (DCOS-53763)
+* DC/OS no longer increases the rate limit for journald logging.  Scale testing demonstrated that raising the limit overloads journald, causing problems for other components that see delayed or lost logs or, worse, hang until log buffers are read. The default of 10000 messages per 30 seconds appears to distinguish well between busy components and excessively verbose components. (DCOS-53763)
 
 
 ### Breaking changes
--- a/gen/cloud-config.yaml
+++ b/gen/cloud-config.yaml
@@ -15,8 +15,6 @@ root:
     content: |
       [Journal]
       MaxLevelConsole=warning
-      RateLimitInterval=1s
-      RateLimitBurst=20000
   - path: /etc/rexray/config.yml
     permissions: "0644"
     content: {{ rexray_config_contents }}


jongiddy:

jgehrcke:
Thanks for taking on the ownership on this one, @jongiddy!

gen/cloud-config.yaml
```diff
@@ -15,8 +15,6 @@ root:
     content: |
       [Journal]
       MaxLevelConsole=warning
-      RateLimitInterval=1s
-      RateLimitBurst=20000
```
jgehrcke: Do we really want to rely on the default or "set our default explicitly"? :) What is the default? Maybe add a code comment here?

CHANGES.md
```diff
@@ -64,6 +64,8 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Prune VIPs with no backends in order to avoid unbounded growth of state and messages exchanged among `dcos-net` processes. (DCOS_OSS-5356)
 
+* DC/OS no longer increases the rate limit for journald logging. (DCOS-53763)
```
jgehrcke: Can we add one or two more sentences explaining your beautiful findings and reasoning?

jgehrcke:
@jongiddy once again thanks for your efforts on this front. 

I have two questions that are based on curiosity, independent of approving this PR:

1) The motivation for this change is to not lose log important log messages as of journal overload. Do we understand how much this change contributes to this goal? Do we _know_ that when journald limits the log ingress of _one_ component that it actually has more ingress capacity for other components? Even better, do we understand how efficiently the journal can protect itself here?

2) Should we start setting lower rates for specific components of which we know, thanks to your measurements, that they burst much more than other components? The Mesos master could be such a candidate: we wouldn't even lose much: the Mesos master logs to file anyway (properly log-rotated).

gen/cloud-config.yaml
jongiddy: I'm wary of adding comments for configuration that we don't override. Like, why do we not have a comment explaining why we're using the journald default for `SyncIntervalSec`?

On the other hand, I'd like to have an explicit comment on why we set `MaxLevelConsole=warning`.

jongiddy:
> Do we understand how much this change contributes to this goal? Do we know that when journald limits the log ingress of one component that it actually has more ingress capacity for other components?

It's circumstantial evidence, but I think I've established that Admin Router drops log messages at the same time that other components are logging excessively. The real test will be the next Scale Test, where we hope to see no instances of Admin Router dropping messages, while also looking at how often the excessive loggers have messages dropped.

> Should we start setting lower rates for specific components of which we know, thanks to your measurements, that they burst much more than other components?

Possibly. I think we should see what we get with this change, and then consider further actions. There are instances in CI where similar problems occur, but the log rate is lower than the default limit, so would not be fixed by this change. Those appear to be due to how disks are allocated in CI. Perhaps more selective log rate management is the answer for those problems.

CHANGES.md
```diff
@@ -64,6 +64,8 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Prune VIPs with no backends in order to avoid unbounded growth of state and messages exchanged among `dcos-net` processes. (DCOS_OSS-5356)
 
+* DC/OS no longer increases the rate limit for journald logging.  Scale testing demonstrated that raising the limit overloads journald, causing problems for other components that see delayed or lost logs or, worse, hang until log buffers are read. The default of 10000 messages per 30 seconds appears to distinguish well between busy components and excessively verbose components. (DCOS-53763)
+
```
jgehrcke: Lovely, thank you!

makkes:
I'm not sure I understand the issue. As far as I understand it, the rate limiting was introduced in August, 2016, wasn't it? Why is this causing problems just now?

gen/cloud-config.yaml
```diff
@@ -15,8 +15,6 @@ root:
     content: |
       [Journal]
       MaxLevelConsole=warning
-      RateLimitInterval=1s
```
makkes: So, in the latest DC/OS version, we're logging so much that we reach the limit of what journald can handle, right? So, we want to fall back to the journald defaults (which probably are tested by the systemd team and journald should be able to cope with).

But isn't the real problem that we are logging way too much? I stumbled the huge amount of log messages even in small clusters as well and it's a pain to rummage through those logs.

jgehrcke:
> Why is this causing problems just now?

It has always caused problems. But we never really discovered them systematically, prioritized working on this. Really this required us to do a couple of MWTs.

jgehrcke:
>  But I'd love to see tickets for tracking the real problem here: That we log way too much.

Example: we lost valuable Admin Router logs. The error message is "send() to syslog failed while logging request". This is tracked by https://jira.mesosphere.com/browse/DCOS-53763 (in description of the PR). 

More context: https://jira.mesosphere.com/browse/DCOS-43956, https://jira.mesosphere.com/browse/DCOS-43592, https://jira.mesosphere.com/browse/DCOS-53208

makkes:
> > But I'd love to see tickets for tracking the real problem here: That we log way too much.
> 
> Example: we lost valuable Admin Router logs. The error message is "send() to syslog failed while logging request". This is tracked by https://jira.mesosphere.com/browse/DCOS-53763 (in description of the PR).
> 
> More context: https://jira.mesosphere.com/browse/DCOS-43956, https://jira.mesosphere.com/browse/DCOS-43592, https://jira.mesosphere.com/browse/DCOS-53208

What I meant is, shouldn't we identify the components that are logging so much and try to decrease the amount of log messages they produce?

jongiddy:
@makkes Before this change, if a component logged too much, then journald slowed down and all components suffered, even those that mostly do the right thing. With this change, only the components that log too much will suffer, by having their log messages dropped.

This puts the pressure on the verbose components to fix their logging. If they don't, or if they have a rare case that exceeds the limit, then at least they won't affect the running cluster. Some issues that are relevant:
https://jira.mesosphere.com/browse/DCOS-42131
https://jira.mesosphere.com/browse/DCOS-56615

jongiddy:
Override please:
```
@mesosphere-mergebot override-status teamcity/dcos/test/terraform/aws/onprem/static/group3 https://jira.mesosphere.com/browse/DCOS-57318
```

jgehrcke:
@mesosphere-mergebot override-status teamcity/dcos/test/terraform/aws/onprem/static/group3 https://jira.mesosphere.com/browse/DCOS-57318

jongiddy:
@mesosphere-mergebot request-review @tillt

PR #6041, opened 2019-08-08
Replace pytest-xdist by execnet

cprovencher:
- The only real dependency needed was execnet, not all of pytest-xdist.
- Adding a README for how to run tests
- Pinning all pytest-xdist dependencies because pytest-xdist package doesn't and it sometimes causes errors, like recently when execnet was updated and broke pytest-xdist, even if pytest-xdist version was pinned!
- Moving packages/dcos-integration-test/extra/requirements.txt to packages/dcos-integration-test/requirements.txt so that the whole virtual environment is not copied to the remote cluster when integration tests are ran with pytest-xdist. The whole "extra" directory is synced to the remote cluster when pytest runs.

### DIFF
--- a/packages/dcos-image-deps/build
+++ b/packages/dcos-image-deps/build
@@ -5,8 +5,8 @@ mkdir -p "$LIB_INSTALL_DIR"
 
 # Install wheels.
 for package in adal analytics-python azure-nspkg azure-common azure-mgmt-nspkg \
-    azure-mgmt-network azure-storage beautifulsoup4 docutils keyring msrest \
-    msrestazure pytest-xdist requests-oauthlib schema webob wheel; do
+    azure-mgmt-network azure-storage beautifulsoup4 docutils execnet keyring msrest \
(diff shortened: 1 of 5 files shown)


cprovencher:

cprovencher:
@mesosphere-mergebot changelog-not-required bug fix for integration tests

packages/dcos-image-deps/build
```diff
@@ -5,8 +5,8 @@ mkdir -p "$LIB_INSTALL_DIR"
 
 # Install wheels.
 for package in adal analytics-python azure-nspkg azure-common azure-mgmt-nspkg \
-    azure-mgmt-network azure-storage beautifulsoup4 docutils keyring msrest \
-    msrestazure pytest-xdist requests-oauthlib schema webob wheel; do
```
orsenthil: Hi @cprovencher 

In this file we are installing only `execnet` while on dcos-integration-test/requirements.txt file, we install `pytest-xdist` - Would it create any problems?

Thanks!

PR #6060, opened 2019-08-12
[Forward port master] Fix Admin Router: Error on invalid Marathon service port label

timaa2k:
## High-level description

Forward port of #6139

Previously, specifying an invalid `DCOS_SERVICE_PORT_INDEX` value in a Marathon app definition led to `Admin Router` crash looping on Marathon state cache refresh.
Now this scenario logs an appropriate error and the cache refresh continues by skipping such apps.

  - [DCOS_OSS-5491](https://jira.mesosphere.com/browse/DCOS_OSS-5491) Admin Router: wrong value in DCOS_SERVICE_PORT_INDEX breaks AR cache.

## Related tickets (optional)

  - [COPS-5147](https://jira.mesosphere.com/browse/COPS-5147) Specifying incorrect DCOS_SERVICE_PORT_INDEX breaks /service routing in adminrouter.

### DIFF
--- a/packages/adminrouter/extra/src/lib/cache.lua
+++ b/packages/adminrouter/extra/src/lib/cache.lua
@@ -255,15 +255,47 @@ local function fetch_and_store_marathon_apps(auth_token)
        -- In "container/bridge" and "host" networking modes we need to use the
        -- host port for routing (available via task's ports array)
        local port
-       if is_container_network(app) and app["container"]["portMappings"][portIdx]["containerPort"] ~= 0 then
-         port = app["container"]["portMappings"][portIdx]["containerPort"]
-       else
-         port = task["ports"][portIdx]
+       if is_container_network(app) then
+
+           -- Special case, meaning no ports defined for app in container networking mode.
+           if next(app["container"]["portMappings"]) == nil then
+               goto continue
+           end
+
+           -- In every other case portMappings exist with at least the default.
+           -- Skip routing if DCOS_SERVICE_PORT_INDEX points out of bounds of existing portMappings.
+           if not app["container"]["portMappings"][portIdx] then
+               ngx.log(ngx.NOTICE, "Cannot find port in container portMappings at Marathon port index '" .. (portIdx - 1) .. "' for app '" .. appId .. "'")
+               goto continue
+           end
+
+           -- If the portMapping exists containerPort always exists.
+           -- https://mesosphere.github.io/marathon/docs/networking.html#port-mappings
+           -- For any other case route to the containerPort in container networking mode.
+           -- NOTE(tweidner): I believe this is unnecessary, containerPort 0 is not a special case.
+           if app["container"]["portMappings"][portIdx]["containerPort"] ~= 0 then
+               port = app["container"]["portMappings"][portIdx]["containerPort"]
+           end
        end
 
+       -- If the containerPort was randomly assigned or any other networking mode is used
+       -- try routing to the task ports assigned by Mesos for the given Marathon app.
        if not port then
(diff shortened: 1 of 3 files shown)


timaa2k:

packages/adminrouter/extra/src/lib/cache.lua
```diff
@@ -255,15 +255,21 @@ local function fetch_and_store_marathon_apps(auth_token)
        -- In "container/bridge" and "host" networking modes we need to use the
        -- host port for routing (available via task's ports array)
        local port
-       if is_container_network(app) and app["container"]["portMappings"][portIdx]["containerPort"] ~= 0 then
-         port = app["container"]["portMappings"][portIdx]["containerPort"]
-       else
-         port = task["ports"][portIdx]
+       if is_container_network(app) then
+           if not app["container"]["portMappings"][portIdx] then
+               ngx.log(ngx.NOTICE, "Cannot find port in container portMappings at Marathon port index '" .. (portIdx - 1) .. "' for app '" .. appId .. "'")
```
jgehrcke: Why did you choose `NOTICE`? (not implying anything else for now, just would love to understand how you thought about it!)

jgehrcke: Is it helpful to have `portIdx` in code and `portIdx - 1` in the log message?

jgehrcke: can we shorten this line?

jgehrcke: Can you quickly summarize in which cases `if not app["container"]["portMappings"][portIdx]` is true? What happens if `app["container"]` does not have the property `["portMappings"]`?

packages/adminrouter/extra/src/test-harness/tests/test_cache.py
```diff
@@ -1132,6 +1132,54 @@ def test_app_without_port_index_label(
         self._assert_filter_regexp_for_invalid_app(
             filter_regexp, app, nginx_class, mocker, valid_user_header)
 
+    def test_app_container_networking_with_invalid_port_mapping_index_label(
```
jgehrcke: I did not review the tests, but thanks for writing them! :rocket: If you want to have feedback also from @vespian just let him know :-)

packages/adminrouter/extra/src/lib/cache.lua
timaa2k: I see it as in general there are two log levels of interest: NOTICE (INFO) and DEBUG.
Since we don't halt on this (ERROR case) and  but rather just skip and continuing cache refreshing, emitting a NOTICE is the reasonable thing to do. One could make it a WARNING, but from the perspective of Nginx nothing went wrong. However the user should get a reasonable response on what went wrong and we the deployment didn't work (at the Marathon level?). This should not be done primarily in the Admin Router logs IMO.

timaa2k: Continue cases:
1. `container` networking mode, `portMappings` defined, `portMappings` list index out of bounds.
2. `container/bridge` networking mode, `ports` list index out of bounds.
3. `host` networking mode, `ports` list index out of bounds.

> What happens if app["container"] does not have the property ["portMappings"]
Then, the first part is skipped without notice and the ports array index port is used instead.
I'm not sure what's the intended behavior here.

timaa2k:
Haha, please I never want to do it again! :dagger:

timaa2k:
@ahairape PTAL

timaa2k:
@mesosphere-mergebot changelog-not-required forward port to unreleased version

PR #6070, opened 2019-08-13
Update test_util/terraform_test.sh

cprovencher:
Make terraform_test.sh run integration tests outside the cluster using pytest-xdist
tested here for open: https://teamcity.mesosphere.io/viewLog.html?buildId=2181649&
tested here for enterprise: https://teamcity.mesosphere.io/viewQueued.html?itemId=2182231

PR #6082, opened 2019-08-15
Add apipkg dependency

cprovencher:
When pytest-xdist feature was tested, it was determined that execnet had to be installed on the remote host. We have since then added execnet as a dependency to DC/OS. However, it appears pytest-xdist's dependencies are not pinned (moving targets). So without any changes to DC/OS, pytest-xdist feature stopped working. The solution was to not only pin pytest-xdist (was already the case), but also all its dependencies. Those changes went into DC/OS, but there was still an issue that wasn't fixed. Pinning all the dependencies wasn't enough. Adding apipkg on the remote host was also necessary. This was not caught because we don't test pytest-xdist feature in CI. This is about to change with this PR: https://github.com/dcos/dcos/pull/6070
The current PR was also tested with the pytest-xdist feature here: https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&buildTypeId=DcOs_Open_Test_IntegrationTest_TerraformBasedIntegrationTests_PytestXdistTestAws&buildId=2180311

### DIFF
--- a/packages/dcos-image-deps/build
+++ b/packages/dcos-image-deps/build
@@ -4,7 +4,7 @@ export LIB_INSTALL_DIR="$PKG_PATH/lib/python3.6/site-packages"
 mkdir -p "$LIB_INSTALL_DIR"
 
 # Install wheels.
-for package in adal analytics-python azure-nspkg azure-common azure-mgmt-nspkg \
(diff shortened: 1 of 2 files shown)


cprovencher:
@mesosphere-mergebot changelog-not-required internal impact only

cprovencher:
@mesosphere-mergebot  label ready for review

PR #6166, opened 2019-08-28
Remove nogroup group creation

drewvanstone:
## High-level description

ATT&T is blocked from upgrading DC/OS because their security team does not allow non-standard linux group creation. We currently create the `nogroup` group, but it appears that it may not actually be used any longer. It looks like it was originally used by nginx (admin router). Depending where the service was started from, it'd set the nginx configuration to use either the `nobody` group (for Redhat based systems), or the `nogroup` group (for Debian based systems). It looks like we started creating the `nogroup` group on all systems to keep things consistent.

  - [COPS-5220](https://jira.mesosphere.com/browse/COPS-5220)


drewvanstone:

janisz:
Refs: https://github.com/dcos/dcos-launch/pull/244

PR #6232, opened 2019-09-09
Implement Winpanda Application core.

anatolii-borysov:
JIRA: DCOS_OSS-5512: Implement Winpanda Application Core

## High-level description

Winpanda application to implement DC/OS packages management for Windows-based agent nodes.

  - [DCOS-ID](https://jira.mesosphere.com/browse/DCOS_OSS-5512) Implement Winpanda Application Core.

Create a core codebase structure of Winpanda Application providing intergation points for the DC/OS distribution storage drivers and Windows service management module being developed in parallel.

  - [ ] This is the initial commit for the Winpanda application  

@jongiddy @timaa2k @DominikDary


DominikDary:
@mesosphere-mergebot request-review jgehrcke

packages/winpanda/extra/src/winpanda/svcm/base.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        pass
+
+    @abc.abstractmethod
+    def start(self, *args, **kwargs):
+        """Start a registered service (immediately).
+        """
+        pass
+
+    def stop(self, *args, **kwargs):
```
jongiddy: These last two methods should also be abstract methods.

packages/winpanda/extra/src/winpanda/utils.py
```diff
@@ -0,0 +1,66 @@
...
+def unpack(tarpath, location):
+    """
+    unpacks tar.xz to  location
+    """
+    
+    _location = os.path.abspath(location)
+
+    if  not os.path.exists(_location):
+        print("no Directory exist creating...\n{}".format(_location))
+        os.mkdir(_location)
```
jongiddy: This assumes that the parent directories exist. Perhaps use `os.makedirs`

packages/winpanda/extra/src/winpanda/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    return path
+
+def unpack(tarpath, location):
+    """
+    unpacks tar.xz to  location
+    """
+    
+    _location = os.path.abspath(location)
+
+    if  not os.path.exists(_location):
```
jongiddy: If the directory does exist, should we check that it is empty?

packages/winpanda/extra/src/winpanda/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    _location = os.path.abspath(location)
+    dl = SmartDL(url, _location)
+    dl.start()
+    path = os.path.abspath(dl.get_dest())
+    print ("Downloaded to {} ".format(path), " from {}".format(url),sep='\n')
+    return path
+
+def unpack(tarpath, location):
+    """
+    unpacks tar.xz to  location
```
jongiddy: It would be worth noting that this `location` is a folder, not a filepath. Perhaps even change the name.

packages/winpanda/extra/src/winpanda/cli.py
```diff
@@ -0,0 +1,93 @@
...
+        InstallationManager().command_.execute()
+    except Exception as e:
+        print('{}: {}'.format(type(e).__name__, e), file=sys.stderr)
+        sys.exit(1)
+    else:
+        sys.exit(0)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
jongiddy: Either `sys.exit` is not required here, or `main` should return an exit status rather than exiting.

packages/winpanda/extra/src/winpanda/cli.py
```diff
@@ -0,0 +1,93 @@
...
+        return command.create(command_name=command_name, **cmd_opts)
+        # The 'unsupported command' case is handled by the docopts module.
+
+
+def main():
+    """"""
+    try:
+        InstallationManager().command_.execute()
+    except Exception as e:
+        print('{}: {}'.format(type(e).__name__, e), file=sys.stderr)
```
jongiddy: It would be useful to have a mode for debugging (maybe an environment variable) that displayed the full exception with traceback

packages/winpanda/extra/src/winpanda/svcm/nssm.py
```diff
@@ -0,0 +1,402 @@
...
+
+class NSSMConfSection(enum.Enum):
+    """NSSM configuration file sections."""
+    SERVICE = 'service'
+
+
+@base.svcm_type('nssm')
+class WinSvcManagerNSSM(base.WindowsServiceManager):
+    """NSSM-based Windows service manager."""
+    __exec_fname__ = 'nssm.exe'  # Executable file name.
```
jongiddy: What is the intent with the underscores here? General Python behaviour is:
- a single leading underscore indicates a private attribute, but it is not enforced by the interpreted
- a double leading underscore indicates a private attribute, somewhat enforced by being obscured through name mangling
- double leading and trailing underscores are for Python language attributes, not for general use (and are not private in any way).

anatolii-borysov: That's correct. Will fix it.

packages/winpanda/extra/src/winpanda/cli.py
anatolii-borysov: Will do.

packages/winpanda/extra/src/winpanda/cli.py
anatolii-borysov: Agree. Will be fixed.

packages/winpanda/extra/src/winpanda/bin/winpanda.py
```diff
@@ -0,0 +1,120 @@
+"""Panda package management for Windows.
+
+CLI entry point definition.
+"""
+from pathlib import Path
+import sys
+
+from docopt import docopt, DocoptExit
+
+sys.path.insert(1, f'{Path(sys.path[0]).parent}')
```
jongiddy: Why is this needed? Can we structure the code so that it is not needed?

packages/winpanda/extra/src/winpanda/bin/winpanda.py
```diff
@@ -0,0 +1,120 @@
...
+            cmd_stop=CLI_COMMAND.STOP,
+            default_cmd_target=CLI_CMDTARGET.PKGALL,
+            default_root_dpath=storage.DCOS_INST_ROOT_DPATH_DFT,
+            default_config_dpath=storage.DCOS_INST_CFG_DPATH_DFT,
+            default_state_dpath=storage.DCOS_INST_STATE_DPATH_DFT,
+            default_repository_dpath=storage.DCOS_INST_PKGREPO_DPATH_DFT,
+            default_var_dpath=storage.DCOS_INST_VAR_DPATH_DFT,
+            default_dstor_url=cm_const.DCOS_DSTOR_URL_DFT,
+            default_dstor_pkgrepo_path=cm_const.DCOS_DSTOR_PKGREPO_PATH_DFT
+        )
```
jongiddy: This is missing `valid_cmd_targets`.

packages/winpanda/extra/src/winpanda/bin/winpanda.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+LOG = logger.get_logger(__name__)
+
+
+class DCOSInstallationManager:
+    """DC/OS installation manager."""
+    def __init__(self):
+        """Constructor."""
+        self.cli_args = self._get_cli_args()
```
jongiddy: We don't need to store `cli_args` as an attribute. Just pass it straight to the `_get_command` method.

packages/winpanda/extra/src/winpanda/common/cli.py
```diff
@@ -0,0 +1,73 @@
...
+Application CLI specification.
+"""
+
+
+class CLI_COMMAND:
+    """CLI command set."""
+    SETUP = 'setup'
+    TEARDOWN = 'teardown'
+    START = 'start'
+    STOP = 'stop'
```
jongiddy: How does this set of commands relate to the `pkgpanda` commands? https://github.com/dcos/dcos/blob/f4899e6d2dbf2a3a3a5d94c5da347796275852ef/pkgpanda/cli.py#L4-L13

packages/winpanda/extra/src/winpanda/common/storage.py
```diff
@@ -0,0 +1,243 @@
+"""Panda package management for Windows.
+
+DC/OS installation local storage management tools.
+
+Default local storage layout for DC/OS installation:
+
+<inst_drive>:/                # DC/OS installation drive
+    +-<inst_root>/            # DC/OS installation root dir
```
jongiddy: In the future, we are likely to want to position these folders in different locations. `<inst_tmp>` for example might map to a folder in `C:\Temp`. Similarly other folders might need to be on separate drives.

We should define default equivalents to:
INSTALL_FOLDER `/opt/mesosphere` -> `C:\dcos\opt`
VARLOG_FOLDER `/var/log/dcos` -> `C:\dcos\log`
VARLIB_FOLDER `/var/lib/dcos` -> `C:\dcos\lib`
... and anything else ...
without breaking them into drive and path components, and leave open the possibility of individually modifying each of these FOLDER variables through command line options.

packages/winpanda/extra/src/winpanda/bin/winpanda.py
anatolii-borysov: Keeping an 'entry point' module (as well as all the other source modules) in a separate directory makes entire aplication's code base structure cleaner. .. to my mind. Code base's root would then contain only helper files, like requirements.txt, README.md, tox.ini, etc

I can move winpanda.py to the code base root, if the consideration above does not make sense for you.

packages/winpanda/extra/src/winpanda/bin/winpanda.py
anatolii-borysov: As winpanda is under active development for the moment I'd keep it as it is for providing better visual perception of which entities the app is operating with. And this approach also follows PEP20's rule of 'explicit is better than implicit' and so suport better code readability ... to my mind.

packages/winpanda/extra/src/winpanda/common/storage.py
anatolii-borysov: This is **default** FS layout. Every individual directory path can be easily redefined via an appropriate CLI option:
--inst-root-dir=<path>         DC/OS installation root directory
--inst-config-dir=<path>       DC/OS installation configuration directory
--inst-state-dir=<path>        DC/OS installation state directory
--inst-repo-dir=<path>         DC/OS local package repository directory.
--inst-var-data-dir=<path>     DC/OS variable data root directory.

Every of the options above accepts both relative and absolute pathes as their values.
Relative pathes will be resolved basing on the value of '--inst-root-dir' option. Absolute pathes will feature the drive part indeed and so we'll have enough flexibility to configure DC/OS installation local FS layout. I.e. this implementation already covers everything you are concerning about?

packages/winpanda/extra/src/winpanda/common/cli.py
anatolii-borysov: Here there is a number of points to take into account:

1) "winpanda setup" - this directly corresponds "pkgpanda setup" functionality.

2) "winpanda teardown" -  this one directly corresponds to the "pkgpanda uninstall", which does completely remove entire DC/OS installation (stops all the system services and removes all the appropriate configs for systemd, removes all the files/dirs related to a DC/OS installation). 
    I just think that the "teardown" is a better fit for the operation name here.  It can be considered as a synonym  for the "uninstall" but we use "install/uninstall" names also at the package and the system service levels and I just wanted to use another word for the whole DC/OS operation level for  better separation. 
    The other point is that we use the "setup", but not the "install" name for  the whole DC/OS deployment operation and the "teardown" better fits as the name for opposite operation. 
    I can rename it to "uninstall", if the considerations above do not make sense for you.

3) "winpanda start/stop" - this will do start/stop all the system service processes corresponding to all the installed DC/OS packages. A kind of DC/OS "on/off" for a windows-based node. This is different to pkgpanda behaviour. 
    The pkgpanda does these operations under the hood for top level DC/OS installation management operations like "setup" and "activate", but does not expose them via CLI. 
    I think that this will be useful to implement a "maintenance" state for windows nodes when we need to repair/investigate issues, and also will facilitate the DC/OS "upgrade" operation.
   
4) As far as I can see, currently only three of the whole bunch of operations which the pkgpanda exposes via CLI are really used. These are "setup", "fetch" and "activate". The former one is used for the clean DC/OS installation on a node in the dcos_install.sh script (/opt/dcos-install/<DC/OS version>/genconf/serve/dcos_install.sh residing on a bootsrap node), two latter are used during DC/OS upgrade operation in the dcos_node_upgrade.sh script (/opt/dcos-install/<DC/OS version>/genconf/serve/upgrade_from_<DC/OS version>/<UUID>/dcos_node_upgrade.sh residing on a bootstrap node).
    I can assume that the others were intended to be used via pkgpanda web-API,  but in accordance with our initial investigation of pkgpanda operation on a node (Artem said that, as far as I can remember) that API is not in use now and unlikely will be used in the future.
Ref: https://docs.d2iq.com/mesosphere/dcos/1.13/administering-clusters/component-management/#routes

    Hence the question: do we need to replicate the complete pkgpanda's  node-level functionality in winpanda?
   
5) We had an internal discussions within the local team and preliminary agreed on the logic of DC/OS upgrade operation. Our vision is that it will be  covered by the "setup" operation as well. 
    If we introduce a notion of DC/OS installation state (i.e. list of packages that are currently active on a node), than both  "upgrade" and **clean** "setup" operations in their essence are just two particular cases of the **generic** "setup" operation. 
    Meaning that the  **clean** "setup" is the case when the state is empty (list of active packages is empty) and so it needs to install all the packages listed as parts of a DC/OS version being deployed (e.g. specified in a special <DC/OS version>--package_list.json file at a  bootstrap node).
    The "upgrade" operation is the case when we have the DC/OS state non-empty (list of active packages is non empty) and so we need to compare it with a list describing a new DC/OS version (the one we want to upgrade to, e.g <New DC/OS version>--package_list.json at a bootstrap node), produce a list of packages to be removed and another list of those to be installed, and then just do remove/install appropriately for the every package in those lists.
    This approach would save us from generating additional helper scripts like the dcos_node_upgrade.sh, avoiding unnecessary complication and sharing responsibility of DC/OS upgrade operation between the "upgrade script" and winpanda app, and alow us to have just a single artifact, describing a set of  packages that are to be present on a node in accordance with desired DC/OS version - <DC/OS version>--package_list.json. 
    Also moving all the responsibility of the "upgrade" operation to the winpanda app would allow us to provide more consistent error reporting and logging for that operation (from the single source).
   
    As far as I can understand now a bootstrap node already provides an equivalent of that imaginary "<DC/OS version>--package_list.json" artifact as the /opt/dcos-install/<DC/OS version>/genconf/serve/package_lists/UUID.package_list.json. 
    So, the idea outlined above is no way a completely new approach to define the   **desired** DC/OS state. And requires very little, or no changes to the logic iplemented at the bootstrap node side. We just implement logic of winpanda app to handle the "upgrade" operation more wisely that the pkgpanda does.

packages/winpanda/extra/src/winpanda/svcm/base.py
anatolii-borysov: Correct.
Fixed.

packages/winpanda/extra/src/winpanda/utils.py
anatolii-borysov: These two utility functions definitely need refactoring. Will do.

packages/winpanda/extra/src/winpanda/utils.py
anatolii-borysov: This utility function definitely needs refactoring. Will do.

packages/winpanda/extra/src/winpanda/utils.py
anatolii-borysov: Unfortunately I can't get your idea here. Could you please elaborate on this matter?

packages/winpanda/extra/src/winpanda/cli.py
anatolii-borysov: Fixed.

packages/winpanda/extra/src/winpanda/cli.py
anatolii-borysov: Partially fixed by introducing application-wide logging sub-system supporting logging levels.
Now all the telemetry of winpanda app is packed into unified format and goes to a winpanda.log file (being rotated on size basis, and maintaining history of fixed size)
But for the moment that logging level is hardcoded to DEBUG and not configurable (as well as any other logging parameters, like log-file name, size before get rotated, history size, etc.)
We can use env variables or introduce winpanda's own config file to get everything configurable.

packages/winpanda/extra/src/winpanda/svcm/nssm.py
anatolii-borysov: Fixed.

packages/winpanda/extra/src/winpanda/bin/winpanda.py
anatolii-borysov: Fixed.

jongiddy:
@mesosphere-mergebot changelog-not-required Part of addition of Windows support

PR #6234, opened 2019-09-09
Add marathon_new_group_enforce_role config option to the DC/OS installer

timcharper:
## High-level description

This adds a new config option to the DC/OS installer, marathon_new_group_enforce_role, and defaults it to 'top'. This causes new services for not-yet-created groups to default to use the group-role, rather than the default mesos role. The same thing happens when we 

  - [DCOS_OSS-5532](https://jira.mesosphere.com/browse/DCOS_OSS-5532) Change the default for root marathon to enable enforceRole by default for new top-level groups


timcharper:

PR #6269, opened 2019-09-17
Check if the returned JSON from marathon actually is an empty list or nil

ANeumann82:
## High-level description
Fix for an edge case where the marathon apps endpoint returns null for the task list instead of an empty array

  - [DCOS_OSS-5541](https://jira.mesosphere.com/browse/DCOS_OSS-5541) AdminRouter Cache Fails When Marathon returns no task field


ANeumann82:

jongiddy:
@ANeumann82 Can you please add a description to the `CHANGES.md` file.

PR #6270, opened 2019-09-17
Bbannier/boost 1.65.0

bbannier:
## High-level description

  - [DCOS_OSS-5555](https://jira.mesosphere.com/browse/DCOS_OSS-5555) DC/OS Boost version is older than the one used and tested by Apache Mesos upstream

## Related tickets (optional)


packages/boost-libs/buildinfo.json
```diff
@@ -1,7 +1,7 @@
 {
   "single_source": {
     "kind": "url_extract",
-    "url": "https://downloads.mesosphere.com/pkgpanda-artifact-cache/boost_1_53_0.tar.gz",
-    "sha1": "0e4ef26cc7780c6bbc63987ef2f29be920e2395b"
+    "url": "https://dl.bintray.com/boostorg/release/1.65.0/source/boost_1_65_0.tar.gz",
```
bbannier: @orsenthil, do we still want to upload the file to downloads.mesosphere.com? This seems to be rarely used by packages, and we e.g., already download a number of Python packages from the outside.

orsenthil: @bbannier  - It is okay to keep at the bintray.  My guess is, we assumed our internally maintained  s3 location will be stabler than bintray.

Also, when we are using in a CI, some domains could block us if there are frequent number of requests in a time-period. 

We could try this again by using bintray directly.

bbannier:
@mesosphere-mergebot bump-ee

packages/boost-libs/buildinfo.json
lava: Hm, per the bintray FAQ:

> The current limit per non-Premium user is 300 queries a day, and 1440 per month. Limits are returned in the X-RateLimit headers.

although I'm not sure to what exactly that limit applies.

Given that the uploading procedure looks relatively straightforward ( https://wiki.mesosphere.com/display/ENG/ROQ+-+Publishing+to+downloads.mesosphere.com ), it might be worth doing it just to keep things consistent.

PR #6447, opened 2019-10-24
Pin CLI version

bamarni:
This pins the DC/OS CLI version to a specific tag in order to reduce
moving dependencies in the CI.

We should bump it to the latest version occasionally, possibly having a
way to trigger DC/OS integration tests from CLI PRs.

Related to https://jira.mesosphere.com/browse/DCOS-60349

### DIFF
--- a/packages/dcos-integration-test/extra/conftest.py
+++ b/packages/dcos-integration-test/extra/conftest.py
@@ -216,8 +216,8 @@ def install_dcos_cli(tmpdir_factory: TempdirFactory):
     """
     tmpdir = tmpdir_factory.mktemp('dcos_cli')
     cli = dcos_cli.DcosCli.new_cli(
-        download_url='https://downloads.dcos.io/cli/releases/binaries/dcos/linux/x86-64/latest/dcos',
-        core_plugin_url='https://downloads.dcos.io/cli/releases/plugins/dcos-core-cli/linux/x86-64/dcos-core-cli-2.0-patch.2.zip',  # noqa: E501
+        download_url='https://downloads.dcos.io/cli/releases/binaries/dcos/linux/x86-64/1.1.1/dcos',
+        core_plugin_url='https://downloads.dcos.io/cli/releases/plugins/dcos-core-cli/linux/x86-64/dcos-core-cli-2.0-patch.3.zip',  # noqa: E501
         ee_plugin_url='https://downloads.mesosphere.io/cli/releases/plugins/dcos-enterprise-cli/linux/x86-64/dcos-enterprise-cli-1.13-patch.1.zip',  # noqa: E501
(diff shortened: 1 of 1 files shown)


cgawron1:
@mesosphere-mergebot sync

cgawron1:
@mesosphere-mergebot bump-ee

bamarni:
@mesosphere-mergebot changelog-not-required just pinning a test dependency

bamarni:
@mesosphere-mergebot sync

bamarni:

bamarni:
@mesosphere-mergebot request-review @drewvanstone @br-lewis @makkes @janisz

packages/dcos-integration-test/extra/conftest.py
```diff
@@ -216,8 +216,8 @@ def install_dcos_cli(tmpdir_factory: TempdirFactory):
     """
     tmpdir = tmpdir_factory.mktemp('dcos_cli')
     cli = dcos_cli.DcosCli.new_cli(
-        download_url='https://downloads.dcos.io/cli/releases/binaries/dcos/linux/x86-64/latest/dcos',
-        core_plugin_url='https://downloads.dcos.io/cli/releases/plugins/dcos-core-cli/linux/x86-64/dcos-core-cli-2.0-patch.2.zip',  # noqa: E501
+        download_url='https://downloads.dcos.io/cli/releases/binaries/dcos/linux/x86-64/1.1.1/dcos',
```
janisz: Why not using latest?

bamarni:
@makkes @janisz : Pinning dependencies is one of the initiatives that's being done in DC/OS integration tests to make them more stable, if we keep a moving dependency here, we are still subject to issues like [DCOS-60349](https://jira.mesosphere.com/browse/DCOS-60349), where CI was failing for a whole day and blocked multiple other teams PRs.

We could for example run DC/OS integration tests as part of the CLI release process, but until it's the case the only way to guarantee stable CI is to pin the version.

orsenthil:
Hi @bamarni  / @makkes  / @janisz  :+1: 

The motivation here was for stability. That said, I understand how *latest* version could help with testing  the software that we own.

- I am fine with using the latest version of CLI - as long as "latest" is always stable and any regression caused is fixed immediately by revert / fix.
- Automated bump of CLI is fine too. It could be worked out if desired. This is similar to how Mesos bumps it's packages.

bamarni:
@orsenthil : `latest` is indeed the latest stable release (latest tag). We are running unit/integration tests against DC/OS before publishing a new release and make sure to add non-regression tests when we find issues.

In general it's stable but there might be occasional failures like this in the future for some edge-cases we are not testing, we can however react quickly to fix / revert the release when needed. So I'll stick to latest as suggested.

PR #6454, opened 2019-10-25
Replace docker-gc with docker system prune

janisz:
## High-level description

docker-gc is inactive and was replaced by `docker system prune`.
See: https://github.com/spotify/docker-gc/commit/6a5734c6f0ae20b3a838d2fed077982141a3b81a#diff-04c6e90faac2675aa89e2176d2eec7d8

  - [DCOS_OSS-5441](https://jira.mesosphere.com/browse/DCOS_OSS-5441)

## Related tickets (optional)


janisz:
@mesosphere-mergebot bump-ee

packages/docker-gc/extra/dcos-docker-gc.service
```diff
@@ -1,16 +1,10 @@
...
-LimitNOFILE=16384
-User=dcos_docker_gc
-EnvironmentFile=/opt/mesosphere/environment
-EnvironmentFile=-/var/lib/dcos/docker-gc.env
-Environment=STATE_DIR=/var/lib/dcos/docker-gc
-Environment=PID_DIR=/var/lib/dcos/docker-gc
-Environment=LOG_TO_SYSLOG=0
-Environment=REMOVE_VOLUMES=1
-ExecStart=$PKG_PATH/docker-gc
+ExecStart=/bin/sh -c 'docker system prune --volumes --force'
```
makkes: This would prune all the stopped containers, even those stopped just a second ago. `docker-gc` only removed old containers (1h?). 

You might want to have a look at https://github.com/spotify/docker-gc/blob/539c30b823f01278a968f6c360c5963e75396f25/docker-gc#L22-L25

packages/docker-gc/extra/dcos-docker-gc.service
```diff
@@ -1,16 +1,10 @@
...
+
+[Install]
+WantedBy=default.target
```
makkes: What do we need this for?

packages/docker-gc/extra/dcos-docker-gc.service
janisz: Yes I know about this but the only solution to this problem is to create a script and manually filter images by container exit time. I think it's OK to remove everything. At least it's clean and we do not any additional logic. I understand that this might increase startup time of failed container but this is the trade off we need to make to keep things simple.

janisz: `dcoker system prune`  removes only "unreferenced-dangling" images not all old images. So it should be safe option. If container if flaking then it probably will not be deleted since it will be still in use.

packages/docker-gc/extra/dcos-docker-gc.service
janisz: Removed

janisz:
@mesosphere-mergebot request-review @drewvanstone @br-lewis @makkes @bamarni @rgo3

janisz:
:wave: @drewvanstone @br-lewis @makkes @bamarni @rgo3  Can you review, please? I added changelog entry

PR #6468, opened 2019-10-31
Winpanda App: Fix Misprocessing Escape Sequences in JSON

anatolii-borysov:
1) add screening stage to processing JSON-formatted data
   to eliminate improper recognition/handling of escape
   sequences  by the standard JSON-parser
2) add loading package info descriptor to package manifest

JIRA: DCOS_OSS-5622 - Winpanda App: Fix Misprocessing
                      Escape Sequences in JSON

## High-level description

## Related tickets (optional)

### DIFF
--- a/packages/winpanda/extra/src/winpanda/core/package.py
+++ b/packages/winpanda/extra/src/winpanda/core/package.py
@@ -102,11 +102,9 @@ class PackageManifest:
         self.pkgrepo_dpath = pkgrepo_dpath
         self.pkgactive_dpath = pkgactive_dpath
 
-        # TODO: Add package info descriptor handling
-        # self.pkg_info = (
-        #     pkg_info if isinstance(pkg_info, dict) else self._load_pkg_info()
-        # )
-        self.pkg_info = {}
+        self.pkg_info = (
+            pkg_info if isinstance(pkg_info, dict) else self._load_pkg_info()
+        )
         # TODO: Add package initialization descriptor handling
         # self.pkg_ini = (
         #     pkg_ini if isinstance(pkg_ini, dict) else self._load_pkg_ini()
--- a/packages/winpanda/extra/src/winpanda/core/utils.py
+++ b/packages/winpanda/extra/src/winpanda/core/utils.py
@@ -3,10 +3,16 @@
 Core utility routine definitions.
 """
 import configparser as cfp
+import io
 import json
 from pathlib import Path
+import re
 
+from common import logger
 from core import exceptions as cr_exc
+
+
+LOG = logger.get_logger(__name__)
 
 
 def rc_load_json(fpath, emheading=None):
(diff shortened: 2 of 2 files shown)


jongiddy:
As I understand this problem, `pkgpanda.build` is adding the string `C:\dcos\thing` into the `pkginfo.json` file as the characters `"C:\dcos\thing"` rather than the correct JSON encoding of `"C:\\dcos\\thing"`? This appears to be a bug in `pkgpanda` and should be fixed there, so that this additional code is not required.

anatolii-borysov:
@mesosphere-mergebot changelog-not-required

anatolii-borysov:

anatolii-borysov:

anatolii-borysov:
@jongiddy wrote on Fri 11/1/2019 11:05 AM
Ref: https://github.com/dcos/dcos/pull/6468#issuecomment-548716546
> As I understand this problem, `pkgpanda.build` is adding the string `C:\dcos\thing` into the `pkginfo.json` file as the characters `"C:\dcos\thing"` rather than the correct JSON encoding of `"C:\\dcos\\thing"`? This appears to be a bug in `pkgpanda` and should be fixed there, so that this additional code is not required.

I believe that a good application should implement two startegies on processing external input:
**1) "resilience"** - to be able to survive and still accomplish a task being provided with incomplete or malformed data by augmenting it to some reasonable degree in some predefined way (in accordance with applications logic). For eaxample FS path normalization operations (os.path.normcase()/normpath(), encoding URL-incompatible characters (urllib.parse.quote()) or use hardcoded defaults for missed params.
**2) "fail fast"** - to fail immediately when reasonable preprocessing is not going to provide acceptable content anyway. For example when data formatting rules are unrecoverably broken (JSON-formatting) or there is no way to provide any reasonable default vaule for a missed part of critical data.

This fix follows the approach described above and allows not to decline the whole DC/OS build (a whole deployment is obviously to fail, if a single DC/OS package is malformed) and makes the whole system more tolerate to minor discrepancies in config data.

If the considerations above still look unreasonable for you we can decline these changes.

jongiddy:
@anatolii-borysov  This code:
- adds complexity to a simple piece of code, without an actual problem being observed.
- is not obvious as to what it does. I know how JSON is parsed, but I don't know with a quick look at this what non-JSON is permitted.
- appears to break valid JSON. How do I provide a tab character (`\t`) or a newline (`\n`) in a JSON string with this code?
- does not appear to be complete - what if the path component starts with a capital letter (`A-Z`)? Or a `_`?

In this case it is much better to fail the installation rather than guess the correct value. We are parsing code-generated JSON here, not typed user input. DC/OS packages are created using an automated process that should not create malformed contents. If they do, we should fix them, and we want to see the failures so we know to fix them.

So I will decline the changes to JSON parsing, although I expect the simple change to `packages/winpanda/extra/src/winpanda/core/package.py` is still required?

anatolii-borysov:
OK.
Just decline the whole PR then.
I'll add that small change to the packages/winpanda/extra/src/winpanda/core/package.py to load the pkginfo.json along with the next PR, addressing the https://jira.mesosphere.com/browse/DCOS_OSS-5570.

jongiddy:
@mesosphere-mergebot label Holding

PR #6473, opened 2019-11-01
docker-gc: don't run `system prune`

jkoelker:
## High-level description

`docker system prune` will remove networks that do not have containers
currently attached to them. This will remove the `dcos` and `dcos6`
network if they are not in use. Instead prune just the objects desired.

  - [DCOS_OSS-5441](https://jira.mesosphere.com/browse/DCOS_OSS-5441) Switch to `docker system prune`.

### DIFF
--- a/packages/docker-gc/extra/dcos-docker-gc.service
+++ b/packages/docker-gc/extra/dcos-docker-gc.service
@@ -4,4 +4,7 @@ ConditionPathExists=/opt/mesosphere/etc/docker_gc_enabled
 
 [Service]
 Type=simple
-ExecStart=/bin/sh -c 'docker system prune --volumes --force'
+ExecStartPre=-/bin/sh -c 'docker container prune --force'
+ExecStartPre=-/bin/sh -c 'docker image prune --force'
+ExecStartPre=-/bin/sh -c 'docker volume prune --force'
+ExecStart=/bin/true


jkoelker:
@mesosphere-mergebot changelog-not-required service ordering fix for regression from #6454

jkoelker:
@mesosphere-mergebot request-review @drewvanstone @br-lewis @makkes @bamarni @rgo3 @janisz @urbanserj

packages/docker-gc/extra/dcos-docker-gc.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Docker GC: periodically garbage collects Docker containers and images
 ConditionPathExists=/opt/mesosphere/etc/docker_gc_enabled
+Before=dcos-mesos-slave.service
```
urbanserj: https://github.com/dcos/dcos/blob/master/docs/systemd-rules.md

urbanserj: > `Before=`, `After=` are discouraged. They are not strong guarantees

janisz: Can we use filter to not remove dcos networks?

jkoelker: :+1: while I agree in general, I would advocate that this be one of the exceptions, as it stands its a coin flip as to if docker apps can use container networks.

urbanserj: @janisz I'd probably run `docker container prune`, `docker image prune`, etc, but not `docker network prune` or `docker system prune`.

jkoelker: > Can we use filter to not remove dcos networks?

Not currently, as no labels are added to the network to be able to filter on

jkoelker: @urbanserj @janisz I like the idea of just running the `container` and `image` `prunes`. If that's amenable to all, I'll change to that here.

urbanserj:
@mesosphere-mergebot changelog-not-required bug fix

PR #6538, opened 2019-11-19
Winpanda App: Introduce Parametrization For Package Config Files

anatolii-borysov:
1) introduce basic unified set of standard placeholder names
   for use in package configuration files
2) implement resource rendering context type to incapsulate
   all the individual rendering context elements (variables)
3) generalize approach to placeholder substitution across all
   package configuration sources
4) improve error handling

JIRA: DCOS_OSS-5570 - Winpanda App: Add Support For Extra
                      DC/OS Package Installation Helpers

## High-level description

## Related tickets (optional)


anatolii-borysov:

anatolii-borysov:

packages/winpanda/extra/src/winpanda/core/command.py
```diff
@@ -212,52 +217,57 @@ def execute(self):
...
+            if pkg_manifest.pkg_svccfg:
+                svc_manager = WinSvcManagerNSSM(
+                    svc_conf=pkg_manifest.pkg_svccfg
                 )
-                svc_status = str(stdout).strip().rstrip('\n')
+            else:
+                svc_manager = None
 
-            if svc_status == SVC_STATUS.STOPPED:
+            if svc_manager:
```
jongiddy: With this change, this `if` body can be merged into the previous `if`.

packages/winpanda/extra/src/winpanda/core/rc_ctx.py
```diff
@@ -0,0 +1,207 @@
...
+        for retriever in retrievers:
+            items.update(retriever(json_ready))
+
+        return items
+
+    def _get_istor_items(self, json_ready=False):
+        """Discover resource rendering context items from DC/OS installation
+        storage configuration.
+
+        :param json_ready: bool, get JSON-compatible context items, if True
```
jongiddy: Why do we need this? Is stripping double-quotes the only extra processing needed? e.g. could there be escapes in the string?

anatolii-borysov: Not exactly... When we do preliminary Jinja rendering for a JSON-formatted template we can face with the situation when our variable for substitution contains backslahes. So after rendering that template would contain character sequences wrongly interpreted by JSON-parser as escape sequences and causing parsing errors. This is the exact case for substituting variable parts of any FS path on windows. So, to handle such cases we have to screen every single back slash in such variables before doing rendering. I suppose that the most correct approach to give this work to the standard JSON serializer. And ater the JSON0-serializer does all the escaping in variables the only thing laft is to strip double quotes.

packages/winpanda/extra/src/winpanda/core/command.py
anatolii-borysov: Good point. Thanks. Will fix it in the next commit to this PR.

packages/winpanda/extra/src/winpanda/common/storage.py
```diff
@@ -371,29 +381,81 @@ def add_package(self, pkg_id, dstor_root_url, dstor_pkgrepo_path):
...
+                    shutil.copytree(
+                        temp_dpath, str(pkg_inst_dpath)
+                    )
+
             LOG.debug(f'{msg_src}: Add package: Extract: {pkg_id}')
         except Exception as e:
-            raise cr_exc.RCExtractError(
-                f'Add package: {pkg_id}: {type(e).__name__}: {e}'
-            )
+            if type(e) is not cr_exc.RCExtractError:
```
jongiddy: Idiomatic Python is to use `not isinstance(e, cr_exc.RCExtractError)`

packages/winpanda/extra/src/winpanda/common/utils.py
```diff
@@ -95,3 +97,36 @@ def rmdir(path, recursive=False):
...
+
+    :param cl_elements: list, elements of command line, beginning with
+                        executable name
+    :param timeout:     int|float, forcibly terminate execution, if this number
+                        seconds passed since execution was started
+    :return:            subprocess.CompletedProcess, results of sub-process
+                        execution
+    """
+    # cl_elements = cl_elements if isinstance(cl_elements, list) else []
+    timeout = timeout if isinstance(timeout, (int, float)) else 30
```
jongiddy: The only way this takes the `else` path is if the caller passes a non-numeric parameter. This should be allowed to fail rather than hiding the problem.

jongiddy:
@mesosphere-mergebot changelog-not-required Part of overall Windows changes

PR #2833, opened 2018-05-01
Add support for CoreOS 1688.4.0

urbanserj:
## High-level description

Add support for CoreOS 1688.4.0

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS_OSS-2417](https://jira.mesosphere.com/browse/DCOS_OSS-2417) Add support for CoreOS 1688.4.0

## Related tickets (optional)

Other tickets related to this change:

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change: let's mention this change only after we migrate all our AIMs on coreos 16xx.
  - [x] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: none.
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and review process

**What is the review process and when will my changes land?**

All PRs require 2 approvals using GitHub's [pull request reviews](https://help.github.com/articles/about-pull-request-reviews/).

Reviewers should be:
* Developers who understand the code being modified.
* Developers responsible for code that interacts with or depends on the code being modified.

It is best to proactively ask for 2 reviews by @mentioning the candidate reviewers in the PR comments area. The responsibility is on the developer submitting the PR to follow-up with reviewers and make sure a PR is reviewed in a timely manner. Once a PR has **2 ship-it's**, **no red reviews**, and **all tests are green** it will be included in the [next train](https://github.com/dcos/dcos/blob/master/contributing.md).

PR #6548, opened 2019-11-21
[master] set interfaces as unmanaged by coreos networkd

mainred:
## High-level description

This was first introduced in https://github.com/dcos/dcos/pull/2833 which set network interfaces used by DC/OS as unmanaged by networkd on all releases of Linux, and before this happens only on coreos. 
This may bring errors when networkd is enabled on other Linux releases, refer to [cops ticket explanation](https://jira.mesosphere.com/browse/COPS-5575?focusedCommentId=303879&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-303879) for detailed explanation.

Here we change the configuration takes effect only on coreos, and since coreos 1409.x.0, unmanaged link is supported in coreos networkd. But the interfaces are unmanaged by default.
```
[root@osboxes2 ~]# networkctl
IDX LINK             TYPE               OPERATIONAL SETUP
  1 lo               loopback           carrier     unmanaged
  2 enp0s3           ether              routable    unmanaged
  3 enp0s8           ether              routable    unmanaged
  4 virbr0           ether              no-carrier  unmanaged
  5 virbr0-nic       ether              off         unmanaged
  6 m-dcos           ether              routable    unmanaged
  7 docker0          ether              routable    unmanaged
  8 d-dcos           ether              no-carrier  unmanaged
 12 vethfe3cdb4      ether              degraded    unmanaged
 13 tunl0            tunnel             routable    unmanaged
[root@osboxes2 ~]# systemctl --version
systemd 219
+PAM +AUDIT +SELINUX +IMA -APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ +LZ4 -SECCOMP +BLKID +ELFUTILS +KMOD +IDN
[root@osboxes2 ~]# cat /etc/*release
CentOS Linux release 7.6.1810 (Core)
... ...
```

Also I tested on Ubuntu 16.04.6 LTS and Ubuntu 18.04.3 LTS, the interfaces are unmanaged by default when networkd is active.

  - [DCOS-60956](https://jira.mesosphere.com/browse/DCOS-60956) COPS-5575: Nodes and applications are not pingable on DC/OS 2.0.

## Related tickets (optional)

### DIFF
--- a/packages/dcos-net/extra/dcos-net-setup.py
+++ b/packages/dcos-net/extra/dcos-net-setup.py
@@ -19,17 +19,20 @@ import datetime
 import filecmp
 import logging
 import os
+import platform
 import shutil
 import subprocess
 import sys
 
 
 def main():
+    return_code = 0
     if sys.argv[1:4] in [['ip', 'link', 'add'], ['ip', 'addr', 'add'], ['ip', '-6', 'addr']]:
         result = subprocess.run(sys.argv[1:], stderr=subprocess.PIPE)
         sys.stderr.buffer.write(result.stderr)
         if result.stderr.strip().endswith(b'File exists'):
             result.returncode = 0
+        return_code = result.returncode
     elif sys.argv[1] == 'iptables':
         # check whether a rule matching the specification does exist
         argv = ['-C' if arg in ['-A', '-I'] else arg for arg in sys.argv[1:]]
@@ -37,20 +40,32 @@ def main():
         if result.returncode != 0:
             # if it doesn't exist append or insert that rules
             result = subprocess.run(sys.argv[1:])
+        return_code = result.returncode
     elif sys.argv[1] == '--ipv6':
         if os.getenv('DCOS_NET_IPV6', 'true') == 'false':
             sys.exit(0)
         else:
             del sys.argv[1]
             result = subprocess.run(sys.argv)
+            return_code = result.returncode
     elif sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
-        result = add_networkd_config(sys.argv[3])
+        return_code = add_networkd_config_for_coreos(sys.argv[3])
     else:
         result = subprocess.run(sys.argv[1:])
-    sys.exit(result.returncode)
+        return_code = result.returncode
+    sys.exit(return_code)
 
 
-def add_networkd_config(src):
+def add_networkd_config_for_coreos(src: str) -> int:
+    # systemd-networkd, when enabled, will wipe the configurations like IP
+    # address of network interfaces and this behavior happens only on coreos
+    # This problem is tracked by:
+    # https://jira.mesosphere.com/browse/DCOS_OSS-1790
+    # We need to mark interfaces managed by DC/OS as unmanaged when networkd is
+    # enabled on coreos
+    if platform.system() != "Linux" or "coreos" not in platform.release():
+        return 0
+
     networkd = b'systemd-networkd.service'
     networkd_path = '/etc/systemd/network'
 
@@ -58,9 +73,9 @@ def add_networkd_config(src):
     result = subprocess.run(['systemctl', 'list-unit-files', networkd],
                             stdout=subprocess.PIPE)
     if result.returncode != 0:
-        return result
+        return result.returncode
     if networkd not in result.stdout:
-        return result
+        return result.returncode
 
     # Copy the configuration
     bname = os.path.basename(src)
@@ -76,7 +91,7 @@ def add_networkd_config(src):
                             stdout=subprocess.PIPE)
     if result.returncode != 0:
         result.returncode = 0
-        return result
+        return result.returncode
 
     # Restart networkd only if the configuration is updated
     mtime = os.path.getmtime(dst)
@@ -90,13 +105,13 @@ def add_networkd_config(src):
                 active_enter_timestamp,
                 '%a %Y-%m-%d %H:%M:%S %Z')
             if started.timestamp() > mtime:
-                return result
+                return result.returncode
         except ValueError:
             logging.warning('Unexpected ActiveEnterTimestamp value: "%s"',
                             active_enter_timestamp)
 
     # Restart networkd
-    return subprocess.run(['systemctl', 'restart', networkd])
+    return subprocess.run(['systemctl', 'restart', networkd]).returncode
 
 
 def safe_filecmp(src, dst):
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,3 +50,6 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 ### Fixed and improved
 
 * Reserve all agent VTEP IPs upon recovering from replicated log. (DCOS_OSS-5626)
+
+* Set network interfaces as unmanaged for networkd only on coreos. (DCOS-60956)
+


mainred:
@mesosphere-mergebot request-review  @ichernetsky @GoelDeepak @urbanserj

mainred:
@GoelDeepak could you please take a look? I searched through the ticket, and find you introduced the networkd management into mesos startup script and @urbanserj move it to dcos-net.

mainred:
In DC/OS 2.0, we are not going to support CoreOS, but it's safe to keep this code. And we need to backport the PR to the N-2 DC/OS branch.

https://docs.d2iq.com/mesosphere/dcos/version-policy/#dcos-platform-version-compatibility-matrix
```
NOTE: Starting with DC/OS 2.0, we have discontinued support for CoreOS and have ceased qualifying it as an Operating System. This chart shows the final qualified versions of DC/OS for the final qualified versions CoreOS.
```

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
+    if sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
+        result_code = add_networkd_config_for_coreos(sys.argv[3])
+        sys.exit(result_code)
+
+    result = subprocess.run(sys.argv[1:])
 
 
-def add_networkd_config(src):
+def add_networkd_config_for_coreos(src: str) -> int:
+    # systemd-netword, when enabled, will wipe the configurations like IP
```
ichernetsky: typo: `netword`

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
             result = subprocess.run(sys.argv)
-    elif sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
-        result = add_networkd_config(sys.argv[3])
-    else:
-        result = subprocess.run(sys.argv[1:])
-    sys.exit(result.returncode)
+        sys.exit(result.returncode)
+
+    if sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
+        result_code = add_networkd_config_for_coreos(sys.argv[3])
```
ichernetsky: How about `return_code` for the sake of consistency in this function?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
     networkd_path = '/etc/systemd/network'
 
     # Check if there is networkd
-    result = subprocess.run(['systemctl', 'list-unit-files', networkd],
-                            stdout=subprocess.PIPE)
-    if result.returncode != 0:
-        return result
-    if networkd not in result.stdout:
-        return result
+    list_unit_ret = subprocess.run(['systemctl', 'list-unit-files', networkd],
```
ichernetsky: `list_unit_files_ret`?

ichernetsky: Or simply `result`?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -75,31 +91,31 @@ def add_networkd_config(src):
...
+        return 0
 
     # Restart networkd only if the configuration is updated
     mtime = os.path.getmtime(dst)
-    result = subprocess.run(['systemctl', 'show', '--value',
-                             '--property', 'ActiveEnterTimestamp',
-                             networkd], stdout=subprocess.PIPE)
-    if result.returncode == 0:
-        active_enter_timestamp = result.stdout.strip().decode()
+    show_property_ret = subprocess.run(['systemctl', 'show', '--value',
```
ichernetsky: `result`?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -75,31 +91,31 @@ def add_networkd_config(src):
...
             if started.timestamp() > mtime:
-                return result
+                return show_property_ret.returncode
         except ValueError:
             logging.warning('Unexpected ActiveEnterTimestamp value: "%s"',
                             active_enter_timestamp)
 
     # Restart networkd
-    return subprocess.run(['systemctl', 'restart', networkd])
+    restart_ret = subprocess.run(['systemctl', 'restart', networkd])
```
ichernetsky: ditto?

packages/dcos-net/extra/dcos-net-setup.py
mainred: We have several different commands, and it's a better practice to give the result different names in case of misuse, what do you think?

packages/dcos-net/extra/dcos-net-setup.py
mainred: please see the above reply.

packages/dcos-net/extra/dcos-net-setup.py
mainred: please see the above reply.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
         sys.stderr.buffer.write(result.stderr)
         if result.stderr.strip().endswith(b'File exists'):
             result.returncode = 0
-    elif sys.argv[1] == 'iptables':
+        sys.exit(result.returncode)
```
GoelDeepak: The reason why the earlier code was consuming the errors was that it was harmless. I think with this new code you will have issues with the restart of `dcos-net`

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
+
+    if sys.argv[1] == 'iptables':
         # check whether a rule matching the specification does exist
         argv = ['-C' if arg in ['-A', '-I'] else arg for arg in sys.argv[1:]]
         result = subprocess.run(argv)
         if result.returncode != 0:
             # if it doesn't exist append or insert that rules
             result = subprocess.run(sys.argv[1:])
-    elif sys.argv[1] == '--ipv6':
+        sys.exit(result.returncode)
```
GoelDeepak: I would still inclined to keep the previous code.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
             sys.exit(0)
         else:
             del sys.argv[1]
             result = subprocess.run(sys.argv)
-    elif sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
-        result = add_networkd_config(sys.argv[3])
-    else:
-        result = subprocess.run(sys.argv[1:])
-    sys.exit(result.returncode)
+        sys.exit(result.returncode)
```
GoelDeepak: ditto

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -30,38 +31,53 @@ def main():
...
+
+    if sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
+        return_code = add_networkd_config_for_coreos(sys.argv[3])
+        sys.exit(return_code)
+
+    result = subprocess.run(sys.argv[1:])
```
mainred: We missed throwing the return code here.

packages/dcos-net/extra/dcos-net-setup.py
mainred: The issues can also be consumed because we remained `result.returncode = 0` before `sys.exit(result.returncode)`?

packages/dcos-net/extra/dcos-net-setup.py
mainred: OK, let's keep the minimal change.

mainred:
rebased changes against master.
Thanks @ichernetsky and @GoelDeepak , could you please give another pass?

packages/dcos-net/extra/dcos-net-setup.py
ichernetsky: AFAIK, there is a best practice for every single way of doing things. To me, it is a matter of writing a blog post and calling it a best practice :)

Personally, I don't think it improves readability and/or makes the code safer. Just my $.02, but it is your call.

GoelDeepak: I am on the fence when I see changes like these. Personally, I like to see minimal code changes as long as there isn't a pressing need to make other changes. Remember any code that you touch has the potential of introducing a regression. Having said that I would be inclined to use the previous code

mainred: I did not say it's the best practice, see the code below
```
ret = a + b
if ret > 10:
    return ret

retu = c + d
if ret > 10:
    return ret
```

mainred: I reverted the code.

mainred:
@GoelDeepak @ichernetsky  could you please take another look? thanks.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -72,34 +89,34 @@ def add_networkd_config(src):
         shutil.copyfile(src, dst)
 
     # Restart networkd only if it's active
-    result = subprocess.run(['systemctl', 'is-active', networkd],
-                            stdout=subprocess.PIPE)
-    if result.returncode != 0:
-        result.returncode = 0
-        return result
+    active_check_ret = subprocess.run(['systemctl', 'is-active', networkd],
```
GoelDeepak: do we really need this change?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -72,34 +89,34 @@ def add_networkd_config(src):
...
+        return 0
 
     # Restart networkd only if the configuration is updated
     mtime = os.path.getmtime(dst)
-    result = subprocess.run(['systemctl', 'show', '--value',
-                             '--property', 'ActiveEnterTimestamp',
-                             networkd], stdout=subprocess.PIPE)
-    if result.returncode == 0:
-        active_enter_timestamp = result.stdout.strip().decode()
+    show_property_ret = subprocess.run(['systemctl', 'show', '--value',
```
GoelDeepak: ditto

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -72,34 +89,34 @@ def add_networkd_config(src):
...
             if started.timestamp() > mtime:
-                return result
+                return show_property_ret.returncode
         except ValueError:
             logging.warning('Unexpected ActiveEnterTimestamp value: "%s"',
                             active_enter_timestamp)
 
     # Restart networkd
-    return subprocess.run(['systemctl', 'restart', networkd])
+    restart_ret = subprocess.run(['systemctl', 'restart', networkd])
```
GoelDeepak: ditto

packages/dcos-net/extra/dcos-net-setup.py
mainred: Will these bring error?

mainred: What I really changed is the return value variable, and the explanation can be found in the above discussion.
> Remember any code that you touch has the potential of introducing a regression
Yeah, we should be as cautious as possible, and you know what, it's just reusing the same variable `result` makes this code much harder to maintain to even let you change a variable.

mainred: @GoelDeepak , what do you think about this?

mainred:
@mesosphere-mergebot request-review @vespian

packages/dcos-net/extra/dcos-net-setup.py
ichernetsky: I think what Deepak meant is that `result` would do the job here too. I am approving the PR, but please consider using just `result` instead of `*_ret`. Thanks.

mainred: Then I am answering the right question.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -19,49 +19,66 @@
...
         result = subprocess.run(sys.argv[1:])
-    sys.exit(result.returncode)
+        return_code = result.returncode
+    sys.exit(return_code)
 
 
-def add_networkd_config(src):
+def add_networkd_config_for_coreos(src: str) -> int:
+    # systemd-networkd, when enabled, will wipe the configurations like IP
+    # address of network interfaces and this behavior happens only on coreos
```
vespian: Please help me understand something - are the interfaces left as `unmanaged` on all distros except coreos? Or does it depend only on the systemd version?

mainred: For non-core distributions supported by us, the default behavior is unmanaged, we don't have to add DC/OS network interfaces into the unmanaged interface list, on the contrary, if the systemd version is lower than expected, the interfaces will misbehave.
For core os, we need to set the interfaces into the unmanaged interface list when networkd is enabled, otherwise, the ip address of DC/OS interfaces will be lost, and as networkd is big features in coreos, we did not add document to disable it. Also, the systemd of all supported coreos support unmanaged parameter.

mainred:
@ichernetsky  @GoelDeepak , I  removed all additional changes in this PR. Looking forward to your another pass.

cgawron1:
@mesosphere-mergebot sync

PR #5376, opened 2019-05-01
Support large uploads and processing times for services via AR

jongiddy:
## High-level description

Support large uploads and long processing times for services accessed through Admin Router's `/service` endpoint. This supports package upload, required for package registries launched using Marathon.

This has been tested manually using the 10GB `beta-mesosphere-jupyter-service-1.3.0-0.35.4-beta.dcos` package.

There is a separate issue [DCOS-51222](https://jira.mesosphere.com/browse/DCOS-51222) for creation of an integration test. I consider this separation to be appropriate in this case, as there are non-trivial problems to be overcome for the test (time and bandwidth usage).

## Corresponding DC/OS tickets (obligatory)

These DC/OS JIRA ticket(s) must be updated (ideally closed) in the moment this PR lands:

  - [DCOS-52768](https://jira.mesosphere.com/browse/DCOS-52768) COPS-4651 repo 413 Entity too large / 504 timeout

## Related tickets (optional)

Other tickets related to this change:

  - [COPS-4651](https://jira.mesosphere.com/browse/COPS-4651) 413 Entity Too Large error when trying to add any packages where the .dcos bundled file is over 1 GB

## Checklist for all PRs

  - [x] Added a comprehensible changelog entry to `CHANGES.md` or explain why this is not a user-facing change:
  - [ ] Included a test which will fail if code is reverted but test is not. If there is no test please explain here: Manually tested. Testing requires a large (10GB) download/upload, which takes significant time and bandwidth.
  - [x] Read the [DC/OS contributing guidelines](https://github.com/dcos/dcos/blob/master/contributing.md)
  - [x] Followed relevant code rules [Rules for Packages and Systemd](https://github.com/dcos/dcos/tree/master/docs)

## Checklist for component/package updates:

If you are changing components or packages in DC/OS (e.g. you are bumping the sha or ref of anything underneath `packages`), then in addition to the above please also include:

  - [ ] Change log from the last version integrated (this should be a link to commits for easy verification and review): [example](https://github.com/dcos/dcos-mesos-modules/compare/f6fa27d7c40f4207ba3bb2274e2cfe79b62a395a...6660b90fbbf69a15ef46d0184e36755881d6a5ae)
  - [ ] Test Results: [link to CI job test results for component]
  - [ ] Code Coverage (if available): [link to code coverage report]
___
**PLEASE FILL IN THE TEMPLATE ABOVE** / **DO NOT REMOVE ANY SECTIONS ABOVE THIS LINE**

## Instructions and review process

[...]

PR #6564, opened 2019-11-25
Further increase adminrouter's client_max_body_size - COPS-5615

jeffwhite530:
## High-level description

The following PR increased the max size to 16GB but that is still too small:  https://github.com/dcos/dcos/pull/5376/files

This change increases it to 32 GB.

  - [DCOS-61233](https://jira.mesosphere.com/browse/DCOS-61233) COPS-5615: Large package cannot be uploaded to package registry

## Related tickets (optional)

  - [DCOS-20269](https://jira.mesosphere.com/browse/DCOS-20269) Request entity too large error


vinodkone:
@jongiddy @vespian  can you review this?

packages/adminrouter/extra/src/includes/server/master.conf
```diff
@@ -285,7 +285,7 @@ location @service_requnbuffered {
     # It takes 3-4 minutes for the DC/OS package registry to validate a 10GB
     # package and respond. We relax the read timeout for all unbuffered
     # services to handle this.
-    proxy_read_timeout 300s;
+    proxy_read_timeout 600s;
```
jongiddy: Have we seen that this extra time is required? Increasing the timeout provides bad usability when the service is never going to respond.

packages/adminrouter/extra/src/includes/server/master.conf
```diff
@@ -231,10 +231,10 @@ location ~ ^/service/(?<service_path>.+) {
     set $original_uri $uri;
 
     # Give services a lot of leeway. A particular extreme is for uploading
-    # packages to a package registry. The Jupyter package is one of our
-    # largest at just under 10GB. Use 16GB as the limit to allow for this to
+    # packages to a package registry. The data-science-engine package is one
+    # of our largest at over 16GB. Use 32GB as the limit to allow for this to
     # expand.
-    client_max_body_size 16384m;
+    client_max_body_size 32768m;
```
jongiddy: I'm tempted to keep this as 24GB. Although it is slightly painful for customers when we hit the limit, it is useful to discover when package sizes actually reach these extremes. And these files can be modified on a per-customer basis when necessary.

packages/adminrouter/extra/src/includes/server/master.conf
jeffwhite530: I was basing the needed value off the comment.  Assuming that comment is based on testing large uploads, wouldn't we need to increase the value to support larger files?

Is the concern here that admin router could seem to be hung for 10 minutes before returning an error?  If so, 5 minutes is already an awfully long wait time so this would only turn an already bad thing into a worse bad thing.  Not sure what the solution for that would be.

jongiddy:
@jeffwhite530 Can you please add a description to the CHANGES.md file?

vespian:
@jeffwhite530 Is this PR still WIP? If not - could you please make the PRs as `ready for review`? 

Thank you!

jeffwhite530:
> @jeffwhite530 Is this PR still WIP? If not - could you please make the PRs as `ready for review`?
> 
> Thank you!

How do I do that?

vespian:
Please issue
```
```

Mergebot also commented this PR with detailed instructions - see https://github.com/dcos/dcos/pull/6564#issuecomment-558244940

jeffwhite530:

jongiddy:
Sorry for taking so long to verify this. I have tested, and it seems we do need to increase the timeout to 10 minutes.

@jeffwhite530 Would you mind if I add my test (https://github.com/mesosphere/dcos-enterprise/pull/7021/files) to the Enterprise branch of this PR? Then we can click the second box in the PR description.

jeffwhite530:
@jongiddy, go right ahead.  Let me know if I need to do anything for that.

jongiddy:
@jeffwhite530 Have you been able to test this in a working cluster? I'm beginning to think that the registry may have a 5 minute timeout for upload.

jeffwhite530:
@jongiddy I just tested it now and I was able to upload the package.  It doesn't seem to be working as 15 minutes later I still don't see it in the catalog.  However, this PR is still needed since when client_max_body_size is 16GB the upload immediately fails.  So we may also need to make a change in package registry but admin router definitely needs this change.

DominikDary:
@vespian & @jongiddy please approve this pr to make sure can deliver the fix to the customer.

jongiddy:
If the package not going to be uploaded successfully, then failing immediately is much better behavior than failing after ten minutes. We need to ensure that we're getting a benefit from making this change, since it may cause undesirable behavior for other services.

@takirala Is there any change needed in the package registry itself to make this work?

jongiddy:

jongiddy:
This change will be merged in #6766, so this PR can be closed. Thanks for everyone's efforts here.

PR #6576, opened 2019-11-28
Add Windows bootstrap package

jongiddy:
## High-level description

Modify DC/OS bootstrap.py to support Windows

  - [DCOS_OSS-5698](https://jira.mesosphere.com/browse/DCOS_OSS-5698) Port Open bootstrap.py to Windows

## Related tickets (optional)

### DIFF
--- a/packages/bootstrap/extra/dcos_internal_utils/utils.py
+++ b/packages/bootstrap/extra/dcos_internal_utils/utils.py
@@ -1,75 +1,139 @@
-try:
-    import fcntl
-except ImportError:
-    pass
 import json
 import logging
 import os
 import platform
+import shutil
 import socket
 import stat
 import subprocess
-import sys
 import tempfile
-
-is_windows = platform.system() == "Windows"
-
-if not is_windows:
-    assert 'fcntl' in sys.modules
+from pathlib import Path
 
 log = logging.getLogger(__name__)
 
+is_windows = platform.system() == "Windows"
+is_linux = not is_windows
 
-# Copied from pkgpanda/util.py#L257-L262
-def load_json(filename):
+if is_windows:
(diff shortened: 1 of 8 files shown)


jongiddy:

packages/bootstrap/extra/dcos_internal_utils/utils.py
```diff
@@ -1,75 +1,139 @@
...
+        tmp_file_dir = os.path.dirname(os.path.realpath(filename))
+        fd, temporary_filename = tempfile.mkstemp(prefix=prefix, dir=tmp_file_dir)
+        # `mkstemp` initially creates file with permissions 0o600
+        try:
+            try:
+                os.write(fd, data)
+            finally:
+                os.close(fd)
+            os.chmod(temporary_filename, stat.S_IMODE(mode))
+            os.replace(temporary_filename, filename)
```
vespian: Would it be a problem to move the file if it was on two different filesystems? According to docs[1]:
```
 The operation may fail if src and dst are on different filesystems. If successful, the renaming will be an atomic operation (this is a POSIX requirement).
```
[1] https://docs.python.org/3.5/library/os.html#os.replace

jongiddy: The file is always on the same filesystem, since `temporary_filename` is created in the same directory as `filename`.

vespian: Ah, sorry - I did not notice the `prefix = os.path.basename(filename)` part.

jongiddy:
Apart from the too few tests in the bootstrap package, the main test for changes to bootstrap is that CI cluster start and work for all the integration and e2e tests.

jongiddy:
@mesosphere-mergebot changelog-not-required change not user visible

PR #6602, opened 2019-12-03
Winpanda App: Add Fetching dcos-config-windows.yaml From Linux Part O…

anatolii-borysov:
…f Distribution

1) implement machanics of discovering (via reference in the
   cluster-package-info.json) and retrievieng the dcos-config-windows.yaml (the
   aggregated DC/OS configuration jbject) and the expanded.config.full.json
   (the source of variable data for substitution in the dcos-config-windows.yaml
   as parts of the dcos-config-win package from the Linux part of the DC/OS
   disctribuion storage at bootstrap node.
2) implement variables rendering (Jinja2-style) and distribution of individual
   parts of the dcos-config-windows.yaml on a target system.
3) integrate sources of variable data for substitution (both locally generated,
   like DC/OS installation local storage directory and those brought by the
   expanded.config.full.json) to be used universally across the system

JIRA: DCOS_OSS-5694 - "Winpanda App: Add Fetching dcos-config-windows.yaml From
      Linux Part Of Distribution"

## High-level description

## Related tickets (optional)

### DIFF
--- a/packages/winpanda/extra/src/winpanda/core/cmdconf.py
+++ b/packages/winpanda/extra/src/winpanda/core/cmdconf.py
@@ -5,13 +5,14 @@ Command configuration object definitions.
 import abc
 from pathlib import Path
 import posixpath
+import re
+import tempfile as tf
 
 from common import constants as cm_const
 from common import logger
 from common.cli import CLI_COMMAND, CLI_CMDOPT, CLI_CMDTARGET
-from common.storage import InstallationStorage, ISTOR_NODE
+from common.storage import InstallationStorage
 from core import exceptions as cr_exc
-from core.rc_ctx import ResourceContext
 from core import utils as cr_utl
 
 from common import utils as cm_utl
@@ -252,13 +253,20 @@ class CmdConfigSetup(CommandConfig):
         """Get the DC/OS aggregated configuration object.
 
         :return: dict, set of DC/OS shared and package specific configuration
-                 objects:
-                     {
-                         'package': {[
-                             {'path': <str>, 'content': <str>},
+                 templates coupled with 'key=value' substitution data
+                 container:
+                 {
+                    'template': {
+                        'package': [
+                            {'path': <str>, 'content': <str>},
                              ...
-                         ]}
-                     }
+                        ]
+                    },
+                    'values': {
+                        key: value,
+                        ...
+                    }
+                 }
         """
 
         dstor_root_url = (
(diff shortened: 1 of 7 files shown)


anatolii-borysov:
@mesosphere-mergebot changelog-not-required Part of owerall Windows changes

anatolii-borysov:

anatolii-borysov:

packages/winpanda/extra/src/winpanda/core/cmdconf.py
```diff
@@ -252,76 +253,173 @@ def get_dcos_conf(self):
...
         # Unblock irrelevant local operations
-        if self.cluster_conf_nop or dstor_dcoscfg_path == 'NOP':
+        if self.cluster_conf_nop or dstor_linux_pkg_index_path == 'NOP':
             LOG.info(f'{self.msg_src}: dcos_conf: NOP')
             return {}
 
-        dcoscfg_url = posixpath.join(dstor_root_url, dstor_dcoscfg_path)
-        dcoscfg_fname = Path(dstor_dcoscfg_path).name
+        # Linux package index direct URL
+        lpi_url = posixpath.join(dstor_root_url, dstor_linux_pkg_index_path)
```
jongiddy: You can use `urlparse.urljoin` to join URL's https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

packages/winpanda/extra/src/winpanda/core/cmdconf.py
```diff
@@ -252,76 +253,173 @@ def get_dcos_conf(self):
...
+                )
+            else:
+                raise
+        else:
+            return {'template': template, 'values': values}
+        finally:
+            dcoscfg_pkg_fpath.unlink()
+
+    @staticmethod
+    def load_dcos_conf_templete(fpath):
```
jongiddy: The file is YAML. You should be able to replace this function with `yaml.load`. 

Also `templete` -> `template`

anatolii-borysov: Sure. It's YAML-formatted, but at the same time it's Jinja-template. holding Jinja formatted parts. And at the same time it incorporates multiple pices that are to become individual JSON-formatted configuration files after they are extracted from the dcos-config-windows.yaml. 
So to get the final *readable* JSON-configs from a source constructed this way we need to:
1) load dcos-config-windows.yaml by YAML-parser
2) process it as Jinja template by Jinja renderer
3) save individual parts as correct JSON-formatted files for JSON-parser to be able to parse them afterwards.

So we have a difficulty here:
1) when we do substitution for final JSON-content we ar to use double backslash for pathes for filese to be valid JSON, otherwise JSON parser would complain
2) when we do substitution for any other kind of content we generally should not do it because other parers (INI, YAML) do not expect double slashes and will possibly complain

This means we can't do substitution by Jinja, until we know which final format is aimed. I.e. before the we read the whole dcos-config-windows.yaml and identified all the individual parts and their desired formatting. So we need to load it first as a whole/ But at this stage besides being YAML formatted it's still a Jinja-template. And when we try to load it by yaml.safe_load() we got complains on Jinja placeholders containing curly braces. This is why we can't use YAML native parser.

jongiddy: OK, the way I expected this to work was:

1. process the template file (using Jinja for now) - the template writer is responsible for dealing with JSON escapes
2. read the output as a YAML file
3. parse the YAML to create individual files

This requires a lot less guessing about the format required in the individual files.

Anyway, @DominikDary has requested that we merge the code to get a working demo, so I will note this as a change for GA.

anatolii-borysov: 1) OK, for the GA plan.

.2) besides substitutions done with values from the expanded.config.full.json? brought along with the dcos-config-windows.yaml we are to support substitutions for local dynamic pathnames and winpanda have to decide it at runtime for which individual part from the dcos-config-windows.yaml to do what. I.e. use path variables with double backslashes or the same variables, but featuring single slashes. This is the main problem here. And fotr the moment I can't see any easy way how do we resolve it having this complicated design (multiformatted content inside a single plain text file) for configuration files delivery.

DominikDary: what @jongiddy said for the ga frame.

DominikDary:
looks like this one is not needed any more. see https://github.com/dcos/dcos/pull/6604

packages/winpanda/extra/src/winpanda/core/cmdconf.py
anatolii-borysov: Good point. Thanks. Will change it in some later commit.

PR #6628, opened 2019-12-06
Winpanda App: Tolerate Service Start Failures

anatolii-borysov:
1) implement operation retrying helper tools
2) rework the winpanda "start" command to support retrying logic for service
   start operation
3) change logic of the winpanda "start" command for not to fail entire
   chain of service start commands, if a single service fails to start

JIRA: DCOS_OSS-5731 - Winpanda App: Tolerate Service Start Failures

### DIFF
--- a/packages/winpanda/extra/src/winpanda/core/command.py
+++ b/packages/winpanda/extra/src/winpanda/core/command.py
@@ -10,6 +10,7 @@ import jinja2 as j2
 from cfgm import exceptions as cfgm_exc
 from common import logger
 from common.cli import CLI_COMMAND, CLI_CMDTARGET, CLI_CMDOPT
+from common import utils as cm_utl
 from core import cmdconf
 from core import exceptions as cr_exc
 from core.package.id import PackageId
@@ -339,11 +340,11 @@ class CmdStart(Command):
     """Start command implementation."""
     def __init__(self, **cmd_opts):
         """Constructor."""
-        msg_src = self.__class__.__name__
+        self.msg_src = self.__class__.__name__
         super(CmdStart, self).__init__(**cmd_opts)
 
         self.config = cmdconf.create(**self.cmd_opts)
-        LOG.debug(f'{msg_src}: cmd_opts: {self.cmd_opts}')
+        LOG.debug(f'{self.msg_src}: cmd_opts: {self.cmd_opts}')
 
     def verify_cmd_options(self):
         """Verify command options."""
@@ -351,8 +352,6 @@ class CmdStart(Command):
 
     def execute(self):
         """Execute command."""
-        msg_src = self.__class__.__name__
-
         pkg_manifests = (
             self.config.inst_storage.get_pkgactive(PackageManifest.load)
         )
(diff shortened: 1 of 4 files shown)


anatolii-borysov:
@mesosphere-mergebot changelog-not-required Part of overall Windows changes

anatolii-borysov:

anatolii-borysov:

packages/winpanda/extra/src/winpanda/core/command.py
```diff
@@ -362,57 +361,76 @@ def execute(self):
...
+            try:
+                ret_code, stdout, stderr = svc_manager.status()
+                LOG.debug(f'Get final service status: {svc_name}:'
+                          f'stdout[{stdout}] stderr[{stderr}]')
+                svc_status = str(stdout).strip().rstrip('\n')
+
+                if svc_status == SVC_STATUS.START_PENDING:
+                    msg = f'Service is starting: {svc_name}'
+                    LOG.debug(msg)
+                    svcm_exc.ServiceFluctuantError(msg)  # Subject to retry
```
jongiddy: Missing `raise`

packages/winpanda/extra/src/winpanda/svcm/exceptions.py
```diff
@@ -43,3 +43,13 @@ class ServiceConfigError(ServiceError):
 class ServiceSetupError(ServiceError):
     """Service setup error."""
     pass
+
+
+class ServiceFluctuantError(ServiceError):
```
jongiddy: I had to look up "fluctuant" - it's not a common word. How about `ServiceTransientError` and `ServicePermanentError`?

packages/winpanda/extra/src/winpanda/core/command.py
anatolii-borysov: Nice shot. Thanks. Will fix it.

packages/winpanda/extra/src/winpanda/svcm/exceptions.py
anatolii-borysov: I was considering several words for this position: intermittent, temporary, unsteady, unstable, soft and ...  transient. OK 'transient' is probably better. But for the opposite I'd rather chose 'persistent' instead of 'permanent'. I like it better. 

Thus, we'd have **'ServiceTransientError'** and **'ServicePersistentError'** instead of the current pair - **'ServiceFluctuantError'** and **'ServiceHardError'**

anatolii-borysov:
> You should use one of the Python packages `retry` or `retrying` instead of writing your own code for retrying. They work very similarly to the code you have written.

I inspected them before I decided to proceed with my own implementation. They both appear to be abandoned projects. They both stuck on supporting Python v3.5  and were not updated for at least three years. I suppose it's better to support our own code than be dependent on issues in abandoned 3rd party libs?

DominikDary:
I did a manual end-to-end test using the universal installer and the mesos-agent service was running as expected:
image

jongiddy:
@anatolii-borysov Sometimes packages stop being modified because they are complete. :) We use `retrying` elsewhere in the code.

Anyway, the main things are to fix the missing `raise` and possibly the trivial exercise of improving the exception names. "Persistent" is fine.

PR #6638, opened 2019-12-09
Removed explicit hostname setup and disabled hostname lookup.

tillt:
Updated the Windows Mesos agent startup to be more in line with the linux variant. 

For linux, we explicitly prevent hostname lookups on the agent - the same is now done on Windows. While the results should be identical to the previous approach (of setting an explicit hostname) - semantically we are now in sync with linux on those specific parameters.

This is the linux variant:
https://github.com/dcos/dcos/blob/2fb94d602546c467f6bb9f6303e0176987b1ee62/gen/dcos-config.yaml#L1355-L1363

## Related tickets (optional)


jongiddy:
@mesosphere-mergebot changelog-not-required fix to unreleased code

DominikDary:
@vadymzagorodko fyi

PR #6674, opened 2019-12-17
Added some extra Mesos flags configurations.

mengzhugithub:
## High-level description

Made the following mesos flags configurable:
- mesos_cgroups_enable_cfs (default stays as `true`)
- mesos_max_completed_tasks_per_framework (default changed from 1000 to 100)
- mesos_max_completed_frameworks (default changed from 50 to 10)
- logrotate_max_stdout_size and logrotate_max_stderr_size (default to 2MB)

  - [DCOS-59021](https://jira.mesosphere.com/browse/DCOS-59021) [Mesos] Config.yml changes to support Mesos Flags.


gen/calc.py
```diff
@@ -206,6 +206,20 @@ def validate_mesos_log_retention_mb(mesos_log_retention_mb):
...
 
+def calculate_has_mesos_logrotate_file_size_mb(mesos_logrotate_file_size_mb):
+    return calculate_set(mesos_logrotate_file_size_mb)
+
+
+def validate_mesos_logrotate_file_size_mb(
+        mesos_logrotate_file_size_mb, has_mesos_logrotate_file_size_mb):
+    if has_mesos_logrotate_file_size_mb == 'true':
+        try:
+            int(has_mesos_logrotate_file_size_mb)
```
vinodkone: s/has_mesos_logrotate_file_size_mb/mesos_logrotate_file_size_mb/ ?

gen/calc.py
```diff
@@ -206,6 +206,20 @@ def validate_mesos_log_retention_mb(mesos_log_retention_mb):
...
+        except ValueError as ex:
+            raise AssertionError("Error parsing 'has_mesos_logrotate_file_size_mb' "
```
vinodkone: s/has_mesos_logrotate_file_size_mb/mesos_logrotate_file_size_mb/ ?

gen/calc.py
```diff
@@ -1217,12 +1248,15 @@ def calculate_fault_domain_detect_contents(fault_domain_detect_filename):
...
         'master_dns_bindall': 'true',
+        'mesos_cgroup_enable_cfs': 'true',
         'mesos_dns_ip_sources': '["host", "netinfo"]',
         'mesos_dns_set_truncate_bit': 'true',
         'master_external_loadbalancer': '',
         'mesos_log_retention_mb': '4000',
+        'mesos_logrotate_file_size_mb': '2',
         'mesos_container_log_sink': 'fluentbit+logrotate',
-        'mesos_max_completed_tasks_per_framework': '',
+        'mesos_max_completed_frameworks': '10',
```
vinodkone: This is changing the default as well right (previous was 50)? I didn't see this captured in the PR description.

vinodkone: I personally would prefer changing the defaults in a separate PR than this one which is about making some flags configurable.

gen/dcos-config.yaml
```diff
@@ -1317,11 +1317,8 @@ package:
       MESOS_WORK_DIR={{ mesos_master_work_dir }}
       MESOS_ZK=zk://zk-1.zk:2181,zk-2.zk:2181,zk-3.zk:2181,zk-4.zk:2181,zk-5.zk:2181/mesos
       SASL_PATH=/opt/mesosphere/lib/sasl2
-{% switch has_mesos_max_completed_tasks_per_framework %}
-{% case "true" %}
+      MESOS_MAX_COMPLETED_FRAMEWORKS={{ mesos_max_completed_frameworks }}
```
vinodkone: Why was this `switch` removed?

mengzhugithub: I thought since we have added a default value to `mesos_max_completed_frameworks `, the switch is no longer needed as it will always be true.

gen/calc.py
mengzhugithub: Updated the PR description about the default change. Putting them together since it is all about DCOS-59021.

gen/dcos-config.yaml
vinodkone: I see. Can a user set the value to empty in ocnfig.yaml? If that's not possible, then can we get rid of `has_mesos_max_completed_frameworks ` and `has_mesos_max_completed_tasks_per_framework ` variables altogether if they are always going to be set?

mengzhugithub: I think it was possible. But that feels like a validation error. Removed the `has_ ` variables and make sure a value always present.

vinodkone:
@mesosphere-mergebot label "Ready for Review"

PR #6706, opened 2019-12-26
Winpanda App: Handle Execution Status Of External Commands

anatolii-borysov:
1) check and handle exit code from package extra installation helper
2) catch and save stdout/stderr streams from package extra installtion helper to facilitate debugging process

JIRA: DCOS_OSS-5785 - Winpanda Is To Handle Subprocess Exit Codes
JIRA: DCOS_OSS-5786 - Winpanda App: Redirect stdout and stderr Streams From
                                       an External Command To Dedicated Log Files

### DIFF
--- a/packages/winpanda/extra/src/winpanda/extm/extm.py
+++ b/packages/winpanda/extra/src/winpanda/extm/extm.py
@@ -2,9 +2,12 @@
 
 DC/OS package extra installation options manager definition.
 """
+import subprocess
+
 from common import logger
 from common import exceptions as cm_exc
 from common import utils as cm_utl
+from core.package.manifest import PackageManifest
 from extm import exceptions as extm_exc
 
 
@@ -24,26 +27,31 @@ class EXTCFG_OPTION:
 
 class PkgInstExtrasManager:
     """DC/OS package extra installation options manager."""
-    def __init__(self, ext_conf):
+    def __init__(self, pkg_manifest):
         """Constructor.
 
-        :param ext_conf: dict, package installation extras config
+        :param pkg_manifest: PackageManifest, DC/OS package manifest object
         """
-        if not isinstance(ext_conf, dict):
-            raise extm_exc.InstExtrasManagerConfigError(
-                f'Invalid configuration container structure: {ext_conf}'
-            )
+        self._msg_src_base = self.__class__.__name__
 
-        self.ext_conf = ext_conf
+        assert isinstance(pkg_manifest, PackageManifest), (
+            f'{self._msg_src_base}: Argument: pkg_manifest:'
+            f' Got {type(pkg_manifest).__name__} instead of PackageManifest'
+        )
+        self._pkg_manifest = pkg_manifest
+        self._pkg_id = self._pkg_manifest.pkg_id
+        self._ext_conf = self._pkg_manifest.pkg_extcfg
+
+        self.msg_src = f'{self._msg_src_base}: {self._pkg_id.pkg_id}'
 
     def __str__(self):
         return str({
-            'ext_conf': self.ext_conf,
+            'ext_conf': self._ext_conf,
         })
 
     def handle_install_extras(self):
         """DC/OS package install extra options handler."""
-        install_sect = self.ext_conf.get(EXTCFG_SECTION.INSTALL, {})
+        install_sect = self._ext_conf.get(EXTCFG_SECTION.INSTALL, {})
 
         if not isinstance(install_sect, dict):
             raise extm_exc.InstExtrasManagerConfigError(
@@ -63,7 +71,7 @@ class PkgInstExtrasManager:
 
     def handle_uninstall_extras(self):
         """DC/OS package uninstall extra options handler."""
-        uninstall_sect = self.ext_conf.get(EXTCFG_SECTION.UNINSTALL, {})
+        uninstall_sect = self._ext_conf.get(EXTCFG_SECTION.UNINSTALL, {})
 
         if not isinstance(uninstall_sect, dict):
             raise extm_exc.InstExtrasManagerConfigError(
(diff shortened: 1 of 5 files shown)


anatolii-borysov:
@mesosphere-mergebot changelog-not-required A part of overall windows changes

anatolii-borysov:

anatolii-borysov:
@mesosphere-mergebot request-review @vinodkone

packages/winpanda/extra/src/winpanda/core/constants.py
```diff
@@ -22,3 +22,13 @@
...
 PKG_INSTCFG_FEXTS = [PKG_EXTCFG_FEXT, PKG_SVCCFG_FEXT]
+
+# Package's extra installation helper files
+PKG_EINSTHLP_STDOUT_FEXT = 'stdout'
+PKG_EINSTHLP_STDERR_FEXT = 'stderr'
+PKG_EINSTHLP_STDOUT_FNAME = '.'.join([
+    'pkg-einst-helper-{id}', PKG_EINSTHLP_STDOUT_FEXT
+])
+PKG_EINSTHLP_STDERR_FNAME = '.'.join([
+    'pkg-einst-helper-{id}', PKG_EINSTHLP_STDERR_FEXT
```
jongiddy: What is an "einst"? Can we just call it `pkg-install-1.stderr`?

packages/winpanda/extra/src/winpanda/extm/extm.py
```diff
@@ -87,13 +98,58 @@ def _exec_external_commands(self, *cmd_cl_defs):
...
-
-        for cmd_cl_def in cmd_cl_defs:
+        for cmd_id, cmd_cl_def in enumerate(cmd_cl_defs, start=1):
             try:
-                cm_utl.run_external_command(cmd_cl_def)
-                LOG.debug(f'{msg_src}: External command: {cmd_cl_def}: OK')
+                cmd_run_result = cm_utl.run_external_command(cmd_cl_def)
+                LOG.debug(f'{self.msg_src}: External command: {cmd_cl_def}:'
+                          f' OK')
+                self._save_extcmd_output(str(cmd_id),
```
jongiddy: The subprocess `communicate()` method needs to be called to fill in stdout and stderr correctly. Since the contents are written to files, it's better to redirect stdout and stderr directly to the files.  This will avoid buffering problems if a command happens to write more than the buffer allows.

packages/winpanda/extra/src/winpanda/core/constants.py
anatolii-borysov: **'pkg-einst-helper'** stands for **'Package extra installation helper'**, which is exactly corresponds functional designation of that component. **'pkg-install'** name won't reflect real meaning/source of that artifact. 
Would it help, if we change it to **'pkg-extra-installation-helper'**?

packages/winpanda/extra/src/winpanda/extm/extm.py
anatolii-borysov: We do not expect huge output from helper scripts and I suppose a buffer should be auto-flushed, if writing the next chunk of data to it whould exceed its capacity, so we should be safe with the current implementation. But probably I'm missing something. Which exact buffers do you mean?

Could you please give me a hint on steps how I could reproduce the issue you are talking?

jongiddy: The problem is the pipe that you set for stdout and stderr in the `run_external_command` command. It has limited capacity, so you need to read both pipes while the subprocess runs, to avoid filling the buffer. I expect to see two problems:

1. If you can start a command that outputs slowly, the current code may not capture all of it.
2. If you run a command that outputs a large file, say `Get-Content "some-large-file"`, the current code may hang, but will almost certainly not save all the data to the log file.

The simplest way to avoid this would be to always redirect the stdout and stderr to files. Since this is what you need anyway, you just need a version of `run_external_command` that accepts filenames.

jongiddy: I've looked at this again, and because you use the `run` call, the pipe handling is done for you. So, this part is OK. I tested the full behavior of the `run_external_command` in #6715 and apart from bad behavior if the command times-out, it is good.

However, two points to consider:
1. Since you want to treat success and error exits the same (saving the outputs), the code would be simpler if `run_external_command` did not raise an exception when the code runs, but fails with a non-zero code. You could do this by removing the `check=True` parameter.
2. The outputs could be written directly to the `winpanda` log, to show the failure happening in-place. To understand what command is failing I need to look in the `winpanda` logs, so there appears to be little benefit to putting the command output into a separate file.

jongiddy:
@anatolii-borysov Can you please rebase this PR onto master, as I have merged https://github.com/dcos/dcos/pull/6715 which fixed a problem with exceptions (e.g. timeout) that do not provide `stderr` as an attribute.

As stated above, I think the code would benefit from having a version of `run_external_command` that does not raise an exception for non-zero exit, but instead returns the `CompletedProcess` object for handling by the caller.

packages/winpanda/extra/src/winpanda/extm/extm.py
anatolii-borysov: 1) JG: Since you want to treat success and error exits the same (saving the outputs), the code would be simpler if run_external_command did not raise an exception when the code runs, but fails with a non-zero code. You could do this by removing the check=True parameter.

We consider any failure of an extra package installation helper as critical, so we need to fail the entire DC/OS installation process in such a case, and this is why we need the run call to raise an exception (check=True). Thus to my mind this implementation should be OK.


2) JG: The outputs could be written directly to the winpanda log, to show the failure happening in-place. To understand what command is failing I need to look in the winpanda logs, so there appears to be little benefit to putting the command output into a separate file.

    The thing is that in debug mode (which is the only mode implemented within the Beta scope) the Winpanda application generates (in the winpanda.log file) lots of textual data, reflecting command (winpanda setup, winpanda start) execution processes along with dumps of all the package configuration/metadata artifacts (dcos-config-windows.yaml, manifests, .nssm, .extra files for every package, etc.). Amount of data for a single command run can be as big as several megabytes, and so it can be really difficult to identify/track down particular parts of the whole process or parts related to a particular issue within this big log file.
    Taking into account  everything above saving stdout/stderr from extra installation helpers to the winpanda.log appears not to be the best decision, as it would make searching throughout it even more difficult. Thus, I tend to stick to the implementation that was suggested initially.

anatolii-borysov:
JG: @anatolii-borysov Can you please rebase this PR onto master, as I have merged #6715 which fixed a problem with exceptions (e.g. timeout) that do not provide `stderr` as an attribute.

As far as I can see from the source of the subprocess module, both the CalledProcessError (https://github.com/python/cpython/blob/3.6/Lib/subprocess.py#L60-L71) and the TimeoutExpired (https://github.com/python/cpython/blob/3.6/Lib/subprocess.py#L97-L108) exception objects provide exactly the same interface? So existing code handles correctly both of them, which is exact behaviour we want there. I can't see (real) needs to change it ...

jongiddy:
@anatolii-borysov The error is due to `returncode` attribute being used in the exception handling. This is available in `CalledProcessError` but not in the timeout error. you can verify this by undoing the code change, and seeing that the timeout test fails.

packages/winpanda/extra/src/winpanda/extm/extm.py
jongiddy: As I said in stand-up, I'm not planning to block this PR on these issues. I am trying to ensure that we get simple, error-free code.

My point with issue 1 is that in the calling code you need to have two points where you call `_save_extcmd_output`, one for zero return codes, and one for non-zero return codes. The Python `subprocess` code will happily return both to you in the same form (a `CompletedProcess` object) which keeps the code simpler. You can still call its `check_returncode` method after writing the logs.

With issue 2, the length of individual log files is not really a problem (not when disk usage will be the same either way). More important is that diagnosing problems is easy, and requiring the user to change files to check information adds extra work for them.

anatolii-borysov:
@jongiddy OK, got it. I was misled by your initial statement "... #6715 which fixed a problem with exceptions (e.g. timeout) that do not provide **stderr** as an attribute." Actually it should have been stating "... #6715 which fixed a problem with exceptions (e.g. timeout) that do not provide **returncode** as an attribute."
So now the PR-6715 looks reasonable. Thanks for adding tests for the run_external_command() function.

jongiddy:
Winpanda tests are failing due to addition of a space character:
```
>       assert 'Exit code[4]' in str(e)
E       assert 'Exit code[4]' in "C:\\BuildAgent\\work\\f76365414248b6de\\packages\\winpanda\\extra\\src\\winpanda\\common\\utils.py:126: common.except...x\\\\py36-winpanda-win32\\\\tmp\\\\test_run_external_command_fail0\\\\script.ps1'): CalledProcessError: Exit code [4]:"
E        +  where "C:\\BuildAgent\\work\\f76365414248b6de\\packages\\winpanda\\extra\\src\\winpanda\\common\\utils.py:126: common.except...x\\\\py36-winpanda-win32\\\\tmp\\\\test_run_external_command_fail0\\\\script.ps1'): CalledProcessError: Exit code [4]:" = str(<ExceptionInfo ExternalCommandError tblen=2>)
```

jongiddy:
@anatolii-borysov From the work being done with `dcos-diagnostics` right now, it is clear that putting the script outputs into separate files will make diagnostics significantly more complicated.

Hence, we do need to add these outputs into a single file. The `winpanda.log` file makes the most sense to me, since this is part of the setup of the packages.

anatolii-borysov:
JG: we do need to add these outputs into a single file. The winpanda.log file makes the most sense to me, since this is part of the setup of the packages.

@jongiddy OK. Will make this change.

PR #6774, opened 2020-01-16
Winpanda App: Add Event Hooks Support For Windows Services

anatolii-borysov:
1) introduce event hooks configuration options for NSSM configuration file

JIRA: DCOS_OSS-5632 - Winpanda App: Add Support For Pre/Post Actions For
                                       Start/Stop Events For Windows Services

### DIFF
--- a/packages/winpanda/extra/src/winpanda/svcm/nssm.py
+++ b/packages/winpanda/extra/src/winpanda/svcm/nssm.py
@@ -36,18 +36,23 @@ class NSSMParameter(enum.Enum):
     APPSTDOUT = 'appstdout'
     APPSTDERR = 'appstderr'
     APPENVIRONMENTEXTRA = 'appenvironmentextra'
+    APPEVENTSSTARTPRE = 'appevents start/pre'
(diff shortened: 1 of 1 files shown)


anatolii-borysov:
@mesosphere-mergebot changelog-not-required A part of overall windows changes

anatolii-borysov:

anatolii-borysov:

anatolii-borysov:
@jongiddy 
OK. I see your point. Those upper/lower conversions look ugly, but the main point was not to rewrite everything but just add required parts. To my mind using dedicated class to hold all the parameter details (name/value) is not that bad idea and syntax of using them in the code is quite natural/idiomatic (ex. NSSMParameter.APPLICATION.name, NSSMParameter.APPLICATION.value). Furthermore, it gives that the feelling that it's a constant object. I do not see big difference against defining it as a dict. Though to avoid those upper/lowercase conversions we can just redefine NSSMParameter attributes in lowercase, e.g. NSSMParameter.application. We could also redefine values as lists of strings instead of strings, but, to my mind, it does not introduce crucial benefits.

PR #6810, opened 2020-01-28
Fix pytest instructions

cprovencher:


cprovencher:

PR #6853, opened 2020-02-11
adminrouter: make grpc proxy port configurable

jkoelker:
## High-level description

Adds a configuration option to change the AdminRouter gRPC proxy port.

  - [D2IQ-61007](https://jira.d2iq.com/browse/D2IQ-61007) Make the gRPC socket port configurable via install config param.

### DIFF
--- a/gen/calc.py
+++ b/gen/calc.py
@@ -658,6 +658,14 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
             "`master_http_load_balancer` then exhibitor_storage_backend must not be static."
 
 
+def validate_adminrouter_grpc_proxy_port(adminrouter_grpc_proxy_port):
+    try:
+        assert 0 < int(adminrouter_grpc_proxy_port) < 65536
+    except ValueError as ex:
+        raise AssertionError("Error parsing 'adminrouter_grpc_proxy_port' "
+                             "parameter as an integer: {}".format(ex)) from ex
+
+
 def calculate_adminrouter_tls_version_override(
         adminrouter_tls_1_0_enabled,
         adminrouter_tls_1_1_enabled,
@@ -1250,6 +1258,7 @@ entry = {
         lambda calico_vxlan_port: validate_int_in_range(calico_vxlan_port, 1025, 65535),
         lambda calico_vxlan_vni: validate_vxlan_vni(calico_vxlan_vni),
         validate_overlay_networks_not_overlap,
+        validate_adminrouter_grpc_proxy_port,
     ],
     'default': {
         'exhibitor_azure_account_key': '',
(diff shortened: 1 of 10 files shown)


jkoelker:
@mesosphere-mergebot changelog-not-required part of grpc/etcd/calico effort

jkoelker:

gen/calc.py
```diff
@@ -658,6 +658,14 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
             "`master_http_load_balancer` then exhibitor_storage_backend must not be static."
 
 
+def validate_adminrouter_grpc_proxy_port(adminrouter_grpc_proxy_port):
+    try:
+        int(adminrouter_grpc_proxy_port)
```
vespian: 0<adminrouter_grpc_proxy_port<65500 ?

jkoelker: fixed.

gen/calc.py
```diff
@@ -658,6 +658,14 @@ def validate_exhibitor_storage_master_discovery(master_discovery, exhibitor_stor
             "`master_http_load_balancer` then exhibitor_storage_backend must not be static."
 
 
+def validate_adminrouter_grpc_proxy_port(adminrouter_grpc_proxy_port):
+    try:
+        assert 0 < int(adminrouter_grpc_proxy_port) < 65536
```
vespian: Should we exclude ports 80 and 443 as well?

test-e2e/test_adminrouter_grpc.py
```diff
@@ -0,0 +1,92 @@
...
+            etcd3.etcdrpc.WatchStub(etcd.channel),
+            timeout=etcd.timeout,
+            call_credentials=etcd.call_credentials,
+            metadata=etcd.metadata
+        )
+
+        try:
+            value, meta = etcd.get('probably-invalid-key')
+        except Exception as e:
+            import time
```
mainred: What's the purpose of the `sleep` here?

jkoelker: Oops, left over from troubleshooting. Good catch!

gen/calc.py
jkoelker: I'm 50/50 on restricting it down, on the one hand most peeps wouldn't want it on those ports, but if for whatever reason they do want it on one of those, then they'd have to hack it up.

If you have a strong opinion to restrict it, let me know and I'll add it, otherwise I'd say lets just let it ride ;).

vespian: Admin Router already uses ports 80 and 443 and this is hardcoded [1] If somebody tries to have gRPC port on port 80/443 then the things will break.

Should we be worried about it?

[1] https://github.com/mesosphere/dcos-enterprise/blob/78fc339ba732111331e1ae08d8c106a3632b85c6/gen_extra/dcos-config.yaml#L8-L28

jongiddy: We have many other ports in this range that will not work. I'm happy to keep it as-is.

wavesoft: I also think that it's not going to be easy to figure out dynamically _what_ to exclude. I am also fine to keep it as-is.

vavy:
Hi! Am I correct, that on the dcos cli side this port number is hardcoded (as 12379)?

jkoelker:
@vavy kind of, for the calico subcommand yes it is: https://github.com/dcos/dcos-core-cli/blob/2.2-patch.x/pkg/cmd/calico/calico.go#L134-L143.

PR #6917, opened 2020-02-28
Compiled mesos with openssl support, introduced mesos-modules.

akornatskyy:
https://jira.d2iq.com/browse/D2IQ-64686


packages/mesos-modules/windows.buildinfo.json
```diff
@@ -0,0 +1,24 @@
...
+  "sources": {
+    "openssl": {
+      "kind": "url",
+      "url": "https://slproweb.com/download/Win64OpenSSL-1_1_1d.exe",
+      "sha1": "df9c40e9e82edf693c15302cbc8e8a057857a13c"
+    },
+    "mesos": {
+      "kind": "git",
+      "git": "https://github.com/apache/mesos",
+      "ref": "54227a33b68d83b97ef1a7c14283351d45322317",
```
janisz: This is another place where we need to change version when we bump mesos

akornatskyy: right... similar place in enterprise modules

akornatskyy:

DominikDary:
@mesosphere-mergebot request-review  @jongiddy

DominikDary:
@mesosphere-mergebot request-review @asekretenko

DominikDary:
@mesosphere-mergebot sync

DominikDary:
@mesosphere-mergebot changelog-not-required Internal change

packages/mesos-modules/extra/metrics.json
```diff
@@ -0,0 +1,30 @@
+{
+  "libraries": [
+    {
+      "file": "metrics.dll",
```
janisz: Can you add this configuration to `dcos-config-windows.yaml` like we do in linux version
https://github.com/dcos/dcos/blob/8c6ec915d7653a0fd288148c391454097f3f4bf2/gen/dcos-config.yaml#L362-L366

packages/mesos-modules/extra/logsink.json
```diff
@@ -0,0 +1,18 @@
+{
```
janisz: Can you add this configuration to dcos-config-windows.yaml like we do in linux version
https://github.com/dcos/dcos/blob/8c6ec915d7653a0fd288148c391454097f3f4bf2/gen/dcos-config.yaml#L256-L278

packages/mesos-modules/extra/journal.json
```diff
@@ -0,0 +1,50 @@
+{
+  "libraries": [
+    {
+      "file": "libjournaldlogger.dll",
```
janisz: https://github.com/dcos/dcos/blob/8c6ec915d7653a0fd288148c391454097f3f4bf2/gen/dcos-config.yaml#L378-L421

packages/windows.treeinfo.json
```diff
@@ -1,14 +1,15 @@
...
+    "mesos-modules": "windows",
+    "mesos": "windows",
+    "nssm": "windows",
+    "python": "windows",
     "telegraf": "windows",
     "vcredist": "windows",
-    "nssm": "windows",
-	"bootstrap": "windows",
-    "python": "windows"
+    "winpanda": "windows"
```
akornatskyy: keep original order

akornatskyy: done

packages/mesos-modules/extra/metrics.json
akornatskyy: the approach has pending review with @jongiddy ; if we want to be consistent with enterprise, I will change this. call is no Jon side for now.

janisz: We need to be consistent with Linux and EE version since it will be soon unmaintainable if setups will differ in the way how we configure it. @DominikDary

akornatskyy: @janisz and @DominikDary there will be a separate pull request to address that issue for linux. at the moment the approach is unknow. this PR is for windows only and doesn't affect anything on linux.

janisz: I know it's windows only but please be consistent with the rest of the system. What I'm asking is to move this files from `packages/mesos-modules/extra` into `dcos/gen/dcos-config-windows.yaml`. Thats all. So we have at least all modules configured in the same place.

akornatskyy: right, as pointed in the other comment, that is on @jongiddy to suggest. both ways work right now. if we want to be consistent, we will make this change, no problem.

packages/mesos-modules/buildinfo.json
```diff
@@ -0,0 +1 @@
+{}
```
jongiddy: Is this to avoid building this package on Linux? This can be done by adding `mesos-modules` to the `exclude` list in `packages/treeinfo.json`.

packages/mesos-modules/extra/journal.json
jongiddy: I guess this comment is asking should these files be created by `dcos-config-windows.yaml`? This would provide more consistency between the two platforms.

packages/mesos-modules/extra/metrics.json
jongiddy: Yes, I think it is reasonable to move this to the `dcos-config-windows.yaml`. The only criteria is that we can install the files in `etc/mesos/modules` without them being deleted by winpanda when it installs the `mesos` module, and I believe that is true now.

packages/mesos-modules/windows.build.ps1
```diff
@@ -0,0 +1,114 @@
+$ErrorActionPreference = "stop"
+
+$PKG_DIR = "c:\pkg"
+
+function Install-OpenSSL {
```
jongiddy: Installing OpenSSL and Mesos seems to repeat a lot of the work done in the `mesos` module. Is there a way to avoid this?

Do we need OpenSSL here, for Open DC/OS?

packages/mesos-modules/extra/metrics.json
akornatskyy: OK. got it.

packages/mesos-modules/extra/journal.json
akornatskyy: right

packages/mesos-modules/buildinfo.json
akornatskyy: that cause an issue on windows... the above recommendation was suggested by @janisz earlier. there is a separate task in jira to migrate linux modules as well.

packages/mesos-modules/windows.build.ps1
akornatskyy: yes, we do need. no, we cannot avoid at the moment. there was a separate discussion before around this since it impacts build time significantly.

packages/mesos-modules/buildinfo.json
janisz: > Is this to avoid building this package on Linux?
@jongiddy `buildinfo.json` is required to mark directory as a package.

gen/dcos-config-windows.yaml
```diff
@@ -255,3 +255,110 @@ package:
...
+        "libraries": [
+          {
+            "file": "logsink.dll",
+            "modules": [
+              {
+                "name": "com_mesosphere_mesos_LogSink",
+                "parameters": [
+                  {
+                    "key": "output_file",
+                    "value": "nul"
```
asekretenko: I'm not quite getting: where does the custom glog logsink (implemented in this module) dump logs on a Windows agent host?

packages/mesos-modules/windows.build.ps1
asekretenko: @jongiddy Yes; specifically, we are considering two options to get rid of redundant Mesos builds in the future:
1) Adding a package `mesos-build` that will be the only one that will actually build Mesos, will have all the Mesos build artifacts as an output, will be a dependency of `mesos`, `mesos-modules` and `dcos-ee-mesos-modules`, and will not be packaged into a final DCOS build. The last part needs support in pkgpanda.

 2) Producing a full-fledged installable package with all Mesos libraries and includes as a result of Mesos build. Needs quite some work in Mesos cmake build. Will make the DCOS build slightly heavier than (1)

Looks like neither of two can be done quickly enough, hence we are implementing this as an intermediate step.

packages/mesos/windows.build.ps1
```diff
@@ -1,184 +1,88 @@
...
-    PROCESS {
-        if($LASTEXITCODE){
-            # Leftover exit code. Some other process failed, and this
-            # function was called before it was resolved.
-            # There is no way to determine if the ScriptBlock contains
-            # a powershell commandlet or a native application. So we clear out
-            # the LASTEXITCODE variable before we execute. By this time, the value of
-            # the variable is not to be trusted for error detection anyway.
-            $LASTEXITCODE = ""
+    Get-ChildItem -Path $PKG_DIR/extra/windows-ee.patches -Filter "*.patch" | `
```
asekretenko: Shouldn't the `windows-ee.patches` be renamed into `windows.patches`? They are not EE anymore.

gen/dcos-config-windows.yaml
akornatskyy: @DominikDary I guess we need to find out someone who can address that question.

packages/mesos/windows.build.ps1
akornatskyy: good catch, will rename to be consistent.

packages/mesos-modules/windows.build.ps1
janisz: Do we have a ticket to track this? If not plese crate one and add TODO here.
I'm fine with merging not ideal but working solution

> Done is better than perfect

gen/dcos-config-windows.yaml
DominikDary: this has been discussed in slack with @asekretenko, @kaysoky and  @akornatskyy and the logsink module should be removed from windows and as well this config file.
This with https://github.com/dcos/dcos/pull/6917/files#r394897271 is the last piece that needs to be done in order to merge this Pr.

PR #6928, opened 2020-03-02
Fix test_systemd_units_are_health on Windows

janisz:
## High-level description

Windows has no systemd but we still thread it's native services as dcos-diagnostics services so they appear in /health output, so we need to thereat windows nodes differently. Unfortunately there is no way to detect if node runs windows so we assume if there is `WinRT` service it must be Windows.

  - [D2IQ-64752](https://jira.d2iq.com/browse/D2IQ-64752) 

## Related tickets (optional)


janisz:
Since there is no CI I tested it manually
```
[gw0] PASSED ../../../extra/test_composition.py::test_systemd_units_are_healthy [2020-03-02 15:36:48,644|dcos_test_utils.helpers|DEBUG]: Request method GET: http://leader.mesos/marathon/v2/apps. Arguments: {}
```

janisz:
@mesosphere-mergebot request-review @jeschkies @DominikDary @jongiddy

packages/dcos-integration-test/extra/test_composition.py
```diff
@@ -227,6 +237,8 @@ def test_systemd_units_are_healthy(dcos_api_session) -> None:
         role = node_health["Role"]  # Is one of master, agent, agent_public
         units_per_node[node] = set(node_health["Output"])
         exp_units_per_node[node] = expected_units[role]
+        if 'WinRM' in units_per_node[node]:
+            exp_units_per_node[node] = windows_units
```
jeschkies: So this just verifies that there is a health report but not its state, correct? Should we also `assert 'WinRM' in units_per_node`?

janisz: It checks state later here we only assign expected state to `windows_units` instead of `expected_units['agent']`

janisz:
:wave: @jeschkies @DominikDary   Can you review, please?

DominikDary:
@mesosphere-mergebot changelog-not-required Internal change

PR #6965, opened 2020-03-06
Introduce Makefile in test_util directory to exercise integration tests for Windows

orsenthil:
* D2IQ-65138  - Enable CI for Windows Tests on Open 

We are going to enable CI for windows tests on Open DC/OS via Makefile.

---

This introduces testing based on a checkined `main.tf` and with Makefile.

The pipeline here setups only these.

* Introduces steps for `make test`
* Setups a CI pipeline for [Open DC/OS](https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcOs_Open_Test_IntegrationTest_WindowsIntegrationTEsts&branch_DcOs_Open_Test_IntegrationTest=pull%2F6965&tab=buildTypeStatusDiv) and [Enterprise DC/OS](https://teamcity.mesosphere.io/viewType.html?buildTypeId=DcOs_Enterprise_Test_Inte_WindowsIntegrationTest&branch_DcOs_Enterprise_Test_Inte=pull%2F7471&tab=buildTypeStatusDiv)
* This will start reporting status to the Open and Enterprise PRs.
* Same `makefile` is used for both Open DC/OS and Enterprise DC/OS
* The variables used in the `Makefile` can be overridden by environment variables.

*status report*

image

---

**NOTE** - This does not bring in tests yet. The test exercise should be added by **dcos-windows** team.

### DIFF
--- a/test_util/Makefile
+++ b/test_util/Makefile
@@ -1,0 +1,70 @@
+SHELL := /bin/bash -euo pipefail
+
+export AWS_REGION ?= us-west-2
+export TF_VAR_custom_dcos_download_path_win ?= https://downloads.dcos.io/dcos/testing/master/windows/dcos_generate_config_win.sh
+export TF_VAR_variant ?= open
+export OUTPUT_FILE ?= /dev/stdout
+
+.DEFAULT_GOAL := test
+
+.PHONY: maws
+maws:
+	$(eval $(shell maws login "eng-devprod"))
+
+.PHONY: get-terraform
+get-terraform:
+	rm -rf ./terraform_0.11.14_linux_amd64.zip
+	wget https://releases.hashicorp.com/terraform/0.11.14/terraform_0.11.14_linux_amd64.zip
+	unzip ./terraform*.zip
+	chmod +x terraform
+
+# TODO (dcos-team): This should not be required Open DC/OS
+.PHONY: dcos-set-license
+dcos-set-license:
+	@echo ${DCOS_LICENSE_CONTENTS} > license.txt
+
+.PHONY: terraform-init
+terraform-init:  get-terraform dcos-set-license
+	./terraform init --upgrade
+
+.PHONY: ssh-agent
+ssh-agent:
+	if [ -z "$$SSH_AUTH_SOCK" ] ; then \
+	  eval $$(ssh-agent -s); \
+	fi
+
+.PHONY: ssh-keygen-add
+ssh-keygen-add: ssh-agent
+	ssh-keygen -t rsa -b 2048 -f ./tf-dcos-rsa.pem -q -N ""; \
+	ssh-add ./tf-dcos-rsa.pem;
+
+.PHONY: test
+test: terraform-init ssh-keygen-add
+	./terraform apply -auto-approve
+
+.PHONY: output
+output:
+	./terraform output > ${OUTPUT_FILE}
+
+.PHONY: destroy
+destroy:
+	./terraform destroy
+
+.PHONY: clean
+clean:
+	rm -rf ./terraform
+	rm -rf ./.terraform/
+	rm -rf ./inventory
+	rm -rf ./terraform.tfstate
+	rm -rf ./terraform.tfstate.backup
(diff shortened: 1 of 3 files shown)


test_util/Makefile
```diff
@@ -0,0 +1,55 @@
...
+export TF_VAR_variant := open
+
+.DEFAULT_GOAL := test
+
+.PHONY: get-terraform
+get-terraform:
+	rm -rf ./terraform_0.11.14_linux_amd64.zip
+	wget https://releases.hashicorp.com/terraform/0.11.14/terraform_0.11.14_linux_amd64.zip
+	unzip ./terraform*.zip
+	chmod +x terraform
```
jeschkies: We should not install Terraform if it is already present. 
```suggestion
    if ! command -v terraform >/dev/null 2>&1; then
        rm -rf ./terraform_*.zip
	    wget https://releases.hashicorp.com/terraform/0.11.14/terraform_0.11.14_linux_amd64.zip
	    unzip ./terraform*.zip
	    chmod +x terraform
    fi
```
Alternatively we could run the whole thing in a Docker container. Check out Tim's work on Marathon's [packager](https://github.com/mesosphere/marathon/blob/master/tools/packager/Makefile#L5).

orsenthil: Hi @jeschkies  - I could easily add some checks to use the existing terraform, however, I think, it might have some disadvantages.
* Terraform is very strict with versioning and 1.11 is not compatible with 1.12 ; So we will have to check with the version.
* We don't want to have variations with minor versions of terraform too.

Using the downloaded, and pinned version of terraform sounds better to me.

----

On using dockerfile, I will check that. For now, since we are running the tests from VM (TC agents) using agents directly seems useful to get both the debug output and access to logs directly.

orsenthil:

DominikDary:
@mesosphere-mergebot changelog-not-required internal change

DominikDary:

jeschkies:
Sorry to be annoying but this should have been accepted by @fatz, @janisz or myself.

test_util/main.tf
```diff
@@ -56,7 +62,7 @@ module "dcos" {
   }
 
   cluster_name        = "${local.cluster_name}"
-  ssh_public_key_file = "~/.ssh/id_rsa.pub"
+  ssh_public_key_file = "./tf-dcos-rsa.pem.pub"
```
janisz: Please do not introduce hardcoded values since it breaks workflow in other parts. I moved it to variable https://github.com/dcos/dcos/pull/6974

test_util/main.tf
```diff
@@ -68,7 +74,7 @@ module "dcos" {
 
   dcos_variant              = "${var.variant}"
   dcos_version              = "2.1.0-beta1"
-  dcos_license_key_contents = "${file("~/license.txt")}"
+  dcos_license_key_contents = "${file("./license.txt")}"
```
janisz: https://github.com/dcos/dcos/pull/6965/files#r390240646

test_util/main.tf
```diff
@@ -24,7 +24,7 @@ variable "owner" {
 
 variable "expiration" {
     type = "string"
-    default = "2h"
+    default = "1h"
```
janisz: Why this change? You can override it with variable

orsenthil:
> Sorry to be annoying but this should have been accepted by @fatz, @janisz or myself.

Hi @jeschkies  - I totally understand this expectation.

It seems to me that DC/OS team had prefered to see a complete CI wired up between Open and Enterprise, using the `main.tf` - so that we could iterate upon this solution.

* The PR is attempt to bring only this. I agree with many refactoring solutions that you might propose. 
* I expect even more changes, what I concentrated was

a ) We could launch clusters using main.tf
b) We could set variables and override variables via environment both locally and in teamcity
c) Finally working with teamcity will be _easy_ if we do everything in Makefile (Note the teamcity steps, it simply calls make and the parameters has env vars. That's it).

That was the focus of this PR, and it accomplished those, so I set it to ready for review and request for it to reviewed and merged.

PR #7028, opened 2020-03-19
Fix `test_dcos_diagnotics` for Windows.

jeschkies:
## High-level description

The Mesos service unit has a different name on Windows so we query that one as well.
The expected files of the diagnostics bundle are different as well.

  - [D2IQ-65707](https://jira.d2iq.com/browse/D2IQ-65707)

### DIFF
--- a/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
+++ b/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
@@ -582,7 +582,7 @@ def _download_bundle_from_master(dcos_api_session, master_index, bundle, diagnos
             unit_ids = [unit['id'] for unit in health_report['units']]
             if 'WinRM' in unit_ids:
                 unit_output = get_file_content(agent_folder + 'C:\\d2iq\\dcos\\var\\log\\mesos\\mesos-agent.log', z)
-                verify_unit_response(unit_output, 100)
+                verify_unit_response(unit_output, 20)
 
                 verify_archived_items(agent_folder, archived_items, expected_windows_agent_files)
             else:
--- a/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
+++ b/packages/dcos-integration-test/extra/test_dcos_diagnostics.py
@@ -579,8 +579,8 @@ def _download_bundle_from_master(dcos_api_session, master_index, bundle, diagnos
             assert health_report['ip'] == slave_ip
 
             # Decide if we have a Windows or Linux agent based on the health report.
-            assert 'units' in health_report
-            if 'WinRM' in health_report['units']:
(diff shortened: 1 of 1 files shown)


packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -262,10 +264,13 @@ def get_nodes_from_response(response):
 
         agent_nodes_response = check_json(
             dcos_api_session.health.get('/units/dcos-mesos-slave.service/nodes', node=master))
-
         agent_nodes = get_nodes_from_response(agent_nodes_response)
 
-        assert len(agent_nodes) == len(dcos_api_session.slaves), '{} != {}'.format(agent_nodes, dcos_api_session.slaves)
+        windows_agent_nodes_response = check_json(
+            dcos_api_session.health.get('/units/mesos-agent/nodes', node=master))
```
jeschkies: @DominikDary and @janisz, this points to inconsistencies between Linux and Windows agents. Ideally the service names would be the same or at least indicate that one is a Linux agent and the other a Windows agent. Do we have a PO for the Windows support?

jeschkies:
It's green!

jeschkies:

jongiddy:
@mesosphere-mergebot changelog-not-required Change to tests

jeschkies:
@mesosphere-mergebot request-review @janisz

packages/dcos-integration-test/extra/test_dcos_diagnostics.py
```diff
@@ -262,10 +264,17 @@ def get_nodes_from_response(response):
...
-        assert len(agent_nodes) == len(dcos_api_session.slaves), '{} != {}'.format(agent_nodes, dcos_api_session.slaves)
+        # Fetch Windows nodes if we have some.
+        if 'WinRM' in pulled_units:
+            windows_agent_nodes_response = check_json(
+                dcos_api_session.health.get('/units/mesos-agent/nodes', node=master))
+            windows_agent_nodes = get_nodes_from_response(windows_agent_nodes_response)
+        else:
+            windows_agent_nodes = list()
+
+        assert set(agent_nodes + windows_agent_nodes) == set(dcos_api_session.slaves)
```
DominikDary: can't we add this to test utils?

janisz: It's not a easy change. We should change the way how we start tests and add new a env variable with windows private nodes. Then we can use it to distinguish IPs in tests.

jeschkies: Ideally the tests would be mostly agnostic to the node types. Eg the service units should have the same name. I think it is ok to make it explicit right now and not hide it in a library.

PR #7115, opened 2020-03-31
Do not require diagnostics on destroy.

janisz:
## High-level description

This leads to weird situations when we want to destroy cluster but first we need to create a diagnostics bundle for it. It's especially annoying when you want to stop cluster creation and free resources but first need to finish cluster creation and then destroy it.

After merging this change we need to add step to our CI to generate diagnostics bundle even if previous step failed.

  - [D2IQ-65565](https://jira.d2iq.com/browse/D2IQ-65565)

## Related tickets (optional)

### DIFF
--- a/test_util/Makefile
+++ b/test_util/Makefile
@@ -67,8 +67,11 @@ test: cluster.json
 ssh: cluster.json
 	ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i $(SSH_KEY) centos@$(shell jq '.masters_public_ip.value' $<)
 
+.PHONY: diagnostics
+diagnostics: $(diagnostics_bundle_file)
+
 .PHONY: destroy
-destroy: $(diagnostics_bundle_file)
+destroy:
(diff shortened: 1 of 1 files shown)


janisz:
> If we do this we won't have a diagnostic bundle after each run. What we can do is remove the cluster.json dependency.

But `cluster.json` is required by `make ssh` which is used by diagnostics. After merging this I'll edit CI steps and add bundle creation as a step after `make test` and `make destroy` after all to run even if there was an error.

jeschkies:
> After merging this I'll edit CI steps and add bundle creation as a step

You'll have the same issue then. I guess we could just hard code `cluster.json` instead of using `$<` as the effect is the same.

janisz:
@mesosphere-mergebot request-review @jongiddy @jeschkies @DominikDary

janisz:
@mesosphere-mergebot request-review @fatz

janisz:
@mesosphere-mergebot changelog-not-required internal

PR #7127, opened 2020-04-01
D2IQ-66446 Upgrade fails on dcos-diagnostics service stop

mediapills:
## High-level description

Upgrade fails on dcos-diagnostics service stop.
Force to kill process after unsuccessful stop.

  - [D2IQ-66446](https://jira.d2iq.com/browse/D2IQ-66446) Upgrade fails on dcos-diagnostics service stop.

## Related tickets (optional)

### DIFF
--- a/packages/winpanda/extra/src/winpanda/svcm/nssm.py
+++ b/packages/winpanda/extra/src/winpanda/svcm/nssm.py
@@ -25,6 +25,8 @@ from typing import List, Tuple
 
 LOG = logger.get_logger(__name__)
 
+CMD_KILL = "taskkill /f /fi \"SERVICES eq {}\""
+
 
 class NSSMParameter(enum.Enum):
     """NSSM parameter set."""
@@ -401,7 +403,10 @@ class WinSvcManagerNSSM(base.WindowsServiceManager):
 
     def stop(self) -> None:
         """Stop a registered service (immediately)."""
-        self._primitive_command(NSSMCommand.STOP.value)
+        try:
+            self._primitive_command(NSSMCommand.STOP.value)
+        except svcm_exc.ServiceManagerCommandError:
+            self.kill()
 
     def restart(self) -> None:
         """Restart a registered service (immediately)."""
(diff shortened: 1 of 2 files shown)


mediapills:
@mesosphere-mergebot request-review jongiddy

mediapills:
@mesosphere-mergebot request-review @sergiimatusEPAM

jongiddy:

jongiddy:
@mesosphere-mergebot changelog-not-required unreleased code

packages/winpanda/extra/src/winpanda/svcm/nssm.py
```diff
@@ -25,6 +25,8 @@
 
 LOG = logger.get_logger(__name__)
 
+CMD_KILL = "taskkill /f /fi \"SERVICES eq {}\""
```
jongiddy, line 28: If the service exits after the timeout but before we run this kill command, will this raise an error? It looks like maybe no, since the command uses a filter, but, for example, does `taskkill /f /fi "SERVICES eq does-not-exist"` exit successfully?

sergiimatusEPAM:
taskkill /f /fi "SERVICES eq does-not-exist"
INFO: No tasks running with the specified criteria.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

packages/winpanda/extra/src/winpanda/svcm/nssm.py
mediapills: In a case of KILL debug message will be added.
Instead of curly brakets service name that can not be stpped successfuly will be placed and we will not have `does-not-exist` case.
When we are talking about filter this command will came from stop where we already have the filter and also we are not using `_primitive_command` method where we do have this filter. As a option I can do this method as a static and will sent service name there then it will be like a function with predefined powershell commad and I will use it in `stop` method.

jongiddy:
@mediapills Can you please rebase once more. Everything that has stopped tests working should be merged now

PR #7129, opened 2020-04-01
Note an important breaking change

timcharper:
Marathon 1.9 deprecated the specification of invalid acceptedResourceRoles. In Marathon 1.10, we're turning it off by default. We should make this clear to users that are upgrading.


jongiddy:

jongiddy:
@mesosphere-mergebot sync

PR #7153, opened 2020-04-06
Reduce Windows machine size

janisz:
## High-level description

We can use smaller machines for our CI, to save some money.

  - [D2IQ-66575](https://jira.d2iq.com/browse/D2IQ-66575)
  - [D2IQ-66580](https://jira.d2iq.com/browse/D2IQ-66580)  

## Related tickets (optional)

  - [D2IQ-66576](https://jira.mesosphere.com/browse/D2IQ-66576)

### DIFF
--- a/test_util/main.tf
+++ b/test_util/main.tf
@@ -50,15 +50,32 @@ variable "dcos_license_key_contents" {
   description = "Defines content of license used for EE."
 }
 
+variable "instance_type" {
+    type = "string"
+    default = "t3.medium"
+    description = "Defines type of used machine."
+}
+
+variable "build_id" {
+    type = "string"
+    default = ""
+    description = "Build ID from CI."
+}
+
+variable "build_type" {
+    type = "string"
+    default = ""
+    description = "Build type from CI."
+}
+
 # Used to determine your public IP for forwarding rules
 data "http" "whatismyip" {
   url = "http://whatismyip.akamai.com/"
 }
 
 resource "random_string" "password" {
-  length = 6
-  special = true
-  override_special = "-"
+  length = 12
+  special = false
 }
 
 locals {
@@ -76,6 +93,8 @@ module "dcos" {
   tags {
     owner = "${var.owner}"
     expiration = "${var.expiration}"
+    build_id = "${var.build_id}"
+    build_type_id = "${var.build_type}"
   }
 
   cluster_name        = "${local.cluster_name}"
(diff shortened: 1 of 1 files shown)


test_util/main.tf
```diff
@@ -87,7 +106,10 @@ module "dcos" {
   num_public_agents  = "1"
 
   dcos_instance_os        = "centos_7.5"
-  bootstrap_instance_type = "m4.xlarge"
```
janisz, line 90: The default is `t2.medium` so let's stick with it.

janisz:
@mesosphere-mergebot request-review @jongiddy @jeschkies @DominikDary

janisz:
@mesosphere-mergebot changelog-not-required OSS internal

test_util/main.tf
```diff
@@ -76,6 +93,8 @@ module "dcos" {
   tags {
     owner = "${var.owner}"
     expiration = "${var.expiration}"
+    build_id = "${var.build_id}"
```
DominikDary, line 96: @orsenthil does this works as expected and can you find the corresponding cost details in cost explorer?

janisz:
Since we can't get dcos build I tests in manually and it's working

```
-- Docs: https://docs.pytest.org/en/latest/warnings.html
======= 53 passed, 1 skipped, 7 xfailed, 127 warnings in 1109.61 seconds =======
```

janisz:
@mesosphere-mergebot request-review @orsenthil

test_util/main.tf
orsenthil: @DominikDary  - I don't how it correlates with cost details. But I will find out.

orsenthil: And this conversation on measurement only indirectly related to the motivation here, which is about reducing the size of the instance. So, +1 with this change if the reduced instance size meets our testing criteria.

test_util/main.tf
```diff
@@ -50,15 +50,32 @@ variable "dcos_license_key_contents" {
...
 # Used to determine your public IP for forwarding rules
 data "http" "whatismyip" {
   url = "http://whatismyip.akamai.com/"
 }
 
 resource "random_string" "password" {
-  length = 6
-  special = true
-  override_special = "-"
+  length = 12
```
orsenthil, line 77: @janisz  - I think, we will run into problems here.

* Terraform / aws has limits on size of the string the characters allowed.

Could you please test this locally before using it?

orsenthil: By testing locally, I meant, try to test it 3-5 times.

janisz: Good point, I'll change to 6 and cluster ID so names will be shorter.

janisz: Updated can you take another look

janisz: Ouch, it was already on train. I'll fix it later but all runs local + 3 on CI passed so I think we do not need to change it.

PR #7178, opened 2020-04-09
Updated cmake in dcos-builder docker image.

akornatskyy:
# High-level description

Updated cmake in dcos-builder docker image.

# Corresponding DC/OS tickets (required)
[D2IQ-6671](https://jira.d2iq.com/browse/D2IQ-66810)

### DIFF
--- a/pkgpanda/docker/dcos-builder/Dockerfile
+++ b/pkgpanda/docker/dcos-builder/Dockerfile
@@ -5,7 +5,6 @@ RUN apt-get -qq update && apt-get -y install \
   autoconf \
   automake \
   autopoint \
-  cmake \
   cpp \
   curl \
   default-jdk \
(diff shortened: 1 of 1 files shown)


akornatskyy:

janisz:
Can you elaborate in the description why do we need this change?

akornatskyy:
it is in the ticket (see link in description).

janisz:
I insist to have it in here and best in the commit message. It's really helpful when you can use only git to reason why change was introduced. Links may not working in the future. Imagine that after merging this commit then only way to get to this link will be to go to github and then to jira. That means additional two steps in order to understand why you made that change.

akornatskyy:
I believe the source of true is JIRA ticket. take you time to have a look there.

jongiddy:
@mesosphere-mergebot changelog-not-required change to build only

DominikDary:
> I believe the source of true is JIRA ticket. take you time to have a look there.

This is wrong @akornatskyy , what @janisz said is my expectation as well.

akornatskyy:
@DominikDary The title of the ticket, this pull request description and commit message reflect the intent to update cmake; there are multiple reason for the change which in short result in the fact of the change (update cmake). Note, pull request template suggests a high level description.

An obvious reason for any version update is that an existing one doesn't suits you any longer. That usually pulls entire thread of reasons why did you arrive to that point... which can be quite lengthy (in terms of time to get to, explain and satisfy) and even doubtful (maybe not?).

If you look at version history of [this](https://github.com/dcos/dcos/commits/master/pkgpanda/docker/dcos-builder/Dockerfile) specific file changed by this PR, you will notice that the previous change was related to [PR 7004](https://github.com/dcos/dcos/pull/7004)... there is no description, no reference to a JIRA ticket (which is required according to pull request template).

PR #7217, opened 2020-04-23
Removed unrelated changelog entries from changelog

timcharper:
Marathon and Metronome changes were listed that were released in 2.0.0; they don't need to be mentioned for 2.1.0 because they are not new changes.


vinodkone:

PR #7311, opened 2020-05-25
add cache-header for dcos-version.json

pierrebeitz:
as during a SOAK update the UI did not consistently report the correct
version of the cluster, we now let the UI know, that it must always
revalidate `dcos-version.json`.

### DIFF
--- a/packages/adminrouter/extra/src/includes/server/common.conf
+++ b/packages/adminrouter/extra/src/includes/server/common.conf
@@ -23,4 +23,7 @@ location /nginx/metrics {
 location /dcos-metadata/dcos-version.json {
     # Allow non-authed access for the UI.
     alias /opt/mesosphere/active/dcos-metadata/etc/dcos-version.json;
+    # Indicate that dcos-version.json must not be cached.
+    # E.g. to have the UI report the correct version after a SOAK-update.
+    add_header Cache-Control no-cache;
 }


jongiddy:
Should this use `no-cache` to force re-validation? It seems `must-revalidate` would make sense if we were setting a non-zero cache time.

> no-cache
>     The response may be stored by any cache, even if the response is normally non-cacheable. However, the stored response MUST always go through validation with the origin server first before using it.
> 
> must-revalidate
>     Indicates that once a resource becomes stale, caches must not use their stale copy without successful validation on the origin server.

pierrebeitz:
@jongiddy it is my understanding that if there's no other headers indicating how long it takes a file to become stale, it is immediately stale when `must-revalidate` is set. setting an additional cache-time or ETags (as in our case) enables browsers to actually make use of caching.

please correct me if i'm wrong here.

jongiddy:
The only info I'm going on is the PR description - is there a JIRA as well? But it sounds like every request needs to go back to the server, whether the ETag matches or not.

`no-cache` forces every request to go back to the server, no matter what expiration time is set or calculated, or whether the cached value is considered to be fresh or stale.

With `must-revalidate` and no other cache headers, the client can use a heuristic to create an expiration time and, during that time, the cached value is fresh so the client does not contact the server. See https://tools.ietf.org/html/rfc7234#section-4.2.2

In either case, the ETag can allow the server to avoid returning the actual response (but in this case, the data is small and cheap enough that it saves very little).

pierrebeitz:
that makes a lot of sense. thanks for educating! 💜

jongiddy:

jongiddy:
@mesosphere-mergebot changelog-not-required insignificant external visibility

pierrebeitz:
@mesosphere-mergebot backport 2.1

pierrebeitz:
@mesosphere-mergebot backport 2.0

pierrebeitz:
@mesosphere-mergebot backport 1.13

PR #7314, opened 2020-05-26
Use ZooKeeper autopurge instead of Exhibitor periodic cleanup

jongiddy:
## High-level description

ZooKeeper 3.4 introduced autopurge to clean up old snapshots and transaction logs. This was also one of the responsibilities handled by Exhibitor. For some time we have had ZK running autopurge every 6 hours, but also Exhibitor calling ZK purge every 5 minutes.

Exhibitor does not pass the configured JNA temporary directory through to its explicit purge, so it causes ZooKeeper to to use `/tmp` for executable files, even if it is mounted `noexec`.

This PR stops the Exhibitor purge, and changes the ZK autopurge to run every hour (the minimum). 

  - [D2IQ-68109](https://jira.d2iq.com/browse/D2IQ-68109) COPS-6111: Regression: Exhibitor writing JNA files to /tmp again

## Related tickets (optional)

### DIFF
--- a/packages/exhibitor/extra/start_exhibitor.py
+++ b/packages/exhibitor/extra/start_exhibitor.py
@@ -99,7 +99,10 @@ zookeeper-log-directory={zookeeper_log_dir}
 zookeeper-config-directory=/var/lib/dcos/exhibitor/conf
 zookeeper-pid-path=/var/lib/dcos/exhibitor/zk.pid
 log-index-directory={zookeeper_log_dir}
-cleanup-period-ms=300000
+# Exhibitor runs PurgeTxnLog every `cleanup-period-ms` ms.  With ZK 3.4+ this
+# is no longer required if `autopurge` is configured for ZK.  Setting
+# `cleanup-period-ms` to 0 prevents Exhibitor from running the purge.
+cleanup-period-ms=0
 check-ms={check_ms}
 backup-period-ms=600000
 client-port=2181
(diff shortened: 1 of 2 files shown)


jongiddy:
@mesosphere-mergebot request-review @wavesoft @jkoelker

packages/exhibitor/extra/start_exhibitor.py
```diff
@@ -108,7 +111,7 @@ def invoke_detect_ip():
 connect-port=2888
 observer-threshold=0
 election-port=3888
-zoo-cfg-extra=tickTime\\=2000&initLimit\\=10&syncLimit\\=5&quorumListenOnAllIPs\\=true&maxClientCnxns\\=0&autopurge.snapRetainCount\\=5&autopurge.purgeInterval\\=6
+zoo-cfg-extra=tickTime\\=2000&initLimit\\=10&syncLimit\\=5&quorumListenOnAllIPs\\=true&maxClientCnxns\\=0&autopurge.purgeInterval\\=1
```
wavesoft, line 114: Do you think for clarity to keep the `autopurge.snapRetainCount\\=3` that is the default?

jongiddy: It's a tricky one. On one hand, as you suggest, explicitly stating the defaults indicates that the value was chosen purposefully, even though it was the default. On the other hand, it could be clearer to keep the config short to specify "the defaults, with these few purposefully chosen exceptions".

In this case, if ZooKeeper found a reason to change the number of retained snapshots, we would probably go with their findings. Hence, we accept the default, whatever it may be, rather than specifically needing 3 snapshots.

PR #7317, opened 2020-05-29
Added patch files for DCOS reservation label offer filtering.

bmahler:
## High-level description

In DC/OS, Marathon and SDK use reservation labels to mark the
reservations that they created and to prevent other frameworks
from using them. However, some frameworks (e.g. Spark, Jenkins)
do not check the labels and can use reservations created by
Marathon / SDK. See D2IQ-68800 for an example.

The patch applied to mesos strips reservations that belong to
another framework, using the framework id label convention. Longer
term, we would want to just use framework specific reservations for
this, see MESOS-7651.

  - [D2IQ-68800](https://jira.d2iq.com/browse/D2IQ-68800) Frameworks can interfere with Marathon pods by launching tasks on "Marathon's" resources reserved to their role.


vinodkone:
cc @asekretenko

greggomann:

vinodkone:
@bmahler Can you update CHANGES.md with this change description?

PR #7457, opened 2020-07-20
Complete skip annotation of e2e tests

jongiddy:
## High-level description

Add annotations for all e2e tests, to allow them to be skipped if only safe files are modified.

Remove `*.txt` files from default safe list, since `requirements.txt` is a common file that does affect Python code.

Rename `e2e_module.py` so it doesn't get mistaken for a test file. Slightly related: update the collection of modified files to check for renames, and return both original and new name.

  - [D2IQ-70203](https://jira.d2iq.com/browse/D2IQ-70203) Annotate remaining e2e tests for conditional skipping

## Related tickets (optional)


jongiddy:

jongiddy:
@mesosphere-mergebot changelog-not-required changes to te.st code

test-e2e/conditional.py
```diff
@@ -7,9 +7,10 @@
 from wcmatch import glob
 
 
-# E2E_SAFE_DEFAULT are files that typically do not affect e2e tests
+# E2E_SAFE_DEFAULT includes files that typically do not affect e2e tests.
+# `*.txt` not included here because `requirements.txt` can affect tests.
 E2E_SAFE_DEFAULT = [
-    '**/*.{md,txt}', '.github/**', 'config/**', 'docs/**', 'flake8_dcos_lint/**',
```
jkoelker, line 12: Could we not add override just for `requirements.txt`? (probably failed flake8 line length ;()
```suggestion
    '**/*.{md,txt}', '!**/requirements.txt', '.github/**', 'config/**', 'docs/**', 'flake8_dcos_lint/**',
```

test-e2e/conftest.py
```diff
@@ -2,8 +2,8 @@
 
 # Hack to exclude test-e2e/conftest.py from the top-level tox config py35-unittests.
 # https://stackoverflow.com/a/37493203
-pytest_plugins = ['test_e2e_module']
-# Actual content of conftest.py can be found in test_e2e_module.py.
+pytest_plugins = ['e2e_module']
```
jkoelker, line 5: \o/ thank you

test-e2e/conditional.py
jongiddy: The problem is that that makes any `requirements.txt` unsafe, including one that is in an otherwise ignored package.

Having looked at the code, there's no top-level `txt` files and most others will be in directories that are ignored.

The exception is `./packages/etcd/extra/README.txt`. Changing that file will run all the Calico and etcd tests, but that would be the case anyway, since a negative `!packages/etcd/**` will override any positive `**/*.txt`.

jkoelker: > The problem is that that makes any `requirements.txt` unsafe, including one that is in an otherwise ignored package.

ahhh, I see, yea ;(.

PR #7470, opened 2020-07-27
[2.1] Use ZooKeeper autopurge instead of Exhibitor periodic cleanup

jongiddy:
## High-level description

Back-port of #7314

ZooKeeper 3.4 introduced autopurge to clean up old snapshots and transaction logs. This was also one of the responsibilities handled by Exhibitor. For some time we have had ZK running autopurge every 6 hours, but also Exhibitor calling ZK purge every 5 minutes.

Exhibitor does not pass the configured JNA temporary directory through to its explicit purge, so it causes ZooKeeper to to use `/tmp` for executable files, even if it is mounted `noexec`.

This PR stops the Exhibitor purge, and changes the ZK autopurge to run every hour (the minimum). 

  - [D2IQ-68109](https://jira.d2iq.com/browse/D2IQ-68109) COPS-6111: Regression: Exhibitor writing JNA files to /tmp again
  - [D2IQ-68868](https://jira.d2iq.com/browse/D2IQ-68868) Back-port dropping Exhibitor purge

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -25,6 +25,8 @@ Please follow the [`CHANGES.md` modification guidelines](https://github.com/dcos
 
 * Display user email address in UI when logging in using external provider. (D2IQ-70199)
 
+* Removed Exhibitor snapshot cleanup and now rely on ZooKeeper autopurge. (D2IQ-68109)
+
 
 ## DC/OS 2.1.0 (2020-06-09)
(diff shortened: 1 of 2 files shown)


jongiddy:
@mesosphere-mergebot request-review @DominikDary @janisz  @jkoelker

PR #7521, opened 2020-08-07
[WIP] Test calico startup with multiple agents

jongiddy:
## High-level description

What features does this change enable? What bugs does this change fix?

## Corresponding DC/OS tickets (required)

  - [D2IQ-ID](https://jira.d2iq.com/browse/D2IQ-ID) JIRA title / short description.

## Related tickets (optional)

  - [D2IQ-ID](https://jira.mesosphere.com/browse/D2IQ-<number>) JIRA title / short description.

PR #7526, opened 2020-08-12
Minimize Docker daemon configuration changes, set Calico timeouts more consistently

jongiddy:
## High-level description

The Calico startup is relatively new, and we've seen some problems with the way timeouts are set. This PR updates the timeouts to make more sense. Also, the setup changes a Docker config file, which interferes with existing config managers. Although we can't "fix" that, we can keep the JSON in the original order to make the structure more predictable. This also ensures that updates to the network setup in future versions will be reflected in the Docker configuration.

This change was developed as WIP in #7521

  - [D2IQ-70674](https://jira.d2iq.com/browse/D2IQ-70674) packages/calico/extra/dcos-calico-libnetwork-plugin.service problems.

### DIFF
--- a/packages/calico/extra/create-calico-docker-network.py
+++ b/packages/calico/extra/create-calico-docker-network.py
@@ -9,9 +9,12 @@ import os
 import shlex
 import signal
 import socket
+import stat
 import subprocess
 import sys
+import tempfile
 import time
+from collections import OrderedDict
 
 from contextlib import contextmanager
 from typing import Generator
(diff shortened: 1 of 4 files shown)


jongiddy:

packages/calico/extra/create-calico-docker-network.py
```diff
@@ -215,21 +218,55 @@ def is_docker_cluster_store_configured():
     return False
 
 
-def config_docker_cluster_store():
-    if is_docker_cluster_store_configured():
-        print("Docker cluster store has already been configured")
-        return
+def write_file_bytes(filename, data, mode):
```
jkoelker, line 221: This would be great as a `contextmanager` as well ;)

packages/calico/extra/create-calico-docker-network.py
```diff
@@ -286,7 +328,7 @@ def _wait_docker_cluster_store_config():
             raise Exception("Cluster store not configured")
         return True
```
jkoelker, line 329: Don't need the return here if its not returned below.

PR #7543, opened 2020-08-13
[2.1] Minimize Docker daemon configuration changes, set Calico timeouts more consistently

jongiddy:
Back-port of #7526 

## High-level description

The Calico startup is relatively new, and we've seen some problems with the way timeouts are set. This PR updates the timeouts to make more sense. Also, the setup changes a Docker config file, which interferes with existing config managers. Although we can't "fix" that, we can keep the JSON in the original order to make the structure more predictable. This also ensures that updates to the network setup in future versions will be reflected in the Docker configuration.

  - [D2IQ-70674](https://jira.d2iq.com/browse/D2IQ-70674) packages/calico/extra/dcos-calico-libnetwork-plugin.service problems.

### DIFF
--- a/packages/calico/extra/dcos-calico-libnetwork-plugin.service
+++ b/packages/calico/extra/dcos-calico-libnetwork-plugin.service
@@ -8,7 +8,7 @@ EnvironmentFile=/opt/mesosphere/etc/calico/calico-node.env
 EnvironmentFile=/opt/mesosphere/etc/calico/calico-node-datastore.env
 ExecStart=/opt/mesosphere/bin/start-calico-libnetwork-plugin.sh
 ExecStartPost=/opt/mesosphere/bin/create-calico-docker-network.py
-TimeoutStartSec=60s
+TimeoutStartSec=180s
 
 Restart=always
 StartLimitInterval=0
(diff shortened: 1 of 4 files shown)


jongiddy:

packages/calico/extra/dcos-calico-libnetwork-plugin.service
```diff
@@ -8,7 +8,7 @@ EnvironmentFile=/opt/mesosphere/etc/calico/calico-node.env
 EnvironmentFile=/opt/mesosphere/etc/calico/calico-node-datastore.env
 ExecStart=/opt/mesosphere/bin/start-calico-libnetwork-plugin.sh
 ExecStartPost=/opt/mesosphere/bin/create-calico-docker-network.py
-TimeoutStartSec=60s
+TimeoutStartSec=180s
```
janisz, line 11: Can we replace it with pre check? I understand now we wait for `cos-calico-libnetwork-plugin.service`, right?
 https://github.com/dcos/dcos/blob/6d4aad5c3b99ba5d7ce4f4d0e0921cf96fb2e901/packages/calico/extra/dcos-calico-libnetwork-plugin.service#L11

jongiddy: `create-calico-docker-network.py` can take 120+ seconds to complete, which is why the timeout needs to be increased.

I think it was created this slightly weird way because the libnetwork plugin is the service that needs to be restarted if it exits. The Docker network cannot be created until the plugin is running, which is why it is done as a 'post' step.

It looks like there must be a better way, but I'm going for the simple fix right now.

PR #7551, opened 2020-08-17
gen: check spartan addresses are not in resolvers

jkoelker:
## High-level description

Setting the spartan addresses (`198.51.100.1-3`) in the resolvers list
will lead to resolution failures as spartan will end up querying itself
in an infinate loop. Check that they are not in the list at install
time.

  - [COPS-4616](https://jira.d2iq.com/browse/COPS-4616) Installer should make sure Spartan IP's are not set as the 'resolvers'


jkoelker:

jkoelker:
@mesosphere-mergebot request-review @vinodkone

PR #7574, opened 2020-08-25
Bump Mesos for the CSI external volume support

qianzhangxa:
## High-level description

Bump Mesos for the CSI external volume support

## Corresponding DC/OS tickets (required)

  - [D2IQ-71267](https://jira.d2iq.com/browse/D2IQ-71267) Bump Mesos in DC/OS for the statically provisioned CSI volumes support

PR #7579, opened 2020-08-27
Add NFS CSI plugin to DC/OS

qianzhangxa:
## High-level description

Add NFS CSI plugin into DC/OS

  - [D2IQ-65490](https://jira.d2iq.com/browse/D2IQ-65490) Add NFS CSI plugin into DC/OS


qianzhangxa:
@mesosphere-mergebot help

qianzhangxa:
@mesosphere-mergebot test all

qianzhangxa:
@mesosphere-mergebot test all

qianzhangxa:
@mesosphere-mergebot test all

qianzhangxa:
@mesosphere-mergebot ready-for-review-label

qianzhangxa:
@mesosphere-mergebot ready-for-review-label

qianzhangxa:

gen/dcos-config.yaml
```diff
@@ -254,6 +254,26 @@ package:
         },
         "max_configuration_attempts": {{ dcos_overlay_config_attempts }}
       }
+  - path: /etc/dcos/storage/csi/nfs.conf
```
greggomann, line 257: I don't think this is the proper location for this file. `/etc/dcos` doesn't currently exist on DC/OS nodes, does it?

I think putting it under `/opt/mesosphere/etc` makes more sense. How about `/opt/mesosphere/etc/storage/csi/nfs.conf`?

qianzhangxa: We did the same for CNI network config file, see https://github.com/dcos/dcos/blob/2.1.0/gen/dcos-config.yaml#L202. Actually during installation, our installer will put that file under /opt/mesosphere/etc/storage/csi/ as you mentioned, I have verified this by installing the DC/OS package built from this PR.

qianzhangxa:
@mesosphere-mergebot test all

qianzhangxa:
> Ideally I think this PR would come after one which has already set up the `csi_plugin_config_dir` flag on an updated Mesos agent to pick up the config file which is added by this PR. That way, we would know that the plugin container is launched successfully on agent startup.
> 
> We could either include the Mesos bump and Mesos agent flags update in this PR, or we could just wait until the Mesos bump lands to test/merge this one, I would be OK with either. WDYT?

Yeah, I would suggest we land #7574 and #7586 and then land this one. WDYT?

greggomann:
> > Ideally I think this PR would come after one which has already set up the `csi_plugin_config_dir` flag on an updated Mesos agent to pick up the config file which is added by this PR. That way, we would know that the plugin container is launched successfully on agent startup.
> > We could either include the Mesos bump and Mesos agent flags update in this PR, or we could just wait until the Mesos bump lands to test/merge this one, I would be OK with either. WDYT?
> 
> @greggomann Yeah, I would suggest we land #7574 and #7586 and then land this one. WDYT?

Yep, sgtm.

greggomann (approved):

qianzhangxa:
@mesosphere-mergebot changelog-not-required internal

DominikDary (approved):

PR #7634, opened 2020-09-21
Stop all services at once during upgrade

janisz:
## High-level description

Currently we stop servies one by one. This may lead to a situation when we stop service before we stop it's socket activator. So if any packet gets sent to that socket, the service will be activated. To prevent this we need to stop all services in a single command.

  - [COPS-6512](https://jira.d2iq.com/browse/COPS-6512) [DC/OS] [1.13.9]dcos-telegraf.socket  down after a patch

## Related tickets (optional)

### DIFF
--- a/pkgpanda/__init__.py
+++ b/pkgpanda/__init__.py
@@ -80,22 +80,21 @@ class Systemd:
         if not os.path.exists(self.__unit_directory):
             log.warning("Do not stop services. %s does not exist", self.__unit_directory)
             return
-        for name in os.listdir(self.__unit_directory):
-            # Skip directories
-            if os.path.isdir(os.path.join(self.__unit_directory, name)):
-                continue
-            try:
-                cmd = ["systemctl", "stop", name]
-                if not self.__block:
-                    cmd.append("--no-block")
-                check_call(cmd)
-            except CalledProcessError as ex:
-                # If the service doesn't exist, don't error. This happens when a
-                # bootstrap tarball has just been extracted but nothing started
-                # yet during first activation.
-                log.warning(ex)
-                if ex.returncode != 5:
-                    raise
+        names = list(filter(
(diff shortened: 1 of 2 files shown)


janisz:
@mesosphere-mergebot request-review @jongiddy @pierrebeitz  @DominikDary

pkgpanda/__init__.py
```diff
@@ -80,22 +80,21 @@ def stop_all(self):
         if not os.path.exists(self.__unit_directory):
             log.warning("Do not stop services. %s does not exist", self.__unit_directory)
             return
-        for name in os.listdir(self.__unit_directory):
```
janisz, line 83: Alternatively we can just sort list A→Z to stop socket before service

janisz:
@jongiddy Do we have a test for it?

cgawron1:
@mesosphere-mergebot test teamcity/dcos/build/tox

jongiddy:
> Do we have a test for it?

For restarting services? We have some e2e tests that perform a same-version upgrade (`EE-Integration-Test/E2E/group-2`), so they should identify any really obvious problems with this change. There's probably other things that could be going wrong that those tests don't check.

They should get picked up when this change is added to Soak, which is unfortunately not very timely.

I can't see any specific reason why this approach should be any worse than the previous approach. There may be some required order of shutdown that happens to be achieved in the previous approach, but, if so, we need to identify it and make it more explicit.

jongiddy:
@mesosphere-mergebot sync

jongiddy (approved):
The right solution is to modify systemd dependencies correctly, not just the `dcos-target.wants` mechanism that we have now. However, that is a non-trivial change.

This faster approach appears to fix the immediate problem, and passes the integration tests following an upgrade.

jongiddy:
@mesosphere-mergebot sync

janisz:
@mesosphere-mergebot backport 2.1 2.0 1.13

PR #7683, opened 2020-09-30
NetworkManager ignores

jkoelker:
## High-level description

Adds an ignore device file for NetworkManager similar to how we do for systemd-networkd. It appears that in RHEL 7.8 sometimes NetworkManager starts after dcos-net and that ends up wiping out the 2nd and 3rd ip (198.51.100.2-3) on the spartan device. 

  - [COPS-6519](https://jira.d2iq.com/browse/COPS-6519) dcos-net not always able to set all Spartan IP addresses after node reboot

### DIFF
--- a/packages/dcos-net/extra/dcos-net-setup.py
+++ b/packages/dcos-net/extra/dcos-net-setup.py
@@ -24,11 +24,32 @@ import shutil
 import subprocess
 import sys
 
+DEBUG = os.environ.get('DCOS_NET_STARTUP_DEBUG')
+
+
+def run(cmd, *args, **kwargs):
+    command = ' '.join(cmd)
+
+    if DEBUG:
+        print('command: `{}`'.format(command))
+
+    result = subprocess.run(cmd, *args, **kwargs)
+    if result.stderr:
+        sys.stderr.buffer.write(result.stderr)
+
+    if DEBUG:
+        print('command: `{}` exited with status `{}`'.format(
+            command,
+            result.returncode,
+        ))
+
+    return result
+
 
 def main():
     return_code = 0
     if sys.argv[1:4] in [['ip', 'link', 'add'], ['ip', 'addr', 'add'], ['ip', '-6', 'addr']]:
-        result = subprocess.run(sys.argv[1:], stderr=subprocess.PIPE)
+        result = run(sys.argv[1:], stderr=subprocess.PIPE)
         sys.stderr.buffer.write(result.stderr)
         if result.stderr.strip().endswith(b'File exists'):
             result.returncode = 0
@@ -36,22 +57,22 @@ def main():
     elif sys.argv[1] == 'iptables':
         # check whether a rule matching the specification does exist
         argv = ['-C' if arg in ['-A', '-I'] else arg for arg in sys.argv[1:]]
-        result = subprocess.run(argv)
+        result = run(argv)
         if result.returncode != 0:
             # if it doesn't exist append or insert that rules
-            result = subprocess.run(sys.argv[1:])
+            result = run(sys.argv[1:])
         return_code = result.returncode
     elif sys.argv[1] == '--ipv6':
         if os.getenv('DCOS_NET_IPV6', 'true') == 'false':
             sys.exit(0)
         else:
             del sys.argv[1]
-            result = subprocess.run(sys.argv)
+            result = run(sys.argv)
             return_code = result.returncode
     elif sys.argv[1:3] == ['networkd', 'add'] and len(sys.argv) == 4:
         return_code = add_networkd_config_for_coreos(sys.argv[3])
     else:
-        result = subprocess.run(sys.argv[1:])
+        result = run(sys.argv[1:])
         return_code = result.returncode
     sys.exit(return_code)
 
@@ -70,8 +91,8 @@ def add_networkd_config_for_coreos(src: str) -> int:
     networkd_path = '/etc/systemd/network'
 
     # Check if there is networkd
-    result = subprocess.run(['systemctl', 'list-unit-files', networkd],
-                            stdout=subprocess.PIPE)
+    result = run(['systemctl', 'list-unit-files', networkd],
+                 stdout=subprocess.PIPE)
     if result.returncode != 0:
         return result.returncode
     if networkd not in result.stdout:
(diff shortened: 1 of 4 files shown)


jkoelker:
@mesosphere-mergebot test all

jkoelker:
@mesosphere-mergebot ready-for-review-label

jkoelker:

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+
+    return result
+
 
 def main():
     return_code = 0
     if sys.argv[1:4] in [['ip', 'link', 'add'], ['ip', 'addr', 'add'], ['ip', '-6', 'addr']]:
-        result = subprocess.run(sys.argv[1:], stderr=subprocess.PIPE)
+        result = run(sys.argv[1:], stderr=subprocess.PIPE)
         sys.stderr.buffer.write(result.stderr)
```
jongiddy, line 53: This line can be removed as the `run` function does this now.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+def check_for_unit(unit: str) -> int:
+    result = run(['systemctl', 'list-unit-files', unit],
+                 stdout=subprocess.PIPE)
+
+    if result.returncode == 0 and unit in result.stdout.decode():
         return 0
 
-    networkd = b'systemd-networkd.service'
-    networkd_path = '/etc/systemd/network'
+    return result.returncode
```
jongiddy, line 89: This function always returns `result.returncode`. The `if` has no effect. Should this return something different if `result.returncode == 0` but `unit **not** in result.stdout.decode()`?

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+    if result.returncode == 0:
+        return True
+
+    return False
+
+
+def copy_file(src, dst) -> bool:
+    try:
+        if not filecmp.cmp(src, dst):
+            return bool(shutil.copyfile(src, dst))
```
jongiddy, line 104: `shutil.copyfile` always return `dst`, so I think this would be clearer as:
```
shutil.copyfile(src, dst)
return True
```

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+    except FileNotFoundError:
+        return bool(shutil.copyfile(src, dst))
+
 
-    # Check if there is networkd
-    result = subprocess.run(['systemctl', 'list-unit-files', networkd],
-                            stdout=subprocess.PIPE)
+def add_config(unit: str, src: str, path: str) -> int:
+    # Check if the unit exists
+    result = check_for_unit(unit)
```
jongiddy, line 113: `check_for_unit` returns an `int`, not a `subprocess` result. The following code can check `result != 0` directly.

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
-        return filecmp.cmp(src, dst)
-    except FileNotFoundError:
-        return False
+    # Restart unit only if it's active
+    if not is_unit_active(unit):
+        return 0
+
+    # Restart the unit only if the configuration is updated
+    if replaced:
+        return run(['systemctl', 'restart', unit]).returncode
```
jongiddy, line 131: :grey_question: This might be simpler as:
```
config_replaced = copy_file(src, dst)
# Restart the unit only if configuration changed and unit is active
if config_replaced and is_unit_active(unit):
    ...
```

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+
+    return 0
+
+
+def add_networkmanager_config(src: str) -> int:
+    if not NETWORK_MANAGER or 'el' not in platform.release():
+        return 0
+
+    return add_config('NetworkManager.service', '/etc/NetworkManager/conf.d', src)
```
jongiddy, line 140: :grey_question: Do you think it's clearer with fewer not's?
```
if NETWORK_MANAGER and 'el' in platform.release():
    return add_config('NetworkManager.service', '/etc/NetworkManager/conf.d', src)
return 0
```

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -15,110 +15,142 @@
...
+    # systemd-networkd, when enabled, will wipe the configurations like IP
+    # address of network interfaces and this behavior happens only on coreos
+    # This problem is tracked by:
+    # https://jira.mesosphere.com/browse/DCOS_OSS-1790
+    # We need to mark interfaces managed by DC/OS as unmanaged when networkd is
+    # enabled on coreos
+    if not NETWORKD or 'coreos' not in platform.release():
+        return 0
+
+    return add_config('systemd-networkd.service', '/etc/systemd/network', src)
```
jongiddy, line 153: Ditto on fewer nots.

jongiddy (commented):
Mostly suggestions. However, there seems to be confusion about the return value for `check_for_unit`.

packages/dcos-net/extra/dcos-net-setup.py
jkoelker: Nice catch!

packages/dcos-net/extra/dcos-net-setup.py
jkoelker: :+1: I was trying to hard to prevent retuning statics instead of the results, its much clearer splitting it up.

packages/dcos-net/extra/dcos-net-setup.py
jkoelker: Much clearer and simpler, thank you!

packages/dcos-net/extra/dcos-net-setup.py
jkoelker: Yep thats nicer, good call to invert the logic.

jkoelker:
Thanks for the detailed review! Fixed the errors and added the suggestions.

jkoelker:
Whoops, it like like inverting the result from check unit caused a failure look since we don't have network manager running on out images by default. Re-factored to add in a negative return to signal unit not found, that will successfully exit the function, and added a few more `if DEBUG` printing statements to catch issues with the unit checking.

DominikDary:
@jkoelker this should be in the "ready to review" state, right?

jongiddy (approved):
LGTM!

jkoelker:
@mesosphere-mergebot backport 2.1 2.0 1.13

PR #7789, opened 2020-10-28
Use retry on test_dcos_cni_l4lb as from centos 7.6 first call will fail

fatz:
## High-level description

D2IQ-43857 is mentioning that this feature doesn't work at all. The truth is that only the first call fails. Doing a second call to the same address works fine. With that in mind this PR adds retry logic to `test_dcos_cni_l4lb` so this test no longer fails on centos >= 7.6

  - [D2IQ-43857](https://jira.d2iq.com/browse/D2IQ-43857) test_dcos_cni_l4lb fails for 1.12.0-rc3 & 1.11.6 on RHEL 7.5

### DIFF
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -753,6 +753,7 @@ def test_l4lb(dcos_api_session: DcosApiSession) -> None:
             app.purge(dcos_api_session)
 
 
+@retrying.retry(wait_fixed=5000, stop_max_attempt_number=3)
 def test_dcos_cni_l4lb(dcos_api_session: DcosApiSession) -> Any:
     '''
     This tests the `dcos - l4lb` CNI plugins:
--- a/packages/dcos-integration-test/extra/test_networking.py
+++ b/packages/dcos-integration-test/extra/test_networking.py
@@ -219,7 +219,8 @@ def lb_enabled() -> Any:
 
 @retrying.retry(wait_fixed=2000,
                 stop_max_delay=5 * 60 * 1000,
-                retry_on_result=lambda ret: ret is None)
+                retry_on_result=lambda ret: ret is None,
+                retry_on_exception=lambda e: isinstance(e, Exception))
 def ensure_routable(cmd: str, host: str, port: str, json_output: bool = True) -> Any:
     proxy_uri = 'http://{}:{}/run_cmd'.format(host, port)
     log.info('Sending {} data: {}'.format(proxy_uri, cmd))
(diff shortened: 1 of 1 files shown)


fatz:
@mesosphere-mergebot backport 2.2 2.1 2.0

DominikDary:
@mesosphere-mergebot request-review  @jkoelker @DominikDary

packages/dcos-integration-test/extra/test_networking.py
```diff
@@ -753,6 +753,7 @@ def getjson(url: str) -> Any:
             app.purge(dcos_api_session)
 
 
+@retrying.retry(wait_fixed=5000, stop_max_attempt_number=3)
```
jkoelker, line 756: Given that its only the first call, should the retry be on the `ensure_routable` function to just retry the call and keep the same app and vip?

fatz: yeah I get your point... this actually works as its a fixed ip... 🤔 

but that `retrying.retry` annotation is only meant for the outer function. Or can it also be called with a lambda?

jkoelker: \<sidenote\>
It could technically wrap a lambda since the `@` syntax is just shorthand for passing in a function:

```python
func = retrying.retry(wait_fixed=5000, stop_max_attempt_number=3)(lambda x: x)
```
\</sidenote\>

I started following the 🐇 a bit today and when I went and looked and it looks like there is already a retry on `ensure_routable`: https://github.com/dcos/dcos/blob/02066c174a3ac5552cc80c5e9b2c81efd52e53aa/packages/dcos-integration-test/extra/test_networking.py#L220-L223

This got me looking back into the error on the OS qualifications https://jenkins.mesosphere.com/service/jenkins/blue/organizations/jenkins/dcos-images/detail/PR-546/11/pipeline#step-93-log-1350 and its raising an exception, which the existing retry decoration of `ensure_routable` doesn't retry on! I think the fix here is to retry on exception for `ensure_routable`:

```python

 @retrying.retry(wait_fixed=2000, 
                 stop_max_delay=5 * 60 * 1000, 
                 retry_on_result=lambda ret: ret is None,
                 retry_on_exception=lambda x: True) 
 def ensure_routable(cmd: str, host: str, port: str, json_output: bool = True) -> Any: 
```

before as soon as the timeout exception happened, it bailed out without ever retrying.

fatz: @jkoelker so i've tried that but it doesn't work....

Checking retrying's doc `retry_on_exception` is just for setting a specific exception to retry on. Without it every exception should be fine.

However I'm not sure why but I get the feeling the decorator is not handled.

fatz:
@jkoelker got a solution please review

jkoelker (approved):

jkoelker:
\o/

jkoelker:
@mesosphere-mergebot changelog-not-required test only change

jkoelker (approved):

fatz:
@mesosphere-mergebot merge-it

jkoelker:
@mesosphere-mergebot backport 2.2 2.1 2.0

PR #7814, opened 2020-11-03
Restart Docker only if restart file exists

jongiddy:
## High-level description

Only restart Docker if a restart file exists. This allows bootstrap to determine when a restart is necessary while keeping the certificate name the same.

  - [D2IQ-72846](https://jira.d2iq.com/browse/D2IQ-72846) upgrade from 2.1.{0,1} to 2.2.0 stops workloads 

## Related tickets (optional)


jongiddy:
@mesosphere-mergebot test all

PR #7831, opened 2020-11-12
dcos-net: fix argument order for `add_config`

jkoelker:
## High-level description

dcos-net: fix argument order for `add_config`

  - [D2IQ-73113](https://jira.d2iq.com/browse/D2IQ-73113) dcos-net fails to start on RHEL/Centos 7.8 in dcos 2.2
  - [COPS-6519](https://jira.d2iq.com/browse/COPS-6519) dcos-net not always able to set all Spartan IP addresses after node reboot

### DIFF
--- a/gen/dcos-config.yaml
+++ b/gen/dcos-config.yaml
@@ -111,9 +111,6 @@ package:
       Unmanaged=yes
   - path: /etc/dcos-net.conf
     content: |
-      [main]
-      plugins=keyfile
-
       [keyfile]
       unmanaged-devices=interface-name:docker*;interface-name:m-*;interface-name:d-*;interface-name:vtep*;interface-name:spartan;interface-name:minuteman;
   - path: /etc/dcos_net
--- a/packages/dcos-net/extra/dcos-net-setup.py
+++ b/packages/dcos-net/extra/dcos-net-setup.py
@@ -111,7 +111,7 @@ def copy_file(src, dst) -> bool:
         return True
 
 
-def add_config(unit: str, src: str, path: str) -> int:
+def add_config(unit: str, path: str, src: str) -> int:
     # Check if the unit exists
     result = check_for_unit(unit)
     if result < 0:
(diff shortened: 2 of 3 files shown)


jkoelker:

jkoelker:
@mesosphere-mergebot request-review  @jongiddy

packages/dcos-net/extra/dcos-net-setup.py
```diff
@@ -111,7 +111,7 @@ def copy_file(src, dst) -> bool:
         return True
 
 
-def add_config(unit: str, src: str, path: str) -> int:
+def add_config(unit: str, path: str, src: str) -> int:
```
jkoelker, line 114: from the calls below:

```python
def add_networkmanager_config(src: str) -> int:
    if NETWORK_MANAGER or 'el' in platform.release():
        return add_config('NetworkManager.service', '/etc/NetworkManager/conf.d', src)
    return 0
def add_networkd_config(src: str) -> int:
    # systemd-networkd, when enabled, will wipe the configurations like IP
    # address of network interfaces and this behavior happens only on coreos
    # This problem is tracked by:
    # https://jira.mesosphere.com/browse/DCOS_OSS-1790
    # We need to mark interfaces managed by DC/OS as unmanaged when networkd is
    # enabled on coreos
    if NETWORKD or 'coreos' in platform.release():
        return add_config('systemd-networkd.service', '/etc/systemd/network', src)
```

CI didn't catch this (and neither did I because I ran in on aws ;( ) because the `check_for_unit` is returning that `NetworkManager.service` is not running, so no need to patch it in.

jongiddy: Does Flatcar have `coreos` in its release string?

jkoelker: Good catch, looks like they changed it as leas as of the most recent ami:

```
core@ip-172-31-25-228 ~ $ docker run python python -c 'import platform; print(platform.release())'
5.4.72-flatcar
```

I've added a commit to also run this on flatcar.

jkoelker:
@mesosphere-mergebot request-review @cneth

gen/dcos-config.yaml
```diff
@@ -111,9 +111,6 @@ package:
       Unmanaged=yes
   - path: /etc/dcos-net.conf
     content: |
-      [main]
-      plugins=keyfile
-
```
jkoelker, line 116: This is per RH, we shouldn't be setting this ref: COPS-6561 ;)

jongiddy (approved):

jkoelker:
@mesosphere-mergebot backport 2.2 2.1 2.0 1.13

PR #7878, opened 2021-02-16
Add notice about DC/OS being abandoned by D2iQ.

waptaff:
## High-level description

Project should warn potential new users about the possibility of DC/OS development seriously slowing down or stopping.

Nothing on https://dcos.io/ mentions this winding down, and the homepage still mentions: “DC/OS is created and maintained by D2iQ”.


waptaff:
This should not be closed.

